PR #18, opened 2026-04-10
feat: add docker-compose profile for frontend-only dev

Matheus083:
"Hi @warren618! I've implemented the requested profile. Now, contributors can run only the frontend and backend containers using docker compose --profile frontend up. I've also verified that the containers are communicating correctly. Let me know if you need any adjustments!"

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,19 @@ services:
       - vibe-runs:/app/agent/runs
       - vibe-sessions:/app/agent/sessions
     restart: unless-stopped
+    profiles: ["frontend", "fullstack"] 
+
+  frontend:
+    image: node:20 
+    working_dir: /app
+    ports:
+      - "5173:5173" 
+    volumes:
+      - ./frontend:/app 
+    command: sh -c "npm install && npm run dev -- --host" 
+    profiles: ["frontend"] 
+    depends_on:
+      - vibe-trading
 
 volumes:
   vibe-runs:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,19 +9,22 @@ services:
       - vibe-runs:/app/agent/runs
       - vibe-sessions:/app/agent/sessions
     restart: unless-stopped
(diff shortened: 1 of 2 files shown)


warren618 (changes requested):
Thanks for the contribution @Matheus083! A few issues to address before merging:

### 1. Profile on `vibe-trading` breaks default `docker compose up`

Adding `profiles: ["frontend", "fullstack"]` to the `vibe-trading` service means a plain `docker compose up` **no longer starts anything** — users must now pass `--profile`. This is a breaking change for existing users.

The fix: **remove the `profiles` line from `vibe-trading`** so it always starts by default. Only the new `frontend` service needs a profile gate.

### 2. Frontend port mismatch

The project's frontend runs on port **5899** (configured in `frontend/vite.config.ts`). The PR exposes `5173:5173` which is Vite's default but not what this project uses. Should be `5899:5899`.

### 3. `package-lock.json` diff is unrelated noise

The removal of `libc` fields is likely caused by a different local npm version. Please revert `frontend/package-lock.json` so the PR only contains the docker-compose change.

### 4. Missing trailing newline

The last line of `docker-compose.yml` lost its trailing newline (`\ No newline at end of file`). Please add it back.

---

After fixing, the docker-compose should work like:
- `docker compose up` → only backend (same as before, no breaking change)
- `docker compose --profile frontend up` → backend + frontend dev server

Matheus083:
Hi @warren618! I've updated the PR with all your requested changes.

Additionally, I've improved the setup by:

Adding VITE_API_URL=http://vibe-trading:8899 to ensure the frontend container can communicate with the backend API via Docker's internal DNS.

Using explicit depends_on syntax with condition: service_started for better reliability.

Verified that docker compose up still works for the backend-only flow.

I also reverted the package-lock.json and added the trailing newline. Ready.

warren618 (approved):
Looks good — all four issues from the initial review are addressed. Clean single-file change, default `docker compose up` behavior is preserved. Thanks @Matheus083!

PR #33, opened 2026-04-14
fix: update MiniMax provider config and fix temperature=0 API error

octo-patch:
## Summary

- **Fix outdated model name**: `.env.example` referenced `MiniMax-Text-01` which has been superseded. Updated to `MiniMax-M2.7` (current production model) and documented `MiniMax-M2.7-highspeed` as the faster alternative.
- **Fix temperature clamping**: MiniMax API rejects `temperature=0.0` (valid range is `(0.0, 1.0]`). When `LANGCHAIN_PROVIDER=minimax` and the default `LANGCHAIN_TEMPERATURE=0.0` is in effect, `build_llm()` now automatically clamps the value to `1.0`, preventing an otherwise-silent API validation error for all users who follow the example config.
- **Add unit tests**: Cover MiniMax env-var → `OPENAI_*` mapping, base URL propagation, temperature clamping from `0.0`, and preservation of explicitly-set positive temperatures.

## Test plan

- [ ] `pytest agent/tests/test_llm.py` — all 19 tests pass (11 pre-existing + 4 new MiniMax tests)
- [ ] Integration verified: `LANGCHAIN_PROVIDER=minimax`, `LANGCHAIN_MODEL_NAME=MiniMax-M2.7`, `LANGCHAIN_TEMPERATURE=0.0` → model responds correctly after auto-clamp

## API reference

- Chat (OpenAI Compatible): https://platform.minimax.io/docs/api-reference/text-openai-api

### DIFF
--- a/agent/.env.example
+++ b/agent/.env.example
@@ -52,9 +52,10 @@ OPENROUTER_BASE_URL=https://openrouter.ai/api/v1
 
 # --- MiniMax ---
 # LANGCHAIN_PROVIDER=minimax
-# LANGCHAIN_MODEL_NAME=MiniMax-Text-01
+# LANGCHAIN_MODEL_NAME=MiniMax-M2.7          # or MiniMax-M2.7-highspeed (faster)
 # MINIMAX_API_KEY=xxx
 # MINIMAX_BASE_URL=https://api.minimax.io/v1
+# Note: MiniMax requires temperature > 0. Set LANGCHAIN_TEMPERATURE=1.0 (default when using MiniMax)
 
 # --- Xiaomi MIMO ---
 # LANGCHAIN_PROVIDER=mimo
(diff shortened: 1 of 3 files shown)


warren618 (commented):
Thanks for the fix! One suggestion on the temperature clamping:

```python
if os.getenv("LANGCHAIN_PROVIDER", "openai").lower() == "minimax" and temperature <= 0.0:
    temperature = 1.0
```

Clamping to `1.0` (max randomness) seems too aggressive — users who set `temperature=0` typically want deterministic output. Since MiniMax's valid range is `(0.0, 1.0]`, could we clamp to `0.01` instead? That stays closest to the user's intent while satisfying the API constraint.

```python
temperature = 0.01
```

warren618:
Thanks for the fix @octo-patch! I tweaked the clamp value to `0.01` instead of `1.0` — keeps it closest to deterministic while satisfying the MiniMax API. Merged now 🎉

Welcome to the contributors list!

Issue #39, opened 2026-04-18
[Bug] Kimi-k2.5 thinking mode is not supported


codewings:
### Description

kimi-k2.5 enable thinking mode as default, but they use 'reasoning_content' field to save the thinking content. So When I use Kimi-k2.5, I often receive the following error message:

> thinking is enabled but reasoning_content is missing in assistant tool call message at index 2

### Steps to Reproduce

1. Use Kimi-k2.5 as the LLM provider
2. Ask something to trigger the thinking

### Error Logs

```shell
Reason: Error code: 400 - {'error': {'message': 'thinking is enabled but reasoning_content is missing in assistant tool call message at index 2', 'type': 'invalid_request_error'}}
```

### Interface

CLI (vibe-trading)

### LLM Provider

kimi-k2.5

### Version

v0.1.0

### Environment

macOS 26.3.1

yule153604:
I opened a PR with a proposed fix. I could not do a full end-to-end verification because I do not currently have a valid official Kimi/Moonshot API key, but the implementation is aligned with the current official docs and includes a regression test. If you have a working official key, could you please test the PR branch and confirm whether it resolves the issue?

warren618:
Fixed in #41 (merged at 5350c88). Thanks for the detailed repro + error log @codewings — the root cause was exactly what the error said: `reasoning_content` was being dropped when we replayed assistant tool-call messages.

Both the main agent loop and the swarm worker path preserve `reasoning_content` now, with LangChain-compatible fallbacks (`additional_kwargs` / `response_metadata`).

Available in the next release (0.1.6, ETA within a few days). If you want to run off main in the meantime, `pip install git+https://github.com/HKUDS/Vibe-Trading.git` works.

If you have a Kimi K2.5 key handy, a real end-to-end confirmation would be much appreciated — our regression suite covers the protocol layer but we didn't do a live-key integration test.

codewings:
Thanks to @yule153604 for the fix. But I still got the same error when I use swarm agent after I pulled the latest code from the main branch.

It seems you need to use ChatMoonshot from the langchain_moonshot package; otherwise, the additional_kwargs field may not include reasoning_content.

warren618:
You're right @codewings — PR #41 was a no-op end-to-end. Thanks for catching it.

[...]

PR #41, opened 2026-04-18
fix(kimi): preserve reasoning_content in thinking tool-call context

yule153604:
## Summary

- Fixes Kimi K2.5 thinking/tool-call compatibility by preserving `reasoning_content` in replayed assistant tool-call messages.

- Extracts `reasoning_content` from provider responses and carries it through the shared `LLMResponse` object.

- Adds regression tests to prevent this compatibility break from recurring.

## Why

Closes (#39 )the Kimi `400 invalid_request_error` where follow-up tool-call requests fail with `thinking is enabled but reasoning_content is missing in assistant tool call message...`. 

## Changes

- Core response parsing: Updated `agent/src/providers/chat.py` to add `reasoning_content` to `LLMResponse` and extract it from direct attributes plus LangChain fallbacks.

- Assistant tool-call formatting: Updated `agent/src/agent/context.py` so assistant messages with `tool_calls` include `reasoning_content` when present.

- Main agent loop: Updated `agent/src/agent/loop.py` to pass `reasoning_content` back into replayed tool-call messages, with `thinking_text` as a fallback.

- Swarm execution path: Updated `agent/src/swarm/worker.py` so swarm runs preserve the same field.

- Regression tests: Added `agent/tests/test_kimi_reasoning_content.py` covering extraction, fallback extraction, preservation, and omission behavior.

## Test Plan

- [x] New regression tests pass (`python -m pytest agent/tests/test_kimi_reasoning_content.py -q`)

- [ ] Integration verified with a valid official Moonshot/Kimi API key against `kimi-k2.5` thinking + tool-calls

- [x] Tested manually: verified local protocol-layer behavior preserves `reasoning_content` in parsed responses and formatted assistant tool-call messages

## Checklist

### DIFF
--- a/agent/src/agent/context.py
+++ b/agent/src/agent/context.py
@@ -209,7 +209,11 @@ class ContextBuilder:
         }
 
     @staticmethod
-    def format_assistant_tool_calls(tool_calls: list, content: Optional[str] = None) -> Dict[str, Any]:
+    def format_assistant_tool_calls(
+        tool_calls: list,
+        content: Optional[str] = None,
+        reasoning_content: Optional[str] = None,
+    ) -> Dict[str, Any]:
         """Format an assistant tool_calls message, preserving thinking text.
 
         Args:
@@ -219,7 +223,7 @@ class ContextBuilder:
         Returns:
             OpenAI-format assistant message.
         """
-        return {
+        message = {
             "role": "assistant",
             "content": content,
             "tool_calls": [
@@ -234,3 +238,6 @@ class ContextBuilder:
                 for tc in tool_calls
             ],
         }
+        if reasoning_content is not None:
+            message["reasoning_content"] = reasoning_content
+        return message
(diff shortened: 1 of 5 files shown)


warren618 (approved):
LGTM. Clean, minimal, and the choice to cover both `loop.py` and `swarm/worker.py` is the right call — otherwise swarm runs would silently regress while normal runs work.

The LangChain-compatible fallback chain (`additional_kwargs` → `response_metadata`) also means this should benefit DeepSeek / Qwen thinking providers, not just Kimi — nice side effect.

Cross-checked against all 13 configured providers: field is only added when `reasoning_content is not None`, so non-thinking providers see zero change. Zero regression risk.

Small nit (I'll land as a follow-up, not blocking):
- `format_assistant_tool_calls` docstring should mention the new `reasoning_content` param.

Merging. Thanks @yule153604 — this is your second shipped contribution after the i18n READMEs.

PR #75, opened 2026-05-07
feat: implement TavilySearchTool for more robust financial information retrieval

panjunchang:
feat: add TavilySearchTool。

### Problem
The current financial information retrieval relies on traditional web scraping, which often fails due to website anti-scraping measures or structure changes, leading to unstable agent performance.

### Solution
I have introduced `TavilySearchTool`, which utilizes the Tavily AI search API. This provides a much more stable and reliable way for the trading agent to fetch real-time financial news and data.

### Changes
- Added `agent/src/tools/tavily_search_tool.py`
- Updated `.env.example` to include `TAVILY_API_KEY`
- Updated `pyproject.toml` dependencies

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,143 @@
+# ============================================================================
+# LLM Provider — uncomment ONE provider block below
+# ============================================================================
+
+# --- OpenRouter (recommended, multi-model gateway) ---
+LANGCHAIN_PROVIDER=openrouter
+LANGCHAIN_MODEL_NAME=deepseek/deepseek-v3.2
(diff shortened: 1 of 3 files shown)


warren618:
Thanks for the PR.

We are closing this because the project already has DuckDuckGo-based `web_search`, so we do not need an additional Tavily search provider.

This PR also has blocking integration issues: the tool is placed at the repository root instead of `agent/src/tools`, it is not covered by tests, and it changes unrelated release metadata.

We are not planning to add this dependency.

PR #79, opened 2026-05-08
feat: 使用oxc进行格式化、增加多语言切换

DvShu:
1. 使用 oxc 进行代码格式化
2. 增加多语言切换，增加了：简体中文、繁体中文、English


warren618:
Thanks for the contribution.

After reviewing this PR, I don’t think we should introduce i18n / multi-language switching at this stage.

Vibe-Trading is still moving quickly, and much of the frontend copy, settings surface, agent workflow, and dashboard UI are changing frequently. Maintaining multiple locale files right now would add extra synchronization cost and make it easy for translations to become stale or inconsistent.

So for now, we prefer to keep the frontend UI text in English only until the product surface is more stable.

There are also a few implementation issues in this PR that make it hard to merge as-is:

- it mixes i18n, formatter setup, lockfile changes, and proxy changes in one PR
- `vite.config.ts` hardcodes the proxy target to a public IP, which conflicts with the `VITE_API_URL` fix merged in #82
- `package.json` and the npm lockfile are out of sync, so `npm ci` fails

The oxc/oxlint/oxfmt part may be useful, though. If you’re still interested, please feel free to open a separate focused PR for frontend lint/format tooling only.

For that PR, please keep the existing npm workflow (`package-lock.json`, `npm ci`), avoid unrelated UI/i18n/proxy changes, and include clear scripts such as `lint` / `format:check` so we can wire them into CI cleanly.

I’m going to close this PR for now. Smaller focused PRs for bug fixes, docs, or narrowly scoped frontend improvements are still very welcome.

PR #80, opened 2026-05-08
fix(api): validate run_id and session_id path parameters

SJoon99:
## Summary

- Validate `run_id` and `session_id` path parameters against a strict
  character-class regex before they flow into `RUNS_DIR / run_id / ...`
  filesystem paths.
- Mirrors the existing `_SHADOW_ID_RE` guard already in this file, applied
  consistently to all 13 endpoints under `/runs/`, `/swarm/runs/`, and
  `/sessions/`.

## Why

`agent/api_server.py` already validates `{shadow_id}` with
`_SHADOW_ID_RE = re.compile(r"^shadow_[0-9a-f]{8}$")` (line 1459). The
neighbouring `{run_id}` and `{session_id}` endpoints did not have an
equivalent guard, so an authenticated caller could submit values containing
`..` or path separators and the handler would happily build a `Path` outside
the intended `RUNS_DIR` / `SESSIONS_DIR`. This change closes the gap with
the same lightweight pattern.

> SECURITY.md asks for vulnerabilities to be reported via private advisory
> first. The exposure here is limited to authenticated callers (every
> affected route is behind `Depends(require_auth)`), so I judged a public
> hardening PR to be appropriate. If you'd prefer this be moved to a
> private advisory, please let me know and I'll close this PR and re-open
> via [Security Advisories](https://github.com/HKUDS/Vibe-Trading/security/advisories/new).

## Changes

- `agent/api_server.py`
  - New helper `_validate_path_param(value, kind)` and the matching
    `_SAFE_PATH_PARAM_RE = re.compile(r"^[A-Za-z0-9_-]{1,128}$")`.
  - Helper called at the top of every handler whose path includes
    `{run_id}` or `{session_id}`. 13 endpoints total.
- `agent/tests/test_security_auth_api.py`
  - Parametrized validator unit tests: 5 valid IDs (covering the three
    real ID formats the codebase produces) and 9 invalid inputs
    (`..`, separators, dot, null byte, oversize).
  - Per-endpoint integration tests asserting `400` + `detail="invalid
    run_id"` or `"invalid session_id"`.

The character class `[A-Za-z0-9_-]` accepts every ID currently produced by
the codebase:

| Source | Format | Example |
|---|---|---|
| `state.create_run_dir` | `{YYYYmmdd}_{HHMMSS}_{2digit}_{6hex}` | `20260105_120342_12_a1b2c3` |
| `swarm/presets.py:271` | `swarm-{ts}-{short_uuid}` | `swarm-20260105_120342-a1b2c3` |
| `session/models.py:45` | `uuid.uuid4().hex[:12]` | `abcdef012345` |

It rejects path separators, dots (so `..` cannot pass), NUL bytes,
whitespace, and oversize input.

## Test Plan

- [x] New tests added (validator unit + per-endpoint integration)
- [x] Tested manually:
  - `curl http://127.0.0.1:8899/runs/foo.bar/pine -H "Authorization: Bearer $KEY"` → `400 {"detail": "invalid run_id"}`
  - Legit run_id (`20260105_120342_12_a1b2c3`) still returns the run
  - All three real ID formats accepted by the new regex (verified
    programmatically in `test_validate_path_param_accepts_known_good_values`)

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) — only `api_server.py` and tests.
- [x] No hardcoded values (regex constant lives in module scope, named).
- [x] Code follows CONTRIBUTING.md guidelines (Google-style docstrings, no comments stating the obvious).
- [x] Documentation updated — N/A (internal hardening, no user-facing behavior change for valid IDs).

## Notes for reviewer

- I deliberately matched the existing `_SHADOW_ID_RE` style (`__import__("re").compile(...)`) so the patch is locally consistent. Happy to switch both to a top-level `import re` in a follow-up if you prefer.
- I did NOT touch `agent/cli.py` or `agent/src/session/store.py`. The CLI is a local trust boundary and `session/store.py` is in the protected `src/session/` area per CONTRIBUTING.md. The validation happens at the HTTP boundary in `api_server.py`, which is the correct layer.
- The `{shadow_id}` endpoint keeps its existing `_SHADOW_ID_RE` (stricter than the new generic regex). I considered consolidating but decided against it to minimise diff scope.

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -870,6 +870,31 @@ def _build_response_from_run_dir(run_dir: Path, elapsed: float, *, include_analy
 
 
 # ============================================================================
+# Path-parameter validation
+# ============================================================================
+
+# ``run_id`` and ``session_id`` flow directly into filesystem paths
+# (``RUNS_DIR / run_id`` etc.). Restrict to a safe character class so that
+# values like ``..`` or ``foo/../bar`` cannot escape the parent directory.
+_SAFE_PATH_PARAM_RE = __import__("re").compile(r"^[A-Za-z0-9_-]{1,128}$")
+
+
+def _validate_path_param(value: str, kind: str) -> None:
+    """Reject path parameters that could escape the parent directory.
+
+    Args:
+        value: User-supplied path-parameter value.
+        kind: Parameter name, used in the error detail.
+
+    Raises:
+        HTTPException: 400 when ``value`` does not match the safe character
+            class, mirroring the existing ``_SHADOW_ID_RE`` check.
+    """
+    if not _SAFE_PATH_PARAM_RE.match(value or ""):
+        raise HTTPException(status_code=400, detail=f"invalid {kind}")
+
+
+# ============================================================================
 # API Endpoints
 # ============================================================================
 
@@ -883,6 +908,7 @@ async def get_run_code(run_id: str):
     Returns:
         Map filename -> source text.
     """
+    _validate_path_param(run_id, "run_id")
     run_dir = RUNS_DIR / run_id / "code"
     if not run_dir.exists():
         raise HTTPException(status_code=404, detail=f"Code directory for run {run_id} not found")
@@ -904,6 +930,7 @@ async def get_run_pine(run_id: str):
     Returns:
         Object with pine script content and exists flag.
     """
+    _validate_path_param(run_id, "run_id")
     pine_path = RUNS_DIR / run_id / "artifacts" / "strategy.pine"
     if not pine_path.exists():
         return {"exists": False, "content": None}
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for the focused hardening PR. I tested this against the latest `main` after #82:

- clean local merge into `origin/main`
- `python -m py_compile agent/api_server.py agent/tests/test_security_auth_api.py`
- `git diff --check`
- `python -m pytest agent/tests/test_security_auth_api.py -q` → 34 passed
- verified all 13 `{run_id}` / `{session_id}` routes call `_validate_path_param`

One small security edge case before merge: `_SAFE_PATH_PARAM_RE.match(...)` with the current `$` anchor accepts a trailing newline, so URL-encoded values like `foo%0A` pass validation.

Example:
- `_validate_path_param("foo\n", "run_id")` is accepted
- `GET /runs/foo%0A/pine` returns 200 instead of 400

Could you switch the validator to `fullmatch` and add newline/whitespace regression coverage?

Suggested change:

```python
if not _SAFE_PATH_PARAM_RE.fullmatch(value or ""):
    raise HTTPException(status_code=400, detail=f"invalid {kind}")
```

Suggested test additions:
- validator rejects `"foo\n"`, `"foo\r"`, `"foo\t"`
- endpoint rejects `/runs/foo%0A/pine` with `400 {"detail": "invalid run_id"}`

After that, I’m happy to merge this.

PR #84, opened 2026-05-09
fix: persist swarm worker reports and improve summary extraction

iampengqian:
Swarm workers frequently produced no usable output because:
- The Execution Rules prompt asked agents to write summaries "directly in your response," but the worker only captured `last_assistant_content` (often a mid-process message like "let me fix the script").
- On timeout/iteration limit, all in-memory messages were discarded.
- Event logging recorded only tool names and elapsed time, not I/O.

Changes to worker.py:
- Force agents to write final reports via `write_file report.md` (Phase 3), decoupling report persistence from the in-memory message lifecycle.
- Add `_best_summary()` to extract the longest assistant message (>100 chars) instead of relying on the last fragment.
- Add `_persist_messages()` to save full conversation to messages.json on timeout/iteration-limit for post-mortem debugging.
- Include tool arguments in tool_call events and result content (up to 5000 chars) in tool_result events for better observability.

Changes to equity_research_team.yaml:
- Increase timeout_seconds from 600 to 900 for all four agents, as the aggregator previously timed out at 615s.

## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/src/swarm/presets/commodity_research_team.yaml
+++ b/agent/src/swarm/presets/commodity_research_team.yaml
@@ -32,7 +32,7 @@ agents:
     tools: [bash, read_file, write_file, load_skill, read_url]
     skills: [commodity-analysis, web-reader, geopolitical-risk]
     max_iterations: 50
-    timeout_seconds: 600
+    timeout_seconds: 1800
     max_retries: 1
 
   - id: demand_analyst
@@ -65,7 +65,7 @@ agents:
     tools: [bash, read_file, write_file, load_skill, read_url]
     skills: [commodity-analysis, seasonal, global-macro]
     max_iterations: 50
-    timeout_seconds: 600
+    timeout_seconds: 1800
     max_retries: 1
 
   - id: cycle_strategist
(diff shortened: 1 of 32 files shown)


warren618 (changes requested):
Thanks for tackling swarm report persistence. Persisting post-mortem messages and moving the durable worker report into an artifact are both useful directions.

I think this needs changes before merge because the current implementation can regress the final swarm report shown by the API/UI, and the expanded event logging may persist or stream large/sensitive tool payloads.

Blocking: the prompt now requires the worker to write the full report to `report.md` and then return only a brief 2-3 sentence response. But the runtime still uses `response.content` as `WorkerResult.summary`, and `SwarmRuntime` later uses that summary as `run.final_report`.

That means the final report exposed by the dashboard/API can become only the short closing response, while the real report is hidden in artifacts. Please read `report.md` after the final `write_file` and use that content as the worker summary/final report, falling back to response text only if the file is missing.

This now persists and streams raw tool arguments plus up to 5000 chars of tool result content. That can include `write_file` content, file/document snippets, command output, or other sensitive/large payloads in `events.jsonl` and SSE.

Please keep this to a short preview, similar to the main AgentLoop, and explicitly redact or omit large/sensitive fields such as `content`, `env`, `headers`, and full file contents.

The PR changes behavior but does not add a regression test for the report handoff. A focused test where a worker writes `report.md` and then returns a short response would pin the intended summary/final_report behavior.

warren618 (approved):
Maintainer follow-up pushed to the PR branch. Verified locally: swarm report regression tests, preset packaging/inspect tests, py_compile, and diff check pass. Thanks @iampengqian for the fix.

Issue #42, opened 2026-04-19
[Bug] The Swarm agent often gets stuck.


codewings:
### Description

When I use the swarm agent, it often gets stuck. I’m not sure whether this is an issue with kimi-2.5 or something caused by background thread scheduling.

<img width="3526" height="1106" alt="Image" src="https://github.com/user-attachments/assets/7f3c141a-8474-412f-a1e3-fd07bff5f261" />

I tried again, and after waiting for more than 10 minutes, the result was somewhat different, but there were still issues.
<img width="3512" height="1068" alt="Image" src="https://github.com/user-attachments/assets/d8d70b72-2b46-4948-b42a-c6b6e5ab5e2c" />

### Steps to Reproduce

1. Use swarm to assemble a team to analyze any problem.
2. Waiting for run_swarm to finish.
3. The terminal had no output for a long time.

### Error Logs

```shell

```

### Interface

CLI (vibe-trading)

### LLM Provider

kimi-2.5

### Version

0.1.5

warren618:
Thanks for the detailed report @codewings — confirmed and fixed in `9b9d147`.

**Root cause.** Swarm workers were using `llm.chat()` (non-streaming `invoke`) for every ReAct iteration. I ran a matrix against moonshot/kimi, deepseek, and openai-style providers with your HK-stocks prompt:

| provider | non-streaming (`invoke`) | streaming |
|---|---|---|
| moonshot/kimi-k2 | **fails @ 183s with `APIConnectionError: Connection error`** | first chunk @ 3.4s, 63s to finish |
| deepseek | 35s ok | ok |

So `invoke` is specifically broken for moonshot — matching your two screenshots exactly: run #1 silent-hangs on a stalled connection; run #2 hits the 120s timeout and every retry spams `LLM call failed at iteration 1: Request timed out`.

**Fix:**
1. `worker.py` — swap `llm.chat()` → `llm.stream_chat(on_text_chunk=…)`. Same path the main Agent loop already uses, so other providers see zero change. Bonus: the dashboard now renders live `worker_text` deltas during kimi's thinking time.
2. `runtime.py` — `as_completed()` now enforces a layer deadline (`timeout_seconds * (max_retries + 1) + 60s`). Defensive belt for any future stuck past the in-loop timeout check.

**Verified end-to-end** with `kimi-k2-0905-preview` + `equity_research_team` preset:
- 724 streaming deltas across 3 parallel workers in 8 min
[...]

Issue #32, opened 2026-04-14
backtest MCP tool fails with "Connection closed"; direct subprocess path works


eltonaguiar:
# `backtest` MCP tool fails with `Connection closed`; subprocess path works

**Package:** `vibe-trading-ai 0.1.4` (pip, Python 3.14.x on Windows 11)
**Transport:** stdio MCP via `vibe-trading-mcp.exe` registered in Claude Code user config
**Affected tool:** `mcp__vibe-trading__backtest` only. All other 16 tools work.

## Symptoms

Every call to the `backtest` MCP tool returns:

```
MCP error -32000: Connection closed
```

Reproduced across:
- 2 separate Claude Code sessions
- 2 separate subagents spawned within a session
- 6+ tool-call attempts total
- 3 different `run_dir` inputs (new run_dir, pre-existing Cursor run_dir in `~/.vibe-trading/backtests/etf_rsi_mean_reversion`, fresh run_dir with identical content)
- Both relative and absolute paths

All six tools returned `Connection closed` with no partial output. No logs written to `<run_dir>/logs/`. The MCP server continues accepting other tool calls normally after the failure -- only the `backtest` tool handler crashes the connection.

## Working reproduction (direct Python, bypassing MCP)

The engine itself is completely fine. Invoking `backtest.runner.main()` directly via subprocess against the exact same `run_dir` succeeds in ~10 seconds and returns full metrics:

```python
from pathlib import Path
from backtest.runner import main
main(Path("/path/to/run_dir"))
# -> prints JSON metrics to stdout, writes artifacts/ and logs/
```

This strongly suggests the bug is in the MCP tool handler wrapper, not the backtest engine. Probably the handler is attempting to capture stdout/stderr in a way that deadlocks under stdio transport, or is returning a result payload that exceeds a transport buffer.

## Minimal reproduction run_dir

```
run_dir/
  config.json
  code/
    signal_engine.py
```

`config.json`:
```json
{
  "source": "yfinance",
  "codes": ["SPY", "QQQ", "XLK"],
  "start_date": "2020-01-01",
  "end_date": "2026-04-13",
  "initial_capital": 100000,
  "commission": 0.001
}
```

`code/signal_engine.py`: any `class SignalEngine` with a `generate(data_map) -> Dict[str, pd.Series]` method (e.g. the RSI example from `~/.vibe-trading/backtests/etf_rsi_mean_reversion/code/signal_engine.py`).

[...]

PR #86, opened 2026-05-09
test(mcp_server): add smoke test for tools/call deadlock regression

Teerapat-Vatpitak:
## Summary

- Add a regression test for [#85](https://github.com/HKUDS/Vibe-Trading/pull/85) (the MCP `tools/call` deadlock fix).
- Spawns `vibe-trading-mcp` over stdio and verifies the JSON-RPC happy path: `initialize` + `tools/list` + `tools/call analyze_options`.
- No network dependency; total runtime ~4s with the fix in place.

## Why

[#85](https://github.com/HKUDS/Vibe-Trading/pull/85) fixes a deadlock where `tools/call` hangs indefinitely if the registry pre-warm in `mcp_server.main()` is removed. Without a regression test, a future contributor could revert that line during a refactor and reintroduce the bug — `initialize` and `tools/list` would still pass, so the breakage would be silent until users hit `tools/call`.

This test fails deterministically (15s timeout on `tools/call`) when the pre-warm is removed, with an assertion message that points back at #85.

## What it covers

1. **`initialize`** — JSON-RPC handshake completes within 30s (covers cold imports).
2. **`tools/list`** — at least 20 tools registered, including `analyze_options`, `get_market_data`, and `list_skills`. Catches tool-registration regressions and missing package data.
3. **`tools/call analyze_options`** — pure CPU Black-Scholes calculation; must return within 15s. This is the call that deadlocks without the fix.
4. **Result shape** — `price > 0`, `delta ∈ [0, 1]`. Catches Black-Scholes formula regressions.

## Differential validation

Verified locally on Windows + Python 3.12 + fastmcp from main:

| Run | `mcp_server.main()` state           | Result                                | Wall time |
|-----|-------------------------------------|---------------------------------------|-----------|
| 1   | with pre-warm (current main)        | PASS                                  | 3.91s     |
| 2   | pre-warm commented out              | **FAIL — timeout 15s on tools/call**  | 17.45s    |
| 3   | pre-warm restored                   | PASS                                  | 3.81s     |

`tools/list` succeeds in all three runs — only `tools/call` exposes the deadlock, which is the exact symptom users hit (e.g. #42, #32).

## Why no network in the test

`analyze_options` is a Black-Scholes calculation with no external service. Picking it over `get_market_data` keeps the test:
- Deterministic (no API rate-limits / outages)
- CI-friendly (no outbound network needed)
- Fast (sub-second after init)

## Test plan

- [x] `pytest agent/tests/test_mcp_server_smoke.py -v` passes locally (~4s)
- [x] Test fails with a timeout when the pre-warm in `mcp_server.main()` is removed
- [x] Restoring the pre-warm makes the test pass again
- [x] Test only spawns one subprocess and cleans it up in `finally`; no shared state with other tests

## Notes for reviewer

Tested locally on Windows + Python 3.12 with the differential validation table above. CI will be the first run on Linux + Python 3.11. The test uses only portable APIs (`subprocess.Popen`, `pathlib.Path`, `os.pathsep`, `sys.executable`, `threading`), so cross-platform issues are unlikely — but I'm happy to follow up with fixes if anything surfaces in CI.

This is also the first subprocess-spawning test in this repo (existing tests use direct imports / FastAPI TestClient). I picked subprocess because the deadlock only manifests across the FastMCP server thread boundary, which is hard to mock faithfully. Happy to refactor toward a fixture-based / mocked approach if that fits the repo's testing philosophy better.

## Checklist

- [x] No changes to protected areas (`agent/src/agent/`, `agent/src/session/`, `agent/src/providers/`) — only adds a test file under `agent/tests/`.
- [x] No hardcoded secrets / paths; uses `sys.executable` and resolves repo root from `__file__`.
- [x] Code follows `CONTRIBUTING.md` (Conventional Commit title, `@pytest.mark.integration`, Google-style docstrings).
- [x] Documentation updated — N/A (test file is self-documenting).

### DIFF
--- a/agent/tests/test_mcp_server_smoke.py
+++ b/agent/tests/test_mcp_server_smoke.py
@@ -1,0 +1,218 @@
+"""Smoke test for the MCP server entry point.
+
+Regression test for PR #85 (deadlock fix). Spawns ``vibe-trading-mcp`` as a
+subprocess and walks through the full JSON-RPC happy path:
+
+1. ``initialize`` handshake.
+2. ``tools/list`` returns the registered tool catalogue.
+3. ``tools/call`` for ``analyze_options`` (a network-free Black-Scholes
+   calculation) returns a sane numeric result.
+
+The combined run cost is one cold spawn (~3-5s) plus sub-second per request.
+The most important assertion is that ``tools/call`` does not hang: without
+the registry pre-warm in ``mcp_server.main()``, the lazy registry build runs
+inside FastMCP's worker thread on the first ``tools/call`` and deadlocks.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+import threading
+import time
+from pathlib import Path
+from queue import Empty, Queue
+
+import pytest
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+AGENT_DIR = REPO_ROOT / "agent"
+
+# Generous bound — covers cold imports of fastmcp + tool registry build on
+# slow CI runners. The patched server typically finishes initialize in 3-9s.
+INIT_TIMEOUT = 30.0
+
+# tools/call analyze_options is a pure CPU calculation. With the fix in place
+# it returns in well under a second; without it, it never returns. 15s is a
+# safe lower bound that still flags the deadlock quickly.
+CALL_TIMEOUT = 15.0
+
+# Tools we rely on as a baseline. The repo currently ships 22 tools; we
+# assert ``>= 20`` so unrelated tool additions / removals don't break the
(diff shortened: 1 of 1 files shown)


warren618:
Thanks for adding this MCP smoke test. I verified the test locally as well, and it is useful for guarding against the tools/call deadlock regression.

One small note for future PRs: since this is an integration-style subprocess test, we may later tune the default pytest selection if the suite gets slower. For now this looks good.

Teerapat-Vatpitak:
Thanks for the local verify and the heads-up on integration-style test categorization — noted for future PRs (likely will add a `@pytest.mark.integration` marker if I introduce more subprocess tests, so default `pytest` runs stay fast).

PR #96, opened 2026-05-10
Improve CLI run UI and preflight startup check

ykykj:
## Summary
  - Refresh CLI welcome/settings/result UI with clearer panels and live run dashboard
  - Keep interactive input on the existing main-branch PromptSession flow
  - Avoid importing akshare during preflight by using importlib discovery

  ## Tests
  - python -m pytest agent/tests/test_preflight.py agent/tests/test_cli_init.py agent/tests/test_llm.py
  - git diff --cached --check


warren618:
Thanks for the CLI polish. The refreshed run UI and startup preflight check should make the first-run experience smoother without touching the protected agent core. Merged.

PR #102, opened 2026-05-13
feat(cli): add memory introspection commands

Teerapat-Vatpitak:
## Summary

Adds `vibe-trading memory list/show/search/forget` so users can inspect
and manage persistent memory without opening `~/.vibe-trading/memory/`
by hand. Surfaces real bug signal: two recent memory fixes (#87
snake_case, #95 CJK slugs) were hard to triage without observability
into what is actually saved and what the recall scorer returns.

Diff vs `main`: **+516 / -5** across `cli.py`, `src/memory/persistent.py`,
and two test files. 81 tests pass. `vibe-trading --version` measured
at 1.0s (no regression on cold startup).

## CLI surface

```
vibe-trading memory list   [--type user|feedback|project|reference]
vibe-trading memory show   <name>
vibe-trading memory search <query> [--limit N]
vibe-trading memory forget <name> [-y|--yes]
```

`<name>` resolves by exact title first, then by on-disk filename stem
(`{type}_{slug}` or the bare `slug`), so users can paste either form
from the index.

Exit codes mirror existing commands:

| scenario              | code                   |
|-----------------------|------------------------|
| success               | `EXIT_SUCCESS` (0)     |
| user declines confirm | `EXIT_SUCCESS` (0)     |
| not found / EOF stdin | `EXIT_USAGE_ERROR` (2) |
| remove failed         | `EXIT_RUN_FAILED` (1)  |

## Commits

The branch is split into four commits so each step is reviewable on
its own:

| Commit | Why |
|--------|-----|
| `892cd16` feat(cli): add memory introspection commands | the actual feature |
| `f8de9d0` perf(cli): avoid loading agent runtime on cli startup | hoisting `PersistentMemory` made `--version` 7.6s because `src.agent.frontmatter` transitively loads `AgentLoop` / `SkillsLoader`. Reverted to lazy imports inside each `cmd_memory_*` and a local `_MEMORY_TYPES` tuple in `cli.py` so non-memory commands stay fast |
| `fe4fca6` style(cli): revert unintended formatter spillover on unrelated functions | a local pre-tool-edit formatter reformatted ~30 unrelated functions in `cli.py`. This commit restores them to upstream form so the diff only shows memory-feature lines |
| `672e2ba` fix(memory): coerce frontmatter values to str in _scan_entries | bug fix discovered during heavy E2E — see below |

### New (no impact on existing callers)

- **`agent/cli.py`** — `memory` subparser; `cmd_memory_list/show/search/forget`;
  `_MEMORY_TYPES` tuple; `_MEMORY_TYPE_STYLES` (with module-load assert);
  `rich.markup.escape` on every user-supplied field; `Confirm.ask`
  wrapped against `EOFError` for non-interactive deletes
- **`agent/src/memory/persistent.py`** — `MEMORY_TYPES` constant; module
  logger; public `list_entries()` / `find(name)` / `remove_entry(entry)`.
  Existing `remove(title)` is unchanged so `remember_tool` is unaffected
- **`agent/tests/test_cli_memory.py`** — 29 new tests (command logic,
  name resolution incl. true stem fallback, parser wiring, Rich markup
  escape, empty body rendering, EOF path, remove-failure path,
  bracketed-description regression)

### One existing-code change: `_scan_entries` coercion

`MemoryEntry` annotates `title`, `description`, and `memory_type` as
`str`, but `_scan_entries` was passing the raw `parse_frontmatter`
output through untouched. The parser returns lists for `[a, b]`
syntax and bools for `true`/`false`, so any markdown file with e.g.
`description: [red]inject[/red]` leaked a list into the dataclass.
The previous CLI never called string ops on these fields, so the
mismatch was latent; the new CLI hits `rich.markup.escape(...)` and
crashes with `TypeError`.

Add a small `_coerce_str` helper and route the three frontmatter
fields through it so the dataclass contract is truthful at the storage
boundary. Existing consumers (`find_relevant`'s tokenization,
`remember_tool`'s rendering, `remove(title)`'s title comparison) all
silently improve on the edge case and behave identically on the happy
path. `tests/test_remember_tool.py` (16 tests) is unchanged and still
green.

## Pre-existing limitations surfaced by the new CLI (out of scope)

Heavy E2E testing with Thai content surfaced two pre-existing bugs in
`persistent.py` that this PR intentionally does **not** fix:

- `_tokenize` accepts only CJK Unified Ideographs (`一-鿿`);
  Thai / Arabic / Hebrew / Cyrillic queries always return zero
  matches because the tokenizer returns an empty set on those scripts
- The slug regex in `add()` strips Thai (and other non-CJK scripts)
  to `_`, producing filenames like `feedback______________.md` for
  Thai titles

These are follow-ups of #95 and worth a separate issue/PR. The new
CLI happens to make them visible (`memory show` renders the Thai
title fine, but the on-disk slug is mangled).

## Test Plan

- [x] `pytest agent/tests/test_cli_memory.py agent/tests/test_persistent_memory.py agent/tests/test_remember_tool.py agent/tests/test_cli_init.py` → 81 pass
- [x] `ruff check agent/cli.py agent/src/memory/persistent.py agent/tests/test_cli_memory.py agent/tests/test_persistent_memory.py` → no new errors (14 pre-existing E402/F541/F401 untouched)
- [x] `time vibe-trading --version` → 1.0s
- [x] File-level E2E on a populated `~/.vibe-trading/memory/`: list (with and without `--type` filter), show by title / by full slug / by bare slug, search (Latin and CJK queries), forget with `-y`, with `n` decline, with closed stdin (EOF)
- [x] End-to-end with a real LLM: `vibe-trading run` → agent invokes `remember` tool → new entry visible in `memory list`, retrievable via `memory show`, matchable by `memory search`, removable via `memory forget -y`

## Checklist

- [x] No changes to protected modules (`src/agent/`, `src/session/`, `src/providers/`)
- [x] No hardcoded values
- [x] Follows `CONTRIBUTING.md`
- [x] Tests added
- [x] No on-disk format changes; existing `remember_tool` workflows are byte-for-byte compatible

### DIFF
--- a/agent/cli.py
+++ b/agent/cli.py
@@ -27,7 +27,6 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 import warnings
-
 warnings.filterwarnings("ignore", message=".*Importing verbose from langchain.*")
 warnings.filterwarnings("ignore", category=DeprecationWarning, module="langchain")
 
@@ -143,7 +142,9 @@ def _print_status_bar(stats: _SessionStats) -> None:
         stats: Session statistics.
     """
     parts = _build_status_parts(stats)
-    bar = "[dim] │ [/dim]".join(f"[bold]{parts[0]}[/bold]" if i == 0 else p for i, p in enumerate(parts))
+    bar = "[dim] │ [/dim]".join(
+        f"[bold]{parts[0]}[/bold]" if i == 0 else p for i, p in enumerate(parts)
+    )
     console.print(bar)
 
 
@@ -254,7 +255,6 @@ def _read_prompt_source(
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
-
 
 def _read_json(path: Path) -> dict:
     """Safely read JSON."""
@@ -595,19 +595,18 @@ class _RunDashboard:
 # Agent execution core
 # ---------------------------------------------------------------------------
 
-
 def _format_tool_call_args(tool: str, args: Dict[str, str]) -> str:
     """Smart-format tool argument summary."""
     if tool == "load_skill":
         return f'("{args.get("name", "")}")'
     if tool in ("write_file", "read_file", "edit_file"):
-        return f" {args.get('path', args.get('file_path', ''))}"
+        return f' {args.get("path", args.get("file_path", ""))}'
     if tool in ("bash", "background_run"):
         cmd = args.get("command", "")[:80]
-        return f" [yellow]{cmd}[/yellow]"
+        return f' [yellow]{cmd}[/yellow]'
     if tool == "check_background":
         tid = args.get("task_id", "")
-        return f" {tid}" if tid else ""
+        return f' {tid}' if tid else ""
     if tool in ("backtest", "compact"):
         return ""
     for v in args.values():
@@ -627,7 +626,7 @@ def _format_tool_result_preview(tool: str, status: str, preview: str) -> str:
         if sharpe:
             parts.append(f"sharpe={sharpe.group(1)}")
         if ret:
-            parts.append(f"return={float(ret.group(1)) * 100:.1f}%")
+            parts.append(f"return={float(ret.group(1))*100:.1f}%")
         return ", ".join(parts) if parts else ""
     if tool == "render_shadow_report":
         url = re.search(r'"report_url":\s*"([^"]+)"', preview)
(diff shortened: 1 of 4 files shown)


warren618:
Thanks @Teerapat-Vatpitak — this is a clean and useful addition. The memory CLI makes the persistent-memory layer much easier to inspect and debug, and the extra frontmatter coercion fix is a nice catch. Merged!

PR #57, opened 2026-04-28
feat: add model and data source settings UI

GTC2080:
## Summary

- Add a Settings page for configuring LLM provider, model, endpoint, API key, generation limits, and reasoning effort.
- Add data source settings for Tushare token management and BaoStock availability/status display.
- Add backend settings endpoints, Tushare placeholder-token handling, and API regression tests.

## Why

Users currently need to edit `agent/.env` manually to switch LLM models or configure market data credentials. This PR exposes those common local settings in the Web UI while keeping secrets masked in API responses.

No local runtime-directory migration is included in this PR.

## Changes

- Added `GET/PUT /settings/llm` for project-local LLM settings persisted to `agent/.env`.
- Added `GET/PUT /settings/data-sources` for Tushare token settings and BaoStock status reporting.
- Added a frontend `/settings` route and sidebar entry.
- Added provider defaults for common OpenAI-compatible providers and Ollama.
- Added `max` as an allowed reasoning-effort option.
- Treat `your-tushare-token` as an unconfigured Tushare placeholder.
- Added `agent/tests/test_settings_api.py` coverage for secret masking, env persistence, and placeholder handling.

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`)
- [x] New tests added (if applicable)
- [x] Tested manually (describe below)

Manual/targeted checks run on Windows:

- `python -S -m py_compile agent\api_server.py agent\backtest\loaders\tushare.py agent\tests\test_settings_api.py`
- `E:\测试\vibe_trading\.venv\Scripts\python.exe -m pytest agent\tests\test_settings_api.py -q` -> `4 passed`
- `npm ci --cache E:\测试\vibe_trading_pr_settings\.npm-cache`
- `npm run build` -> passed, with the existing Vite large-chunk warning
- Full `pytest --ignore=agent/tests/e2e_backtest --tb=short -q` on Windows reached `669 passed, 1 skipped`, with 3 existing Windows path assertion failures in `agent/tests/test_loop_helpers.py::TestNormalizeToolRunDir` where `/tmp/...` is normalized to `E:\tmp\...`.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion
- [x] No hardcoded values (API keys, file paths, magic numbers)
- [x] Code follows `CONTRIBUTING.md` guidelines
- [ ] Documentation updated (if user-facing change)

## Related issue

Closes #58

PR #105, opened 2026-05-14
feat(providers): add Anthropic Claude provider via langchain-anthropic

Tezigudo:
## Summary

Wires Anthropic Claude as a first-class LLM provider alongside the existing 13 (OpenRouter, OpenAI, OpenAI Codex, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama). Selected via `LANGCHAIN_PROVIDER=anthropic` (or the `claude` alias).

**Why a separate code path and not the OpenAI-compat shim?** Anthropic's OpenAI-compat endpoint does not surface extended thinking and has subtle tool-calling differences. The native `langchain-anthropic` adapter gives us first-class tool calls, `usage_metadata`, and extended thinking. This matches the precedent set by the `openai-codex` provider (also a separate code path because it's not really OpenAI).

## What it does

- Adds `langchain-anthropic>=0.3.0` to `requirements.txt` + `pyproject.toml`.
- `ChatAnthropicWithReasoning` subclasses `ChatAnthropic` and normalises:
  - **Content list-of-blocks** (emitted when extended thinking is enabled) → flat string. Thinking blocks are surfaced via `additional_kwargs["reasoning_content"]` (same key Moonshot/DeepSeek use, so the existing `ChatLLM._parse_response` parser works without changes).
  - **`stop_reason` → `finish_reason`** (`end_turn`/`tool_use`/`stop_sequence`/`max_tokens`/`pause_turn`/`refusal`). When `tool_calls` are present, the content list is preserved so LangChain's tool-call extraction continues to work.
- Hooks at `_generate` / `_agenerate` / `_stream` / `_astream` because `langchain-anthropic` 0.3.x emits the message via `_format_output` inside `_generate` — it does **not** call `_create_chat_result` (verified empirically before relying on the hook).
- `LANGCHAIN_REASONING_EFFORT=low|medium|high|max` opts in to extended thinking with `budget_tokens` 1024 / 4096 / 12288 / 24576, forces `temperature=1.0` (Anthropic requirement), and raises `max_tokens` above the budget.
- `ANTHROPIC_API_KEY` is never projected into `OPENAI_API_KEY` so a later provider switch within the same process cannot leak it.

## Test plan

**25 new tests in `agent/tests/test_anthropic_provider.py`, all green; 978 total in the suite.**

- [x] `llm_providers.json` contains the anthropic entry with required fields
- [x] `_sync_provider_env` does not leak `ANTHROPIC_API_KEY` into `OPENAI_API_KEY`
- [x] `claude` alias routes to the anthropic branch
- [x] `build_llm` wiring: api_key, base_url, max_tokens, callbacks, error paths
- [x] `LANGCHAIN_REASONING_EFFORT` → budget_tokens / temperature / max_tokens
- [x] Explicit `ANTHROPIC_MAX_TOKENS` overrides the thinking-budget default
- [x] Content-block normalisation (text, thinking, redacted_thinking)
- [x] `stop_reason` → `finish_reason` mapping incl. unknown pass-through
- [x] `tool_calls` present → content list preserved, `finish_reason="tool_calls"`
- [x] **End-to-end**: wrapper's `_generate` fires through real `ChatAnthropic._format_output` (HTTP boundary mocked) for both text-only and `tool_use` responses

```bash
python -m pytest agent/tests/test_anthropic_provider.py -v   # 25 passed
python -m pytest agent/tests/ -q                              # 978 passed
```

## Docs

- README + 4 localized READMEs updated to list Anthropic Claude in the supported-providers line
- `agent/.env.example` gains an Anthropic block with model / key / base url / max tokens / extended-thinking opt-in

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - **Docker** for Path A
 - OpenAI Codex can also be used with ChatGPT OAuth: set `LANGCHAIN_PROVIDER=openai-codex`, then run `vibe-trading provider login openai-codex`. This does not use `OPENAI_API_KEY`.
 
-> **Supported LLM providers:** OpenRouter, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). See `.env.example` for config.
+> **Supported LLM providers:** OpenRouter, OpenAI, OpenAI Codex (ChatGPT OAuth), Anthropic Claude, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). See `.env.example` for config.
 
 > **Tip:** All markets work without any API keys thanks to automatic fallback. yfinance (HK/US), OKX (crypto), and AKShare (A-shares, US, HK, futures, forex) are all free. Tushare token is optional — AKShare covers A-shares as a free fallback.
 
--- a/README_ar.md
+++ b/README_ar.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - **Docker** للمسار A
 - يمكن استخدام OpenAI Codex أيضاً عبر ChatGPT OAuth: اضبط `LANGCHAIN_PROVIDER=openai-codex`، ثم شغل `vibe-trading provider login openai-codex`. هذا لا يستخدم `OPENAI_API_KEY`.
 
-> **مزودو LLM المدعومون:** OpenRouter, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). راجع `.env.example` للإعداد.
+> **مزودو LLM المدعومون:** OpenRouter, OpenAI, OpenAI Codex (ChatGPT OAuth), Anthropic Claude, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). راجع `.env.example` للإعداد.
 
 > **نصيحة:** تعمل كل الأسواق دون مفاتيح API بفضل fallback التلقائي. yfinance (HK/US)، وOKX (crypto)، وAKShare (A-shares, US, HK, futures, forex) مجانية. رمز Tushare اختياري، إذ يغطي AKShare أسهم A كبديل مجاني.
 
--- a/README_ja.md
+++ b/README_ja.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - Path A では **Docker**
 - OpenAI Codex は ChatGPT OAuth でも利用できます。`LANGCHAIN_PROVIDER=openai-codex` を設定し、`vibe-trading provider login openai-codex` を実行してください。`OPENAI_API_KEY` は使いません。
 
-> **Supported LLM providers:** OpenRouter、OpenAI、DeepSeek、Gemini、Groq、DashScope/Qwen、Zhipu、Moonshot/Kimi、MiniMax、Xiaomi MIMO、Z.ai、Ollama（local）。設定は `.env.example` を参照してください。
+> **Supported LLM providers:** OpenRouter、OpenAI、OpenAI Codex（ChatGPT OAuth）、Anthropic Claude、DeepSeek、Gemini、Groq、DashScope/Qwen、Zhipu、Moonshot/Kimi、MiniMax、Xiaomi MIMO、Z.ai、Ollama（local）。設定は `.env.example` を参照してください。
 
 > **Tip:** 自動フォールバックにより、すべての市場は API key なしで利用できます。yfinance（HK/US）、OKX（crypto）、AKShare（A-shares、US、HK、futures、forex）はすべて無料です。Tushare token は任意で、A-shares は AKShare が無料 fallback としてカバーします。
 
--- a/README_ko.md
+++ b/README_ko.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - 경로 A용 **Docker**
 - OpenAI Codex도 ChatGPT OAuth로 사용할 수 있습니다. `LANGCHAIN_PROVIDER=openai-codex`를 설정한 뒤 `vibe-trading provider login openai-codex`를 실행하세요. 이 방식은 `OPENAI_API_KEY`를 사용하지 않습니다.
 
-> **지원 LLM provider:** OpenRouter, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama(local). 설정은 `.env.example`을 참고하세요.
+> **지원 LLM provider:** OpenRouter, OpenAI, OpenAI Codex(ChatGPT OAuth), Anthropic Claude, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama(local). 설정은 `.env.example`을 참고하세요.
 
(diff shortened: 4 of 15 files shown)


Tezigudo:
### Code review

Found 2 issues:

1. `_normalize_anthropic_message` preserves the content list when `tool_calls` are present, but `swarm/worker.py:411` calls `response.content.strip()` — which `AttributeError`s on a list. ChatAnthropic `_format_output` lifts `tool_use` blocks into `msg.tool_calls` AND keeps them in the content list, so this path triggers any time Claude calls a tool with text-only commentary (the default ReAct shape). Drop the `not has_tool_calls` guard and flatten unconditionally; tool calls already survive via `msg.tool_calls`, and on the next ReAct turn LangChain rebuilds the Anthropic content blocks from string content + `tool_calls`.

https://github.com/HKUDS/Vibe-Trading/blob/6814640/agent/src/providers/llm.py#L130-L150

2. `sk-ant-xxx` is added as the documented `.env.example` placeholder but is missing from `LLM_API_KEY_PLACEHOLDERS`. The Settings UI uses that set to decide whether the key is configured (`_is_configured_secret` reads it); without registering the new placeholder, a fresh Settings load reports Anthropic as configured and a `PUT /settings/llm` write-back would persist the literal `sk-ant-xxx`. This is the same gap flagged on #57 for other vendor placeholders. Add `"sk-ant-xxx"` to the set.

https://github.com/HKUDS/Vibe-Trading/blob/6814640/agent/api_server.py#L508-L512

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

warren618:
Thanks for the work here. Native Anthropic support is useful, but because this touches the provider core path, I’d like to see real smoke evidence before merging.

Default CI should stay mock-only and should not require Anthropic credentials. But please add or document an opt-in smoke path that runs only when `ANTHROPIC_API_KEY` is present, and attach the output from a real run.

Before merge, I’d like evidence for:

1. a basic Claude response through `LANGCHAIN_PROVIDER=anthropic`;
2. one agent-loop tool call using Claude;
3. extended thinking enabled, or a clear guard/documentation if thinking + tool use is not supported yet;
4. verified model IDs in `.env.example`, especially Haiku 4.5.

Once CI is green and the real smoke evidence is attached, I’m happy to review again.

Tezigudo:
Thanks for the review — addressed all four asks in `597cdcf`.

### 1. Opt-in smoke harness (default CI still mock-only)

New folder `agent/tests/integration/` — every smoke is gated on `ANTHROPIC_API_KEY` via `pytest.mark.skipif`, so CI without credentials shows `3 skipped` rather than failing or requiring secrets.

- `agent/tests/integration/test_anthropic_smoke.py` — three live tests, one per contract you listed.
- `agent/tests/integration/README.md` — how to run + estimated cost (<\$0.05 per full run on Sonnet 4.6).

https://github.com/HKUDS/Vibe-Trading/blob/597cdcfb5800b6b62c85c29b5c78608bf4cb4b2b/agent/tests/integration/test_anthropic_smoke.py#L1-L40

Verified locally with no key set: `0 passed, 0 failed, 3 skipped`. Default suite is still 981 passing, mock-only.

### 2. Live smoke evidence

The smoke harness covers all three contracts you asked for. I do not have the upstream `ANTHROPIC_API_KEY` available in this environment to attach captured output to this comment, so I'm flagging that explicitly: I built the runnable harness and will paste real-run output as soon as I can run it against a key. If you'd prefer to run it yourself before merging, the one-liner is:

```bash
export ANTHROPIC_API_KEY=sk-ant-...
python -m pytest agent/tests/integration/ -v --no-header -s
```

The `-s` keeps the `print(...)` lines in each smoke so you get `usage`, `finish_reason`, and a content snippet per test — grep-friendly for attaching here.

### 3. Extended thinking + tool use position

Confirmed via [docs.anthropic.com extended-thinking docs](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking):

- **Single-turn thinking + tool use works fully** on Sonnet 4.6 + Haiku 4.5. The wrapper flattens content to a string, surfaces thinking via `additional_kwargs[\"reasoning_content\"]`, and `msg.tool_calls` survives. Smoke #3 + the existing mock test `test_invoke_with_tool_use_flattens_content_and_preserves_tool_calls` cover this end-to-end.
- **Multi-turn thinking across tool hops has a documented gap**: Anthropic requires the signed thinking block to be passed back unchanged on subsequent turns to maintain reasoning continuity. The wrapper flattens content to a string, so the thinking block is not re-sent on the next ReAct turn — the model re-reasons from scratch each tool hop. Functional but degraded.

This is now called out in `.env.example`:

https://github.com/HKUDS/Vibe-Trading/blob/597cdcfb5800b6b62c85c29b5c78608bf4cb4b2b/agent/.env.example#L102-L106

The competing alternative — keeping `content` as a list to preserve thinking blocks — was the original PR shape and the bug `/code-review` caught: `swarm/worker.py:411` calls `response.content.strip()` which crashes on a list. So this is a real trade-off, not an oversight; flagging it as a known limitation rather than silently shipping a broken multi-turn path.

### 4. Verified model IDs

Re-verified from [docs.anthropic.com Models overview](https://docs.anthropic.com/en/docs/about-claude/models/overview) (2026-05-14):

| Model | Canonical API ID | Alias | Extended thinking |
|---|---|---|---|
| Opus 4.7 | `claude-opus-4-7` | same | **No** (adaptive thinking only) |
| Sonnet 4.6 | `claude-sonnet-4-6` | same | Yes |
| Haiku 4.5 | `claude-haiku-4-5-20251001` | `claude-haiku-4-5` | Yes |

For Haiku 4.5 specifically, `claude-haiku-4-5` is officially documented as an alias and resolves to the date-stamped pinned snapshot. The `.env.example` block lists all three with these notes inline.

Because Opus 4.7 doesn't support extended thinking, setting `LANGCHAIN_REASONING_EFFORT` with `claude-opus-4-7` would silently produce wrong behavior — the wrapper now raises a clear `RuntimeError` at `build_llm()` time:

https://github.com/HKUDS/Vibe-Trading/blob/597cdcfb5800b6b62c85c29b5c78608bf4cb4b2b/agent/src/providers/llm.py#L370-L387

Two new mock tests cover the guard and the unguarded Opus-4.7 path (`test_extended_thinking_blocked_for_opus_4_7`, `test_opus_4_7_without_reasoning_effort_works`).

### Summary

| Ask | Status |
|---|---|
| Default CI stays mock-only | yes, suite is 981 passed without any Anthropic key |
| Opt-in smoke path with `ANTHROPIC_API_KEY` gate | new `agent/tests/integration/` folder |
| Real-run output attached | harness is ready; awaiting access to attach output (see above) |
| Basic Claude response evidence | smoke 1 in harness |
| Agent-loop tool call evidence | smoke 2 in harness |
| Thinking + tool use | works single-turn; multi-turn caveat documented in `.env.example` |
| Verified model IDs | all three confirmed against docs; Haiku alias verified |

Happy to iterate if you'd like a different shape for the multi-turn thinking case — e.g. an `ANTHROPIC_PRESERVE_THINKING_BLOCKS=1` opt-in that keeps content as a list and patches `swarm/worker.py` to be list-tolerant. That felt out of scope for this PR but I'm fine breaking it into a follow-up.

warren618:
Thanks for the iteration on the four asks — the Opus-4.7 guard and the model-ID verification are clean improvements, and the wrapper itself is nicely shaped.

After living with this version for a bit, I'm leaning toward closing it. The reason is mainly a scope call, plus a verifiability gap that's hard to bridge from where things sit today.

On the verifiability side, the 25 mock tests cover the wrapper's internal contract well, but there isn't a live smoke output against the real `api.anthropic.com` to anchor that against — for a PR that lands +189 lines into `agent/src/providers/llm.py` (which is core-path), I'm not personally comfortable merging without that. This isn't a knock on the PR; it's more of a baseline for changes into the dispatch path.

On the scope side, holding a 14th provider up against where the project sits today:

- The marginal value over the existing OpenRouter path is modest. `LANGCHAIN_PROVIDER=openrouter` + `anthropic/claude-sonnet-4-6` already gives users first-class Claude access — extended thinking, native tool calls — with no new dependency. The unique surface this PR adds is "talk to `api.anthropic.com` directly instead of through OpenRouter," which feels more like a preference than a capability gap for most users.
- A `ChatAnthropic` subclass introduces a long-lived maintenance edge. LangChain chat-model subclasses tend to drop custom fields at `_generate` / `_agenerate` / `_stream` / `_astream` across version bumps; we've already paid that patch cycle once, and each new reasoning-class provider adds another set of overrides to keep alive through future `langchain-anthropic` releases.

Taken together, adding a parallel channel next to a path we already serve doesn't quite clear the bar at this point in time.

None of this is a critique of the craft — the test structure is solid and the docs are careful. It just hasn't landed on the "yes, take this in" side for me.

If you later hit a class of Anthropic usage that OpenRouter genuinely can't deliver (a specific header, a feature that doesn't round-trip, a concrete billing/quota concern), I'd be very happy to look at a narrower PR scoped to that gap.

PR #113, opened 2026-05-14
feat(providers): add Claude Code (Pro/Max subscription) provider via claude-agent-sdk

Tezigudo:
## Summary

Adds a new `LANGCHAIN_PROVIDER=claude-code` provider that routes LLM calls through the user's **Claude Pro/Max subscription** instead of a paid Anthropic API key. Uses the official `claude-agent-sdk` Python package, which delegates auth to `claude login`. Independent of #105 — sits alongside the existing 13 providers as a fresh code path.

**Why a separate provider?** Anthropic's Feb-2026 ToS update prohibits direct OAuth-token-as-API-bearer reuse, and `claude-agent-sdk` is the supported path for programmatic Claude Max usage. Auth, billing, and the transport layer are all different from the regular Anthropic API path — analogous to the existing `openai-codex` precedent in this repo (also OAuth, also a separate adapter).

## v1 scope (deliberate)

- **Text completion only.** `bind_tools()` raises `NotImplementedError` with a clear pointer.
- **Tool calling deferred.** Claude Agent SDK auto-executes tools inside its own agent loop, which doesn't compose with Vibe-Trading's ReAct loop. Inverting that to surface tool calls as `response.tool_calls` needs the SDK's PreToolUse hook + a custom MCP server — non-trivial, follow-up PR.
- **Recommendation in docs:** Workflows that need backtests/skills/etc. via tool calls should stay on `LANGCHAIN_PROVIDER=anthropic` or `=openrouter`. The `claude-code` provider is useful today for swarm reports, doc summarization, single-turn chat — anywhere tools aren't required.

## Isolation guarantees

The SDK is built to *be* Claude Code, not to be a thin LLM wrapper. The adapter clamps every escape hatch:

- `_sync_provider_env` short-circuits for `claude-code`/`claude_code` **before** touching `OPENAI_*`, so the user's standard OpenAI key can't leak into a Claude Code call path on a later provider switch within the same process (same pattern as `openai-codex`).
- `ClaudeCodeLLM._build_options` sets `tools=[]`, `allowed_tools=[]`, `mcp_servers={}`, `strict_mcp_config=True`, `permission_mode=\"dontAsk\"`, `max_turns=1`, and isolates `cwd` (defaults to a fresh temp dir). Together this prevents the user's Claude Code plugins, MCP servers, and project-level `CLAUDE.md` from bleeding into Vibe-Trading agent responses.
- The system message from the OpenAI-style message list is hoisted into the SDK's `system_prompt` so the model sees exactly the prompt Vibe-Trading constructed.

## Response-shape contract

The adapter returns a `ClaudeCodeMessage` shaped exactly like what `ChatLLM._parse_response` consumes:

- `content` flattened to a string (concatenated `TextBlock` content).
- Thinking surfaces via `additional_kwargs[\"reasoning_content\"]` (same key Moonshot/DeepSeek use — parser works unchanged).
- `stop_reason` mapped to OpenAI-style `finish_reason`: `end_turn`→`stop`, `tool_use`→`tool_calls`, `stop_sequence`→`stop`, `max_tokens`→`length`, `pause_turn`→`stop`, `refusal`→`content_filter`.
- `usage_metadata` normalized to LangChain's `{input_tokens, output_tokens, total_tokens}` shape.

## Optional dependency

`claude-agent-sdk` is ~58 MB. To keep the default install lightweight, it lives in a new `claude-code` extra:

```bash
pip install 'vibe-trading-ai[claude-code]'  # or: pip install claude-agent-sdk
claude login                                  # then this
export LANGCHAIN_PROVIDER=claude-code
```

`agent/.env.example` documents the model knob (`LANGCHAIN_MODEL_NAME` is optional — empty defers to the Claude Code CLI default; `claude-opus-4-7` / `claude-sonnet-4-6` / `claude-haiku-4-5` are valid).

## Test plan

**20 new mock tests + 1 opt-in live smoke. Full default suite: 973 passed.**

- [x] `llm_providers.json` claude-code entry shape (subscription auth + no base URL)
- [x] `_sync_provider_env` does not touch `OPENAI_*` (3 cases incl. preserving pre-existing OpenAI key)
- [x] `build_llm` dispatch + missing-SDK error path + empty-`LANGCHAIN_MODEL_NAME` allowed for claude-code (still required for every other provider)
- [x] `invoke()`: flatten contract, reasoning_content extraction, all 4 main stop_reason mappings, empty-messages short-circuit
- [x] Options isolation: `tools=[]`, `strict_mcp_config=True`, `max_turns=1`, `permission_mode=dontAsk`, system_prompt hoist, cwd defaults + override
- [x] `bind_tools` rejects non-empty tools; no-ops on `[]`
- [x] `stream()` yields one aggregated message (v1 simplification — real token-level streaming is a follow-up)

**Live smoke** at `agent/tests/integration/test_claude_code_smoke.py`. Gated on `VIBE_TRADING_CLAUDE_CODE_SMOKE=1` + `claude-agent-sdk` importable, so default CI stays mock-only with no credentials.

Verified locally against the contributor's Claude Max plan:

```
[claude-code smoke] finish=stop usage={'input_tokens': 6, 'output_tokens': 16, 'total_tokens': 22} content[:140]='2 + 2 equals 4.'
```

```bash
# Default CI suite (no API keys, no claude login):
python -m pytest agent/tests/ --ignore=agent/tests/integration  # 973 passed

# Live smoke (requires claude login + the [claude-code] extra):
pip install 'vibe-trading-ai[claude-code]'
claude login
VIBE_TRADING_CLAUDE_CODE_SMOKE=1 python -m pytest agent/tests/integration/test_claude_code_smoke.py -v -s
```

## Docs

- README + 4 localized READMEs updated to list Claude Code in the supported-providers line.
- `.env.example` block with prerequisites + the v1 tool-calling limitation called out explicitly.
- `pyproject.toml` `claude-code` extra pinning `claude-agent-sdk>=0.1.81`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - **Docker** for Path A
 - OpenAI Codex can also be used with ChatGPT OAuth: set `LANGCHAIN_PROVIDER=openai-codex`, then run `vibe-trading provider login openai-codex`. This does not use `OPENAI_API_KEY`.
 
-> **Supported LLM providers:** OpenRouter, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). See `.env.example` for config.
+> **Supported LLM providers:** OpenRouter, OpenAI, OpenAI Codex (ChatGPT OAuth), Claude Code (Pro/Max subscription), DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). See `.env.example` for config.
 
 > **Tip:** All markets work without any API keys thanks to automatic fallback. yfinance (HK/US), OKX (crypto), and AKShare (A-shares, US, HK, futures, forex) are all free. Tushare token is optional — AKShare covers A-shares as a free fallback.
 
--- a/README_ar.md
+++ b/README_ar.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - **Docker** للمسار A
 - يمكن استخدام OpenAI Codex أيضاً عبر ChatGPT OAuth: اضبط `LANGCHAIN_PROVIDER=openai-codex`، ثم شغل `vibe-trading provider login openai-codex`. هذا لا يستخدم `OPENAI_API_KEY`.
 
-> **مزودو LLM المدعومون:** OpenRouter, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). راجع `.env.example` للإعداد.
+> **مزودو LLM المدعومون:** OpenRouter, OpenAI, OpenAI Codex (ChatGPT OAuth), Claude Code (اشتراك Pro/Max), DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). راجع `.env.example` للإعداد.
 
 > **نصيحة:** تعمل كل الأسواق دون مفاتيح API بفضل fallback التلقائي. yfinance (HK/US)، وOKX (crypto)، وAKShare (A-shares, US, HK, futures, forex) مجانية. رمز Tushare اختياري، إذ يغطي AKShare أسهم A كبديل مجاني.
 
--- a/README_ja.md
+++ b/README_ja.md
@@ -321,7 +321,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - Path A では **Docker**
 - OpenAI Codex は ChatGPT OAuth でも利用できます。`LANGCHAIN_PROVIDER=openai-codex` を設定し、`vibe-trading provider login openai-codex` を実行してください。`OPENAI_API_KEY` は使いません。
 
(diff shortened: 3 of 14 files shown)


warren618:
Thanks for putting this together — the isolation work is thoughtful, and the full set of clamps (`tools=[]` / `mcp_servers={}` / `strict_mcp_config=True` / `permission_mode=dontAsk` / `max_turns=1` / isolated `cwd`) is exactly the homework I'd expect for this surface.

After turning this over a few times against where the project is right now, I'm leaning toward closing this for fit rather than implementation. A few observations behind that:

`claude-agent-sdk` sits closer to being an agent SDK (a Node CLI wrapping its own agent loop) than an LLM SDK. The set of clamps you had to apply to compress it down into a single LLM call is, I think, a reflection of that mismatch rather than something the PR itself can resolve — the tension comes from the SDK's current shape.

The downstream consequence is what `bind_tools()` already calls out: in its v1 form, this provider doesn't participate in the ReAct loop, and ReAct + tools + swarm is the most frequently exercised path in Vibe-Trading. On the verifiability side, that means the `2 + 2 = 4` style single-turn smoke is the entire surface this provider can exercise — the paths that would actually stress provider behaviour in this project (ReAct loop, tool batching, swarm DAG, context compression) are excluded by design, so there isn't a natural way to validate integration health.

Combined with a 58 MB Node-CLI dependency and a 14th provider entry, the acceptance cost feels a bit high for the project at this point in time.

For Pro/Max subscribers who want programmatic Claude access today, the smoother paths are still (a) an Anthropic API key, or (b) OpenRouter with `anthropic/claude-*` models — both cover the agent-loop case this PR explicitly steps back from.

If `claude-agent-sdk` later grows a more LLM-only shape (e.g. tool calls surfaced first-class through a `PreToolUse` hook), I'd be very happy to look at a fresh version. Either way the care that went into this is visible.

PR #119, opened 2026-05-16
fix(swarm): surface errors, output contract, Windows-safe store, redact paths

Teerapat-Vatpitak:
## Summary
- Surface swarm task/run errors at every MCP read boundary (incl. the event stream) instead of folding timeout/empty into a `completed` status.
- Enforce a worker output contract (reject plan-only / unparsed-tool-markup / fabricated deliverables).
- Make `SwarmStore` atomic write/read Windows-safe (WinError-scoped bounded retry; POSIX behavior unchanged).
- Add `src/tools/redaction.py` (anchored, idempotent, no regex) and redact internal absolute paths from user-facing errors and events (CWE-209/497).

## Why
A run could report `completed` with no grounded deliverable; raised errors and `task_failed`/`run_error` event payloads leaked the OS user and venv/install topology; on Windows a concurrent `os.replace` race crashed the swarm.

## Changes
- `swarm/{serialization,runtime,store,worker,models}`, `tools/{redaction,read_file,write_file,edit_file}`
- redaction wired at swarm read boundaries, file-tool errors, and event payloads

## Test Plan
- [x] `pytest --ignore=agent/tests/e2e_backtest -q` -> 1080 passed, 1 skipped
- [x] New regression tests; fail-before/pass-after verified per fix

## Checklist
- [x] No changes to protected areas (`src/agent|session|providers`)
- [x] No hardcoded values
- [x] Follows CONTRIBUTING.md

Note: follow-up will harden `_classify_deliverable`/`_is_error_result` (substring -> JSON parse); out of scope here.


warren618:

warren618:
Merged after a quick origin/main merge into the branch (purely style conflicts in `mcp_server.py` from #120's reformatting — your `serialize_task` version preserved end-to-end).

Three things worth calling out:

1. The `serialization.py` single-source extraction is a textbook fix for P04 — three read boundaries × hand-maintained field allowlists × one omitted `error` field is exactly the bug class that hides forever. Centralizing the projection means a future field can't be silently dropped by one of the three.

2. `_classify_deliverable`'s Hybrid policy is the right shape — content sanity for everyone, tool-evidence only for data agents — and the explicit synthesis/editor false-reject guard test (`test_synthesis_agent_prose_is_accepted`) shows the policy was designed around the failure mode, not just around the bug. The substring heuristics for plan-only / fabrication markers are pragmatic; I'll harden `_classify_deliverable` / `_is_error_result` toward JSON parse as you flagged (out of scope here was the right call).

3. `redaction.py` is correct in every axis: anchored to known internal roots only (no over-redaction of caller paths — the `test_no_over_redaction` matrix proves it), idempotent (sentinel is not itself a root), no regex (zero ReDoS surface), None-safe.

The Windows-vs-POSIX split in `_replace_with_retry` (`getattr(exc, "winerror", None) in {5, 32}` → POSIX `OSError` never matches → loop runs exactly once off-Windows) is the right shape too. Nice work overall.

PR #120, opened 2026-05-16
fix(mcp): unresolved symbols, finite options validation, row cap

Teerapat-Vatpitak:
## Summary
- Surface every requested market-data symbol that resolved to nothing under a reserved `_unresolved` key; log the swallowed loader exception instead of vanishing silently.
- Reject NaN/Inf/non-finite options inputs (incl. `risk_free_rate`) at the validation boundary before pricing; stop defaulting a missing Greek to a confident `0.0`.
- Cap `get_market_data` output via an even-stride downsample (last bar pinned) and reject negative `max_rows`; add a `DEFAULT_MAX_ROWS` constant.

## Why
Silent data-drop, confident Greeks on NaN volatility, and mid-series amputation are numeric-integrity hazards for an agent acting on the output.

## Changes
- `mcp_server.py` (`_unresolved` logging, `DEFAULT_MAX_ROWS`, even-stride `_cap_rows`)
- `tools/options_pricing_tool.py` (finite validation incl. `r`; no authoritative `0.0`)

## Test Plan
- [x] `pytest --ignore=agent/tests/e2e_backtest -q` -> 1069 passed, 1 skipped
- [x] New tests; fail-before/pass-after verified

## Checklist
- [x] No hardcoded values (`DEFAULT_MAX_ROWS` constant)
- [x] Follows CONTRIBUTING.md

Note (by design): `_unresolved` still groups a transient loader outage with bad tickers in the response; observability is added now, a per-symbol status contract is a larger follow-up.

### DIFF
--- a/agent/mcp_server.py
+++ b/agent/mcp_server.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import json
 import logging
+import math
 import os
 import re
 import sys
@@ -416,6 +417,8 @@ def run_swarm(preset_name: str, variables: dict[str, str]) -> str:
 # Market data tool
 # ---------------------------------------------------------------------------
 
+DEFAULT_MAX_ROWS = 250
+
 _SOURCE_PATTERNS = [
     (re.compile(r"^\d{6}\.(SZ|SH|BJ)$", re.I), "tushare"),
     (re.compile(r"^[A-Z]+\.US$", re.I), "yfinance"),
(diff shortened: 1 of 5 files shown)


warren618:

warren618:
Merged. Three independent fixes, each disciplined to stay additive:
- `_unresolved` omitted when all codes resolve (happy path byte-identical)
- `max_rows=N` returns plain list when under cap (byte-identical for small queries)
- `T==0` switched from confident `status:"ok"` with zero Greeks to `status:"degenerate"` with the warning surfaced — that's the kind of numeric-integrity fix the agent layer compounds across many calls

The `allow_nan=False` + try/except envelope around `json.dumps` is the right belt-and-suspenders for non-finite leaks from extreme inputs.

PR #121, opened 2026-05-16
fix(loaders): bound ccxt + okx fetch (timeout, retry, budget)

Teerapat-Vatpitak:
## Summary
Bound `ccxt` and `okx` data fetches: explicit per-call timeout, bounded transient-error retry with backoff, and a hard wall-clock budget that raises a clear `TimeoutError` instead of hanging unbounded on a transient disconnect.

## Why
A transient network disconnect could hang a fetch 10+ minutes (no timeout, unbounded retry), stalling the agent.

## Changes
- `backtest/loaders/ccxt_loader.py`, `backtest/loaders/okx.py`

## Test Plan
- [x] `pytest --ignore=agent/tests/e2e_backtest -q` green; transport mocked (no network)
- [x] Asserts bounded retry count + wall-clock budget enforced

## Checklist
- [x] No hardcoded values (env-overridable named constants)
- [x] Follows CONTRIBUTING.md

Follow-up: extract the shared retry/budget into `loaders/base.py`; treat a non-JSON okx body as transient.

### DIFF
--- a/agent/backtest/loaders/ccxt_loader.py
+++ b/agent/backtest/loaders/ccxt_loader.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import logging
 import os
+import time
 from typing import Dict, List, Optional
 
 import pandas as pd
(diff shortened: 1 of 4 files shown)


warren618:

warren618:
Merged. Bounded retry with the NetworkError-only filter (vs blanket except) is exactly the right call — happy path stays one call per page, non-transient errors keep their original type, wall-clock budget gives a clear ceiling. The five-test matrix (transient→success, persistent bounded, non-network not retried, happy path unchanged, budget enforced) catches every regression I'd worry about.

On the follow-up to extract the shared retry/budget into loaders/base.py you flagged — I'll take that on shortly so it doesn't fall through.

PR #122, opened 2026-05-16
fix(tools): read_url third-party disclosure + cache opt-out

Teerapat-Vatpitak:
## Summary
Disclose the `read_url` third-party dependency (r.jina.ai) in the skill doc, add a `no_cache` opt-out, and sanitize upstream errors so the vendor name and response body are not leaked to the caller (kept in server logs).

## Why
`read_url` silently routed URLs through a third party and could return stale cached snapshots with no disclosure or opt-out.

## Changes
- `src/skills/web-reader/SKILL.md`, `src/tools/web_reader_tool.py`

## Test Plan
- [x] `pytest --ignore=agent/tests/e2e_backtest -q` green; requests mocked
- [x] Asserts default path byte-identical + opt-out header behavior

## Checklist
- [x] No hardcoded values
- [x] Follows CONTRIBUTING.md

Follow-up: anchor the cached-marker detection to the header region; verify the bypass header against current Jina docs.

### DIFF
--- a/agent/src/skills/web-reader/SKILL.md
+++ b/agent/src/skills/web-reader/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: web-reader
-description: Read web pages, articles, and document links by converting URLs into Markdown text. Use the `read_url` tool directly, without bash.
+description: Read web pages, articles, and document links by converting URLs into Markdown text. Use the `read_url` tool directly, without bash. Sends the full URL to the third-party Jina Reader (r.jina.ai).
 category: tool
 ---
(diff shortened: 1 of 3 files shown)


warren618:
Merged — thanks.

One tweak on top before squash: rolled back the HTTP/exc error sanitization in `web_reader_tool.py` (kept your `logger.warning`). Vibe-Trading runs as a local stdio MCP, so the caller is the user's own agent — scrubbing the body / exc text just removes debug signal. Tests updated to match.

PR #126, opened 2026-05-17
[codex] Pass proxy environment to CCXT loader

ruok808:
## Summary

- Read conventional proxy environment variables when constructing the CCXT exchange client.
- Pass the resulting `proxies` config to CCXT so public exchange data can be fetched from restricted networks.
- Add regression coverage for `ALL_PROXY` and for the no-proxy case.

## Why

Requests-based loaders already inherit proxy environment variables, but CCXT does not reliably do so through this loader's current constructor path. In restricted network environments, Binance/OKX public market data can time out unless the proxy is passed explicitly to CCXT.

This change only uses environment variable names and never logs proxy values or credentials.

## Tests

```bash
/Users/barry/Desktop/alevinno/src/vibe-trading-research/.venv/bin/python -m pytest agent/tests/test_ccxt_loader_proxy.py agent/tests/test_ccxt_loader_bounded.py agent/tests/test_get_market_data_unresolved.py -q
```

Result: `13 passed`.

### DIFF
--- a/agent/backtest/loaders/ccxt_loader.py
+++ b/agent/backtest/loaders/ccxt_loader.py
@@ -37,6 +37,28 @@ _CCXT_TIMEOUT_MS = int(os.getenv("CCXT_TIMEOUT_MS", "15000"))
 _CCXT_FETCH_BUDGET_S = float(os.getenv("CCXT_FETCH_BUDGET_S", "60"))
 
 
+def _first_proxy_env(*names: str) -> str:
+    for name in names:
+        value = os.getenv(name, "").strip()
+        if value:
+            return value
+    return ""
+
+
+def _ccxt_proxy_config() -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for the focused fix. I reviewed the diff and the approach looks reasonable to me.

I also verified the declared regression suite locally:

```bash
python -m pytest agent/tests/test_ccxt_loader_proxy.py agent/tests/test_ccxt_loader_bounded.py agent/tests/test_get_market_data_unresolved.py -q
```

Result: `13 passed`.

One small process note: this PR is still marked as Draft, so we cannot treat it as ready to merge yet. When you are ready, please mark it as ready for review. After that I think this should be straightforward to merge.

Optional follow-up, not a blocker: it would be nice to add one more regression covering explicit `HTTP_PROXY` / `HTTPS_PROXY` taking precedence over `ALL_PROXY`, since the implementation already supports that behavior.

PR #129, opened 2026-05-18
fix(providers): normalize remote Ollama base URLs without /v1

faizack:
Append /v1 automatically when `LANGCHAIN_PROVIDER=ollama` so remote hosts like `http://host:11434` work with ChatOpenAI without manual path suffixes.

## Summary

- Auto-append `/v1` to `OLLAMA_BASE_URL` when the path is missing, so Ollama’s OpenAI-compatible API is used correctly.
- Update Ollama defaults and `.env.example` to use host:port only (e.g. `http://localhost:11434`).
- Add unit tests for local and remote Ollama base URLs.

## Why

Bare Ollama URLs (e.g. `http://host:11434/`) cause ChatOpenAI requests to hit the wrong path and return `404 page not found`. Users should not need to remember the `/v1` suffix for remote Ollama servers.

## Changes

- `agent/src/providers/llm.py`: `_normalize_ollama_base_url()` + apply in `_sync_provider_env()` for `ollama`
- `agent/cli.py`, `agent/src/providers/llm_providers.json`, `agent/.env.example`: default URL without `/v1`
- `agent/tests/test_llm.py`, `agent/tests/test_cli_init.py`: coverage for URL normalization

## Test Plan

- [x] New tests added (`test_ollama_base_url_appends_v1`, updated init/ollama tests)
- [ ] Tested manually: `OLLAMA_BASE_URL=http://<remote-host>:11434` (no `/v1`), run `vibe-trading`, confirm chat works

## Checklist

- [x] Minimal change in `src/providers/llm.py` only (provider env mapping)
- [x] No hardcoded API keys or secrets
- [x] Documentation updated (`.env.example`, provider defaults)

### DIFF
--- a/agent/.env.example
+++ b/agent/.env.example
@@ -79,7 +79,7 @@ OPENROUTER_BASE_URL=https://openrouter.ai/api/v1
 # --- Ollama (local) ---
 # LANGCHAIN_PROVIDER=ollama
 # LANGCHAIN_MODEL_NAME=qwen2.5:32b
-# OLLAMA_BASE_URL=http://localhost:11434/v1
+# OLLAMA_BASE_URL=http://localhost:11434
 
 # LLM parameters
 LANGCHAIN_TEMPERATURE=0.0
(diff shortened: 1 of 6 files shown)


warren618:
Thanks for the clean fix and tests! Merged — this also unblocks anyone running Ollama behind a reverse proxy without thinking about the `/v1` suffix.

PR #136, opened 2026-05-23
Fix agent session history loading

fightZy:
## Background
Opening an existing Agent session can leave the conversation area stuck on loading skeletons in development. The session-loading effect increments its generation counter on every effect run, so a no-op rerun can mark the in-flight history request as stale before it clears the loading state.

## Changes
- Increment the session load generation only when actually switching sessions or resetting the current session.
- Preserve the existing cached-session and SSE setup behavior.


warren618:
@fightZy Thanks for the fix! Merged. I rebased the branch onto the latest main to resolve the Agent.tsx conflict from the newer goal workflow, kept your authorship on the commit, and verified the frontend build passes. Great catch on the stale generation/loading skeleton issue.

PR #146, opened 2026-05-29
Fix IME Enter submission handling

ArthurXi:
## Summary

- Fix IME Enter handling in the Agent chat input.
- Prevent accidental session submission while text composition is active or has just ended.

## Why

When using Chinese/Japanese/Korean IMEs, pressing Enter can be used to confirm composition. The chat input previously treated that Enter as a submit action, which could accidentally trigger a session before the user finished typing.

## Changes

- Track IME composition state with `onCompositionStart` and `onCompositionEnd`.
- Ignore Enter submit while composition is active.
- Add a short guard after composition ends to avoid browser event-order edge cases.

## Test Plan

Manual test:
- Verified that pressing Enter while using an IME to confirm text does not submit the Agent chat.
- Verified that normal Enter still submits after composition has completed.

## Checklist

### DIFF
--- a/frontend/src/pages/Agent.tsx
+++ b/frontend/src/pages/Agent.tsx
@@ -84,6 +84,8 @@ export function Agent() {
   const [searchParams, setSearchParams] = useSearchParams();
   const listRef = useRef<HTMLDivElement>(null);
   const inputRef = useRef<HTMLTextAreaElement>(null);
+  const isComposingRef = useRef(false);
+  const lastCompositionEndRef = useRef(0);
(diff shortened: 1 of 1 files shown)


warren618:
Merged — thanks @ArthurXi! 🙏

This is a real fix for CJK users: confirming a composition with Enter no longer fires an accidental submit. Verified the composition-state tracking (`isComposing` / `keyCode 229` + the post-composition guard) behaves correctly and normal Enter-to-submit still works; TypeScript typecheck is clean. Appreciate the clear write-up of the browser event-order edge case.

PR #148, opened 2026-05-30
fix(loop): graceful exit on max iterations with wrap-up nudge + forced text-only

toanalien:
## Summary

- **Problem:** `AgentLoop` could exhaust all 50 iterations when the LLM kept calling tools endlessly, resulting in `status="failed"` with no useful output returned to the user.
- **Fix:** Mirror the swarm worker's proven approach — inject a wrap-up nudge at 80% of the iteration budget and drop tool definitions on the last iteration to force a text response.
- **Test:** Added `test_force_text_only_on_last_iteration` validating that an LLM which always returns tool calls now produces `status="success"` instead of `status="failed"`.

### `agent/src/agent/loop.py`
1. Calculate `wrap_up_at = max(1, int(self.max_iterations * 0.8))` before the main loop
2. At 80% of iteration budget, inject a `[SYSTEM]` user message nudging the LLM to stop calling tools and produce a final answer
3. On the last iteration, pass `tools=None` to `stream_chat` — the LLM cannot request tool calls and must produce text output, guaranteeing a final answer

### `agent/tests/test_agent_loop_terminal_state.py`
- Added `_StubLLMAlwaysToolCalls` — LLM stub that returns `compact` tool calls until `tools=None` forces text
- Added `test_force_text_only_on_last_iteration` — verifies the forced text-only last iteration produces `status="success"` with content

## How it works

The swarm worker (`agent/src/swarm/worker.py`) already had both mechanisms (lines 363-426). This PR brings the same pattern to `AgentLoop`:

| Mechanism | Swarm Worker | AgentLoop (before) | AgentLoop (after) |
|-----------|-------------|--------------------|--------------------|
| Wrap-up nudge at 80% | Yes | No | Yes |
| Force text-only on last iter | Yes | No | Yes |

## Test plan

- [x] Existing 3 tests in `test_agent_loop_terminal_state.py` pass unchanged
- [x] New `test_force_text_only_on_last_iteration` passes
- [x] `ruff check` passes
- [x] `python -m py_compile` passes

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -365,6 +365,7 @@ class AgentLoop:
         final_content = ""
         goal_continuations = 0
         goal_last_progress: tuple[int, int] | None = None
+        wrap_up_at = max(1, int(self.max_iterations * 0.8))
 
         try:
             while iteration < self.max_iterations:
@@ -399,6 +400,19 @@ class AgentLoop:
 
                 logger.info(f"ReAct iteration {iteration}/{self.max_iterations}")
 
+                # Inject wrap-up nudge when approaching iteration limit
+                if iteration == wrap_up_at:
+                    remaining = self.max_iterations - iteration
+                    messages.append({
+                        "role": "user",
+                        "content": (
+                            f"[SYSTEM] You have {remaining} iterations remaining out of "
+                            f"{self.max_iterations}. Please wrap up your work. "
+                            "Stop calling tools and provide your final answer as plain text. "
+                            "If you have partial results, summarize what you have so far."
+                        ),
+                    })
+
                 # Streaming output + collect thinking text
                 thinking_chunks: List[str] = []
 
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks @toanalien! 🙏 Mirroring the swarm worker's wrap-up + forced-text-only pattern into `AgentLoop` is a nice consistency win, and the forced-text last iteration is verified to turn a tool-spamming run into a `success` with a final answer (confirmed against the real LLM e2e for `max_iter`).

Heads-up: I pushed a small follow-up on top (`7fd8c56`). The nudge fired at `iteration == wrap_up_at` unconditionally, and on tiny budgets (`max_iter <= 2` → `wrap_up_at == 1`) it injected the `[SYSTEM]` wrap-up message on the **first** iteration. That displaced the active research-goal context as the most recent user message — breaking `test_agent_goal_context` — and could tell the agent to stop early while a goal was still continuing. The follow-up gates the nudge to `1 < iteration < max_iterations` (never on the first iteration, never on the last, where the forced-text path already guarantees an answer). Full suite is green again. Thanks for the well-structured PR and tests!

PR #149, opened 2026-05-30
fix(backtest): pre-flight validation for LLM-generated signal engines

toanalien:
## Summary
- Add pre-flight validation for LLM-generated `SignalEngine` classes in the backtest runner, catching interface errors before execution with actionable JSON error messages instead of raw Python tracebacks
- Detect circular self-imports (`from signal_engine import SignalEngine`), required `__init__` args, missing `generate()` method, and wrong return types (`list` instead of `Dict[str, pd.Series]`)
- Fixes 5 distinct recurring crash patterns observed in production

## Test plan
- [x] All 2844 existing tests pass (`pytest`, excluding e2e)
- [x] `py_compile` passes on both modified files
- [x] `ruff check` passes on both modified files
- [ ] Deploy and verify LLM agent receives actionable JSON errors on malformed signal engines instead of tracebacks
- [ ] Verify valid signal engines still instantiate and run backtests normally

## Changes
| File | Change |
|------|--------|
| `agent/backtest/runner.py` | Circular import detection in AST validator, `_validate_signal_engine_class()` for init/generate checks, wired into `main()` |
| `agent/backtest/engines/base.py` | Return-type validation after `signal_engine.generate()` call |

## Errors addressed
| Error | Root Cause |
|-------|-----------|
| `SignalEngine.__init__() missing 1 required positional argument: 'config'` | LLM wrote `__init__(self, config)` without default |
| `'SignalEngine' object has no attribute 'generate'` | LLM omitted required method |
| `'list' object has no attribute 'reindex'` | `generate()` returned list instead of Dict[str, pd.Series] |
| `ImportError: cannot import name 'SignalEngine' from 'signal_engine'` | Circular self-import |
| `BacktestConfigSchema: start_date/end_date Field required` | Already caught by existing Pydantic validation; no code change needed |

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -351,6 +351,20 @@ class BaseEngine(ABC):
 
         # 2. Generate signals
         signal_map = signal_engine.generate(data_map)
+        if not isinstance(signal_map, dict):
+            print(json.dumps({"error": (
+                f"SignalEngine.generate() must return Dict[str, pd.Series], "
+                f"got {type(signal_map).__name__}. "
+                "Return a dict mapping symbol codes to pandas Series of signals."
+            )}))
+            sys.exit(1)
+        for _code, _sig in signal_map.items():
+            if not isinstance(_sig, pd.Series):
+                print(json.dumps({"error": (
+                    f"SignalEngine.generate() returned {type(_sig).__name__} for '{_code}', "
+                    "expected pd.Series. Each value must be a pandas Series with DatetimeIndex."
+                )}))
+                sys.exit(1)
         valid_codes = sorted(c for c in signal_map if c in data_map)
         if not valid_codes:
             print(json.dumps({"error": "No valid signals generated"}))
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks @toanalien! 🙏 Turning those five recurring LLM-generated-engine crashes into actionable messages is a real quality-of-life improvement for the agent loop. Verified all four interface cases (required `__init__` arg / missing `generate` / wrong return type / circular import) plus a valid engine running end-to-end.

One thing I tightened in a follow-up (`82ff522`): the circular-import check raises `ValueError` from `_validate_signal_engine_source`, but its call site (`_load_module_from_file`) was unguarded, so that case actually surfaced as a raw traceback rather than the JSON envelope the PR describes. I wrapped the call to emit `{"error": "SignalEngine source error: ..."}`, which also unifies the other source-level AST rejections (unsafe imports, decorators, top-level statements) with the runner's JSON contract. Everything else landed as-is.

PR #150, opened 2026-05-30
fix(ui): always show Full Report link when runId exists in session history

toanalien:
### Summary
- Fix "Full Report" link disappearing when loading sessions from a different browser or remote device
- `loadSessionMessages` and SSE handler now always render the `run_complete` card when `runId` exists in message metadata, even if `api.getRun()` fails (auth error, 404, network issue)
- Improve RunDetail error page with diagnostic hints (API auth, run directory) and a back button

### Root Cause
When reloading a session from history, if `metrics` was missing from message metadata, the code called `api.getRun(runId)` to fetch run data. If that call failed (common for remote/cross-browser access due to auth), the error was silently caught and the card was never added — so the "Full Report" link disappeared entirely. The original browser still showed it because the card was cached in Zustand's in-memory store from the live SSE session.

### Test plan
- [x] Open a session with completed backtest runs in the original browser — verify "Full Report" link still appears with metrics/chart
- [x] Open the same session in a new browser (or incognito) — verify "Full Report" link now appears
- [x] Click "Full Report" when run data is available — verify RunDetail page loads normally
- [x] Click "Full Report" when run data is unavailable — verify improved error page with hints and back button
- [x] `cd frontend && npx vite build` passes

CleanShot 2026-05-30 at 11 55 24@2x

### DIFF
--- a/frontend/src/pages/Agent.tsx
+++ b/frontend/src/pages/Agent.tsx
@@ -338,7 +338,7 @@ export function Agent() {
               const runData = await api.getRun(runId);
               if (isReportWorthyRun(runData)) {
                 fetchedMetrics = runData.metrics;
-                fetchedCurve = runData.equity_curve?.map((e) => ({ time: e.time, equity: e.equity }));
+                fetchedCurve = runData.equity_curve?.map((e) => ({ time: e.time, equity: Number(e.equity) }));
               }
             } catch { /* run may be unavailable — still show link */ }
             agentMsgs.push({
@@ -476,7 +476,7 @@ export function Agent() {
             const runData = await api.getRun(runId);
             if (isReportWorthyRun(runData)) {
               runMetrics = runData.metrics;
-              runCurve = runData.equity_curve?.map(e => ({ time: e.time, equity: e.equity }));
+              runCurve = runData.equity_curve?.map(e => ({ time: e.time, equity: Number(e.equity) }));
             }
           } catch { /* run data unavailable — still show link */ }
           s.addMessage({
--- a/frontend/src/pages/Agent.tsx
+++ b/frontend/src/pages/Agent.tsx
@@ -331,25 +331,33 @@ export function Agent() {
           if (metrics && Object.keys(metrics).length > 0) {
             agentMsgs.push({ id: m.message_id, type: "run_complete", content: "", runId, metrics, timestamp: ts + 1 });
           } else {
-            // Always show card when runId exists; enrich with fetched data if available
+            // Fetch run data to check report-worthiness; show fallback card if fetch fails
             let fetchedMetrics: Record<string, number> | undefined;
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for the fix — the root cause is real and the `RunDetail` not-found diagnostic page is a nice improvement. But the "always render the card when `runId` exists" approach is too broad and introduces a regression. Requesting changes before merge.

### The regression

Every assistant turn carries a `run_id`, not just backtests. In `service.py`, every `agent.run()` creates a run directory (`loop.py` `state_store.create_run_dir`), and the reply metadata gets `run_id` whenever a run dir exists:

```python
# session/service.py
if attempt.run_dir:
    reply_metadata["run_id"] = Path(attempt.run_dir).name
```

There is no separate lightweight chat path — a plain conversational answer also gets a `run_id`. Verified end-to-end against a local server:

```
# sent: "what is a moving average? answer in one sentence, do not run any backtest"
assistant metadata: {"run_id": "20260530_062432_41_d7d2d9", "status": "completed"}  # run_id present, no metrics
getRun(run_id) -> status=success, metrics=None, equity_curve=None, artifacts=[]
isReportWorthyRun(...) === false
```

On `main`, the `else` branch (`if (isReportWorthyRun)`) and the SSE branch (`if (hasReport || shadowId)`) correctly skip the card for these turns. With this PR, **every plain chat answer now renders a spurious "Full Report →" link**, and clicking it lands on a RunDetail page with no metrics / chart / trades.

### The case worth preserving

The genuine bug is narrower: a run that **is** report-worthy but has **no metrics in metadata** (e.g. a Pine-only or equity-curve-only run). Metrics-bearing backtests already render from metadata alone (first branch, no `getRun`, so they're fine cross-browser). The link only disappears cross-browser when `getRun` *fails* (401/404/network) and we can't confirm report-worthiness.

### Suggested fix

Only fall back to a bare card when `getRun` actually throws — i.e. distinguish "succeeded but not report-worthy" (plain chat → render nothing) from "couldn't fetch" (unknown → show the link):

```ts
let fetchedMetrics, fetchedCurve, runUnavailable = false;
try {
  const runData = await api.getRun(runId);
  if (isReportWorthyRun(runData)) {
    fetchedMetrics = runData.metrics;
    fetchedCurve = runData.equity_curve?.map(e => ({ time: e.time, equity: Number(e.equity) }));
  } else {
    // succeeded but not report-worthy (plain chat) -> no card
    continue; // or skip the addMessage in the SSE path
  }
} catch {
  runUnavailable = true; // can't tell -> still show the link
}
// push run_complete card only when we have data OR the run couldn't be fetched
```

This keeps the cross-browser fix (auth failure → `catch` → link still shows) without putting a "Full Report" link under every conversational reply. The `RunDetail.tsx` changes are good as-is.

PR #168, opened 2026-06-03
fix(loop): enrich trace entries with call_id for tool-call correlation

zwrong:
## Summary

- Add `call_id` to `tool_call` and `tool_result` trace entries, enabling clear correlation between tool calls and their results in `trace.jsonl` via `call_id`
- Remove 200-char arg truncation in trace writes for better post-hoc debugging

## Why

When replaying agent sessions from `trace.jsonl`, there is currently no way to correlate a `tool_call` entry with its corresponding `tool_result`. The existing `iter` field is **not sufficient** because `_execute_parallel` runs multiple tool calls within the same iteration:

```
iteration=3:
  tool_call   { iter: 3, tool: "web_search", call_id: "call_abc" }
  tool_call   { iter: 3, tool: "read_file",  call_id: "call_def" }
  tool_call   { iter: 3, tool: "web_search", call_id: "call_ghi" }   ← same iter, same tool
  tool_result { iter: 3, tool: "read_file",  call_id: "call_def" }
  tool_result { iter: 3, tool: "web_search", call_id: "call_abc" }
  tool_result { iter: 3, tool: "web_search", call_id: "call_ghi" }
```

In this scenario, `iter` alone (or even `iter + tool`) cannot distinguish the two `web_search` calls. `call_id` (the LLM-assigned tool-call identifier) provides a 1:1 join key for request ↔ response correlation regardless of parallelism or duplicate tool names.

Separately, the 200-char arg truncation in trace makes post-hoc debugging difficult for tools with large payloads (e.g. code generation).

## Changes

**`agent/src/agent/loop.py`** (3 lines changed):

| Location | Change |
|---|---|
| `_execute_parallel` (L749) | Add `call_id: tc.id` to trace, remove `[:200]` on args |
| `_execute_single` (L794) | Same as above |
| `_finalize_tool_result` (L873) | Add `call_id: tc.id` to `tool_result` trace entry |

### Before
```python
trace.write({"type": "tool_call", ..., "args": {k: str(v)[:200] for ...}})
trace.write({"type": "tool_result", ..., "preview": result[:200]})
```

### After
```python
trace.write({"type": "tool_call", ..., "call_id": tc.id, "args": {k: str(v) for ...}})
trace.write({"type": "tool_result", ..., "call_id": tc.id, "preview": result[:200]})
```

> **Note:** `black` reformatting of `loop.py` is a pre-existing issue on `main`, not addressed in this PR to keep the diff minimal.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`) — 2845 passed, 2 skipped, 0 failed. The 2 skipped tests are pre-existing (alpha101_096 NaN ratio, TUSHARE_TOKEN not set), unrelated to this PR.
- [ ] New tests added (if applicable) — N/A, no new behavior, only trace metadata enrichment
- [x] Tested manually: ran an agent session, inspected `trace.jsonl` and confirmed `call_id` appears on both `tool_call` and `tool_result` entries, and args are untruncated in trace

## Checklist

- [ ] Documentation updated (if user-facing change) — N/A, internal trace format only

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -870,7 +870,7 @@ class AgentLoop:
         truncated = result[:TOOL_RESULT_LIMIT]
         messages.append(context.format_tool_result(tc.id, tc.name, truncated))
 
-        trace.write({"type": "tool_result", "iter": iteration, "tool": tc.name, "call_id": tc.id, "status": status, "elapsed_ms": elapsed_ms, "result": result})
+        trace.write({"type": "tool_result", "iter": iteration, "tool": tc.name, "call_id": tc.id, "status": status, "elapsed_ms": elapsed_ms, "preview": result[:200]})
         react_trace.append({"type": "tool_call", "tool": tc.name, "result_preview": result[:200]})
         self._emit("tool_result", {"tool": tc.name, "status": status, "elapsed_ms": elapsed_ms, "preview": result[:200]})
 
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -746,7 +746,7 @@ class AgentLoop:
         for tc in tool_calls:
             args = _normalize_tool_run_dir(tc.arguments, self.memory.run_dir)
             self._emit("tool_call", {"tool": tc.name, "arguments": {k: str(v)[:200] for k, v in args.items()}, "iter": iteration})
-            trace.write({"type": "tool_call", "iter": iteration, "tool": tc.name, "args": {k: str(v)[:200] for k, v in args.items()}})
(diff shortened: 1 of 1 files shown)


warren618:
Thanks for spotting this, @zwrong — correlating `tool_call` ↔ `tool_result` via `call_id` is a genuinely useful trace improvement. We've landed an equivalent change in abbbab5, crediting you via `Co-authored-by`. The only tweak from your patch: we kept the existing `[:200]` truncation on the serialized tool args, since dropping it would write full untruncated inputs into every trace entry and bloat trace files for tools with large arguments. Closing in favor of that commit — thanks again for the contribution! 🙏

PR #176, opened 2026-06-04
fix: round-trip Gemini thoughtSignature through OpenAI-compat tool calls

### LINKED ISSUE #170: thoughtSignature for Gemini 2.5 and 3.0 - AgentLoop error
### Problem or Motivation

When using thinking models (such as the Gemini 3 and 2.5 series), the API returns a  field within the [content parts](https://ai.google.dev/api/caching#Part) of the response (e.g., text or functionCall parts) - thoughtSignature

As a general rule, if you receive a thought signature in a model response, you should pass it back exactly as received when sending the conversation history in the next turn. When using Gemini 3 models, you must pass back thought signatures during function calling, otherwise you will get a validation error (4xx status code). This includes when using the minimal [thinking level](https://ai.google.dev/gemini-api/docs/thinking#thinking-levels) setting for Gemini 3 Flash.

right now, getting an error for src/agent/loop.py,  AgentLoop class
'Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:load_skill` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.', 'status': 'INVALID_ARGUMENT'}}]

### Proposed Solution

if you get response containing "thoughtSignature", add to your next call

### Area

CLI / TUI

### Alternatives Considered

_No response_

### Contribution

mvanhorn:
## Summary

Gemini's per-function-call thought signature now round-trips through the OpenAI-compatible tool-call path, so multi-turn function calling against Gemini 2.5/3.x thinking models no longer fails with `INVALID_ARGUMENT` ("Function call is missing a thought_signature").

## Why

Closes #170. `ChatOpenAIWithReasoning` in `agent/src/providers/llm.py` already preserved `reasoning_content` at message granularity, but LangChain's fixed `{id, name, args, type}` tool-call schema dropped the per-tool-call signature at message reconstruction. `_extract_tool_call_thought_signature` now reads the signature from each inbound tool call (including the `extra_content.google` extension) and re-injects it onto the matching outbound `tool_calls[j]`. `ToolCallRequest` in `agent/src/providers/chat.py` gains an optional `thought_signature: str | None`, threaded through the assistant-message rebuild in `agent/src/agent/loop.py`. The field is default-absent, so OpenAI / DeepSeek / OpenRouter providers never populate it and stay byte-identical.

## Changes

`agent/src/providers/llm.py` captures and re-emits the thought signature per tool call during assistant-message reconstruction. `agent/src/providers/chat.py` adds an optional `thought_signature` to `ToolCallRequest`, and `agent/src/agent/loop.py` threads that signature through the rebuild path.

## Test Plan

- [x] Tested manually (describe below) (N/A: covered by the automated round-trip tests below; no live Gemini call was made)

`agent/tests/test_kimi_reasoning_content.py` is extended with the round-trip cases: a multi-turn exchange re-emits each signature on the matching tool call, mixed calls preserve per-call mapping, and a no-signature response serializes identically to today (field stays `None`).

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — issue #170 is the maintainer's `help wanted` scoping of this exact change ("Happy to review a PR along these lines"), following the preferred Route A.

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -163,6 +163,37 @@ def _fix_tool_pairs(messages: list) -> None:
         messages.insert(pos, stub)
 
 
+def _attach_tool_call_thought_signatures(message: dict[str, Any], tool_calls: list) -> None:
+    """Attach Gemini thought signatures to replayed assistant tool calls."""
+    outbound_tool_calls = message.get("tool_calls")
+    if not isinstance(outbound_tool_calls, list):
+        return
+
+    signatures_by_id = {
+        tc.id: tc.thought_signature
+        for tc in tool_calls
+        if getattr(tc, "thought_signature", None)
+    }
+    for index, outbound_tool_call in enumerate(outbound_tool_calls):
+        if not isinstance(outbound_tool_call, dict):
+            continue
+        signature = signatures_by_id.get(outbound_tool_call.get("id"))
+        if not signature and index < len(tool_calls):
+            signature = getattr(tool_calls[index], "thought_signature", None)
+        if not signature:
+            continue
+
+        extra_content = outbound_tool_call.get("extra_content")
+        if not isinstance(extra_content, dict):
+            extra_content = {}
+            outbound_tool_call["extra_content"] = extra_content
+        google = extra_content.get("google")
+        if not isinstance(google, dict):
+            google = {}
+            extra_content["google"] = google
+        google["thought_signature"] = signature
+
+
 # -- Structured summary templates ------------------------------------------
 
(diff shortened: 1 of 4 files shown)


mvanhorn:
Thanks again @warren618 for the Gemini thoughtSignature round-trip fix.

mvanhorn:
Thanks @warren618 for landing this. Round-tripping the Gemini thoughtSignature keeps tool calls intact through the OpenAI-compat layer.

Issue #62, opened 2026-04-30
[Feature] Tushare fundamental coverage is insufficient for pre-filter strategies (need DataProvider abstraction)


shadowinlife:
### Background
Many JoinQuant-style strategies rely on financial-statement-based pre-filtering before signal generation.
In current Vibe-Trading, Tushare integration appears limited to market bars plus a subset of daily_basic fields, which blocks a large class of fundamental strategies.

### Current behavior
- Tushare loader fetches daily/minute bars, then optionally merges only daily_basic fields
- Backtest pipeline only passes extra_fields through, without a broader financial-table query/join layer
- Loader protocol is OHLCV-centric
- Existing fundamental example is limited to PE/PB/ROE-like fields

### Impact
- Fundamental pre-screening strategies cannot be migrated directly.
- Strategies requiring income/balance-sheet/cashflow/fina indicators are hard to implement.
- Reduced compatibility with community quant recipes.

### Introduce a DataProvider abstraction layer with a stable contract for LLM/tools/backtest:

1. Metadata introspection
- Expose machine-readable metadata (table, columns, types, keys, time semantics, update frequency).
2. Standard DataFrame output
- All data queries return standardized pandas DataFrame objects with schema checks.
3. Virtual view mechanism (optional but recommended)
- Allow reusable derived views from multiple DataFrames (join/transform/cache), reducing repeated expensive recomputation.
4. Point-in-time safety
- Enforce publish-date/announcement-date alignment to avoid look-ahead bias.

### Minimal first milestone

- Extend Tushare support beyond daily_basic to include commonly used financial datasets (for example: fina_indicator, income, balancesheet, cashflow) under a unified provider contract.
- Add regression tests for schema contract, PIT correctness, and representative pre-filter strategy runs.

### Area

Data Sources

### Discussion questions

Should DataProvider live under backtest/loaders or as a separate laye
Should we support local DB backends (DuckDB/MySQL/Postgres) behind the same provider contract

### What we have done
We currently have a production solution based on opencode + nanobot + tushare local data in DuckDB, which helps our clients perform preprocessing for various quantitative analyses through natural language. The entire system is built around the open-source architecture from The University of Hong Kong (HKU), with additional integration of conceptual frameworks from Microsoft's RDAgent for in-depth exploration.

[...]

PR #177, opened 2026-06-04
feat: opt-in local data cache layer for backtest loaders

### LINKED ISSUE #171: [Feature] Local data cache layer for long-horizon / cross-market backtests
Spun off from #62 to track it on its own.

When running long-horizon or cross-market backtests (e.g. 10-year A-share, or US → A-share / US → crypto mapping), fetching the same data from Tushare / yfinance / etc. at runtime on every run is slow, and can trip provider rate limits or IP bans (reported in #62 — falling back to akshare and getting a cloud IP blocked).

A local cache layer would let the same bars / fundamentals be fetched once and reused across runs, without breaking the "no market data cached in the repo" rule — the cache lives under the user's `~/.vibe-trading/`, never committed.

Scope to discuss:
- storage backend (DuckDB / parquet under `~/.vibe-trading/cache/`)
- cache key: source + symbol + timeframe + date range + fields
- opt-in flag, default off, so existing behavior is unchanged
- TTL / invalidation, and point-in-time correctness for fundamentals
- which providers to cover first (Tushare fundamentals + daily bars are the obvious ones)

Context: requested by @shadowinlife in #62 for complex multi-market backtests.

mvanhorn:
## Summary

Adds an opt-in local data cache for backtest loaders so repeated long-horizon and cross-market runs stop re-fetching identical bars and fundamentals from the provider. It is off by default and wired into the Tushare daily-bars loader as the reference integration.

## Why

Closes #171. Long-horizon and cross-market backtests re-fetch the same data on every run, which is slow and can trip provider rate limits or IP bans (the akshare-fallback block in #62). The cache lives under `~/.vibe-trading/cache/`, is created lazily, and is never written into the repo tree, preserving the "no market data cached in the repo" rule.

## Changes

`agent/backtest/loaders/base.py` gains a cache helper alongside the existing retry helpers, content-addressed by source, symbol, timeframe, date range, and fields, and gated behind the `VIBE_TRADING_DATA_CACHE` env var (default off). The as-of date is part of the key for fundamentals so revised figures never overwrite an earlier point-in-time snapshot. In `agent/backtest/loaders/tushare.py` the daily-bars loader reads through the cache when enabled; other loaders are left untouched in this PR.

## Test Plan

- [x] Tested manually (describe below) (N/A: covered by the automated cache tests below; no live provider fetch was made)

`agent/tests/test_loader_retry_helpers.py` adds cache cases: a second identical fetch reads from cache with the provider called once, differing symbol/range/fields produce distinct keys with no false hits, disabled mode never touches `~/.vibe-trading/`, and a corrupt cache entry falls back to a live fetch instead of raising.

## Checklist

### DIFF
--- a/agent/.env.example
+++ b/agent/.env.example
@@ -105,6 +105,13 @@ TUSHARE_TOKEN=your-tushare-token
 # FUTU_HOST=127.0.0.1
 # FUTU_PORT=11111
 
+# Opt-in local market-data cache for backtest loaders (off by default).
+# When enabled, every data source caches settled historical bars under
+# ~/.vibe-trading/cache/loaders/ so repeated/long-horizon backtests skip the
+# network. Only fully-elapsed days are cached (ranges ending today are always
+# re-fetched). Clear it any time with: rm -rf ~/.vibe-trading/cache
+# VIBE_TRADING_DATA_CACHE=1
+
 # ============================================================================
 # API Server (optional)
 # ============================================================================
--- a/agent/backtest/loaders/akshare_loader.py
+++ b/agent/backtest/loaders/akshare_loader.py
@@ -11,7 +11,7 @@ from typing import Dict, List, Optional
 
 import pandas as pd
 
-from backtest.loaders.base import validate_date_range
+from backtest.loaders.base import cached_loader_fetch, validate_date_range
 from backtest.loaders.registry import register
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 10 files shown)


mvanhorn:
Appreciate the merge @warren618, hope the opt-in backtest cache saves folks some loader time.

mvanhorn:
Grateful for the merge @warren618 - the local cache should make repeated long-horizon backtests much faster. Thanks!

PR #184, opened 2026-06-06
fix(llm): preserve Gemini thought_signature through the AgentLoop dict path

ngoanpv:
## Problem

Configuring a Gemini 3.x model (e.g. `gemini-3.5-flash`) breaks multi-turn tool calling:

> 400 - Function call is missing a thought_signature in functionCall parts ... `default_api:load_skill`, position 2

#176 added the thought_signature round-trip, but it only works when history is held as in-memory LangChain `AIMessage` objects — the path its unit tests exercise. The real `AgentLoop` replays history as OpenAI-format **dicts**, stamping the signature into `tool_calls[i].extra_content.google.thought_signature` (`loop.py:_attach_tool_call_thought_signatures`).

## Root cause

LangChain's `_convert_dict_to_message` discards `extra_content` during the early `_convert_input` step of `invoke()`/`stream()` — *before* `_get_request_payload` re-injects. So `_signature_maps` reads an empty `additional_kwargs` and the outbound payload goes out **unsigned** → Gemini 400 on the next tool-using turn. This affects every tool-using loop turn on gemini-3.x, not just parallel calls. (Parallel isn't special: Gemini signs only the first of N parallel calls and that one signature covers the block — the bug was that even the first never reached the wire.) Verified by diffing the outbound wire payload: the in-memory path carries `extra_content`, the dict path does not.

## Fix

Override `ChatOpenAIWithReasoning._convert_input` — the single chokepoint both `invoke` and `stream` call while `input` is still raw dicts — to re-attach the dict's signatures into `additional_kwargs["tool_call_thought_signatures"]` (the exact shape the in-memory path produces). The existing `_signature_maps`/`_inject_tool_call_thought_signatures` machinery then handles both paths uniformly. Idempotent via an `isinstance(raw, dict)` guard (no-op on already-converted `BaseMessage`s). No parallel-specific code needed.

## Tests

- Two regression tests in `TestChatOpenAIWithReasoningOutboundPayload` driving the real dict path (the gap #176's tests missed): confirmed **red** without the fix (`KeyError: 'extra_content'`), **green** with it.
- Full suite: **3072 passed, 2 skipped**.
- Verified end-to-end on a live server: a real session loaded `load_skill` ×2 and ran a backtest on `gemini-3.5-flash`, completing with no 400.

### DIFF
--- a/agent/src/providers/llm.py
+++ b/agent/src/providers/llm.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import logging
 import os
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, Optional
 from urllib.parse import urlsplit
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thank you. This was the missing half of the Gemini fix: #176 only covered the in-memory `AIMessage` path, but the real `AgentLoop` replays history as OpenAI-format dicts where the signature was dropped before the payload was built. Re-attaching at the `_convert_input` chokepoint is exactly the right single place, and the parallel-call test (first-of-N signed, rest absent) nails the tricky case. Verified green here too. 🙏

Small heads-up for future readers: like the rest of this class it leans on langchain-openai's private `_convert_input`, so worth a re-check whenever we bump that dep — not a blocker, just noting.

PR #189, opened 2026-06-07
fix(swarm): honor explicit preset names

BillDin:
## Summary

- Honor explicit swarm preset names before falling back to keyword scoring.
- Tighten derivatives keyword matching so ordinary words like "given" no longer trigger the `IV` rule.
- Add regression coverage for the `investment_committee` prompt routing case.

## Why

The SwarmTool router currently treats the entire user prompt as natural language and scores keyword matches only. A prompt such as `Use the investment_committee preset ... given current market conditions` can miss `investment_committee` because of the underscore, while the bare `IV` derivatives keyword matches the `iv` inside `given`. That routes an explicitly requested investment committee run to `derivatives_strategy_desk`.

## Changes

- Normalize spaces and hyphens to underscores, then prefer an explicitly mentioned preset name before keyword scoring.
- Restrict derivatives English patterns to whole option-related phrases and whole-word `IV`.
- Add `agent/tests/test_swarm_tool_preset_matching.py` covering explicit underscore, explicit spaced preset name, and the `given` false-positive.

## Affected Areas

- `agent/src/tools/swarm_tool.py`
- `agent/tests/test_swarm_tool_preset_matching.py`

## Out of Scope

- No changes to preset YAML definitions.
- No changes to swarm runtime execution, tool permissions, broker/live trading, MCP server configuration, or frontend UI.

## Test Plan

  - Ran in local `.venv` with the CI-style backend command:
    `./.venv/Scripts/python.exe -m pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q`
  - Result: `3056 passed, 2 skipped, 8 failed, 9 errors`.
  - The remaining failures appear local/Windows-specific rather than caused by this PR:
    - `agent/tests/factors/test_registry.py`: symlink fixture fails with `WinError 1314` because this Windows session lacks symlink privilege/developer-mode elevation.
    - `agent/tests/test_oauth_token_cache.py`: `0700` directory mode assertions see `0o777` on Windows filesystem semantics.
    - `agent/tests/test_loader_retry_helpers.py`: opt-in loader cache tests observed local `Path.home()` / cache isolation differences on Windows.
  - `.venv` already had the project dependencies; `pip install -e ".[dev]"` reached package reinstall but could not replace `vibe-trading.exe` because another local process had it open. Test collection still proceeded with installed dependencies.
  - `./.venv/Scripts/python.exe -m pytest agent/tests/test_swarm_tool_preset_matching.py` -> `3 passed`
  - Verified the screenshot prompt now matches `investment_committee` via the new regression test.

## Risk Notes

- Live/broker risk: none. This only changes prompt-to-preset routing before a swarm run starts.
- MCP/network/secret risk: none. No credentials, external MCP configuration, or network behavior changed.
- Behavioral risk: low. Explicit preset names now take precedence; keyword fallback remains for prompts without a preset name.

## Rollback / Close Path

Revert this commit to restore the prior keyword-only routing behavior. The new regression test should be removed with the revert.

## Checklist

  - Not required: this is an internal routing bug fix with regression coverage.

### DIFF
--- a/agent/src/tools/swarm_tool.py
+++ b/agent/src/tools/swarm_tool.py
@@ -123,12 +123,12 @@ _PRESET_KEYWORDS: list[tuple[str, list[str], float]] = [
     (
         "derivatives_strategy_desk",
         [
-            r"\boption\b",
-            "call",
-            "put",
-            "Greeks",
+            r"\boptions?\b",
+            r"\bcall\s+options?\b",
+            r"\bput\s+options?\b",
+            r"\bGreeks?\b",
             r"implied\s+vol",
-            "IV",
+            r"\bIV\b",
             "期权",
             "衍生品",
         ],
@@ -374,6 +374,11 @@ def _match_preset(prompt: str) -> str:
     Returns:
         Best matching preset name.
     """
+    normalized_prompt = re.sub(r"[\s-]+", "_", prompt.strip().lower())
+    for preset_name, _, _ in _PRESET_KEYWORDS:
+        if re.search(rf"(?<![a-z0-9]){re.escape(preset_name)}(?![a-z0-9])", normalized_prompt):
+            return preset_name
+
     scores: dict[str, float] = {}
     for preset_name, keywords, boost in _PRESET_KEYWORDS:
         score = 0.0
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks. Preferring an explicitly named preset before keyword scoring is the right fix, and the `IV`-inside-"given" false match was a real sharp edge. Appreciate the regression coverage.

Issue #198, opened 2026-06-10
[Bug] Swarm workers can bypass market data loader and misroute continuation runs


BillDin:
## Description

**What happened:**

While running the Web UI swarm flow for an NVDA investment committee request, I saw two related issues:

1. The requested `investment_committee` swarm completed, but some workers generated their own ad-hoc `yfinance` scripts instead of using the existing market data loader / `get_market_data` path. One worker hit a dirty latest daily bar from Yahoo/yfinance: the row had `Volume` but empty `Open/High/Low/Close`, which propagated into `NaN` metrics and non-strict JSON output.
2. After the completed committee run, the agent launched an additional continuation swarm with a prompt like `Continue and finish report...`. Because that continuation prompt did not contain the original preset/ticker context, preset matching fell back to `equity_research_team`. That run then failed on `task-stock` with `output contract not met: plan-only stub (no executed analysis / conclusion)`, and the aggregate task was blocked.

**What I expected:**

- Swarm workers that need OHLCV data should prefer the repository's normalized loader / `get_market_data` interface instead of generating raw `yf.Ticker(...).history(...)` / `yf.download(...)` code for core price bars.
- Dirty or partial latest bars should be dropped or repaired before metrics/JSON serialization.
- Follow-up/continuation swarm calls should preserve the original preset and key variables, or avoid launching a new inferred-preset swarm when the prior swarm already completed.

## Steps to Reproduce

1. Start the Web UI/backend with `vibe-trading serve`.
2. Submit a prompt similar to:

   ```text
   [Swarm Team Mode] Use the investment_committee preset to evaluate whether to go long or short on NVDA given current market conditions
   ```

3. Inspect the generated swarm runs/artifacts.
4. In my local run:
   - First swarm: `investment_committee`, status `completed`.
   - A worker-generated script used raw `yfinance` and selected the latest row without validating OHLC fields.
   - The generated CSV contained a latest row like `2026-06-09 ... ,,,,,179602708,...` where OHLC fields were empty but volume existed.
   - A second swarm was launched with `preset=equity_research_team` from a continuation prompt, and failed with `plan-only stub` on the stock task.

## Error Logs

```shell
failed continuation swarm:
  preset: equity_research_team
  task-stock: failed
  error: output contract not met: plan-only stub (no executed analysis / conclusion)
  task-aggregate: blocked
  error: Blocked: upstream not completed (task-stock=failed)

raw yfinance artifact symptom:
[...]

PR #199, opened 2026-06-10
fix(swarm): expose loader-backed market data tool

BillDin:
## Summary

Expose a loader-backed local `get_market_data` tool to swarm workers and steer yfinance/OHLCV workflows toward the existing normalized loader path.

## Why

Swarm workers could previously load the `yfinance` skill and generate raw Yahoo/yfinance scripts for OHLCV bars. Those scripts could bypass loader safeguards, trust malformed latest bars, and emit non-strict JSON containing `NaN`. This made market-data-heavy presets fragile even though the repository already had safer loader behavior.

## Changes

- Added `src.market_data` as a shared helper for MCP and local tools.
- Added a local `get_market_data` `BaseTool` backed by the existing loader registry.
- Kept the MCP `get_market_data` function on the same behavior path by delegating to the shared helper.
- Added strict JSON handling that converts non-finite floats to `null` before serialization.
- Added `get_market_data` to the data-heavy workers in `investment_committee` and `equity_research_team`.
- Added a worker prompt policy that tells agents to use `get_market_data` first for OHLCV/current price bars.
- Updated the yfinance skill so OHLCV examples are loader/tool-first and raw yfinance is reserved for fields outside loader coverage.
- Added regression tests for strict JSON, swarm registry exposure, and worker prompt guidance.

## Affected Areas

- MCP/local market data helper path
- Swarm local tool discovery
- `investment_committee` and `equity_research_team` presets
- yfinance skill prompt/documentation
- Swarm worker prompt construction

## Out of Scope

- No changes to broker/live trading connectors.
- No changes to order placement, account selection, or trading gates.
- No attempt to change all non-yfinance skills or every market-data preset in this PR.
- No continuation/preset routing changes; that is handled separately.

## Risk Notes

Live/broker risk: none expected. This PR only adds read-only market data access and prompt/preset guidance.

MCP/network/secret risk: no new secrets or external servers are introduced. The local tool reuses the existing loader path. It may call external market data providers through existing loaders, as MCP `get_market_data` already does.

Behavior risk: workers may spend an extra tool call on `get_market_data` before scripting OHLCV calculations. This is intentional and should reduce malformed raw-yfinance artifacts.

## Test Plan

- `cd agent; ..\.venv\Scripts\python.exe -m pytest tests\test_market_data_tool.py tests\test_get_market_data_size.py tests\test_get_market_data_unresolved.py`
- `cd agent; ..\.venv\Scripts\python.exe -m py_compile mcp_server.py src\market_data.py src\tools\market_data_tool.py src\swarm\worker.py`
- `git diff --check`

## Rollback / Close Path

Revert this PR to remove the local swarm `get_market_data` tool and restore MCP-only market data behavior. Existing MCP callers should remain isolated because the public MCP tool signature is unchanged.

## Checklist

- [x] Tests added/updated
- [x] Documentation/prompt guidance updated where relevant
- [x] No secrets, credentials, or token cache changes
- [x] No live trading or broker write flows exercised
- [x] DCO sign-off included

Fixes part of #198.

### DIFF
--- a/agent/mcp_server.py
+++ b/agent/mcp_server.py
@@ -32,9 +32,7 @@ from __future__ import annotations
 
 import json
 import logging
-import math
 import os
-import re
 import sys
 from pathlib import Path
 from typing import Any
@@ -45,6 +43,13 @@ if str(AGENT_DIR) not in sys.path:
     sys.path.insert(0, str(AGENT_DIR))
 
 from fastmcp import Context, FastMCP
+from src.market_data import (
+    DEFAULT_MAX_ROWS,
+    cap_rows,
+    detect_source,
+    fetch_market_data_json,
+    get_loader,
+)
 
 mcp = FastMCP("Vibe-Trading")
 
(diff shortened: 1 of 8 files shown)


warren618:
Verified locally — targeted tests plus the swarm/MCP-related suite pass, and an end-to-end NVDA fetch through the new tool returns strict loader-normalized JSON. Really thorough work on both the diagnosis in #198 and the fix, thank you! Merging.

PR #200, opened 2026-06-10
fix(swarm): preserve preset context for continuations

BillDin:
## Summary

Make in-process swarm launches preserve explicit preset context and reject ambiguous continuation prompts before they silently fall back to `equity_research_team`.

## Why

A completed `investment_committee` run can be followed by an agent-generated continuation prompt such as `Continue and finish report...`. That fragment no longer contains the original preset/ticker context, so the current auto-routing fallback can start a fresh `equity_research_team` run instead of continuing the intended committee result. This creates confusing extra swarm failures and unrelated A-share/equity-team variables.

## Changes

- Added optional `preset_name` to the local `run_swarm` agent tool schema.
- Added preset-name normalization so values like `Investment Committee` resolve to `investment_committee`.
- Added continuation-prompt detection for prompts like `continue`, `finish report`, `continue from ...`, and similar Chinese prefixes.
- Refuse ambiguous continuation prompts that lack any preset/routing signal instead of defaulting to `equity_research_team`.
- Updated the main agent routing prompt to pass `preset_name` when the user names a team/preset and to avoid launching fresh swarms from continuation fragments.
- Added regression coverage for explicit preset normalization and the ambiguous-continuation guard.

## Affected Areas

- In-process `run_swarm` tool used by the agent loop
- Main agent system prompt routing guidance
- Swarm preset matching tests

## Out of Scope

- No changes to MCP `run_swarm` wire schema.
- No changes to market data loading or yfinance handling; that is handled separately in #199.
- No changes to swarm runtime DAG execution or worker output contracts.
- No live/broker/order behavior changes.

## Risk Notes

Live/broker risk: none expected. This only affects research swarm launch routing.

MCP/network/secret risk: none expected. No new external endpoints, MCP config fields, or secrets are introduced.

Behavior risk: a prompt that starts with vague continuation wording may now return an explicit error instead of launching the default equity team. That is intentional to avoid misrouted and contextless runs. Fresh non-continuation prompts still retain the existing `equity_research_team` fallback.

## Test Plan

- `cd agent; ..\.venv\Scripts\python.exe -m pytest tests\test_swarm_tool_preset_matching.py tests\test_swarm_status_hydration.py`
- `cd agent; ..\.venv\Scripts\python.exe -m py_compile src\tools\swarm_tool.py src\agent\context.py`
- `git diff --check`

## Rollback / Close Path

Revert this PR to restore the prior always-auto-routed local `run_swarm(prompt=...)` behavior. Existing ordinary swarm prompts should remain compatible because the new `preset_name` parameter is optional.

## Checklist

- [x] Tests added/updated
- [x] Prompt/routing guidance updated
- [x] No secrets, credentials, or token cache changes
- [x] No live trading or broker write flows exercised
- [x] DCO sign-off included

Fixes part of #198.

### DIFF
--- a/agent/src/agent/context.py
+++ b/agent/src/agent/context.py
@@ -43,7 +43,9 @@ Decide which workflow to use based on the request:
 5. Do NOT write run_backtest.py. The engine is built-in.
 
 **Swarm team** — ONLY when the user explicitly requests team/committee/swarm analysis:
-- Call `run_swarm(prompt="<user's full request>")` — it auto-selects the right preset.
+- Call `run_swarm(prompt="<user's full request>", preset_name="<explicit preset>")` when the user names a preset/team, e.g. `investment_committee`.
+- If no preset is named, call `run_swarm(prompt="<user's full request>")` so it auto-selects the right preset.
+- For follow-up wording like "continue", "finish the report", or "continue from ...", do NOT start a fresh swarm from that fragment. Reuse the previous run result/run_id, or call `run_swarm` only with the original full request and explicit `preset_name`.
 - Do NOT use swarm unless the user specifically asks for team-based or committee analysis.
 
 **Analysis / research** — user wants factor analysis, options pricing, market data, or general research:
--- a/agent/src/tools/swarm_tool.py
+++ b/agent/src/tools/swarm_tool.py
@@ -394,6 +394,65 @@ def _match_preset(prompt: str) -> str:
     return "equity_research_team"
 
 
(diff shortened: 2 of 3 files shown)


warren618:
Verified locally — preset matching tests pass and I probed the continuation guard for false positives (prompts with any routing signal, including Chinese ones, still auto-route; only context-free continuation fragments get refused, with a recoverable error). Combined run with #199 is green across the full suite. Thanks again — merging.

PR #201, opened 2026-06-10
feat(loaders): add RSSHub event/sentiment provider

### LINKED ISSUE #172: [Feature] Event / sentiment factor provider (RSSHub)
Spun off from #62 as the "phase 2" provider track, kept separate from the fundamental data layer so each stays reviewable.

The fundamental data layer (#74 / #76) now covers financial statements + indicators for pre-filtering. A separate track is news / announcement / sentiment / event factors — e.g. RSSHub as an entry point — for strategies that key off catalysts and sentiment rather than fundamentals alone.

Scope to discuss:
- RSSHub (or similar) as a source for news / filings / sentiment feeds
- a `sentiment` / `event` provider contract, parallel to the fundamental provider
- mapping feeds to symbols + timestamps with point-in-time correctness (no look-ahead)
- a representative event/sentiment-driven pre-filter or signal example

Context: raised by @shadowinlife in #62 (RSSHub-driven sentiment factors).

Shizoqua:
## Summary

- Add an RSSHub-backed event/sentiment provider, parallel to the Tushare
  fundamental data layer, feeding the existing `event-driven` skill schema.
- Point-in-time-safe enrichment (knowable-date gating + exponential decay)
  plus a representative event-driven signal example.

## Why

Closes #172. Enables catalyst/sentiment strategies alongside fundamentals,
the "phase 2" provider track requested in the issue.

## Changes

- `agent/backtest/loaders/rsshub_events.py`: `RSSHubEventProvider`
  (`is_available`/`list_feeds`/`query_events`), `enrich_price_frames_with_events`
  (`event_score`/`event_count` via knowable-date gating), pluggable `score_event`
  with a dependency-free lexicon default.
- `agent/backtest/engines/base.py` + `runner.py`: optional `event_feeds` config,
  wired beside `fundamental_fields`.
- `agent/src/skills/event-driven/example_signal_engine.py`: long-only sentiment
  pre-filter example.
- Tests: provider parsing/PIT/dedup + a no-look-ahead test.
- `pyproject.toml`: add `defusedxml` for safe feed-XML parsing.

## Test Plan

- [x] `pytest agent/tests/test_rsshub_events_provider.py agent/tests/test_rsshub_events_lookahead.py -q` (passing)
- [x] New tests added (provider + no-look-ahead)
- [x] `python -m py_compile backtest/runner.py` (CI syntax gate) + import smoke test

## Checklist

- [x] No hardcoded values (RSSHub base URL via `$RSSHUB_BASE_URL`)
- [x] Documentation updated (skill example + docstrings)

## Risk note (per AGENT_CONTRIBUTOR_GUIDE.md)

Additive provider; no broker/order/MCP/credential surfaces touched. Network
only via opt-in `$RSSHUB_BASE_URL` at fetch time; unit tests mock HTTP (no live
calls). Untrusted feed XML parsed with `defusedxml` (XXE / entity-expansion
safe). One new dependency (`defusedxml`). Rollback: revert the commit.

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -20,6 +20,10 @@ from typing import Any, Callable, Dict, List, Optional
 
 import pandas as pd
 
+from backtest.loaders.rsshub_events import (
+    RSSHubEventProvider,
+    enrich_price_frames_with_events,
+)
 from backtest.loaders.tushare_fundamentals import (
     TushareFundamentalProvider,
     enrich_price_frames_with_fundamentals,
@@ -202,6 +206,49 @@ def _maybe_enrich_fundamentals(
         ) from exc
 
 
+def _normalise_event_feeds(config: Dict[str, Any]) -> list[str]:
+    """Read the optional RSSHub feed-name list from backtest config."""
+    raw_feeds = config.get("event_feeds")
+    if raw_feeds in (None, [], {}):
+        return []
+    if isinstance(raw_feeds, str):
+        raw_feeds = [raw_feeds]
+    if not isinstance(raw_feeds, Iterable):
+        raise ValueError("event_feeds must be a list of RSSHub feed names")
+
+    feeds = [str(feed).strip() for feed in raw_feeds]
+    if any(not feed for feed in feeds):
+        raise ValueError("event_feeds must not contain empty feed names")
+    return feeds
+
+
+def _maybe_enrich_events(
+    data_map: Dict[str, pd.DataFrame],
+    config: Dict[str, Any],
+) -> Dict[str, pd.DataFrame]:
+    """Attach a point-in-time-safe ``event_score`` column before signal generation."""
+    feeds = _normalise_event_feeds(config)
+    if not feeds:
+        return data_map
+
+    try:
+        provider = RSSHubEventProvider()
+        if not provider.is_available():
+            raise RuntimeError(f"RSSHub base URL not configured (set ${'RSSHUB_BASE_URL'})")
+        return enrich_price_frames_with_events(
+            data_map,
+            provider,
+            as_of=config.get("end_date", ""),
+            feeds=feeds,
+            decay_lambda=float(config.get("event_decay_lambda", 0.1)),
+            lookback=int(config.get("event_lookback", 30)),
+        )
+    except Exception as exc:
+        raise RuntimeError(
+            f"event_feeds requested but RSSHub enrichment failed: {exc}"
+        ) from exc
+
+
 # ─── Base Engine ───
 
 
@@ -348,6 +395,7 @@ class BaseEngine(ABC):
             print(json.dumps({"error": "No data fetched"}))
             sys.exit(1)
         data_map = _maybe_enrich_fundamentals(data_map, config)
+        data_map = _maybe_enrich_events(data_map, config)
 
         # 2. Generate signals
         signal_map = signal_engine.generate(data_map)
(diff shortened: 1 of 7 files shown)


warren618 (changes requested):
Thanks @Shizoqua — the core of this is genuinely well-built. I verified the provider machinery end-to-end against a live self-hosted RSSHub (docker `diygod/rsshub`): fed a real RSS route, it correctly fetched and parsed 76 items, applied knowable-date gating, and scored them. The PIT discipline, the `defusedxml` hardening, and the pluggable scorer are exactly what #172 asked for, and the engine/runner wiring mirrors the fundamentals path cleanly.

But it can't merge as-is: the default feed catalogue doesn't survive an open-box test. Three blocking items, all verified against a live instance:

**1. `DEFAULT_FEEDS` routes don't exist in RSSHub**

`/stock/notice/{code}` and `/stock/news/{code}` aren't registered routes — there is no `stock` namespace in RSSHub's route tree (checked `lib/routes/` on DIYgod/RSSHub), and a healthy local instance rejects them immediately:

```
info: --> GET /stock/notice/600519 503 16ms
```

End-to-end with the shipped defaults: `query_events(["600519.SH"], as_of=now)` against that instance returns **0 rows after ~12s/symbol of retry backoff**, and a backtest with `event_feeds` set then runs with `event_score` silently all-zero.

**2. Per-symbol routes need code-format conversion**

The realistic per-symbol candidates (e.g. `/xueqiu/stock_info/:id`, default type = announcements) take exchange-prefixed codes (`SH600519`), while the provider formats backtest codes (`600519.SH`) straight into the template — so even with correct route names the default catalogue can't resolve. (Note xueqiu routes also need `XUEQIU_COOKIES` configured on the instance; worth documenting for whatever route you settle on. An alternative worth considering: ship **no** default catalogue and require explicit `FeedSpec`s/config, so there's nothing to silently rot as RSSHub routes evolve.)

**3. Fetch failures are swallowed silently**

`_fetch_feed` catches the budget `TimeoutError` and returns `""` with no logging. Combined with (1), the failure mode is the worst kind: the user believes sentiment is factored into the backtest, but every score is 0.0. At minimum `logger.warning` per failed feed/code; better, have `_maybe_enrich_events` raise when every requested feed returned nothing — it already fail-fasts on a missing base URL, so "URL set but every feed unreachable" passing silently is inconsistent with its own philosophy.

Happy to re-review quickly once these are addressed — the provider core itself I'd consider verified.

warren618:
Landed on `main` 🎉 — thank you @Shizoqua; the provider core was well-built and shipped largely as-is (`14ca60b`). The three open-box blockers from the review are resolved:

- **No built-in feed catalogue**: the default `/stock/notice` & `/stock/news` routes don't exist in RSSHub, so `DEFAULT_FEEDS` is now empty and feeds are declared explicitly in config (`event_feeds` = list of `{name, route_template, event_type, code_style}`) — nothing to silently rot as routes evolve.
- **Per-symbol code formatting**: a feed declares a `code_style` (`raw`/`bare`/`exchange_prefix`), so `600519.SH` becomes e.g. `SH600519` for xueqiu-style routes.
- **No silent all-zero scores**: fetch failures are logged per feed/code and `query_events` raises `EventProviderError` when every fetch failed (configured-but-unreachable), while a reachable-but-empty feed stays a legitimate empty result.
- Your PIT discipline, `defusedxml` hardening, pluggable scorer and engine/runner wiring are unchanged; +12 tests cover the new paths.

Thanks for the solid core!

PR #206, opened 2026-06-12
feat(agent): complete trace.jsonl overhaul — zero truncation, offload, message entries, iter consistency

zwrong:
## Summary

- Comprehensive trace.jsonl improvements: zero truncation, message entries, tool result offload, session persistence, iter consistency
- Every entry now carries `iter` (immediately after `type`), and iter increments continuously across resume turns

## Why

The current trace.jsonl on `main` has several limitations:

1. **Tool result truncated** — results are stored as `result[:200]` preview only; full content is lost
2. **Entry content truncated** — prompt, thinking, answer, args, and summary are all truncated by `[:N]`, making replay incomplete
3. **No message-type entries** — user/assistant messages are not recorded, so the conversation flow cannot be fully reconstructed
4. **start/message/end missing iter** — multi-turn sessions cannot attribute entries to specific turns; start and end lack `iter` (error end only has `iterations`)
5. **Session traces in runs/** — multi-turn session traces are scattered across individual `runs/` directories, not aggregated by session
6. **Iter resets on resume** — the local `iteration` variable starts from 0 on each `run()` call, breaking the iter continuity across resumed sessions

## Changes

**TraceWriter core (`agent/src/agent/trace.py`)**
- New `write_tool_result()` method: results ≤50K are inlined in full; larger results offload to `tool-results/<call_id>.txt` with path + 500-char preview + original size in the entry
- `read()` auto-resolves offloaded results, returning self-contained entries
- New `find_trace_dir()` static method: `sessions/` first → `runs/` fallback
- Constructor parameter renamed `run_dir` → `dir_path` (unified for session and run directories)

**Agent loop (`agent/src/agent/loop.py`)**
- Trace location: `sessions/<session_id>/trace.jsonl` (when session_id is valid); `-p` one-off mode still uses `runs/`
- Removed all `[:N]` truncation (prompt, thinking, answer, args, summary)
- Added message-type entries (user/assistant) for full conversation replay
- Unified iteration counter via `self._run_iteration` (persists across `run()` calls)
- Resume iter recovery moved **before** `start`/`message` writes
- `end` entry now carries `iter` (alongside the existing `iterations` count)
- `auto_compact` transcript path uses `trace.dir_path` instead of `run_dir`

**CLI compatibility (`agent/cli/_legacy.py`)**
- `_build_history_from_trace`, `cmd_show`, `cmd_trace`, `cmd_continue` now use `find_trace_dir()` to locate trace
- `cmd_trace` display supports new fields: `result_path`, `result_preview`, `result_size`

**Tests (`agent/tests/test_trace_writer.py`, 398 lines, 21 tests)**
- Write/read, offload threshold boundaries (just above/below 50K), mixed mode
- Unicode, empty directory, malformed line tolerance, non-existent path
- `find_trace_dir()` sessions-first / runs-fallback / both-missing returns None
- Entry field order consistency (type → iter) — covers all 10 entry types

## Test Plan

- [x] Existing tests pass (3114 passed, 2 skipped; 1 failure is a pre-existing hypothesis CLI color issue on main)
- [x] New tests added (21 trace-specific tests covering inline/offload/find/resume/field-order)
- [ ] Tested manually

## Checklist

- [ ] No changes to protected areas without prior discussion
- [ ] No hardcoded values
- [ ] Documentation updated (internal trace format change, not directly user-facing)


warren618:
Thanks @zwrong — this was a really valuable trace overhaul.

The core contribution here is yours: preserving full trace content, adding message entries, making session traces discoverable, keeping iter values consistent, and offloading large tool results instead of truncating them.

I pushed a maintainer follow-up directly to this PR branch so we could rebase it onto the current `main` without asking you to redo the conflict resolution. That follow-up keeps the recent provider retry / timeout / reasoning-stream protections from `main`, and adds a few safety hardenings around sidecar paths, selective offload resolution, and structured trace redaction.

Merged this via a merge commit so your original commits and PR authorship are preserved. Thanks again for the contribution!

PR #207, opened 2026-06-12
[codex] Persist AgentLoop usage artifacts

LemonCANDY42:
## Summary
- Persist provider-reported AgentLoop token usage to `artifacts/llm_usage.json` for each run.
- Expose the persisted usage payload from `/runs/{id}` as `llm_usage`.
- Show historical Agent Usage in Run Detail and treat usage-only runs as report-worthy.

## Why
The agent already emits live `llm_usage` SSE events and the CLI rail can accumulate usage during an active session, but the data disappears after the run. Persisting a compact artifact makes cost and usage auditing possible from historical run artifacts without recording prompts, responses, API keys, or hidden reasoning.

## Notes
- Uses provider `usage_metadata` only; no token estimation is persisted.
- Writes nothing when the provider omits usage metadata.
- This is a general AgentLoop observability change and does not include Moirix, MiniMax, IBKR, or authority-guard changes.

## Validation
- `pytest agent/tests/test_loop_helpers.py agent/tests/test_agent_loop_terminal_state.py agent/tests/test_run_card.py -q`
- `npm test -- --run src/lib/__tests__/runReports.test.ts`
- `npm run build`
- `git diff --cached --check`

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -114,6 +114,7 @@ class RunResponse(BaseModel):
     metrics: Optional[BacktestMetrics] = Field(None, description="Backtest metrics")
     artifacts: List[Artifact] = Field(default_factory=list, description="Run artifacts")
     run_card: Optional[Dict[str, Any]] = Field(None, description="Trust Layer run card payload")
+    llm_usage: Optional[Dict[str, Any]] = Field(None, description="Persisted AgentLoop LLM usage summary")
 
     equity_curve: Optional[List[Dict[str, Any]]] = Field(None, description="Equity preview")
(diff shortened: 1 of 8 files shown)


warren618:
Thanks for the contribution. I'm closing this for now as a maintainer queue/scope decision.

This changes AgentLoop/core run artifact behavior and is currently conflicted with \. For core observability or usage-persistence changes, please open a small design issue first so we can agree on the data contract and where the artifact should live before code is submitted.

Smaller PRs after the contract is agreed will be much easier for us to review.

warren618:
Clarification: the previous closing note should say this PR is currently conflicted with `main`.

PR #209, opened 2026-06-12
Add home run library dashboard

LemonCANDY42:
## Summary
- replace the home-only landing view with a lightweight research workbench
- load recent runs from `/runs`, fetch displayed run details, and show status, metrics, artifact/report badges, and links to Run Detail / Compare
- add empty and unavailable states so the page stays useful when no runs or no backend are available

## Scope
- upstream-friendly frontend only
- no Moirix, IBKR, broker, live-order, MiniMax, or usage-dashboard behavior
- no backend contract changes

## Tests
- `npm run test:run -- src/pages/__tests__/Home.test.tsx src/lib/__tests__/runReports.test.ts`
- `npm run build`
- `git diff --check`
- local browser smoke on `http://127.0.0.1:5907/` and `/runs/safe_run`

### DIFF
--- a/frontend/src/pages/Home.tsx
+++ b/frontend/src/pages/Home.tsx
@@ -1,7 +1,31 @@
+import { useEffect, useState } from "react";
 import { Link } from "react-router-dom";
-import { ArrowRight, Bot, BarChart3, Zap, UserCircle2 } from "lucide-react";
+import {
+  ArrowRight,
+  BarChart3,
+  Bot,
+  CheckCircle2,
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for the contribution. I'm closing this for now because it overlaps with the newer run/report-library direction and changes the Home page into a dashboard, which is a product decision we should not take through an unsolicited PR.

If we revisit this area, we'd prefer starting from a design issue and keeping the Home page separate from any report/run library surface.

PR #212, opened 2026-06-12
Optimize Run Detail chart payloads

LemonCANDY42:
## Summary
- add a `chart_symbol` query parameter for `GET /runs/{id}` so Run Detail can request one symbol's chart payload at a time
- return `chart_symbols` for symbol discovery while avoiding full equity/trades artifact CSV payloads in run detail responses
- update Run Detail with symbol selection, additive multi-select, progressive Load all, cancelable progress, and bounded chart rendering

## Tests
- `uv run --extra dev python -m pytest tests/test_run_card.py -q`
- `npm run test:run -- src/pages/__tests__/RunDetail.test.tsx`
- `npm run build`
- `git diff --check`

## Notes
- This PR is upstream-friendly and does not include Moirix, MiniMax, IBKR, Reports Library, or Agent Usage Dashboard changes.

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -126,6 +126,7 @@ class RunResponse(BaseModel):
     run_directory: str = Field(..., description="Run directory path")
     run_stage: Optional[str] = Field(None, description="UI-facing run stage")
     run_context: Optional[Dict[str, Any]] = Field(None, description="Normalized request context")
+    chart_symbols: List[str] = Field(default_factory=list, description="Symbols available for chart loading")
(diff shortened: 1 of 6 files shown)


warren618:
Thanks for looking at Run Detail payload size. The performance concern is valid, but I'm closing this implementation for now because it changes the default \ response semantics and risks breaking existing consumers that expect the full run detail payload.

If you'd like to continue this work, please open a design issue or a smaller non-breaking PR that preserves the default response and adds an explicitly opt-in payload mode.

warren618:
Clarification: the previous closing note should say this changes the default `/runs/{id}` response semantics.

Issue #214, opened 2026-06-12
Design: AgentLoop usage artifact contract


LemonCANDY42:
## Problem
AgentLoop currently streams provider usage live, but historical runs do not have a stable, reviewable usage artifact contract. This makes run-level cost/usage auditing difficult and makes dashboard work depend on ephemeral SSE state.

## Proposed contract
- Persist provider-reported usage only; do not estimate prices or tokens.
- Write a small JSON artifact under a maintainer-approved location, for example `artifacts/llm_usage.json` or another agreed run-metadata path.
- Include provider, model, input/output/total tokens, calls, per-iteration deltas, and `updated_at`.
- Keep prompt/content out of the artifact.
- Expose it through run detail and, optionally, compact run summaries only after the artifact location/schema is agreed.

## Constraints
- Must not change normal AgentLoop behavior when providers omit usage metadata.
- Must preserve existing live `llm_usage` SSE behavior.
- Must be backward compatible for existing run artifacts.

## Acceptance criteria for a future PR
- Documented schema and artifact location.
- AgentLoop test proving usage is persisted from provider metadata.
- API test proving the artifact is surfaced without requiring a full expensive run-detail scan.
- Frontend dashboard PR submitted separately after the contract lands.

## Prior closed PRs
- #207
- #211

LemonCANDY42:
Thanks for the clarification. That makes sense.

I agree this should be discussed as a small contract first, especially because it touches AgentLoop/core run artifacts. The intent is not to add provider-specific behavior or pricing logic, just to persist provider-reported usage in a lightweight run-scoped artifact so past runs can be audited after the live stream is gone.

A reasonable contract could be:
- run-scoped only, e.g. artifacts/llm_usage.json
- provider-reported token usage only
- no prompt/content capture
- no price estimation in the core artifact
- additive API/UI exposure after the schema is agreed

I’ll wait for maintainer direction here before opening a smaller PR based on latest main.

warren618:
Approved in principle — this is additive and the constraints you listed are exactly right. Two specifics to lock before you open the PR:

**Artifact location:** don't introduce a new top-level `artifacts/` tree. Agent runs already get a per-run directory via `state_store` (the same `run_dir` that holds the request/state files, under the runs root). Persist there as `<run_dir>/llm_usage.json` so it follows the existing run-artifact convention.

**Schema** (provider-reported only, no prompt/content, no price):
```json
{
[...]

PR #223, opened 2026-06-14
feat(agent): persist AgentLoop usage artifacts

LemonCANDY42:
## Summary

Persist provider-reported AgentLoop usage as a lightweight run-scoped artifact and expose it on the existing single-run response.

This follows the contract discussed in #214:
- writes `<run_dir>/llm_usage.json`
- records provider/model, aggregate totals, per-iteration usage, and `updated_at`
- only writes when the provider returns `usage_metadata`
- does not capture prompts/content
- does not estimate prices
- does not change the live `llm_usage` SSE event

## Validation

- `python -m pytest agent/tests/test_agent_loop_terminal_state.py agent/tests/test_run_card.py -q`
- `python -m py_compile agent/src/agent/loop.py agent/api_server.py`
- `git diff --check`

Refs #214.

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -114,6 +114,7 @@ class RunResponse(BaseModel):
     metrics: Optional[BacktestMetrics] = Field(None, description="Backtest metrics")
     artifacts: List[Artifact] = Field(default_factory=list, description="Run artifacts")
     run_card: Optional[Dict[str, Any]] = Field(None, description="Trust Layer run card payload")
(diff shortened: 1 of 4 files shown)


warren618:
Merged 🎉 Thanks @LemonCANDY42 — clean implementation that followed the #214 contract exactly: run-scoped `llm_usage.json`, provider-reported tokens only, behind the existing usage guard, live SSE untouched. Appreciated the tight diff and the two tests.

Issue #215, opened 2026-06-12
Design: opt-in Run Detail chart payload mode


LemonCANDY42:
## Problem
Large Run Detail pages can become slow when `/runs/{id}` returns all chart symbols, all price bars, and all trade markers by default. The performance concern is real, but a previous PR changed default response semantics and was closed as too risky for existing consumers.

## Proposed non-breaking direction
- Preserve the current default `/runs/{id}` response semantics.
- Add an explicitly opt-in payload mode, for example:
  - `GET /runs/{id}?chart_symbol=SYMBOL&chart_payload=selected`, or
  - `GET /runs/{id}/chart?symbol=SYMBOL`.
- Return `chart_symbols` for discovery in the opt-in mode.
- Let the frontend request one symbol at a time and optionally load more symbols progressively.
- Avoid changing existing consumers that expect the full run detail payload.

## Frontend behavior to discuss
- Default to one selected chart for responsiveness.
- Allow manual multi-select and an explicit Load all action.
- Show progress and allow cancel when loading many symbols.
- Render loaded charts in visible batches rather than instantiating all chart canvases at once.

## Acceptance criteria for a future PR
- Backward-compatible default endpoint behavior.
- Dedicated opt-in endpoint/query covered by backend tests.
- Run Detail frontend test covering add-symbol and progressive load-all behavior.
- Measured payload reduction documented in the PR.

## Prior closed PR
- #212

LemonCANDY42:
Thanks, that’s fair. I agree the default /runs/{id} response should stay backward compatible.

I pushed an update to the branch that keeps the default response unchanged and moves the Run Detail UI to an explicit progressive-loading path:
- default /runs/{id} remains the full payload for existing consumers
- chart_payload=summary is opt-in for a lightweight first render
- chart_symbol=... is opt-in for selected-symbol chart payloads
- the frontend uses that opt-in path to show the run summary first, then load chart data progressively with loading state/progress

So the optimization is non-breaking: existing consumers keep the current full default response, while the Run Detail UI can choose the progressive path.

warren618:
The non-breaking approach is right and the default `/runs/{id}` response must stay byte-for-byte unchanged. Confirming the contract so we're aligned before the PR:

- Default `/runs/{id}` (no new params) → unchanged full payload. Existing consumers untouched.
- `?chart_payload=summary` → metadata + a `chart_symbols` discovery list, **no** price bars / trade markers.
[...]

PR #225, opened 2026-06-14
perf(run-detail): add opt-in chart payloads

LemonCANDY42:
## Summary

Add opt-in chart payload modes for the Run Detail endpoint and UI.

This follows the contract discussed in #215:
- default `/runs/{id}` stays on the existing full response path
- `?chart_payload=summary` returns metadata plus `chart_symbols`, with no chart bars or trade markers
- `?chart_symbol=SYM` returns that symbol's chart bars and markers only
- the Run Detail page first renders summary metadata, then progressively loads selected symbols
- chart controls and progress/loading copy are localized in English and zh-CN

## Validation

- `python -m pytest agent/tests/test_run_card.py -q`
- `python -m py_compile agent/api_server.py agent/src/ui_services.py`
- `npm run build`
- `git diff --check`

Refs #215.


warren618:
Merged 🎉 Thanks @LemonCANDY42! The non-breaking opt-in design from #215 came through cleanly. Two heads-ups: I merged main in to resolve a test-file conflict with #223, and added a small follow-up commit so the bare default `/runs/{id}` path skips chart-symbol discovery (it was doing a redundant scan whose result got discarded). Nice work on the summary-first progressive loading.

PR #226, opened 2026-06-14
fix(loaders): include requested yfinance end date

gyx09212214-prog:
## Summary

Include the requested `end_date` when loading daily yfinance data.

## Why

The project-facing loader API accepts a `start_date` / `end_date` range that is expected to include the requested end date. yfinance's `download(end=...)` parameter is exclusive, so passing the requested `end_date` directly can omit the final trading day.

## What changed

- Adjust yfinance provider calls to pass `end_date + 1 day`.
- Preserve cache lookups and writes using the original requested date range.
- Apply the same boundary adjustment to the fallback single-symbol download path.

## Tests

- `python -m pytest agent/tests/test_yfinance_crypto.py -q`
- `python -m py_compile agent/backtest/loaders/yfinance_loader.py agent/tests/test_yfinance_crypto.py`
- `git diff --check`

## Risk notes

This only changes the yfinance provider call boundary. Cache keys remain based on the user-requested range, and the tests monkeypatch the download path without making network calls.

### DIFF
--- a/agent/backtest/loaders/yfinance_loader.py
+++ b/agent/backtest/loaders/yfinance_loader.py
@@ -68,6 +68,11 @@ def _to_yfinance_interval(interval: str) -> str:
     return _INTERVAL_MAP.get(normalized, normalized.lower())
 
 
+def _to_yfinance_exclusive_end(end_date: str) -> str:
+    """Convert the project-inclusive end date to yfinance's exclusive end."""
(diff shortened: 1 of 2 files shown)


warren618:
Merged 🎉 Thanks @gyx09212214-prog, and welcome — great first contribution! Correct catch on yfinance's exclusive `end` boundary, and I like that the cache keys stay on the original requested range. Verified the +1-day boundary end-to-end before merging.

PR #227, opened 2026-06-14
fix(loaders): tolerate malformed crypto timeout env vars

gyx09212214-prog:
## Summary

Make crypto loader timeout environment variables robust to malformed values.

## Why

Currently malformed values such as `CCXT_TIMEOUT_MS=abc` or `OKX_TIMEOUT_S=abc` can fail while importing the loader modules. That makes configuration mistakes harder to recover from and can prevent the app from starting.

## What changed

- Added shared positive numeric env helpers for loader configuration.
- Invalid, blank, or non-positive timeout/budget values now log a warning and fall back to existing defaults.
- Valid values are still honored for CCXT and OKX loaders.

## Tests

- `python -m pytest agent/tests/test_ccxt_loader_bounded.py agent/tests/test_okx_loader_bounded.py -q`
- `python -m pytest agent/tests/test_loader_retry_helpers.py -q` for the relevant retry helper tests
- `python -m py_compile agent/backtest/loaders/base.py agent/backtest/loaders/ccxt_loader.py agent/backtest/loaders/okx.py agent/tests/test_ccxt_loader_bounded.py agent/tests/test_okx_loader_bounded.py`
- `git diff --check`

## Risk notes

This keeps the existing default timeout and fetch budget values. The change only affects malformed or non-positive env values, which now degrade to defaults instead of failing at import time.

### DIFF
--- a/agent/backtest/loaders/base.py
+++ b/agent/backtest/loaders/base.py
@@ -56,6 +56,38 @@ DEFAULT_BACKOFF: tuple[float, ...] = (0.5, 1.5, 4.0)
 DEFAULT_MAX_RETRIES = 3
 
 
+def positive_env_int(name: str, default: int) -> int:
+    """Read a positive integer env var, warning and falling back on invalid values."""
+    raw = os.getenv(name)
+    if raw is None or not raw.strip():
+        return default
+    try:
+        value = int(raw)
+    except ValueError:
+        logger.warning("invalid %s=%r, using default %s", name, raw, default)
+        return default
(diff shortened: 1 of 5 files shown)


warren618:
Merged 🎉 Thanks again @gyx09212214-prog! This one actually prevented an import-time crash — I reproduced `OKX_TIMEOUT_S=abc` raising `ValueError` at import on `main`, and confirmed your shared `positive_env_*` helpers degrade it to a warning + default. Clean, well-tested fix.

PR #230, opened 2026-06-14
fix(baostock): support baostock-native code format (sh.601398)

bhlt:
The baostock loader previously only accepted tushare-style codes (601398.SH, 000001.SZ) but rejected baostock-native codes (sh.601398, sz.000001) — which is the format baostock uses in its own documentation and examples.

This caused users passing sh.601398 to:
- Have _is_a_share() return False → _fetch_one() immediately returns None
- Have _fetch_one() split "sh.601398" as ["SH", "601398"] then walk the dead branch (suffix "601398" never matches "SH" or "SZ")

Fix:
- _is_a_share() now accepts both formats (case-insensitive)
- _fetch_one() detects baostock-native format first and passes it through unchanged, only converting tushare-style suffixes

Tests:
- New test file: agent/tests/test_baostock_loader.py
- 10 unit tests covering both code formats and conversion logic
- All 10 pass

Closes: #TBD

## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/backtest/loaders/baostock_loader.py
+++ b/agent/backtest/loaders/baostock_loader.py
@@ -21,7 +21,12 @@ logger = logging.getLogger(__name__)
 
 
 def _is_a_share(code: str) -> bool:
-    return code.upper().endswith((".SZ", ".SH"))
+    # Support both baostock native format (sh.601398) and tushare-style suffix (601398.SH)
+    code_lower = code.lower()
+    return (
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks @bhlt! Clean fix with solid test coverage. The native `sh.601398` / `sz.000001` format now works alongside the tushare-style `601398.SH`. 🙏

Issue #229, opened 2026-06-14
[Bug] Web: 切换页面停止运作 + 运行中点停止无效


warren618:
> Migrated from Discussion #228 — reported by @kalkinj

## 原始反馈(@kalkinj)
> 1. 切换页面经常就停止运作
> 2. ai助手运行过程中点击停止无效

## 需要补充的信息(便于复现定位)

**环境**
- 用的是 **Web 前端** 还是 **CLI(`vibe-trading`)**?
- 版本:CLI 跑 `vibe-trading --version`,或 `pip show vibe-trading-ai`;Web 的话另外告诉我们浏览器。

**1. 切换页面就停止运作**
Agent 的运行其实是在**服务端**跑的,前端只是通过 SSE 实时订阅进度。切换页面会断开这个订阅,但服务端的运行**本身不会停**。所以想确认:切回该会话页面后,对话是**继续往下走 / 补上了刚才的进度**,还是**一直卡住不动**?
- 切回来能继续 → 前端重连/回放的体验问题,我们来修;
- 切回来彻底没反应 → 可能是另一类问题,麻烦贴一下切换前最后停在哪一步。

**2. 运行中点停止无效**
停止是给当前循环发一个取消信号,会在**下一个步骤边界**生效。如果此刻 Agent 正卡在一次很慢/卡住的模型调用里,取消要等这次调用返回才会落地——表现就像"点了没反应"。想确认:点停止后是**完全不停**,还是**过一会儿才停**?如果是前者,大概率撞上了我们已知的"首个模型调用卡死"问题。

如果方便,Web 端按 F12 → Console 和 Network(看 `/events`、`/cancel` 这两条请求)截个图,我们复现会快很多 🙏

cc @kalkinj

warren618:
Both symptoms are fixed and merged to `main`:

**1. 切换页面就停止运作 — fixed in #234.** The run executes server-side and the page only subscribes via SSE; navigating away tore down that stream, and the page never re-subscribed when you returned to the same session, so progress stopped arriving and the run looked frozen. The frontend now reconnects on return and replays missed progress from the backend's event buffer (and refreshes history for a run that finished while you were away).

[...]

PR #236, opened 2026-06-15
fix(frontend): sync completed chat attempts

sambazhu:
## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/frontend/src/pages/Agent.tsx
+++ b/frontend/src/pages/Agent.tsx
@@ -407,6 +407,37 @@ export function Agent() {
     }
   }, [forceScrollToBottom]);
 
+  const refreshSessionMessages = useCallback(async (sid: string) => {
+    const gen = genRef.current + 1;
+    genRef.current = gen;
+    await loadSessionMessages(sid, gen);
+  }, [loadSessionMessages]);
+
+  const syncCompletedAttempt = useCallback(async (sid: string, attemptId?: string) => {
+    if (!attemptId) return false;
+    for (let i = 0; i < 3; i += 1) {
+      try {
+        const storedMessages = await api.getSessionMessages(sid);
+        const completed = storedMessages.some(
+          (message) => message.role === "assistant" && message.linked_attempt_id === attemptId,
+        );
+        if (completed) {
+          if (act().sessionId !== sid) return true;
+          setReasoningActive(false);
+          act().clearStreaming();
+          act().setStatus("idle");
+          useAgentStore.setState({ toolCalls: [] });
+          await refreshSessionMessages(sid);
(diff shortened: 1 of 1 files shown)


warren618 (commented):
Thanks @sambazhu — this is a useful, orthogonal addition. For context, two fixes for the same issue (#229) just merged: #234 (re-subscribe + replay SSE when returning to a running session) and #235 (responsive cancel). This PR targets a third, distinct gap — the UI staying in "streaming" after a run already committed because the SSE `attempt.completed` was missed (the long-standing replay race) — so it composes rather than overlaps. Verified the contract holds: `sendMessage` returns `attempt_id` (service emits `attempt.created` then runs in the background) and the assistant reply is stored with `linked_attempt_id`, so the poll key is correct.

A few things before merge:

1. **Make it a true fallback (avoid firing on the happy path).** `syncCompletedAttempt` runs after *every* send and polls `getSessionMessages` 3× even when SSE finalized the turn normally; on a sub-2.4s run it then re-forces `idle` + a second `getSessionMessages` refresh. Could you gate the reconcile on the UI still being mid-stream — e.g. only `clearStreaming()/setStatus("idle")/refresh` when `act().status === "streaming"` (and skip the poll loop entirely once it's already `idle`)? That keeps it a no-op whenever SSE worked.

2. **Window scope.** The 3×800ms (~2.4s) window only catches attempts that finish within it; a longer run whose completion event is genuinely missed stays stuck. Is the 2.4s the scenario you actually hit (a fast turn)? If so, a one-line note in the PR body is enough; if you meant to cover longer runs too, the reconnect/replay path from #234 is probably the better hook than a fixed poll budget.

3. **Note `attempt.completed` vs failed/cancelled.** A failed or cancelled attempt also appends an assistant reply with `linked_attempt_id`, so this will reconcile those too — which is fine (and desirable), just confirming it's intended.

4. **Please fill in the PR description** (what scenario/repro this fixes) and **add a frontend test** (vitest) for the sync path — e.g. a completed attempt present in `getSessionMessages` while the store is still `streaming` → status flips to `idle` and history refreshes. That locks the behavior in.

Happy to merge once the fallback guard + a test are in. 🙏

warren618:
Merged, thanks! This is the SSE-completion-drop sync (poll for the persisted assistant message after `sendMessage`, then reconcile to idle). Heads up: the raw 2-way diff looked like it deleted the #234 reconnect block, but that's only because the branch forked from an older `main` — the 3-way merge preserved #234. Verified `tsc` + `vite build` + full vitest suite green on the merged tree. A one-line PR description would help reviewers next time. 🙏

PR #238, opened 2026-06-15
fix(run-card): write strict JSON for non-finite metrics

gyx09212214-prog:
﻿## Summary

Make backtest run-card JSON strict when metrics contain non-finite floating-point values.

## Why

Python's default `json.dumps` can emit `NaN`, `Infinity`, and `-Infinity`, which are not valid JSON tokens for strict parsers. Backtest metrics can naturally produce non-finite values for degenerate or insufficient-return windows, so run-card artifacts should normalize those values before writing `run_card.json` and the Markdown companion.

## What Changed

- Normalize non-finite float values to `null` before serializing the run card.
- Write `run_card.json` with `allow_nan=False` so future regressions fail fast.
- Add a regression test covering scalar metrics and nested validation payloads.

## Validation

- `python -m pytest agent/tests/test_run_card.py agent/tests/test_run_card_strict_json.py -q`
- `python -m py_compile agent/backtest/run_card.py agent/tests/test_run_card_strict_json.py`

## Risk Notes

This keeps the change scoped to run-card artifact serialization. Finite metric values are preserved, while only non-finite values are represented as `null` in JSON/Markdown output.

### DIFF
--- a/agent/backtest/run_card.py
+++ b/agent/backtest/run_card.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+import math
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Mapping, Sequence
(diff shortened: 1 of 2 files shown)


warren618:
Merged, thanks! Strict-JSON run cards (non-finite floats → `null`, `allow_nan=False`) is a clean fix, and the sanitizer correctly covers the nested validation payload too. CI green.

PR #239, opened 2026-06-15
test(search): cover ddgs retry fallback behavior

gyx09212214-prog:
## Summary
- add regression coverage for the current `web_search` ddgs multi-engine behavior
- cover transient retry with the configured backend list
- cover compatibility when ddgs/duckduckgo_search rejects the `backend` kwarg
- cover `No results found` as an empty successful response

## Validation
- `python -m pytest agent/tests/test_security_scanner.py -q`
- `python -m py_compile agent/src/tools/web_search_tool.py agent/tests/test_security_scanner.py`

## Notes
- Narrowed to tests after rebasing onto upstream `main`, which already includes the multi-engine implementation from #232.

### DIFF
--- a/agent/tests/test_security_scanner.py
+++ b/agent/tests/test_security_scanner.py
@@ -95,6 +95,104 @@ def test_web_search_adds_security_warning_to_snippets(
     assert result["security_warnings"][0]["field"] == "results.0.snippet"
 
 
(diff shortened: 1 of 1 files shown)


warren618:
Merged, thanks for the regression coverage on the ddgs multi-engine retry/fallback paths — good follow-up to #232.

PR #240, opened 2026-06-15
fix(loaders): tolerate malformed RSSHub timeout env vars

gyx09212214-prog:
﻿## Summary

Make RSSHub event-provider timeout and fetch-budget environment variables robust to malformed values.

## Why

The crypto loaders already fall back cleanly when timeout env vars are blank, invalid, or non-positive. RSSHub event loading still parsed `RSSHUB_TIMEOUT_S` and `RSSHUB_FETCH_BUDGET_S` with direct `float(...)`, so a typo such as `RSSHUB_TIMEOUT_S=abc` could raise before the provider attempted a fetch.

## What Changed

- Reuse the shared positive float env helper for RSSHub timeout and budget settings.
- Add regression coverage for malformed RSSHub timeout and budget env values.
- Preserve existing defaults and fetch behavior when env vars are valid.

## Validation

- `python -m pytest agent/tests/test_rsshub_events_provider.py agent/tests/test_rsshub_events_lookahead.py -q`
- `python -m py_compile agent/backtest/loaders/rsshub_events.py agent/tests/test_rsshub_events_provider.py`
- `git diff --check`

## Risk Notes

This keeps the change scoped to RSSHub configuration parsing. Valid positive env values are still honored; malformed, blank, or non-positive values now fall back to existing defaults with the shared warning path.

### DIFF
--- a/agent/backtest/loaders/rsshub_events.py
+++ b/agent/backtest/loaders/rsshub_events.py
@@ -33,7 +33,7 @@ import pandas as pd
 from defusedxml import ElementTree as ET
 from defusedxml.common import DefusedXmlException
 
-from backtest.loaders.base import retry_with_budget
+from backtest.loaders.base import positive_env_float, retry_with_budget
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


warren618:
Merged, thanks! Reusing `positive_env_float` for the RSSHub timeout/budget envs matches the crypto-loader pattern. CI green.

Issue #231, opened 2026-06-14
[Bug]


Ethan-sun01:
### Description

**Web Search failure**
when I raise any questions, I constantly run into web seach failure. is it my issue? How can I fix it?
<img width="1025" height="580" alt="Image" src="https://github.com/user-attachments/assets/0fcaf9e4-d6c4-4b82-ab22-2aba9ffbb161" />

Thanks& regards

### Steps to Reproduce

1. Start Vibe-Trading Web UI.
2. Enter a prompt that requires web search, e.g.
   "search me AAPL recent release"
   or
   "Find the latest NVIDIA news."
3. Run the task.
4. Observe the execution steps.

Expected:
- web_search completes successfully.

Actual:
- web_search shows a red ❌ failure.
- Read webpage steps succeed.
- Final answer is still generated.
- Run status is SUCCESS.

### Error Logs

```shell

```

### Interface

Web UI (vibe-trading serve)

warren618:
Hi @Ethan-sun01 — thanks for the clear report and the screenshot. This isn't your setup; it's a known weakness in how `web_search` was wired, and it's now fixed.

**Root cause.** `web_search` used `ddgs`'s default `auto` backend, which leads with DuckDuckGo. DuckDuckGo rate-limits aggressively from cloud / shared IPs, so when it threw, the whole tool returned ❌ — even though your run still finished, because the agent fell back to reading pages directly (that's why `read_url` succeeded and a final answer was produced, and the run status stayed SUCCESS).

**Fix ([#232](https://github.com/HKUDS/Vibe-Trading/pull/232), merged to `main`).** `web_search` now queries several free, no-key engines in order — DuckDuckGo, Google, Bing, Brave, Mojeek, Yahoo — so one throttled engine no longer fails the search. It also retries transient failures with backoff, treats "no results" as an empty (not failed) answer, and returns a clear, actionable message instead of a bare ❌ when every engine is rate-limited.

**Get it now**, before the next release:
```
pip install "git+https://github.com/HKUDS/Vibe-Trading.git"
```
After updating you can also pin specific engines if you ever want to skip DuckDuckGo entirely — in your `.env`:
```
VIBE_TRADING_SEARCH_BACKENDS=google, bing
```
(That env knob only exists in the fix, so it has no effect on the released version — installing from `main` is what brings the fix in.)

Closing as fixed — please reopen if you still hit failures across *all* engines after updating. 🙏

Ethan-sun01:
[...]

PR #255, opened 2026-06-18
Copilot/add tavily source

panjunchang:
## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/.env.example
+++ b/agent/.env.example
@@ -110,6 +110,13 @@ TUSHARE_TOKEN=your-tushare-token
 # FUTU_HOST=127.0.0.1
 # FUTU_PORT=11111
 
+# Web search: Tavily (non-preferred fallback, requires tavily-python package)
+# Free engines (DuckDuckGo, Google, Bing, etc.) are always tried first.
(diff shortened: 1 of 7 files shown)


warren618:
Thanks for the PR, and for the thoughtful Tavily-preferred / ddgs-fallback design — it's a clear improvement over the earlier attempt.

We're closing this one for the same reason we closed #75: web_search already runs as a keyless multi-engine aggregator via ddgs (DuckDuckGo, Google, Bing, Brave, Mojeek, Yahoo), hardened in #231/#232 with an ordered backend list, retry/backoff, and the VIBE_TRADING_SEARCH_BACKENDS override. We want web search to keep working out of the box with no paid API key, so we'd prefer not to add tavily-python as a dependency or a Tavily key as a required-for-best-results setting.

This isn't a quality issue with your code — it's a standing decision about not taking on a paid search dependency. If you think there's a gap the current ddgs path doesn't cover (specific queries where it consistently fails), we'd genuinely welcome an issue with those cases so we can improve the free path. Thanks again for contributing.

PR #267, opened 2026-06-19
feat(autopilot): scaffold signal engine + auto-link backtest metrics (Phase 3)

Robin1987China:
## Summary

Research Autopilot Phase 1 (#260) bridged hypothesis → goal and hypothesis → config.json, but left the rest of the loop wired by hand. Two manual steps remained the most error-prone parts of an autopilot run:

1. Writing `code/signal_engine.py` to the runner contract (a no-arg-constructible `SignalEngine` with `generate(self, data_map)`). Getting it wrong is the most common reason a backtest fails its pre-flight check.
2. Recording results — `link_backtest` needs the agent to open `run_card.json` and hand-extract the metrics dict.

This PR adds two deterministic, backend-only tools that close the loop.

## Changes

- `agent/src/tools/autopilot_tool.py`:
  - **`scaffold_signal_engine`** — writes a contract-correct `code/signal_engine.py` stub with a runnable flat (no-position) default and the hypothesis `signal_definition` embedded as a docstring for the agent to implement against. Refuses to overwrite an existing engine unless `overwrite=true`.
  - **`link_autopilot_backtest`** — reads `run_card.json` from a run dir, extracts the scalar `metrics`, and links them to the hypothesis via the existing `link_backtest`, returning the metrics for thesis evaluation. Degrades to empty metrics with a warning if the key is absent.
  - Both resolve `run_dir` through `safe_run_dir`, inheriting the allowed-run-root boundary (fail-closed on path traversal, consistent with the backtest tool).
- `agent/tests/test_autopilot_phase3.py`: 12 tests.

## Out of scope (deliberate)

- Compiling the free-text `signal_definition` into working signal logic (LLM territory; the stub is a starting point only).
- Auto-invoking the `backtest` tool — execution stays an explicit agent step so run cost and data-source side effects remain visible.
- Writing `goal_evidence` rows — the existing `add_goal_evidence` tool handles that and needs the per-turn `expected_goal_id`.

## Test Plan

- [x] `ruff check agent/src/tools/autopilot_tool.py agent/tests/test_autopilot_phase3.py` — clean
- [x] `pytest agent/tests/test_autopilot_phase3.py agent/tests/test_autopilot_tool.py -q` — 19 passed
- [x] Full suite excl. e2e — 3268 passed, 1 skipped, no regressions
- [x] Stub-contract test imports the generated `signal_engine.py` and runs `generate()`, asserting no-arg construction + dict-of-Series output

## Risk / rollback

Additive only; both tools auto-register via `BaseTool.__subclasses__()`. No schema, config, or behavior changes to existing tools. Revertable with a single `git revert`.

### DIFF
--- a/agent/src/tools/autopilot_tool.py
+++ b/agent/src/tools/autopilot_tool.py
@@ -2,6 +2,9 @@
 
 Phase 1: Connects the Hypothesis Registry to the Research Goal runtime.
 Phase 2: Auto-generates backtest config.json from hypothesis metadata.
+Phase 3: Scaffolds a contract-correct signal_engine.py stub and links
+    backtest run-card metrics back to the hypothesis, closing the
+    hypothesis -> backtest -> evidence loop.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks @Robin1987China! 🙏 `scaffold_signal_engine` + `link_autopilot_backtest` close the two most error-prone manual steps of an autopilot run, and the generated stub matches the runner's `SignalEngine` contract exactly.

Issue #254, opened 2026-06-17
[Feature] Strongly recommend enabling the scheduled feature in advance


Penn-Live:
### Problem or Motivation

Strongly recommend enabling the scheduled feature in advance

### Proposed Solution

Strongly recommend enabling the scheduled feature in advance

### Area

Skills (new finance skill)

### Contribution

- [ ] I'm willing to submit a PR for this feature

warren618:
Thanks for raising this. To turn this into an implementable v1, can you clarify which scheduled workflow you want first?

The key choices are:
- scheduled backtests/research runs vs scheduled market scans/alerts vs scheduled agent prompts
- where it should be configured: CLI, API, or dashboard
- cadence/timezone needs, e.g. cron-style, daily market open/close, or one-shot
- required output/notification: saved run only, report artifact, webhook/email, or in-app list
- failure behavior: retry count, timeout, and whether missed runs should catch up

Once that is pinned down, the smallest safe implementation is likely a scheduler that records jobs, runs them through the existing run/backtest path, and exposes status/history in the UI/API.

mvanhorn:
Started on this in https://github.com/HKUDS/Vibe-Trading/pull/272 - adds the ScheduledResearchJob data model, crash-safe store, and POST/GET/DELETE /scheduled-runs endpoints; execution wiring deferred to a follow-up.

warren618:
#272 is merged, landing the foundation for scheduled research jobs:
- `ScheduledResearchJob` model + a crash-safe, atomically-written store (parse-failure quarantine, state kept under `~/.vibe-trading/`)
- auth-gated `POST` / `GET` / `DELETE /scheduled-runs` to record, list, and cancel jobs

Still open before this is end-to-end: the executor that fires jobs on their schedule. Leaving this issue open to track that follow-up.

mvanhorn:
Implemented in https://github.com/HKUDS/Vibe-Trading/pull/278 - adds the background executor that polls the existing scheduled-research store and fires due jobs (interval-ms and cron), gated behind VIBE_TRADING_ENABLE_SCHEDULER and default off.

warren618:
[...]

PR #272, opened 2026-06-19
feat: add scheduled research job data model and CRUD API

mvanhorn:
## Summary

Implements the data model and CRUD API for scheduled research jobs, the foundational slice of #254 (scheduled execution of agent prompts and backtests). Execution wiring is deliberately deferred to a follow-up PR once the schedule types and output channels are confirmed.

## Why this matters

Issue #254 asks for a scheduler that records jobs, runs them through the existing run/backtest path, and exposes status/history in the API. Before wiring execution, the shape of the persisted job needs to be established and confirmed. This PR lands that shape.

## Changes

- `agent/src/scheduled_research/__init__.py` - new module root
- `agent/src/scheduled_research/models.py` - `ScheduledResearchJob` dataclass with `id`, `prompt`, `schedule` (interval-ms or cron string), `next_run_at`, `status`, `created_at`, and `config`; `validate_schedule` rejects malformed specs with a clear error
- `agent/src/scheduled_research/store.py` - `ScheduledResearchJobStore` with crash-safe atomic persistence (write-temp + fsync + os.replace + fsync-dir), matching the pattern in `src/live/runtime/jobstore.py`; corrupt stores are quarantined rather than silently cleared
- `agent/api_server.py` - three new endpoints: `POST /scheduled-runs` (create/replace), `GET /scheduled-runs` (list with status filter + limit), `DELETE /scheduled-runs/{job_id}` (cancel); all behind `require_auth`

**Deferred:** execution wiring (actually triggering runs at schedule time) belongs in a follow-up once the maintainer confirms which schedule types and output channels to support.

## Testing

- 24 unit tests in `agent/tests/test_scheduled_research_store.py`
- Happy path: POST -> GET returns pending job -> DELETE removes it
- Atomic persistence: temp file cleaned up on success; fresh store instance reads persisted data correctly
- Corrupt store: quarantined file present, `CorruptStoreError` raised, original gone
- Idempotent upsert: same job_id twice replaces without duplicating
- Invalid schedule: `upsert` raises `ValueError` on malformed spec
- Empty store: `list_jobs` returns `[]` when no file exists
- All tests pass (`pytest tests/test_scheduled_research_store.py`)

Refs #254

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -3269,6 +3269,112 @@ register_alpha_routes(app)
 
 
 # ============================================================================
+# Scheduled Research Routes
+# ============================================================================
+#
+# Three lightweight CRUD endpoints backed by ScheduledResearchJobStore.
+# Execution wiring is intentionally deferred: these endpoints only record
+# and expose scheduled-research jobs; no run is triggered here.
+
+
+_scheduled_research_store: Optional["ScheduledResearchJobStore"] = None
+
+
+def _get_scheduled_research_store() -> "ScheduledResearchJobStore":
+    """Return the singleton ScheduledResearchJobStore, creating it on first call."""
+    global _scheduled_research_store
+    if _scheduled_research_store is None:
+        from src.scheduled_research.store import ScheduledResearchJobStore
+
+        _scheduled_research_store = ScheduledResearchJobStore()
+    return _scheduled_research_store
+
+
+class CreateScheduledRunRequest(BaseModel):
+    """Request body for POST /scheduled-runs."""
+
+    id: Optional[str] = Field(None, description="Job id; auto-generated UUID when omitted")
+    prompt: str = Field(..., min_length=1, description="Research prompt or backtest description")
+    schedule: str = Field(..., min_length=1, description="Interval-ms or 5-field cron expression")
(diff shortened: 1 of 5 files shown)


warren618:
Merged — thanks @mvanhorn! 🙏 Clean foundation: the atomic-write + corruption-quarantine store and the auth-gated CRUD endpoints are solid.

Two small things I adjusted before merge:
1. `_default_store_path()` now roots the store under `~/.vibe-trading/scheduled_research/` (via `get_runtime_root()`) instead of `agent/data/`. That dir is inside the repo working tree and isn't gitignored, so a running server would have left job state in the checkout — this matches how the live runtime / swarm / memory resolve their roots.
2. Added cron field-range validation (minute 0-59, hour 0-23, dom 1-31, month 1-12, dow 0-6), so an out-of-range expression like `99 * * * *` is now rejected.

Left #254 open to track the executor follow-up. CI green.

mvanhorn:
Glad to get this in, thanks @warren618. The scheduled research job data model and CRUD API is a solid foundation to build on.

mvanhorn:
Appreciate this slice landing, @warren618. The scheduled-research-job data model and CRUD API is a solid foundation for the scheduled execution work.

PR #274, opened 2026-06-19
feat(loaders): validate OHLC invariants at the loader boundary

Shizoqua:
### Problem
The loaders only drop NaN rows. Bars that are present but structurally invalid — `high < low`, a non-positive price, or a high/low that does not bracket open/close — pass straight through into the backtest, where they silently corrupt downstream metrics (and can yield non-finite values that break strict JSON serialization).

### Change
- Add a reusable `validate_ohlc(frame, strategy="drop")` helper in `agent/backtest/loaders/base.py` (next to `validate_date_range`).
- Call it after the existing `dropna` in the yfinance and local loaders.
- Default drops offending rows and logs a count (matching the loaders' existing silent-clean convention); `"warn"` and `"raise"` strategies are available for stricter callers.

### Tests
`agent/tests/test_ohlc_validation.py` covers drop/warn/raise behaviour, pass-through of valid frames (including flat/doji bars), and an end-to-end check that a dirty bar in a local CSV is dropped by the loader.

```
python -m pytest agent/tests/test_ohlc_validation.py agent/tests/test_local_loader.py -q
# 7 passed, 1 skipped
```

### DIFF
--- a/agent/backtest/loaders/base.py
+++ b/agent/backtest/loaders/base.py
@@ -47,6 +47,59 @@ def validate_date_range(start_date: str, end_date: str) -> None:
         raise ValueError(f"start_date ({start_date}) > end_date ({end_date})")
 
 
+def validate_ohlc(frame: pd.DataFrame, *, strategy: str = "drop") -> pd.DataFrame:
+    """Drop, flag, or reject bars that violate OHLC invariants.
+
+    Loaders only drop NaN rows, so structurally dirty bars — ``high < low``,
+    a non-positive price, or a high/low that fails to bracket open/close —
+    flow straight into the backtest and surface downstream as NaN/inf metrics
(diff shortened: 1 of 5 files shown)


warren618:
Merged — thanks @Shizoqua! 🙏 The `validate_ohlc` helper with its drop/warn/raise strategies is exactly the right boundary check.

Before merging I added one follow-up commit so it guards *every* source, not just local + yfinance. Instead of wiring the call into all ~18 loaders one by one, it now also runs once centrally in `runner.py` (`_sanitize_data_map`) at the single point every fetch path converges — auto / single-source / runtime-fallback — so any current or future loader is covered. Your per-loader calls stay (they also guard the `get_market_data` path). Added a test for the central pass; full suite + CI green.

PR #278, opened 2026-06-20
feat: scheduled research job executor that fires jobs on their schedule

mvanhorn:
## Summary

This adds the background executor that actually fires scheduled research jobs on their schedule. Until now a user could create a scheduled run through the API, but nothing ever ran it. The executor polls the existing job store on a tick interval, finds jobs whose time has come, and runs each one, moving it through pending, running, and then completed or failed. It is wired into the API server's startup and shutdown but stays dormant unless `VIBE_TRADING_ENABLE_SCHEDULER` is set, so existing deployments behave exactly as before.

## Why

PR #272 landed the data model, durable store, and CRUD endpoints for scheduled research jobs, but deliberately stopped short of execution. Maintainer warren618 left issue #254 open to track the remaining piece, "the executor that fires jobs on their schedule." This change is that follow-up. Refs #254.

## Changes

The new executor lives alongside the model and store that #272 already shipped. Its schedule math is split into small pure functions that take the current time as an argument, so cron and interval-millisecond due-time logic can be unit tested without sleeping or freezing the clock, the same shape the live-runtime scheduler already uses in this repo.

Most of the work here is about surviving the messy cases. If the process dies mid-run a job can be left marked running forever, so on startup the executor resets any stale running jobs back to pending. Because dispatch is awaited, a user can delete or replace a job while its run is in flight, so the executor re-reads the record before it writes anything back and skips the write if the job is gone or has been replaced, identified by its creation timestamp. An impossible calendar schedule such as the 31st of February is rejected by walking candidate days rather than scanning minutes, so it fails immediately instead of blocking the event loop. A model field records when each job last ran, written so older stored jobs without it still load.

Dispatch itself stays behind a small injected callable, which keeps the executor testable in isolation and lets the API server route real runs through the existing session service. That dispatch path receives a copy of the job's config so the session runtime cannot mutate the user's saved configuration. No protected area is modified; the session service is only called through its existing public methods.

## Test Plan

New unit tests cover interval and cron due-time math, the full status lifecycle, a dispatch failure not crashing the loop, cancelled and running jobs being skipped, stale-running recovery, no double-fire after a missed window, the delete and replace races during dispatch, impossible-cron handling, and the disabled no-op lifecycle. Running the executor and store suites together gives 40 passing tests, and the API server module imports cleanly with the wiring in place.

## Checklist

AI was used for assistance.

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -631,6 +631,13 @@ async def _run_startup_preflight() -> None:
     from src.preflight import run_preflight
 
     run_preflight(console)
+    _start_scheduled_research_executor()
+
+
+@app.on_event("shutdown")
+async def _stop_scheduled_research_on_shutdown() -> None:
+    """Stop the scheduled research executor on server shutdown."""
+    await _stop_scheduled_research_executor()
 
 
 # ============================================================================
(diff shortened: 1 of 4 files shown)


warren618:
Merged 🎉 Thanks @mvanhorn — this closes out scheduled research end-to-end (the executor was the missing half after #272 landed the model + store + CRUD). Verified before merge: the 40 executor/store tests are green, `api_server` imports cleanly with the wiring in place, and no protected area is touched. The default-off `VIBE_TRADING_ENABLE_SCHEDULER` gate keeps existing deployments unchanged, and the robustness work — stale-`RUNNING` recovery on startup, re-read-before-write on delete/replace races, impossible-cron rejection, and dispatching with a config copy so saved jobs aren't mutated — is exactly what this needed. Closes #254.

mvanhorn:
Thank you @warren618. The executor that actually fires scheduled research jobs on schedule rounds out the feature nicely.

PR #293, opened 2026-06-23
fix: reject cross-site unsafe local API requests

Hinotoi-agent:
## Summary

- Reject unsafe cross-site browser requests before applying loopback dev-mode auth trust.
- Add upload API regressions covering cross-site loopback rejection and the local non-browser upload control.

## Why

Loopback requests are trusted by the local API so CLI/local UI workflows can work in dev mode. However, browsers can also issue simple cross-origin requests to `127.0.0.1`; CORS prevents response reads but does not prevent side effects. This change makes unsafe browser-originated requests hit the existing cross-site guard before loopback auth bypass is honored.

## Changes

- Add `_SAFE_BROWSER_METHODS` and call `_reject_cross_site_browser_request()` from `_validate_api_auth()` for unsafe methods.
- Preserve existing safe-method behavior and local non-browser upload behavior.
- Add regression coverage for `POST /upload` with cross-site browser headers.

## Test Plan

- [x] Existing targeted tests pass: `python3 -m pytest -q agent/tests/test_upload_api.py agent/tests/test_security_auth_api.py agent/tests/test_settings_api.py`
- [x] New tests added (cross-site loopback upload rejection and local non-browser upload control)
- [x] Tested manually in Docker/container:
  - `docker run --rm -v "$PWD:/work" -w /work python:3.11-slim ... python -m pytest -q agent/tests/test_upload_api.py agent/tests/test_security_auth_api.py agent/tests/test_settings_api.py`
  - Result: `70 passed, 5 warnings in 0.54s`
  - Container proof after fix: cross-site upload returned `403` with no created files; local non-browser upload returned `200`.

## Checklist

- [x] Documentation updated (not user-facing; regression tests added instead)

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -764,6 +764,9 @@ def _require_shutdown_authorization(
         )
 
 
+_SAFE_BROWSER_METHODS = {"GET", "HEAD", "OPTIONS"}
+
+
 def _validate_api_auth(
     *,
     request: Request,
@@ -772,6 +775,12 @@ def _validate_api_auth(
     allow_query: bool = False,
 ) -> None:
     """Validate configured auth, preserving loopback-only dev mode."""
+    # CORS protects response reads, not blind side effects. Reject unsafe
+    # browser-originated cross-site requests before honoring loopback dev-mode
+    # trust, otherwise a malicious page can drive local POST/PUT/DELETE routes.
+    if request.method.upper() not in _SAFE_BROWSER_METHODS:
+        _reject_cross_site_browser_request(request)
+
     # Loopback clients are always trusted, even when API_AUTH_KEY is set.
     # The key only gates non-local (LAN/remote) access.
     if _is_local_client(request):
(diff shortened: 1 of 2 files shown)


Penn-Live:
After completing this merge, I found that I cannot access it from another machine on the internal network. My access machine is a Mac, the server is Windows, and the service is running on WSL. How to resolve?

Hinotoi-agent:
Thanks for reporting this. I traced it to the #293 cross-site guard rejecting all non-loopback browser Origin headers, including legitimate same-origin requests from a LAN URL like http://<windows-ip>:8899. I opened a follow-up fix in #304 that allows authenticated same-origin remote Web UI requests while preserving cross-site rejection.\n\nAfter that fix, the expected setup for Mac -> Windows/WSL access is still: set a strong API_AUTH_KEY on the server, restart, browse to http://<windows-ip>:8899, and enter the same key once in Settings. If WSL itself is not reachable from the LAN, Windows firewall / WSL port forwarding may also need to allow port 8899.

PR #300, opened 2026-06-24
feat(settings): expose FRED/iWenCai API keys and visualize Feature Flags in Settings UI

skloxo:
This PR exposes essential environment parameters (like macroeconomic data keys and semantic screening keys) to the frontend Settings page, and adds status indicator cards for core runtime components.

### Key Changes
- Added Settings input fields for FRED API key and iWenCai API key in Settings.tsx.
- Introduced /settings/feature-flags API endpoint on the backend to detect active environment flags.
- Rendered status indicator cards for Shell Tools, Scheduler, and Session Runtime in the settings UI.
- Restored standard secure behavior in _is_local_client (reverting the insecure ip.is_private check in developer mode).

### Verification
- Tested settings API actions using uv run pytest agent/tests/test_settings_api.py (100% success).
- Built frontend cleanly (npm run build completed with zero errors).


warren618:
Thanks for the PR — we’ve seen it and will review as soon as we can.

For settings changes, we’ll check that secrets stay local/config-driven, UI strings go through i18n, and feature-flag display does not expose anything unsafe or misleading.

Issue #295, opened 2026-06-23
[Feature] Trade Journal -> strategy extraction -> backtest bridge (Phase 4c) - planning question


Robin1987China:
## Problem or Motivation

The Trade Journal skill explicitly marks the strategy-extraction -> backtest bridge as not-yet-implemented:

- `agent/src/tools/trade_journal_tool.py:9` — `Strategy extraction -> backtest bridge still pending (Phase 4c).`
- `agent/src/skills/trade-journal/SKILL.md:209-213` — "Phase 4c preview (not yet implemented). Strategy extraction -> SignalEngine code gen -> auto-backtest lands in Phase 4c."

Instead of hypothesis → backtest (Research Autopilot), this goes the other way: a user's own broker journal → extracted strategy rules → backtest the shadow. It reuses the same `SignalEngine` contract that `scaffold_signal_engine` produces.

A direct blocker for extraction quality: `agent/src/shadow_account/extractor.py:10-13` stubs price-dependent features (`prior_N_return`, `rsi`) as `None` because v1 makes no external price calls — so the feature matrix the rules are learned from is missing price context.

## Proposed Solution

Not asking to land all of Phase 4c at once. Asking about the planning boundary:

1. Is Phase 4c on the near-term roadmap, or is it parked behind the messaging layer / Research Autopilot closure?
2. Would a small, self-contained first step be welcome — e.g. implementing the price-dependent features in `shadow_account/extractor.py` against the existing loader registry (a non-protected-area change) — or would you prefer to keep `shadow_account/` frozen until the full 4c design lands?

## Alternatives Considered

- Wait for an official Phase 4c design doc before touching anything: safe, but the gap is concrete and the blocker is localized.
- Submit a full Phase 4c PR blind: too large without scope agreement, given it's near the protected area.

## Contribution

Willing to contribute that first step (price-dependent features + tests) if maintainers confirm it fits. Not opening a PR until scope is agreed.

Area: Other

warren618:
[...]

PR #302, opened 2026-06-24
feat(shadow-account): add as-of price-context features (entry RSI, prior return)

Robin1987China:
## Summary

Shadow Account's `_compute_features()` only learned from journal columns
(`holding_days`, `pnl_pct`, `entry_hour`, `entry_weekday`, `market`), so the
extracted rules were purely behavioral with no market context. This adds two
price-context features per profitable roundtrip, read as-of the buy date:

- `entry_rsi14` — 14-period Wilder RSI at entry
- `prior_5d_return` — trailing 5-bar close-to-close return ending at `buy_dt`

This is the first step you green-lit in #295. I followed the three scoping notes
from that thread:

1. **Reused existing pieces.** RSI mirrors the causal `compute_rsi` shape in
   `agent/src/skills/technical-basic/example_signal_engine.py:13` (re-implemented
   inline because that hyphenated skills dir isn't importable). Prices go through
   the loader registry (`resolve_loader` + `fetch`), the same path
   `backtest/runner.py` uses.
2. **As-of `buy_dt`, no look-ahead.** Both features read only bars dated
   `<= buy_dt`. The fetch window ends at `buy_dt`, so the `(buy_dt, sell_dt]` exit
   window is never touched — look-ahead is structurally impossible, not just
   guarded after the fact.
3. **Kept v1 degradation + offline determinism.** Any failure (unmapped market,
   no source, empty result, insufficient history) drops the feature to NaN and the
   pipeline still runs. All tests use mocked loaders — no network.

A price feature only joins the KMeans feature set when it's present for
`>= min_support` roundtrips; otherwise clustering stays journal-only, exactly as
before. Partial NaNs are median-imputed (that affects grouping only — rule bounds
never read price features).

## Why

Shadow Account extracts rules from journal columns alone, so extracted rules
are purely behavioral with no market context. Adding price-context features
lets clustering see whether profitable roundtrips share common entry conditions
(overbought RSI, recent momentum), making extracted rules actionable rather
than just descriptive.

## Changes

- `agent/src/shadow_account/extractor.py`: `_compute_rsi`, `_fetch_price_history`,
  `_as_of_index`, `_price_features_as_of`, `_attach_price_features` (batches one
  fetch per symbol); `_auto_cluster` takes a `numeric_features` arg instead of the
  module constant; `_promoted_numeric_features` gates promotion.
- `agent/tests/test_shadow_account.py`: 15 new tests + an autouse offline fixture
  so the suite stays network-free.

## Test Plan

- [x] `ruff check` — clean
- [x] `pytest agent/tests/test_shadow_account.py` — 43 passed
- [x] full suite (excl e2e) — 4385 passed, 1 skipped
- [x] RSI causality, look-ahead guard (bars after `buy_dt` don't change values),
  tz-aware `buy_dt`, all three degradation branches, batch-once-per-symbol,
  median-impute / all-NaN-drop in clustering

## Checklist

- [x] No changes to protected areas without prior discussion (#295 green-lit)
- [x] No hardcoded values
- [x] Documentation updated (N/A — no user-facing change)

Refs #295

### DIFF
--- a/agent/src/shadow_account/extractor.py
+++ b/agent/src/shadow_account/extractor.py
@@ -7,10 +7,11 @@ Pipeline:
     → LLM-light natural-language translation (template fallback if no LLM)
 
 Design constraints:
-    * No external price-data calls in v1. All features are derivable from
-      the journal itself (holding_days, pnl_pct, entry hour/weekday, market).
-      Price-dependent features (prior_N_return, rsi) are stubbed with None
-      and dropped from the feature matrix if unavailable.
+    * No *mandatory* external price-data calls. Journal-derived features
+      (holding_days, pnl_pct, entry hour/weekday, market) always work offline.
+      Price-context features (entry_rsi14, prior_5d_return) are read as-of the
+      buy date via the backtest loader registry and degrade to NaN — dropped
+      from the feature matrix — whenever price data is unavailable.
     * Must survive tiny samples: <5 profitable roundtrips → explicit error.
       <2 clusters → degrade to a single-cluster heuristic rule.
     * Rules are immutable ShadowRule objects — codegen's only input.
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @Robin1987China! Merged as `b175c25`. Nicely PIT-safe — `entry_rsi14` / `prior_5d_return` are read as-of `buy_dt` through the loader registry, the per-symbol batch fetch still re-slices to each roundtrip's own buy date, and sparse price data cleanly degrades to the journal-only baseline. 🙏

Issue #283, opened 2026-06-21
[Feature] Add WhatsApp messaging bridges for agent sessions


MarkfuGod:
### Problem or Motivation

I want to interact with Vibe-Trading from WhatsApp, so I can send finance research or backtest requests from mobile and receive the agent's response back in the same conversation.

There are two practical WhatsApp workflows worth supporting:

1. Production/commercial deployments should use Meta's official WhatsApp Cloud API with webhook verification, signature validation, and Graph API replies.
2. Personal/local workflows should support a QR-code WhatsApp Web flow, similar to OpenClaw's Baileys-style bridge, so a user can link their own WhatsApp account and message the agent without setting up Meta Business infrastructure.

This would make Vibe-Trading easier to use as a personal trading/research assistant outside the web UI or CLI.

### Proposed Solution

Add optional WhatsApp bridge support that routes inbound WhatsApp text messages into the existing Vibe-Trading session runtime.

## Draft

I currently have the draft right here: https://github.com/MarkfuGod/Vibe-Trading/tree/feature/whatsapp-bridges

## Performance

Some screenshots here:

<img width="1206" height="2307" alt="Image" src="https://github.com/user-attachments/assets/510099a2-8e66-46dd-a2be-ff3e5f677424" />
<img width="1206" height="2320" alt="Image" src="https://github.com/user-attachments/assets/cc96d2af-4fdb-491d-a81b-b1121e1917a0" />
<img width="1206" height="2233" alt="Image" src="https://github.com/user-attachments/assets/0160031c-53d1-4498-b5cb-6ae3b60bd2f6" />

### Area

Other

### Contribution

- [x] I'm willing to submit a PR for this feature

warren618:
Thanks for the thoughtful proposal and for putting a draft together, @MarkfuGod — messaging access is something we really want, and your write-up (Cloud API vs. personal QR flow) framed the two modes well.

Rather than land it as a standalone WhatsApp bridge, we're going to build messaging in as a **first-class, multi-channel layer ourselves** — one pluggable interface so WhatsApp, Telegram, Discord, Slack and others all share the same session routing and safety model (a messaging channel stays an **inbound prompt path only**; it never becomes a route around the live-trading mandate gate). Both modes you described — the official Cloud API and a personal QR flow — are on the plan.

[...]

PR #306, opened 2026-06-25
[codex] fix(validation): write strict JSON outputs

gyx09212214-prog:
﻿## Summary

Make the standalone backtest validation CLI write strict JSON when validation metrics contain non-finite floating point values.

## Why

`backtest.validation` can produce nested validation payloads that include floating point metrics. Python's default `json.dumps` can emit `NaN`, `Infinity`, or `-Infinity`, which are not valid JSON tokens for strict parsers. The run-card writer already normalizes this class of value before writing JSON; this applies the same defensive behavior to `artifacts/validation.json` and the CLI stdout path.

## What changed

- Add a small JSON-safe normalization helper for validation results.
- Write `validation.json` with `allow_nan=False` and a trailing newline.
- Print the normalized payload to stdout.
- Add a regression test covering nested `NaN`/`Infinity` values from the validation routines.

## Duplicate check

I checked the current open issues and PRs before opening this draft. The newest open issues are already covered by existing work or are planning/draft-branch discussions: #295 is covered by #302, #282 is covered by #296, and #283 already has a contributor draft branch. I also searched open PRs for validation/strict JSON/non-finite work and found no overlap.

## Validation

- `PYTHONPATH=agent python -m pytest agent/tests/test_validation.py agent/tests/test_validation_cli.py -q`
- `PYTHONPATH=agent python -m py_compile agent/backtest/validation.py agent/tests/test_validation_cli.py`
- `python -m ruff check agent/backtest/validation.py agent/tests/test_validation_cli.py`
- `python -m ruff format --check agent/backtest/validation.py agent/tests/test_validation_cli.py`
- `git diff --check`

### DIFF
--- a/agent/backtest/validation.py
+++ b/agent/backtest/validation.py
@@ -11,6 +11,8 @@ is present, or invoked directly on backtest outputs.
 
 from __future__ import annotations
 
+import json
+import math
 from pathlib import Path
 from typing import Any, Dict, List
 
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @gyx09212214-prog! Merged as `6d1406b`. Good catch — `validation.json` and CLI stdout now normalize non-finite values (including numpy floats) to `null` and write with `allow_nan=False`, so strict JSON parsers stop choking. 🙏

PR #319, opened 2026-06-27
fix: resolve default OpenAI model to an available id (gpt-5.5-instant unavailable)

### LINKED ISSUE #311: [Bug] default openai LLM model doesn't resolve (gpt-5.5-instant)
### Description

What happened:
❯ hello! what model are you using?
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
AgentLoop error: provider_stream_error provider=openai model=gpt-5.5-instant: NotFoundError: Error code: 404 - {'error': {'message': 'The
model `gpt-5.5-instant` does not exist or you do not have access to it.', 'type': 'invalid_request_error', 'param': None, 'code':
'model_not_found'}}

I fixed it by editing .env and changing 'gpt-5.5-instant' to 'gpt-5.5'. When I check the available models on my api interface I don't see instant. Maybe they removed it.

### Steps to Reproduce

init vibe trading with openai as your provider, choice #2

### Error Logs

```shell

```

### Interface

CLI (vibe-trading)

### LLM Provider

open ai

### Version

0.1.10

### Environment

_No response_

mvanhorn:
## Summary

Change the OpenAI provider entry's `default_model` in `agent/src/providers/llm_providers.json` from `gpt-5.5-instant` to `gpt-5.5` (an available, generally-resolvable id). Update the matching onboarding default in `agent/cli/onboard.py` where `"gpt-5.5-instant"` is both the seeded default and the head of the suggestion tuple `("gpt-5.5-instant", "gpt-5.5-pro", "gpt-5.5")`, so the prompted default lands on the resolvable model while keeping the others as alternatives. Update `agent/tests/test_llm_provider_defaults.py`, which currently asserts the OpenAI default equals `gpt-5.5-instant`, to assert the new default and to guard against regressing back to an unavailable id.

## Why

On a fresh install, the default OpenAI provider model is `gpt-5.5-instant`, which does not resolve against the OpenAI API for new setups, so the first run fails. The maintainer (warren618, COLLABORATOR) confirmed on 2026-06-27 that the default OpenAI model should resolve cleanly out of the box and that falling back to an available default such as `gpt-5.5` is the right direction unless the provider exposes `instant` again. A focused PR was explicitly welcomed, ideally updating the default config plus a small regression test so a fresh OpenAI setup never points at an unavailable model.

Closes #311

## Changes

- Change the OpenAI provider entry's `default_model` in `agent/src/providers/llm_providers.json` from `gpt-5.5-instant` to `gpt-5.5` (an available, generally-resolvable id). Update the matching onboarding default in `agent/cli/onboard.py` where `"gpt-5.5-instant"` is both the seeded default and the head of the suggestion tuple `("gpt-5.5-instant", "gpt-5.5-pro", "gpt-5.5")`, so the prompted default lands on the resolvable model while keeping the others as alternatives. Update `agent/tests/test_llm_provider_defaults.py`, which currently asserts the OpenAI default equals `gpt-5.5-instant`, to assert the new default and to guard against regressing back to an unavailable id.

## Test Plan

- Happy path: a fresh OpenAI provider config loads with `gpt-5.5` as default and the provider-defaults test passes against the new value. - Edge case: onboarding still offers `gpt-5.5-pro` and `gpt-5.5-instant` as selectable alternatives (no option removed), only the default head changes. - Regression: test pins the default so future edits cannot silently restore `gpt-5.5-instant` as the fresh-install default.

## Checklist

AI was used for assistance.

### DIFF
--- a/agent/cli/onboard.py
+++ b/agent/cli/onboard.py
@@ -58,9 +58,9 @@ PROVIDERS: Final[tuple[Provider, ...]] = (
              ("deepseek/deepseek-v4-pro", "deepseek/deepseek-v4-flash",
               "openai/gpt-5.5-pro", "google/gemini-3.5-flash")),
     Provider("openai", "OpenAI", "GPT-5.5 direct",
-             "gpt-5.5-instant", "OPENAI_API_KEY", "OPENAI_BASE_URL",
+             "gpt-5.5", "OPENAI_API_KEY", "OPENAI_BASE_URL",
              "https://api.openai.com/v1", "sk-",
-             ("gpt-5.5-instant", "gpt-5.5-pro", "gpt-5.5")),
+             ("gpt-5.5", "gpt-5.5-pro", "gpt-5.5-instant")),
     Provider("deepseek", "DeepSeek",
              "cheapest tier — good for batch backtest research",
              "deepseek-v4-pro", "DEEPSEEK_API_KEY", "DEEPSEEK_BASE_URL",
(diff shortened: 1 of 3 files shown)


warren618:
Merged in `24df3d9`. Thanks @mvanhorn — this fixes the unavailable default OpenAI model path by moving the default to an available model and keeping the fresh-install regression covered.

Issue #333, opened 2026-06-29
API server default bind is 0.0.0.0; default to 127.0.0.1 and warn when binding all interfaces without API_AUTH_KEY


tg12:
> Filed as a good-faith hardening suggestion from an external code review. Severity is Low — the server has strong loopback/DNS-rebinding defenses; this is about making the *default* safer-by-default and better documented.

## Summary
The API server's default bind address is `0.0.0.0` (all interfaces), while the no-API-key "dev mode" security posture relies on a peer-IP loopback check plus Host-header validation. The combination is defensible, but defaulting to `0.0.0.0` means a user who runs `vibe-trading serve` on a non-isolated network, without setting `API_AUTH_KEY`, is one misconfiguration away from exposing the surface; a 127.0.0.1 default would be safer-by-default with no loss of normal functionality.

## Evidence
- `agent/api_server.py:3496` — `parser.add_argument("--host", default="0.0.0.0", help="Bind address")`.
- `agent/api_server.py:3532` — `uvicorn.run(app, host=args.host, port=args.port, ...)`.
- Compensating controls (good, and the reason this is Low not High):
  - `require_auth` / `require_local_or_auth` reject non-loopback clients when no key is set (`agent/api_server.py:894-936`).
  - `_reject_untrusted_loopback_host` middleware blocks DNS-rebinding Host headers (`agent/api_server.py:562-570`).
  - `hmac.compare_digest` is used for key comparison (`agent/api_server.py:821`).
- The official `docker-compose.yml` already does the right thing by binding `127.0.0.1:8899:8899`, which shows loopback-only is the intended posture — but the bare CLI default does not match that intent.

## Why this matters
"Secure by default" matters most for the path users take without reading docs. The Docker path is locked to loopback; the direct `serve` path is not. The current safety net is the peer-IP + Host checks, which are good but are a second line of defense — defense-in-depth argues the bind itself should default to loopback, and binding to all interfaces should be a conscious opt-in.

## Attack or failure scenario
A user runs the server directly (not via compose) on a laptop joined to a coffee-shop / corporate LAN, does not set `API_AUTH_KEY`, and assumes "it's just local." The listener is reachable from the LAN at `<lan-ip>:8899`. Remote requests are currently rejected by the loopback peer-IP check — so this is not an immediate breach — but the exposure surface exists, and any future regression in the loopback/Host logic (see the monolithic-file issue) would turn it into one.

## Root cause
[...]

PR #338, opened 2026-06-30
fix(api): default bind to 127.0.0.1 and warn on non-loopback without API_AUTH_KEY

Robin1987China:
## Summary

Default the `--host` bind address from `0.0.0.0` to `127.0.0.1`, and log a startup warning when binding to a non-loopback address without `API_AUTH_KEY` configured.

## Why

The Docker compose already binds to `127.0.0.1:8899:8899`, so loopback-only is the intended secure-by-default posture — but the bare CLI default of `0.0.0.0` did not match. The existing loopback peer-IP and DNS-rebinding checks still reject remote requests, but defense-in-depth argues the bind itself should default to loopback. Binding to all interfaces should be a conscious opt-in, with a warning when done without `API_AUTH_KEY`.

## Changes

- `agent/api_server.py`: `--host` default `"0.0.0.0"` → `"127.0.0.1"`; added `_is_loopback_bind_host()` (`ipaddress.is_loopback` + `"localhost"` fallback); startup warning when binding non-loopback without `API_AUTH_KEY`.
- `agent/cli/main.py`: Typer `serve` command `--host` default also set to `127.0.0.1`, so every serve entry point agrees. The console script and `python -m cli` already bound loopback (they route through the argparse path); this covers the `python -m cli.main` Typer path too.
- `agent/tests/test_serve_bind.py`: new test coverage.

Docker is unaffected: the Dockerfile passes `--host 0.0.0.0` explicitly and compose binds loopback on the host side — neither relies on the default.

I left out one reviewer-style nit — only computing the warning state lazily — because the check is a single comparison at startup, not a hot path.

## Test Plan

- [x] ruff check (changed files) — clean (the 5 F821 in `api_server.py` pre-exist on origin/main; none introduced)
- [x] pytest tests/test_serve_bind.py — 14 passed (`_is_loopback_bind_host` truth table incl. IPv6/hostname/edge, loopback default, warning only on non-loopback without a key)
- [x] pytest (full, excl e2e) — passes, zero failures

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (if user-facing change) — n/a, default/behavior change

Refs #333

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -547,6 +547,14 @@ def _is_allowed_loopback_host(host: str) -> bool:
     return normalized in _DEFAULT_LOOPBACK_HOSTS or normalized in _EXTRA_LOOPBACK_HOSTS
 
 
+def _is_loopback_bind_host(host: str) -> bool:
+    """Return whether ``host`` resolves to a loopback interface."""
+    try:
+        return ipaddress.ip_address(host).is_loopback
+    except ValueError:
+        return host == "localhost"
+
+
 # CORS: override with CORS_ORIGINS (comma-separated explicit origins)
 _CORS_ORIGINS = _parse_cors_origins(os.getenv("CORS_ORIGINS"))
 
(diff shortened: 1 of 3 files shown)


shadowinlife:
LGTM

warren618:
Thanks for the clean secure-by-default fix. Merged in #338.

I verified the focused bind tests locally and also checked the combined merge with #349/#351. This closes the API default-bind hardening from #333.

PR #340, opened 2026-06-30
feat: add agent identity verification gate for multi-agent swarms

saneGuy:
## Summary

Adds a pluggable identity verification gate that sits alongside the existing mandate gate and order gate. Each agent in the swarm proves its identity and permissions before executing trading actions.

**The gap:** Agents currently authenticate via broker credentials but there's no per-agent identity verification within the swarm. A compromised agent process can execute any action. This gate adds cryptographic agent identity checks as defense-in-depth.

**Files added:**
- `agent/src/agent/identity/verifier.py` — `AgentVerifier` ABC + `StructuralVerifier` dev stub
- `agent/src/agent/identity/gate.py` — `IdentityGate` with role-based action permissions
- `agent/src/agent/identity/__init__.py` — module exports
- `agent/tests/test_identity_gate.py` — 9 tests

**Design:**
- Mirrors existing gate pattern (mandate gate, order gate)
- Role-based permissions: read/trade/approve/admin per action type
- Disabled by default, fails closed when enabled
- StructuralVerifier logs explicit security warning on construction
- Decision history for audit ledger integration

## Test plan

- [x] 9 tests: verifier (valid, invalid JSON, missing role, missing expiry, expired) + gate (disabled, no verifier, missing token, valid trade, denied trade, denied admin, audit log)

Signed-off-by: Viswanadha Pratap Kondoju <kondojuviswanadha@gmail.com>

### DIFF
--- a/agent/src/agent/identity/__init__.py
+++ b/agent/src/agent/identity/__init__.py
@@ -1,0 +1,22 @@
+"""Agent identity verification for multi-agent trading workflows.
+
+Provides pluggable identity verification so each agent in a swarm can
+prove its identity and permissions before executing trading actions.
+Integrates with the existing mandate gate and audit ledger systems.
+"""
+
+from agent.identity.verifier import (
+    AgentCredential,
+    AgentVerifier,
+    StructuralVerifier,
+    VerificationResult,
+)
+from agent.identity.gate import IdentityGate
+
+__all__ = [
+    "AgentCredential",
+    "AgentVerifier",
+    "IdentityGate",
+    "StructuralVerifier",
+    "VerificationResult",
+]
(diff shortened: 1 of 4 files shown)


warren618:
Appreciate the thought that went into this. We're keeping trading safety on a single enforcement point — the mandate gate (`sdk_order_gate.py` / `order_guard.py`) — rather than adding parallel governance layers (see also #31/#144, closed as not-planned for the same reason). Closing, but the door stays open if a concrete attack scenario shows the mandate gate itself needs per-agent identity.

saneGuy:
That makes sense. I agree the mandate gate should stay the single enforcement point; parallel governance layers would be the wrong shape. The concrete scenario I'd watch for is a swarm where multiple agents share the same mandate authority, but the gate cannot distinguish which agent is issuing an order. If one agent is compromised, spoofed, or prompt-injected, it can inherit the full swarm/order mandate unless the gate verifies the actual per-agent identity or role before accepting the order. So the issue is not adding another gate, but whether the existing mandate gate eventually needs agent identity as one of its inputs.

PR #341, opened 2026-06-30
feat: wire IM channel runtime

warren618:
## Summary
- add shared IM channel discovery, dependency-gated status, manager, message bus, pairing store, and SessionService runtime wiring
- expose channel controls through REST, CLI, and Settings UI with i18n coverage
- declare channel dependency extras and add focused backend/frontend regression tests
- remove legacy-name references from the committed tree

## Verification
- `PYTHONPATH=agent pytest agent/tests/test_channels_runtime.py agent/tests/test_channels_api.py agent/tests/test_cli_channels.py agent/tests/test_packaging_dependencies.py -q`
- `python -m compileall -q agent/src/channels agent/src/channelsui agent/src/security/workspace_access.py agent/src/security/network.py agent/src/security/workspace_policy.py agent/src/session/goal_state.py agent/src/session/webui_turns.py agent/src/utils/media_decode.py agent/api_server.py agent/cli/_legacy.py`
- `cd frontend && npm test -- SettingsChannels --run && npm test -- i18n --run`
- `cd frontend && npm run build`
- final committed tree legacy-name scan returned no matches

PR #364, opened 2026-07-01
fix(frontend): proxy /channels through Vite dev server

kwp3:
## Summary

The Settings page (`frontend/src/pages/Settings.tsx`) loads three endpoints together in a single `Promise.all`:

```ts
Promise.all([api.getLLMSettings(), api.getDataSourceSettings(), api.getChannelStatus()])
```

`getChannelStatus()` requests `/channels/status`, but `/channels` is **not** in the Vite dev proxy `PROXY_PATHS` (`frontend/vite.config.ts`). In `npm run dev`, that request therefore falls through to the SPA `index.html` fallback and returns HTML instead of JSON. Parsing fails with:

```
Unexpected token '<', "<!doctype "... is not valid JSON
```

Because all three requests share one `Promise.all`, the entire Settings page renders **"Settings are unavailable"** — not just the channels section.

## Fix

Add `/channels` to `PROXY_PATHS` so every `/channels/*` route (`status`, `start`, `stop`, `pairing/command`) reaches the backend in local dev.

```diff
   "/settings/llm",
   "/settings/data-sources",
+  "/channels",
   "/mandate",
```

## Why it surfaced now

The IM channel runtime + Settings UI shipped recently (#341, #343). The backend already serves `/channels/status` as JSON (verified locally — returns `{"running":false,...}` with HTTP 200); only the dev proxy list was missed.

## Scope / impact

- **Local dev only.** Production/Docker is unaffected because the frontend proxies to the backend container there, not through the Vite dev server.
- No backend changes; one line in `vite.config.ts`.

## Testing

- Repro: `vibe-trading dev` → open `http://localhost:5899/settings` → "Settings are unavailable" with the JSON parse error.
- After fix (restart dev server so Vite reloads its config): Settings page loads LLM, data-source, and channel status cleanly.
- `curl http://127.0.0.1:8899/channels/status` returns valid JSON (200), confirming the backend route is correct and only the proxy was missing.

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -8,6 +8,7 @@ const PROXY_PATHS = [
   "/swarm/runs",
   "/settings/llm",
   "/settings/data-sources",
+  "/channels",
   "/mandate",
   "/live",
   "/upload",


warren618:
Thanks - this proxy fix is now covered on main by the integration stack, so I’m closing this as superseded to avoid merging the same /channels proxy change twice.

PR #367, opened 2026-07-01
fix(i18n): fix zh-CN language switching and LanguageSwitcher display

gg686-jkl:
## Summary

Fixes zh-CN (Chinese) language switching that was silently falling back to English, fixes the LanguageSwitcher always displaying "English" regardless of the active language, and fixes Settings page JSON parse error caused by missing Vite proxy rule.

## Why

Three separate i18n bugs combined to make the Chinese UI non-functional:

1. `nonExplicitSupportedLngs: true` in i18next config causes `isSupportedCode()` to strip region codes (`zh-CN` → `zh`) before checking `supportedLngs`. Since `"zh"` is not in the supported list, `changeLanguage("zh-CN")` is rejected and falls back to English. Other languages (`ja`, `ko`, `ar`) are unaffected because they are simple two-letter codes.

2. `LanguageSwitcher` used `i18n.languages` (plural) to detect the current language. This array always includes the fallback `"en"`, and since `"en"` is first in `SUPPORTED_LANGUAGES`, `.find()` always matched English regardless of the active language.

3. Settings page calls `/channels/status` via `Promise.all` alongside `/settings/llm` and `/settings/data-sources`. The Vite proxy config was missing `/channels`, so the request fell through to the static file handler which returned `index.html`, causing `Unexpected token "<", "<!doctype..." is not valid JSON`.

## Changes

- `frontend/src/i18n/index.ts`: Removed `nonExplicitSupportedLngs: true` from i18next init config.
- `frontend/vite.config.ts`: Added `"/channels"` to `PROXY_PATHS` array.
- `frontend/src/components/layout/Layout.tsx`: Changed `LanguageSwitcher` current language detection to check `i18n.language` (singular, primary active language) first, then fall back to `i18n.languages` (plural).

## Test Plan

Manual testing:
- Verified `npm run build` completes with zero TypeScript errors.
- Verified `changeLanguage("zh-CN")` resolves to `["zh-CN", "en"]` hierarchy (not `["en"]`).
- Verified `LanguageSwitcher` shows correct language label after switching.
- Verified Settings page loads without JSON parse error.

## Checklist

- [x] Code follows [CONTRIBUTING.md](CONTRIBUTING.md) guidelines

### DIFF
--- a/frontend/src/components/layout/Layout.tsx
+++ b/frontend/src/components/layout/Layout.tsx
@@ -337,11 +337,15 @@ function LanguageSwitcher() {
     };
   }, [open]);
 
-  // i18n.language may be a regional variant like "ja-JP" even though only
-  // "ja" is in SUPPORTED_LANGUAGES (nonExplicitSupportedLngs resolves the
-  // resources but keeps the full code). The i18n.languages array contains
-  // both the detected and resolved codes, so we match against any entry.
+  // i18n.language (singular) is the primary active language. We try an exact
+  // match first against SUPPORTED_LANGUAGES. If that fails (e.g. a regional
+  // variant like "ja-JP"), we fall back to i18n.languages (plural) which
+  // includes both the detected and resolved codes. NOTE: i18n.languages
+  // always contains the fallback language ("en"), so it must NOT be the
+  // primary match — otherwise "en" being first in SUPPORTED_LANGUAGES
+  // would always win and the switcher would never show any other language.
   const current =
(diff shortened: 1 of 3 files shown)


warren618:
Thanks - this has landed on main via the maintainer integration stack, preserving your zh-CN/i18n fix as commit e06e6c6 and adding explicit zh-CN regression coverage in 0e952ef. Closing this PR to avoid a duplicate merge.

PR #377, opened 2026-07-02
feat(trading): opt-in TAP mode — full Alpaca key isolation (reads + orders + cancel)

0xZKnw:
## Summary

- Opt-in "TAP mode": the **entire** Alpaca egress — order placement, cancel, and all five reads — routes through the TAP credential proxy instead of the broker SDK.
- With TAP on, the agent process holds **no Alpaca key** (and doesn't even need `alpaca-py`): the secret lives in TAP, referenced by `<CREDENTIAL:...>` placeholders and injected server-side. Writes (place/cancel) block on human approval; reads are GETs that TAP auto-approves — isolation, ~zero friction.
- Off by default — unset `TAP_*` env vars keep the existing direct-SDK path. Alpaca-only for now (the one connector that's both order-enabled and header-auth).

## Why

The Alpaca connector loads the API key/secret into the agent process, so a prompt-injected agent can place orders on its own **and** read the raw key out of process to exfiltrate it. TAP mode keeps the secret in TAP and routes every broker call through it: order placement and cancel require an out-of-band human approval, and the reads no longer keep the key in-process — so "the agent holds no Alpaca key" is fully true, not just true on the order path.

## Changes

- `agent/src/trading/tap_forward.py` (new): stdlib-only forwarder — POSTs to TAP `/forward`, polls the approval, returns a structured result; config loaded from every `.env` candidate (`TAP_`-prefixed keys only); fails closed.
- `agent/src/trading/connectors/alpaca/sdk.py`:
  - **order placement** routes through TAP when configured; a deterministic `client_order_id` makes an approval-race / poll-timeout retry idempotent (Alpaca dedups it instead of double-placing).
  - **`cancel_order`** routes through TAP as a human-approved write (`DELETE /v2/orders/{id}`).
  - the **five reads** route through TAP (account/positions/orders on the trading host; quote/bars on `data.alpaca.markets`); market-data's abbreviated keys are aliased back to the SDK model names.
  - SDK/key made optional when TAP is on (`check_status` / `_missing_fields`).
- `README.md`: "TAP Mode" section (full-egress + two allowed hosts). `agent/.env.example`: commented `TAP_*` block.
- `agent/tests/test_alpaca_tap_routing.py`: 14 cases — place/cancel gating + deny fails-closed, idempotent `client_order_id`, each read routes to the correct host with placeholder-only headers, market-data key normalization, credential name overridable, direct-SDK path untouched when disabled.

## Test Plan

- [x] Full suite passes with TAP off (direct-SDK path unchanged): 4594 passed (the only failures are pre-existing Windows-only loader/oauth/factor env issues, unrelated to this change)
- [x] New tests: `agent/tests/test_alpaca_tap_routing.py` (14 cases)
- [x] Tested manually against Alpaca paper, through TAP, with the agent process holding no key: an approved order lands; a denied order/cancel returns an error and never reaches Alpaca; `get_open_orders` (read) auto-approves with no human step; a placed order is then cancelled via TAP

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) — changes live in `src/trading/`
- [x] No hardcoded values (secrets via TAP placeholders + `.env`; upstream hosts from the connector's existing constants)
- [x] Documentation updated (README "TAP Mode" + `.env.example`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1056,6 +1056,48 @@ The agent exposes connector-scoped tools named `trading_connections`,
 Live-broker raw MCP tools are not registered directly as `mcp_<broker>_*`.
 No IBKR order-placement tool is registered.
 
+### 🔐 TAP Mode — credential isolation & human-approved orders
+
+**Opt-in, off by default.** If the `TAP_*` variables below are unset, order
+placement behaves exactly as before (direct broker SDK) — nothing changes.
+
+[TAP](https://tap.human.tech) (Tool Authorization Protocol) is a credential
+proxy: the agent never holds the raw broker API secret, and consequential writes
+are gated on **human approval**. With TAP mode on, an Alpaca order is sent to the
+TAP proxy's `/forward` endpoint instead of the broker SDK; TAP injects the real
+key server-side **after a human approves**, then forwards the order to Alpaca.
+
+- The agent process holds **no Alpaca key** on the order path — it references the
+  secret by name (`<CREDENTIAL:alpaca.key_id>`) and TAP substitutes it.
+- An order **cannot reach the broker without a human approving it**. Even a
+  prompt-injected "buy now" is held for approval; deny it and the order never
+  reaches Alpaca.
+- `allowed_hosts` on the TAP credential pins where the key may be sent, so a
+  tampered target is rejected (403) before injection.
+
+**Enable it:**
+
+1. In the TAP dashboard, create a **multi-secret** credential named `alpaca`
+   holding your Alpaca key pair as fields `key_id` and `secret_key`, with allowed
+   host `paper-api.alpaca.markets` (or the live host), assigned to your agent.
+2. Add to `agent/.env`:
+
+| Variable | Required | Description |
+|----------|:--------:|-------------|
+| `TAP_PROXY_URL` | Yes | TAP proxy base URL (e.g. `https://proxy.tap.human.tech`) |
+| `TAP_AGENT_KEY` | Yes | Your TAP agent API key (secret) |
+| `TAP_ALPACA_CREDENTIAL` | No | TAP credential name for Alpaca (default `alpaca`) |
+| `TAP_APPROVAL_TIMEOUT` | No | Seconds to wait for a human decision (default `300`) |
+
+When an order is placed, approve or deny it in your TAP channel (Telegram /
+dashboard). An approved order is forwarded to Alpaca; a denied or timed-out order
+returns an error and is **never sent**.
+
+**Scope:** currently covers Alpaca **order placement** (the write path). Read
+calls (positions/account/quote) still use the broker SDK, and HMAC-signed brokers
+are follow-ups. The hook is additive — it lives inside the Alpaca connector's
+`place_order()` and leaves the live mandate gate unchanged.
+
 ### Config reference
 
 | Field | Type | Default | Description |
--- a/README.md
+++ b/README.md
@@ -1056,30 +1056,39 @@ The agent exposes connector-scoped tools named `trading_connections`,
 Live-broker raw MCP tools are not registered directly as `mcp_<broker>_*`.
 No IBKR order-placement tool is registered.
 
-### 🔐 TAP Mode — credential isolation & human-approved orders
+### 🔐 TAP Mode — full credential isolation & human-approved writes
 
-**Opt-in, off by default.** If the `TAP_*` variables below are unset, order
-placement behaves exactly as before (direct broker SDK) — nothing changes.
+**Opt-in, off by default.** If the `TAP_*` variables below are unset, the
+connector behaves exactly as before (direct broker SDK) — nothing changes.
 
 [TAP](https://tap.human.tech) (Tool Authorization Protocol) is a credential
 proxy: the agent never holds the raw broker API secret, and consequential writes
-are gated on **human approval**. With TAP mode on, an Alpaca order is sent to the
-TAP proxy's `/forward` endpoint instead of the broker SDK; TAP injects the real
-key server-side **after a human approves**, then forwards the order to Alpaca.
+are gated on **human approval**. With TAP mode on, **every** Alpaca call — order
+placement, cancel, and the reads (account/positions/orders/quote/bars) — is sent
+to the TAP proxy's `/forward` endpoint instead of the broker SDK; TAP injects the
+real key server-side, then forwards upstream.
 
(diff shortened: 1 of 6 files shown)


warren618:
Thanks — the threat model here (a prompt-injected agent reading broker keys out of process memory) is real, and key isolation is a direction we care about. Before code review we need to settle the trust model, so three questions:

1. **Who operates TAP?** If it's a third-party hosted service, routing broker credentials through it conflicts with our no-custody stance — nothing between the user and the broker should hold keys or funds. If it's self-hosted / local to the user, that objection largely goes away. Please clarify the intended deployment.
2. **Autonomy model.** TAP's per-write human approval is per-order confirmation, which is the opposite of the bounded-autonomy mandate design we use for live trading (mandate set once up front, no per-order clicks, kill switch). As an opt-in mode for paper accounts or very conservative users it can coexist, but it must stay opt-in and clearly separate from the mandate-gate path.
3. **Gate integrity.** Please confirm the TAP path does not bypass the fail-closed mandate gate (`sdk_order_gate.py`) — every order should pass the same gate before egress regardless of transport, TAP or direct SDK.

Answers to 1 and 3 unblock the code review.

0xZKnw:
Thanks for the thoughtful questions — answers below.

**1. Who operates TAP?** TAP is built by [Human Tech](https://tap.human.tech) and is open source: [holonym-foundation/tap-oss](https://github.com/holonym-foundation/tap-oss). It self-hosts via Docker (a single compose: proxy + Postgres), so it can run entirely local to the user — broker keys then never leave the user's own infrastructure (stored AES-256-GCM encrypted, injected at egress, scrubbed from responses). A managed instance exists for convenience, but for the no-custody stance the self-hosted mode is the intended fit. The integration itself is deployment-agnostic: `TAP_PROXY_URL` points wherever the user runs it. And TAP never holds funds and never initiates anything — it's a policy engine the agent's outbound request passes through; it can only forward, gate, or deny what the agent already asked for.

**2. Autonomy model** — agreed, and per-write approval is TAP's *default*, not its only mode. Approval policy is per-credential configuration: order writes can be auto-approved (method rules / URL whitelist), in which case TAP contributes pure key isolation with zero per-order clicks — composing with your mandate design rather than replacing it (the mandate decides what's allowed; TAP just keeps the keys out of the agent process). The PR keeps everything opt-in and off by default, orthogonal to the mandate path.

**3. Gate integrity** — confirmed by construction: the TAP branch lives *inside* `connector.place_order`, and `sdk_order_gate.execute_live_order` only reaches `place_order` on ALLOW (mandate / expiry / halt flag / notional caps are all evaluated before any egress). Same gate, same ordering, regardless of transport — TAP is just the wire underneath the connector call.

warren618:
Thanks for the clear answers — self-hosted deployment resolves the custody question, and the code confirms your Q3 claim: the TAP branch is the first statement of `connector.place_order`, strictly downstream of the mandate gate, and with the env vars unset the direct path is byte-for-byte unchanged. The SSRF guard on the poll URL is a nice touch.

Code review done. Two test additions before merge:

1. **Direct tests for `tap_forward.py`.** All 14 current tests mock `tap_forward.forward`, so the 222-line module — including the poll-URL exfil guard and the fail-closed error paths — has zero direct coverage. Please add unit tests with mocked `urllib`: connection error → `ok=False`; 202 → poll → `forwarded` → `ok=True`; `denied`/`timed_out` → `ok=False`; an absolute or non-`/`-prefixed `poll_url` in the response → rejected.
2. **A regression test that the live mandate gate precedes TAP.** Live profile + TAP enabled: a mandate DENY must return blocked with `tap_forward.forward` never called; an ALLOW routes the order through TAP. This is the exact property the red line protects — we want it locked by a test, not just by construction.

Two smaller items, fine in the same push: (3) `_load_env_into_environ` only catches `OSError` — a non-UTF-8 `.env` raises `UnicodeDecodeError` through every read/order call, please broaden the except; (4) document the approval-race edge: if TAP forwards an order but the poll times out, the gate returns error and `max_trades_per_day` under-counts by one — a known-limitation note (or reconcile-on-retry) is fine.

One README note: recommend separate TAP credentials for paper vs live (or `allowed_hosts` pinned per host) so the paper/live separation stays crisp end to end.

0xZKnw:
All four items are in, pushed as 1b0c266 — 60/60 tests green.

1. **Direct `tap_forward.py` coverage** — new `agent/tests/test_tap_forward.py` (13 tests, mocked `urllib`, no network): connection error → `ok=False`; unconfigured TAP → fail-closed with no HTTP call; 202 → poll → `forwarded` → `ok=True` (and `forwarded` with an upstream non-2xx still `ok=False`); `denied`/`timed_out`/`error` → `ok=False`; poll deadline → `ok=False`. The exfil guard is covered explicitly: absolute, protocol-relative (`//host/path`) and non-`/`-prefixed poll URLs are refused, with an assertion that the agent key never travels anywhere but the configured base. Writing those tests surfaced that the old `startswith("/")` check *accepted* protocol-relative `//host/path` values — harmless today (string concat keeps the configured host) but a footgun under any future URL-joining, so the guard now requires a single leading `/`.

2. **Gate-precedes-TAP regression** — three tests at the bottom of `test_alpaca_tap_routing.py` run the *real* Alpaca connector module through the *real* `sdk_order_gate.execute_live_order` with TAP enabled: a mandate DENY (no mandate on file) and a tripped halt flag both return `blocked` with `tap_forward.forward` never called — not even for a read; an in-bounds ALLOW routes the order through TAP to the live host and carries the gate's `live_action` audit record. The red line is now locked by tests, not just by construction.

3. **`_load_env_into_environ`** now also catches `UnicodeDecodeError`, treating a non-UTF-8 `.env` like a missing file (covered by a test with a `\xff\xfe` file).

4. **Approval-race edge documented** — a "Known limitation" note in the README TAP section (poll timeout after a forward → the gate reports an error and `max_trades_per_day` under-counts by one; the deterministic `client_order_id` prevents a retry from double-placing, and we recommend checking open orders after a TAP timeout before retrying), plus a comment at the exact spot in `tap_forward.py`. Kept as a documented limitation for now; reconcile-on-retry is a natural follow-up if you want it.

README also picked up your credential note: separate `alpaca-paper` / `alpaca-live` TAP credentials (selected via `TAP_ALPACA_CREDENTIAL`) with `allowed_hosts` pinned per host, so TAP structurally refuses to send the paper key to the live host and vice versa.

PR #389, opened 2026-07-03
fix(security): block CGNAT/mesh ranges in SSRF guard; close QQ media redirect-to-internal

hobostay:
## Summary

Two SSRF gaps in the channel media-fetch path, both verified against current `main`.

### 1. Central SSRF IP guard misses RFC 6598 (100.64.0.0/10) — `channels/utils.py`

`_is_private()` (the helper under `validate_url_target`, used by **every** channel: QQ, DingTalk, NapCat, Telegram, and `channelsui`) only checked `is_loopback | is_link_local | is_private`. `ipaddress.is_private` is **`False`** for RFC 6598 shared address space (`100.64.0.0/10`) — the default range used by **Tailscale** and other CGNAT/mesh networks — so those hosts passed validation and were fetched directly as channel media.

```
>>> ipaddress.ip_address("100.64.0.1").is_private
False          # slipped through the old guard
```

**Fix:** `return not addr.is_global or addr.is_multicast`. `is_global` is `False` for every non-globally-routable range (loopback, link-local, RFC 1918, unspecified, reserved, 100.64/10), and multicast is added explicitly because `is_global` is `True` for multicast (`224.0.0.0/4`, `ff00::/8`) — the old code only caught IPv6 multicast, so this also closes an IPv4-multicast gap. This matches the check already used by `web_reader_tool._url_allowed`.

The narrow `allow_loopback=True` opt-in still works — that allowance is evaluated before `_is_private` runs.

### 2. QQ outbound media follows redirects without re-validation — `qq.py`

`_read_media_bytes()` calls `validate_url_target(media_ref)` once, then fetches with `allow_redirects=True`. `media_ref` is **agent-controlled** (it comes from `msg.media`, which is routinely influenced by untrusted external content — web pages via `read_url`, MCP tool output, news). A public URL that returns `302 Location: http://169.254.169.254/latest/meta-data/...` is followed unchecked, and the internal body is read (`await resp.read()`) and uploaded as a media attachment — exfiltrating cloud credentials / internal content.

DingTalk (`follow_redirects=False` + per-hop `validate_resolved_url`) and NapCat (`allow_redirects=False`, reject 3xx) already defend against this; **QQ was the outlier**.

**Fix:** `allow_redirects=False` and fail closed on `3xx`, mirroring NapCat. The separate inbound download at `qq.py:618` fetches a trusted QQ CDN (`multimedia.nt.qq.com`) URL and is intentionally left as-is.

## Test plan

- New `agent/tests/test_url_target_security.py` (29 cases) — the central guard (`validate_url_target` / `_is_private`) previously had **no tests at all**.
  - `100.64.0.0/10` and the rest of the non-global space are blocked; public IPs / IPv6 are allowed; loopback is blocked by default and allowed only via the opt-in.
  - QQ outbound media rejects `301/302/303/307/308`, allows `2xx` (still without following redirects), and rejects `4xx/5xx`.
- Existing `agent/tests/test_web_reader_security.py` still passes (48 total).
- All assertions use IP-literal hosts, so `getaddrinfo` is a local parse — **no network required**.

```
$ python -m pytest tests/test_url_target_security.py tests/test_web_reader_security.py -q
................................................                [100%]
48 passed
```

## Notes / out of scope

- I deliberately did **not** change `web_reader_tool._url_allowed` (also flagged in research as a weaker self-rolled check for hostnames). It forwards to the third-party Jina reader rather than fetching directly, and it is intentionally network-free; routing it through `validate_url_target` (which does DNS resolution) would break `test_read_url_allows_public_http_targets` and add a DNS round-trip to every read. Worth a separate discussion.
- DNS-rebinding (TOCTOU between `validate_url_target`'s `getaddrinfo` and the client's connect) affects all channels and needs a pinned-IP fetch helper — left for a follow-up.

### DIFF
--- a/agent/src/channels/qq.py
+++ b/agent/src/channels/qq.py
@@ -409,7 +409,20 @@ class QQChannel(BaseChannel):
         if not self._http:
             self._http = aiohttp.ClientSession(timeout=aiohttp.ClientTimeout(total=120))
         try:
-            async with self._http.get(media_ref, allow_redirects=True) as resp:
+            # Do not auto-follow redirects: ``media_ref`` is agent-controlled
+            # (taken from ``msg.media``), and a public URL that 302s to an
+            # internal address (169.254.169.254, 100.64/10, …) would bypass the
+            # pre-flight ``validate_url_target`` check and exfiltrate internal
+            # content as a media attachment. Fail closed on 3xx instead — this
+            # mirrors napcat.py's inbound image download.
+            async with self._http.get(media_ref, allow_redirects=False) as resp:
+                if 300 <= resp.status < 400:
+                    self.logger.warning(
+                        "outbound media download redirect rejected url={} status={}",
+                        media_ref,
+                        resp.status,
+                    )
+                    return None, None
                 if resp.status >= 400:
                     self.logger.warning(
                         "outbound media download failed status={} url={}",
--- a/agent/src/channels/utils.py
+++ b/agent/src/channels/utils.py
@@ -146,11 +146,21 @@ def _is_allowed_loopback_target(
 
 
 def _is_private(addr: ipaddress.IPv4Address | ipaddress.IPv6Address) -> bool:
-    """Check whether *addr* is private, link-local, or loopback."""
-    if addr.is_loopback or addr.is_link_local or addr.is_private:
-        return True
-    # Also block well-known metadata endpoints
-    if isinstance(addr, ipaddress.IPv4Address):
-        return False
-    # IPv6: block unspecified, multicast, unique-local
-    return addr.is_multicast or addr.is_unspecified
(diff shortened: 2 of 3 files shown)


warren618:
Thanks @hobostay, merged this via normal merge so the original commit authorship and co-author credits are preserved.

Local verification before merge:
- `PYTHONPATH=agent python -m pytest agent/tests/test_url_target_security.py agent/tests/test_web_reader_security.py -q` -> 48 passed
- `python -m compileall -q agent/src/channels/qq.py agent/src/channels/utils.py agent/tests/test_url_target_security.py`

I accepted the focused local verification here despite no GitHub checks on the branch because the security fix is compact and directly covers the SSRF gaps.

PR #409, opened 2026-07-06
feat(data): add audited loader and PIT provenance reports

Elfsa-Miranda:
## Summary

This PR adds data reliability and point-in-time provenance support in observe mode.

It introduces data contracts, source/access provenance, PIT timestamp models, data audit reports, source circuit breaker state, and market-rule warnings while preserving the existing data loader public interface.

## Motivation

Backtests and research outputs are only useful when the data path is auditable.

This PR makes data access more transparent by recording source selection, fallback paths, field coverage, PIT warnings, quality warnings, and circuit-breaker state. It also preserves the existing `local:` no-fallback behavior.

## Scope

This PR adds:

- `DataSetContract`
- `DataAccessContract`
- `DataAuditReport`
- source manifest support
- `AuditedDataLoader`
- PIT timestamp models
- PIT checker
- data quality validators
- source `CircuitBreaker`
- initial market-rule warning support
- tests for audit models, PIT checks, local no-fallback behavior, loader compatibility, and circuit breaker behavior

It also integrates audit metadata into:

- market data access
- backtest runner data access
- reliability test fixtures

## Non-goals

This PR does **not**:

- change `DataLoaderProtocol.fetch(...)`
- replace the existing loader registry
- change the public `get_market_data` tool schema
- silently change source selection behavior
- allow `local:` sources to fall back to network sources
- enforce PIT failures by default
- change live trading behavior
- introduce real external network I/O into tests

## Compatibility and safety

The data audit layer is implemented as an additive wrapper around the existing data loading path.

Important compatibility guarantees:

- `DataLoaderProtocol.fetch(...)` remains unchanged
- explicit local sources do not fall back to network sources
- observe mode records warnings and provenance without changing normal low-risk flows
- data audit reports are schema-versioned
- circuit breaker state is recorded but should not alter explicit local semantics

## Tests

Targeted tests:

```powershell
.\.venv\Scripts\python.exe -m pytest agent/tests/reliability -q --tb=short


warren618:
Closed as superseded by the merged IRR-AGL stack PR #416, which carries this phase forward into main under the same contributor credit.

PR #429, opened 2026-07-07
fix(backtest): create artifacts dir before writing validation.json

### LINKED ISSUE #427: Backtest engine writes `artifacts/validation.json` before the artifacts dir is created → FileNotFoundError when validation is enabled and artifacts/ doesn't pre-exist (hit via swarm portfolio_manager)
## Summary

In `backtest/engines/base.py`, when `config["validation"]` is truthy the engine writes `run_dir/artifacts/validation.json` **before** the `artifacts/` directory is created. The directory is only created later by `_write_artifacts()`. If `run_dir/artifacts/` doesn't already exist, the run crashes with `FileNotFoundError`.

This is not platform-specific (the missing-parent-dir write fails everywhere), but I hit it on Windows via an `investment_committee` swarm run, where the `portfolio_manager` sub-agent's `run_dir` had no pre-created `artifacts/`.

## Where

`backtest/engines/base.py`, `run_backtest()`:

```python
# 7. Validation (optional — triggered by config["validation"])
if config.get("validation"):
    from backtest.validation import run_validation
    v_results = run_validation(...)
    m["validation"] = v_results
    # Write validation.json artifact
    v_path = run_dir / "artifacts" / "validation.json"
    v_path.write_text(json.dumps(v_results, ...), encoding="utf-8")   # <-- parent may not exist

# 8. Artifacts
self._write_artifacts(...)   # <-- this is what does run_dir/"artifacts".mkdir(parents=True, exist_ok=True)
```

`_write_artifacts()` (same file) is the only place that creates `run_dir/artifacts`:

```python
out = run_dir / "artifacts"
out.mkdir(parents=True, exist_ok=True)
```

Because step 7 runs before step 8, the `validation.json` write can hit a non-existent parent directory.

## Observed traceback

```
Runner stderr:
  File ".../backtest/runner.py", line 673, in <module>
    main(Path(sys.argv[1]))
  File ".../backtest/runner.py", line 529, in main
    market_engine.run_backtest(config, loader, signal_engine, run_dir, bars_per_year=bars_per_year)
  File ".../backtest/engines/base.py", line 478, in run_backtest
    v_path.write_text(json.dumps(v_results, indent=2, ensure_ascii=False), encoding="utf-8")
  File ".../pathlib/__init__.py", line 809, in write_text
    ...
FileNotFoundError: [Errno 2] No such file or directory:
'...\\.swarm\\runs\\swarm-...\\artifacts\\portfolio_manager\\artifacts\\validation.json'
```

## Reproduce

Run any backtest through the market engine with `config["validation"]` set (truthy) and a `run_dir` whose `artifacts/` subdir does not already exist. In my case:

```
vibe-trading --swarm-run investment_committee '{"target":"600519.SH","market":"a_share"}'
```

The `portfolio_manager` agent generated a config with validation enabled and ran the backtest; its workspace `run_dir` had no pre-created `artifacts/`, so the validation write raised `FileNotFoundError`. (The swarm itself was fault-tolerant and still produced a final report, but the PM's backtest artifact was lost.)

## Fix

Ensure the parent directory exists before writing, e.g. right before the write:

```python
v_path = run_dir / "artifacts" / "validation.json"
v_path.parent.mkdir(parents=True, exist_ok=True)
v_path.write_text(json.dumps(v_results, indent=2, ensure_ascii=False), encoding="utf-8")
```

(or move the validation write to after `_write_artifacts()`).

## Environment

- vibe-trading-ai 0.1.10
- Windows 10, Python 3.14.5

isaveall:
## What

When `config["validation"]` is set, `BaseEngine.run_backtest()` writes `run_dir/artifacts/validation.json` (step 7) **before** `_write_artifacts()` creates `run_dir/artifacts` (step 8). If that directory doesn't already exist, the write raises `FileNotFoundError` and aborts the backtest.

This surfaced when running an `investment_committee` swarm — the `portfolio_manager` sub-agent's `run_dir` has no pre-created `artifacts/`, so its validation-enabled backtest crashed (the swarm was fault-tolerant and still finished, but the PM's backtest artifact was lost).

## Fix

Ensure the parent directory exists immediately before the write:

```python
v_path = run_dir / "artifacts" / "validation.json"
v_path.parent.mkdir(parents=True, exist_ok=True)
v_path.write_text(...)
```

One line + a short explanatory comment. The bug is not platform-specific.

## Test

Adds `TestValidationArtifactDir` to `tests/test_engine_robustness.py`: runs a backtest with validation enabled against a `run_dir` that has no pre-created `artifacts/`, and asserts `validation.json` is written.

- With the fix: **passes**.
- Reverting only the fix: **fails** with the exact `FileNotFoundError`.
- Full `tests/test_engine_robustness.py`: **32 passed**.

Fixes #427

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -474,8 +474,11 @@ class BaseEngine(ABC):
                 config, equity_series, self.trades, self.initial_capital, bars_per_year,
             )
             m["validation"] = v_results
-            # Write validation.json artifact
+            # Write validation.json artifact. The artifacts dir is normally
+            # created by _write_artifacts() below (step 8), so ensure it exists
+            # here to avoid a FileNotFoundError when run_dir/artifacts is absent.
             v_path = run_dir / "artifacts" / "validation.json"
+            v_path.parent.mkdir(parents=True, exist_ok=True)
             v_path.write_text(json.dumps(v_results, indent=2, ensure_ascii=False), encoding="utf-8")
 
         # 8. Artifacts
--- a/agent/tests/test_engine_robustness.py
+++ b/agent/tests/test_engine_robustness.py
@@ -517,3 +517,62 @@ class TestFullBacktestRobustness:
         engine = ChinaAEngine({"initial_cash": 1_000_000})
         engine._execute_bars(dates, data_map, close_df, target_pos, valid_codes)
         assert len(engine.equity_snapshots) == 20
+
+
+# ---------------------------------------------------------------------------
+# 6. Validation artifact — write must not assume artifacts/ already exists
+# ---------------------------------------------------------------------------
+
+
+class TestValidationArtifactDir:
+    def test_validation_json_written_when_artifacts_dir_absent(
+        self, tmp_path: Path
+    ) -> None:
+        """Enabling validation must not crash when run_dir/artifacts is absent.
+
+        The validation.json write (step 7) runs before _write_artifacts()
+        (step 8) creates run_dir/artifacts, so the write must create the
+        directory itself. Regression test for the FileNotFoundError hit when
+        run_dir has no pre-created artifacts/ (e.g. a swarm agent workspace).
+        """
+        dates = pd.bdate_range("2024-04-01", periods=3)
+        bars = pd.DataFrame(
+            {
+                "open": [10.0, 11.0, 12.0],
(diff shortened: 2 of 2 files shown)


warren618:
Belated note for the record: merged as-is (7311d6b) — clean 4-line fix, verified locally with a regression test before landing. Thanks!

Issue #431, opened 2026-07-07
Agent fails with GovernedToolRegistry object has no attribute _tools


Jaha-who:
I tested both official README setup paths:

Path A Docker:
- docker-compose up --build
- App starts
- Agent page fails with:
  'GovernedToolRegistry' object has no attribute '_tools'

Path B Local Python:
- python3 -m venv .venv
- source .venv/bin/activate
- pip install -e .
- cp agent/.env.example agent/.env
- vibe-trading serve --port 8899
- App starts
- Agent page fails with same error:
  'GovernedToolRegistry' object has no attribute '_tools'

Commit:
d007d5a5de7c3bd7140f636793715d5fe39e2bf0

Repo status:
clean / up to date with origin/main

Expected:
Agent should respond normally after following README setup.

Actual:
Agent returns runtime error.

Penn-Live:
+1

newgo:
I have also encountered the same issue.
Path B Local Python

<img width="1349" height="537" alt="Image" src="https://github.com/user-attachments/assets/f4b1b528-e991-4421-a6d2-82e45166ce76" />

newgo:
> I tested both official README setup paths:
> 
> Path A Docker:
> 
> * docker-compose up --build
> * App starts
> * Agent page fails with:
>   'GovernedToolRegistry' object has no attribute '_tools'
> 
> Path B Local Python:
> 
> * python3 -m venv .venv
> * source .venv/bin/activate
> * pip install -e .
> * cp agent/.env.example agent/.env
> * vibe-trading serve --port 8899
> * App starts
> * Agent page fails with same error:
>   'GovernedToolRegistry' object has no attribute '_tools'
> 
> Commit: [d007d5a](https://github.com/HKUDS/Vibe-Trading/commit/d007d5a5de7c3bd7140f636793715d5fe39e2bf0)
> 
> Repo status: clean / up to date with origin/main
> 
> Expected: Agent should respond normally after following README setup.
> 
> Actual: Agent returns runtime error.

I switched the code from the main branch to the tag v0.1.10, and that solved the problem.

<img width="1489" height="659" alt="Image" src="https://github.com/user-attachments/assets/9285fce2-54df-4024-a76b-08e9345141aa" />

warren618:
Duplicate of #433 — same root cause (`GovernedToolRegistry` not exposing `_tools` to `AgentContext`). Tracking the fix there. Thanks for the report!

Issue #420, opened 2026-07-07
[Bug] GovernedToolRegistry bypass paths + circuit breaker TOCTOU + path traversal HTTP status


shadowinlife:
## Description

**What happened:**

The `GovernedToolRegistry` in `agent/src/governance/runtime.py` exposes two design-level bypass paths that allow the governance policy layer (P10–P999) to be skipped entirely. Additionally, two implementation-level bugs exist in the circuit breaker and research card API.

**What I expected:**

All tool access paths through `GovernedToolRegistry` should be governed. Security violations (path traversal) should return appropriate HTTP status codes. The circuit breaker should use atomic read-write to prevent race conditions.

---

### Issue 1 — `get()` returns raw unwrapped tool (defense-in-depth)

**File:** `agent/src/governance/runtime.py` L41-42

```python
def get(self, name: str) -> Any:
    return self.inner.get(name)  # returns raw BaseTool
```

`get()` returns the raw `BaseTool` object from the inner registry. Any caller that obtains a tool via `registry.get(name)` and calls `tool.execute(**params)` would bypass all P10–P999 policy rules.

**Current impact (low):** All current call sites use `get()` only for metadata inspection (checking `is_readonly`, `repeatable`, `MCPRemoteTool` type), NOT for execution. Both `agent/src/agent/loop.py` (L1327) and `agent/src/swarm/worker.py` (L691) use `registry.execute()` for actual tool execution — the governed path.

**Risk:** A future contributor could accidentally call `tool.execute()` on the object returned by `get()`, creating an ungoverned execution path. This is a defense-in-depth gap, not an active exploit.

### Issue 2 — `self.inner` is a public attribute (defense-in-depth)

**File:** `agent/src/governance/runtime.py` L31

```python
self.inner = inner  # public attribute
```

`self.inner` is publicly accessible. Any code with a reference to the `GovernedToolRegistry` can call `registry.inner.execute(name, params)` to bypass all governance, or `registry.inner.get(name)` to obtain raw tools.

**Current impact (low):** No current code path exploits this. But it provides a trivial one-line bypass for any future code or community-contributed skill that gains access to the registry object.

### Issue 3 — Path traversal attempt returns HTTP 500 instead of 404

**File:** `agent/src/research_card/api.py` L158-159

```python
except ValueError as exc:
    raise HTTPException(status_code=500, detail="artifact path escapes artifact root") from exc
```

[...]

PR #405, opened 2026-07-06
docs: add IRR-AGL reliability RFC and tool inventory

Elfsa-Miranda:
## Summary

This PR adds an initial reliability/governance RFC and a read-only tool inventory script.

The goal is to document the current tool surface and establish a small, reviewable baseline for future reliability and governance improvements.

This PR intentionally keeps the scope narrow: it does not change runtime behavior, wrap tools, modify data loaders, alter MCP schemas, or touch live trading behavior.

## Motivation

Vibe-Trading already has a broad execution surface across local tools, MCP, API, backtesting, scheduled research, swarm workers, and live trading safety components.

Before adding any reliability or governance behavior, this PR introduces a lightweight RFC and a tool inventory script so that future changes can be reviewed against the existing tool surface and safety boundaries.

## Scope

This PR adds:

- `docs/reliability-governance-rfc.md`
- `docs/irr-agl-readiness-check.md`
- `agent/scripts/dump_tool_inventory.py`
- tool inventory smoke tests
- readiness / environment contract tests

The inventory script is read-only and reports tool metadata such as:

- tool name
- module
- readonly / repeatable metadata
- surface/risk guesses where available

## Non-goals

This PR does **not**:

- change `BaseTool`
- change `ToolRegistry`
- change tool execution behavior
- intercept or deny any tool call
- change data loader behavior
- change `local:` data source behavior
- change MCP tool schemas
- change API behavior
- change live trading safety behavior
- add new runtime dependencies
- introduce external network I/O into tests

## Compatibility and safety

This is intended to be a documentation and inventory-only baseline.

It should not affect:

- README quickstart flows
- existing tool registry behavior
- existing loader behavior
- existing backtest behavior
- existing live mandate / kill switch / order guard behavior

## Tests

Targeted tests:

```powershell
.\.venv\Scripts\python.exe -m pytest agent/tests/test_tool_inventory_smoke.py agent/tests/reliability/test_environment_contract.py -q --tb=short

PR #434, opened 2026-07-08
fix(governance): expose _tools passthrough on GovernedToolRegistry

### LINKED ISSUE #433: [Bug] Agent crashes with "'GovernedToolRegistry' object has no attribute '_tools'" in session runtime
Hit this right after pulling latest main (the IRR-AGL governance stack from #405/#416). Every agent message in the Web UI fails immediately.

## Description

`GovernedToolRegistry` in `src/governance/runtime.py` wraps the ToolRegistry but does not expose `_tools`. `AgentContext` still reads it directly when it builds the system prompt:

- `src/agent/context.py:189` — `len(self.registry._tools)`
- `src/agent/context.py:256` — `for tool in self.registry._tools.values()`

`src/session/service.py:259` wraps the registry with `govern_registry(...)` before handing it to `AgentContext` (`:271`), so the session runtime always trips over this. It is not mode-gated either: `govern_registry` returns a `GovernedToolRegistry` even when the governance mode is `off`.

The class docstring says it "preserves the existing ToolRegistry surface", so `_tools` looks like it was just missed.

## Steps to Reproduce

1. Start the server with the session runtime (Web UI agent chat, or `POST /sessions/{id}/messages`).
2. Send any message to the agent.
3. It fails immediately while building the system prompt.

## Error Logs

```
GovernedToolRegistry object has no attribute _tools
```

Shows up in the run card as "执行失败 / 点击重试".

## Interface

Web UI (and any other session-runtime path, e.g. REST `/sessions/*/messages`).

## LLM Provider

Provider-independent — it fails before the model call, while constructing the prompt. Saw it on Moonshot / Kimi.

## Version

main @ `d007d5a` (right after the 2026-07-07 governance landing).

## Environment

WSL2, Python 3.13, session runtime enabled.

---

The fix I am running locally: a read-only `_tools` property on `GovernedToolRegistry` that returns `self.inner._tools`, plus a regression test. Can open a PR if that direction looks right.

yxhuang:
## Summary

- Restore the `_tools` map on `GovernedToolRegistry` so the agent stops crashing in the session runtime.

## Why

Closes #433. After the IRR-AGL governance stack, the registry gets wrapped in `GovernedToolRegistry` before it reaches `AgentContext`, which reads `self.registry._tools` while building the system prompt (`context.py:189`, `:256`). The wrapper doesn't expose `_tools`, so every agent turn in the session runtime dies with `'GovernedToolRegistry' object has no attribute '_tools'`. It's not mode-gated — `govern_registry` wraps the registry even when governance is `off`.

## Changes

- `src/governance/runtime.py`: add a read-only `_tools` property delegating to `self.inner._tools` (the class already documents that it "preserves the existing ToolRegistry surface").
- `tests/governance/test_governed_registry.py`: regression test for the passthrough (exercises `len()` and `.values()` the way `context.py` uses it).

## Test Plan

- [x] Existing tests pass (`pytest agent/tests/governance -q`)
- [x] New test added
- [x] Tested manually — agent replies again in the Web UI session runtime instead of erroring out

## Checklist

- [x] No changes to protected areas — this only touches `src/governance/`
- [ ] Documentation updated (n/a — internal fix, no user-facing surface change)

### DIFF
--- a/agent/src/governance/runtime.py
+++ b/agent/src/governance/runtime.py
@@ -38,6 +38,13 @@ class GovernedToolRegistry:
     def tool_names(self) -> list[str]:
         return list(getattr(self.inner, "tool_names", []))
 
+    @property
+    def _tools(self) -> Any:
+        """Preserve the wrapped registry's private tool map for legacy read
+        access (e.g. AgentContext iterates ``registry._tools`` when rendering the
+        system prompt). Read-only passthrough to the inner ToolRegistry."""
+        return getattr(self.inner, "_tools", {})
+
     def get(self, name: str) -> Any:
         return self.inner.get(name)
 
--- a/agent/tests/governance/test_governed_registry.py
+++ b/agent/tests/governance/test_governed_registry.py
@@ -63,6 +63,22 @@ def _cache(risk: RiskLevel, *, surface: ToolSurface = ToolSurface.CLI) -> Manife
     return ManifestCache({"counting": manifest}, surface=surface)
 
 
+def test_tools_property_passthrough_preserves_legacy_surface() -> None:
+    """AgentContext renders the system prompt via ``registry._tools`` (len +
+    .values()); the governance wrapper must preserve that private-map surface
+    by delegating to the inner ToolRegistry."""
+    inner, tool = _registry()
+    governed = GovernedToolRegistry(
+        inner,
+        manifest_cache=_cache(RiskLevel.R0_READ),
+        context=RuntimeContext(surface=ToolSurface.CLI, mode="off"),
+    )
+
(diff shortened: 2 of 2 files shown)


warren618:
After more time with the IRR-AGL stack on main, we've decided to revert it (#405/#416) rather than fix forward: it wrapped the tool registry in all session paths (even `mode=off`), which broke every session-runtime chat within a day (#433, #431), modified protected core files, and its enforcement layer already had known bypass paths (#420). Your fix was correct for the wrapper — this isn't a quality judgment on it, the whole layer is gone. Thanks for the quick turnaround and the precise issue report.

PR #451, opened 2026-07-09
fix(api): tolerate path-less routes when probing for a root mount

### LINKED ISSUE #450: [Bug] docker can't start
### Description

AttributeError: '_IncludedRouter' object has no attribute 'path'

### Steps to Reproduce

AttributeError: '_IncludedRouter' object has no attribute 'path'

### Error Logs
File "/usr/local/bin/vibe-trading", line 8, in
sys.exit(main())
^^^^^^

```shell
File "/app/agent/cli/main.py", line 1302, in main
return int(_legacy.main(raw_argv))
^^^^^^^^^^^^^^^^^^^^^^
File "/app/agent/cli/_legacy.py", line 5370, in main
return serve_main(raw_argv[1:])
^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/agent/cli/_legacy.py", line 246, in serve_main
return api_serve_main(argv)
^^^^^^^^^^^^^^^^^^^^
File "/app/agent/api_server.py", line 349, in serve_main
if not any(route.path == "/" for route in app.routes):
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/agent/api_server.py", line 349, in
if not any(route.path == "/" for route in app.routes):
^^^^^^^^^^
AttributeError: '_IncludedRouter' object has no attribute 'path'
```

### Interface

dcoker
### LLM Provider

_No response_

### Version

_No response_

### Environment

_No response_

Robin1987China:
## Summary

`serve_main` crashes on startup with `AttributeError: '_IncludedRouter' object has no attribute 'path'` on newer FastAPI, so the server (notably the Docker image) never comes up.

## Why

Before mounting the SPA static files, `serve_main` checks whether a `/` route already exists:

```python
if not any(route.path == "/" for route in app.routes):
```

Older FastAPI flattened `app.include_router(...)` into individual `APIRoute` objects, each with a `.path`. Newer FastAPI instead appends an `_IncludedRouter` marker to `app.routes` that has **no** `.path` attribute. This project calls `app.include_router(qveris_router)`, so on a newer FastAPI the probe hits that marker and raises `AttributeError`, aborting startup.

The dependency floor is unpinned (`fastapi>=0.104.0`), so the Docker build pulls a newer FastAPI than local dev — which is why it reproduces there and not locally.

## Changes

- `agent/api_server.py`: read the route path defensively with `getattr(route, "path", None)` so marker routes without a path are skipped instead of crashing.
- `agent/tests/test_serve_bind.py`: add a regression test that injects a path-less route ahead of a real `/` route and asserts `serve_main` starts.

## Test Plan

- [x] ruff check — clean
- [x] pytest (相关测试) — `tests/test_serve_bind.py` 15 passed (new test fails on the pre-fix code with the exact `AttributeError`, passes after)
- [x] pytest (全量 excl e2e) — 4871 passed, 9 skipped (2 unrelated failures preexist on main: a fundamental-factor projection test and a `bottleneck` backend test)

## Checklist

- [x] No hardcoded values

Closes #450

### DIFF
--- a/agent/api_server.py
+++ b/agent/api_server.py
@@ -346,7 +346,7 @@ def serve_main(argv: list[str] | None = None) -> int:
         print("[dev] Frontend: http://localhost:5173")
         print(f"[dev] API: http://localhost:{args.port}")
     elif frontend_dist.exists():
-        if not any(route.path == "/" for route in app.routes):
+        if not any(getattr(route, "path", None) == "/" for route in app.routes):
             app.mount("/", SPAStaticFiles(directory=str(frontend_dist), html=True), name="frontend")
         print(f"[prod] Frontend served from {frontend_dist}")
     else:
--- a/agent/tests/test_serve_bind.py
+++ b/agent/tests/test_serve_bind.py
@@ -13,6 +13,7 @@ build in CI) cannot satisfy or break them.
 
 from __future__ import annotations
 
+import types
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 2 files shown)


Robin1987China:
Looks like #450 was already fixed in 04ccb21 (same `getattr(route, "path", None)` guard), so this is redundant now. Closing — thanks!

warren618:
Right — 04ccb21 shipped the same guard a few hours earlier; race condition of a fast-moving day. Appreciated you jumping on #450 quickly, and the scheduled-routes test PR (#452) was merged. Thanks!

PR #452, opened 2026-07-09
test(api): cover the scheduled research routes

Robin1987China:
## Summary

Add route-level tests for the scheduled research endpoints (`POST` / `GET` / `DELETE /scheduled-runs`), which had no coverage after the api_server modular split.

## Why

The recent api_server refactor extracted the route modules into `agent/src/api/`. The scheduled research routes came through with only executor-level tests — the REST surface itself (status codes, validation, store side effects) was untested, so a regression in the route wiring would slip through. This fills that gap without touching any behavior.

## Changes

- `agent/tests/test_scheduled_routes.py`: new `TestClient` contracts driving each endpoint and asserting persisted store state:
  - create returns 201 and persists the job; id and `next_run_at` default when omitted; a malformed schedule is rejected with 422
  - list is newest-first, filterable by status, with `limit` bounds enforced (422 outside 1–200)
  - delete returns 204 and removes the job; unknown id returns 404; an unsafe id is rejected with 400
  - the store singleton is redirected to a temp file so the tests never touch the runtime root

## Test Plan

- [x] ruff check — clean
- [x] pytest (相关测试) — `tests/test_scheduled_routes.py` 9 passed
- [x] pytest (全量 excl e2e) — 4879 passed, 9 skipped (2 unrelated failures preexist on main: a fundamental-factor projection test and a `bottleneck` backend test)

## Checklist

- [x] No hardcoded values

### DIFF
--- a/agent/tests/test_scheduled_routes.py
+++ b/agent/tests/test_scheduled_routes.py
@@ -1,0 +1,169 @@
+"""Route-level contracts for the scheduled research endpoints.
+
+Exercises the REST surface mounted by ``register_scheduled_routes``:
+``POST /scheduled-runs`` (create), ``GET /scheduled-runs`` (list + filter),
+and ``DELETE /scheduled-runs/{job_id}`` (cancel). Each test drives the app
+through ``TestClient`` and asserts the persisted store state, so the route
+wiring, validation, and status codes are covered end to end.
+
+The store singleton is redirected to a per-test ``tmp_path`` file so nothing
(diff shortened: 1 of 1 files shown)


warren618:
Merged as 1af6346 — verified locally before merge (fetched the branch, ran the 9 new route contracts against current main: all green). Exactly the kind of coverage the api_server split needed; the temp-file store redirect is the right pattern. Thanks!

PR #217, opened 2026-06-12
feat(i18n): add Chinese (zh-CN) frontend localization + A-share data sources

skloxo:
## Summary

This PR adds complete Chinese (zh-CN) internationalization to the Vibe-Trading frontend, plus new A-share data sources.

### i18n Changes

**Infrastructure:**
- Integrated `i18next` + `react-i18next` + `i18next-browser-languagedetector`
- Created `src/i18n/` directory with configuration and locale files
- Translation files: `zh-CN.json` (449 keys) and `en.json` (449 keys)

**Component Updates:**
- Updated 20+ components and pages to use `useTranslation()` hooks
- All user-facing strings now use translation keys instead of hardcoded English
- Reactive language switching via sidebar toggle button

**Coverage:**
- Navigation: Home, Agent, Alpha Zoo, Settings, Correlation
- Chat: Welcome screen, run status, message bubbles, thinking timeline
- Settings: All configuration options
- Charts: Validation panel, metrics
- Error messages and loading states

### A-Share Data Sources

- Added `baostock` and `tencent` data source providers for China A-shares
- Bypasses CDN restrictions for reliable data access in mainland China

### Testing

- ✅ Build passes (`npm run build`)
- ✅ No hardcoded Chinese strings in components (all via `zh-CN.json`)
- ✅ Language toggle button works reactively
- ✅ Default language: `zh-CN`, fallback: `en`

### Files Changed

- `src/i18n/index.ts` - i18next configuration
- `src/i18n/locales/zh-CN.json` - Chinese translations
- `src/i18n/locales/en.json` - English translations
- `src/components/layout/Layout.tsx` - Language toggle + reactive labels
- 20+ component files with `useTranslation()` integration
- `agent/skills/ashare-mootdx/` - A-share data fetcher skill

PR #453, opened 2026-07-09
feat(frontend): clarify inputs, unify error feedback, guard mandate commit

wison1717-maker:
## Summary
- Add aria-labels and an Enter/Shift+Enter hint to the Agent chat input, goal editor, and the previously label-less send button.
- Add a lightweight, dependency-free ConfirmDialog and require a second confirmation step (with a limits summary) before committing a real trading mandate.
- Surface previously swallowed errors in Compare.tsx - api.listRuns/api.getRun failures now toast instead of silently showing an empty state.
- Remove a duplicate toast+inline error in QVerisSettings save flow.
- Add a 'How it works' 4-step section to the Home page.
- Give the Agent page kill-switch row a border to visually separate it from lighter decorative status chips.
- Route several already-translated mandate strings (Cancel, Send adjustment, Committing..., error messages) through i18n instead of hardcoded English; add matching keys across all 5 locales.
- Fix a flex min-width bug causing long Home page step titles to overflow instead of wrap.

## Notes
- No new dependencies.
- No changes to trading/API/auth logic beyond the mandate confirm step and the two error-surfacing fixes noted above.

## Test plan
- [x] npm run build (tsc + vite) passes with zero TypeScript errors
- [x] npm run test:run - 243/243 existing tests pass
- [x] Manually exercised Home, Agent, Compare, and Settings pages via a local dev server (including a real HTTP 500 backend-failure scenario to confirm the new Compare.tsx error toast fires)


warren618:
Merged the UX commit to main as 0d0cf47 (your authorship preserved) — verified locally: `npm run build` green, 243/243 tests pass. The ConfirmDialog mandate guard, Compare error surfacing, aria-labels, i18n routing of mandate strings, and the Home flex fix are all in. Thanks, this is solid work.

The second commit (55608f3, replace zh-CN with zh-TW) is **declined**, for a reason independent of its quality: Simplified Chinese is a shipped feature (#217) with an active user base, and removing it is a product decision, not a polish item — it also wasn't mentioned anywhere in the PR description, which made it easy to miss. If you'd like Traditional Chinese in the app, a PR that **adds zh-TW alongside zh-CN** (locale file + switcher entry + parity tests extended to 6 locales) would be very welcome — the OpenCC s2twp conversion plus your three manual term corrections (帳/連接器/品質) looked careful.

Closing since the first commit is now on main.

PR #467, opened 2026-07-10
fix(loaders): honor interval in local loader by resampling

Shizoqua:
The local loader accepted an `interval` argument but passed it only to the cache key, so requesting e.g. `4H` against hourly data silently returned the file's native bars.

**Changes**
- Resample to the requested interval using standard OHLCV aggregation (open=first, high=max, low=min, close=last, volume=sum)
- Warn when a finer-than-source interval is requested (cannot fabricate bars)
- Warn on unsupported interval strings
- Make end-date filter inclusive of the whole end day so intraday bars on the final date survive for sub-daily intervals
- Bump loader cache version (2→3) so the end-date and resampling fixes cannot be masked by stale cache entries

**Testing**
- Test verifies 4H request against hourly bars produces correct aggregated buckets
- Test verifies upsampling returns source bars with a warning


warren618:
Merged — thanks. The cache-version bump so stale entries can't mask the fix is exactly right, and the inclusive end-day filter closes the sub-daily edge case. Verified locally: local-loader + cache helper tests green.

PR #418, opened 2026-07-07
Add FXMacroData integration

roberttidball:
## Summary
- Add FXMacroData as a first-class Vibe-Trading data source: REST client, numeric OHLCV loader, auto-routing for FX/macro symbols, fallback-chain integration, and generated-runner env pass-through for `FXMD_API_KEY`.
- Add 11 read-only FXMacroData agent/MCP tools for catalogue, indicators, release calendar, predictions, COT, commodities, rates, curves, central-bank news, market sessions, and risk sentiment.
- Add Settings UI/API support for saving and clearing `FXMD_API_KEY` locally with redacted responses only.
- Document REST/OpenAPI/MCP URLs, supported symbol forms, current audited coverage notes, and a remote MCP seed that does not embed secrets.
- Add focused regression coverage and small compatibility fixes for FastAPI 204 responses and FastMCP transport import layouts.

## Validation
- `python -m ruff check agent/backtest/loaders/fxmacrodata_client.py agent/backtest/loaders/fxmacrodata_loader.py agent/backtest/loaders/registry.py agent/mcp_server.py agent/src/api/scheduled_routes.py agent/src/api/settings_routes.py agent/src/config/schema.py agent/src/core/runner.py agent/src/market_data.py agent/src/tools/fxmacrodata_tools.py agent/src/tools/market_data_tool.py agent/src/tools/mcp.py agent/tests/test_fxmacrodata_loader.py agent/tests/test_fxmacrodata_tools.py agent/tests/test_registry.py agent/tests/test_market_data.py agent/tests/test_market_data_tool.py agent/tests/test_mcp_server_smoke.py agent/tests/test_runner_env.py agent/tests/test_settings_api.py agent/tests/test_agent_config.py agent/tests/test_data_routing_sources_subset.py`
- `python -m pytest agent/tests/test_fxmacrodata_loader.py agent/tests/test_fxmacrodata_tools.py agent/tests/test_registry.py agent/tests/test_market_data.py agent/tests/test_market_data_tool.py agent/tests/test_settings_api.py agent/tests/test_runner_env.py agent/tests/test_agent_config.py agent/tests/test_data_routing_sources_subset.py agent/tests/test_swarm_presets_packaging.py agent/tests/test_mcp_server_smoke.py`
- `npm run build` in `frontend/`
- `git diff --cached --check`

## Notes
- Credentials are read from environment or local settings only and are sent to FXMacroData as `X-API-Key`, never as query parameters.
[...]

PR #469, opened 2026-07-11
fix: support legacy FastMCP transport imports

roberttidball:
## Summary
- keep MCP transport imports compatible with both FastMCP module layouts
- preserves support for environments where fastmcp.client.transports is a module rather than a package

## Context
This was split out from #418 after review feedback to keep the FXMacroData integration scoped. The scheduled-routes 204 fix mentioned there is already present on current main; this PR only carries the remaining FastMCP compatibility fix.

## Validation
- python -m pytest agent/tests/test_market_data_tool.py agent/tests/test_mcp_server_smoke.py -q (5 passed)
- git diff --check


warren618:
Merged — thanks for splitting this out cleanly. Verified against our installed FastMCP: both layouts import, MCP smoke tests green.

PR #472, opened 2026-07-11
fix(correlation): resolve bare tickers so the matrix stops failing with "Fetched: []"

### LINKED ISSUE #471: [Bug] Correlation Matrix fails with "Fetched: []" for the exact ticker format its own hint suggests (AAPL,SPY)
Tried the Correlation tab with `AAPL,SPY` — the format the input hint itself suggests (zh-CN locale: "逗号分隔的标的代码，如 BTC-USDT,ETH-USDT,AAPL,SPY", same in ja/ko/ar) — and every request fails with:

```
Could not fetch price data for at least 2 assets. Fetched: []
```

## Description

`compute_correlation_matrix` (`agent/backtest/correlation.py`) passes the user's raw codes straight to the data loaders, but the loaders key US/HK/A-share instruments by the exchange-suffixed symbol (`AAPL.US` / `0700.HK` / `600000.SH`). A bare `AAPL` fails the yahoo loader's `_is_supported()` check (requires `.US`/`.HK`/`.NS`/`.BO`), returns an empty dict — and since that's not an exception, the yfinance fallback never fires either. Verified: `AAPL.US` → 103 rows, bare `AAPL` → nothing.

Digging further, two more layers make this worse:

1. **`infer_market`'s prefix heuristics misroute many symbols**: bare `0700` (Tencent) → `a_share`, `9988` (Alibaba) → `us_equity`, `300750` (CATL, ChiNext) → `hk_equity`, and even suffixed `830799.BJ` → `us_equity` (no `.SH/.SZ/.BJ` suffix handling at all).
2. **The fetch loop stops at the first *available* loader**, not the first that returns data — when eastmoney (first in the HK chain) hits a network error, the asset silently vanishes even though yahoo (next in chain) can serve it.
3. All failures are swallowed by bare `except: continue`, so the only symptom is `Fetched: []` with no clue why.

## Steps to Reproduce

1. Open the Correlation tab (`/correlation`).
2. Enter `AAPL,SPY` (as the hint text suggests) and compute.
3. Request fails with `Could not fetch price data for at least 2 assets. Fetched: []`.

Same happens for bare A-share codes (`600000,000001`) and bare HK codes (`0700,9988`).

## Interface

Web UI Correlation tab / `GET /correlation?codes=AAPL,SPY` (the API docstring example `BTC-USDT,ETH-USDT,SPY` fails the same way).

Note: `en.json`'s hint was at some point changed to suffixed A-share examples (`000001.SZ,600519.SH,000858.SZ`), but zh-CN/ja/ko/ar still show the bare form — and either way, bare tickers arguably *should* work. Happy to submit a fix (have one working).

yxhuang:
### Summary

Fixes the Correlation tab failing with `Could not fetch price data for at least 2 assets. Fetched: []` when given bare tickers like `AAPL,SPY` — the exact format the UI hint suggests. Root cause: user input was passed straight to loaders that key instruments by exchange-suffixed symbols (`AAPL.US` / `600000.SH`), plus two compounding issues in market inference and the fetch loop.

### What's in it

Three layered fixes in `agent/backtest/correlation.py`:

- **`infer_market` rewritten** — explicit suffix is now authoritative (`.HK`, `.SH/.SZ/.BJ`, `.US`), then bare numeric codes classify by digit length (exactly 6 digits = A-share; ≤5 = HK). The old prefix heuristics misrouted bare `0700`→SZ, `9988`→US, `300750`→HK, and even suffixed `830799.BJ`→US. Prefix alone can't separate HK from A-share (both use leading 0/3); digit length can.
- **`_normalize_symbol` (new)** — canonicalizes bare tickers to the loader symbol form before fetching (`AAPL`→`AAPL.US`, `600000`→`600000.SH`, `0700`→`0700.HK`), while crypto pairs and already-suffixed codes pass through untouched. The user's original input is kept as the matrix label.
- **Fetch loop walks the full `FALLBACK_CHAINS`** until a loader actually returns data, instead of stopping at the first *available* one — a loader can be available yet serve nothing (network error), which silently dropped assets a later loader could serve. Every per-source failure is now logged (`logger.warning`) instead of swallowed by bare `except: continue`, so the next person debugging this sees *why*.

**Tests** (`agent/tests/test_correlation.py`, +10 cases): bare/suffixed classification for all markets incl. BJ, symbol normalization, and a chain fall-through regression test (first loader available-but-empty → second loader serves). 22 pass.

### Notes

- `000001`-style true ambiguity (Ping An Bank vs SSE Composite index) can't be resolved by inference — bare 6-digit codes default to the stock reading; the index still needs an explicit `000001.SH`.
- Behavior for already-suffixed symbols and crypto pairs is unchanged.
- Touches only `correlation.py` + its tests; verified end-to-end: `AAPL,SPY` → 2×2 matrix (corr 0.4057), `0700,9988,AAPL,SPY` → full 4×4 with eastmoney down (falls through to yahoo, failure logged).

Closes #471.

### DIFF
--- a/agent/backtest/correlation.py
+++ b/agent/backtest/correlation.py
@@ -6,30 +6,82 @@ over a configurable lookback window. Used by the /correlation API endpoint.
 
 from __future__ import annotations
 
+import logging
 from typing import Dict, Literal
 
 import pandas as pd
 import numpy as np
 from scipy.stats import spearmanr
 
+logger = logging.getLogger(__name__)
+
 
 def infer_market(code: str) -> str:
-    """Infer market key from a ticker symbol."""
-    code_upper = code.upper()
+    """Infer market key from a ticker symbol.
+
+    Resolution order:
+
+    1. Crypto pair spellings (``BTC-USDT``, ``ETH/USD`` …).
+    2. Explicit exchange suffix — always authoritative (``.HK``, ``.SH``/
+       ``.SZ``/``.BJ``, ``.US``). Bare HK and A-share codes are both purely
+       numeric, so the suffix is the only reliable disambiguator.
+    3. Bare numeric codes by digit length: A-share codes are exactly 6 digits
+       (600000, 000001, 300750, 688981, 830799); HK codes are at most 5
+       (700, 0700, 9988, 3690). Prefix alone cannot tell them apart — both
+       markets use leading 0 and 3.
+    4. Anything else (alphabetic tickers) is a US equity.
+    """
+    code_upper = code.strip().upper()
     crypto_suffixes = ("USDT", "BTC", "ETH", "BNB", "SOL", "ADA", "DOGE")
     if any(code_upper.endswith(s) for s in crypto_suffixes) or "/" in code:
         return "crypto"
-    # Check .HK suffix FIRST so leading-zero tickers like 0700.HK / 0005.HK
-    # are correctly classified before the A-share prefix checks
     if code_upper.endswith(".HK"):
         return "hk_equity"
-    if code_upper.startswith(("6", "000", "001", "002")):
+    if code_upper.endswith((".SH", ".SZ", ".BJ")):
         return "a_share"
-    if code_upper.startswith(("0", "399")):
-        return "a_share"
-    if code_upper.startswith(("0", "1", "2", "3", "4")):
-        return "hk_equity"
+    if code_upper.endswith(".US"):
+        return "us_equity"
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks for the root-cause work. The digit-length classification is the right disambiguator, and the available-but-empty chain fall-through closes a real silent-drop bug. Verified locally: 22 correlation tests green, `AAPL,SPY` end-to-end.

PR #474, opened 2026-07-12
feat(providers): add Requesty as an OpenAI-compatible provider

Thibaultjaigu:
Adds Requesty as a provider, mirroring the existing OpenRouter wiring. Requesty is an OpenAI-compatible LLM gateway that uses the same `provider/model` naming convention as OpenRouter and the same opt-in `extra_body.reasoning` option, so it shares OpenRouter's capability shape.

Changes:
- `agent/src/providers/capabilities.py` — `requesty` `ProviderCapabilities` entry mirroring openrouter (`capture_reasoning`, `openrouter_reasoning_body`), on the generic OpenAI-compatible path with base URL `https://router.requesty.ai/v1` and env `REQUESTY_API_KEY` / `REQUESTY_BASE_URL`.
- `agent/src/providers/llm_providers.json` — requesty registry entry (default model `openai/gpt-4o-mini`).
- `agent/cli/onboard.py` and `agent/cli/_legacy.py` — Requesty added to the onboarding provider list and the legacy CLI provider key/base env maps + choices.
- `agent/.env.example` — commented Requesty provider block.
- `README.md` — Requesty added to the supported LLM providers list.
- Tests: `test_requesty_capabilities_mirror_openrouter` plus the provider-defaults regression map and the CLI provider-index updates.

Testing: `ruff check` clean on edited files; `pytest` passes (provider diagnostics, defaults, CLI init, and the LLM suite). Live-tested against the real endpoint: a chat/completions call to `https://router.requesty.ai/v1` with `openai/gpt-4o-mini` returned successfully.

Docs: https://requesty.ai · https://docs.requesty.ai · https://app.requesty.ai/api-keys · https://app.requesty.ai/router/list

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.


warren618:
Merged — thanks for the clean addition. The capability wiring exactly mirrors OpenRouter's shape (both are OpenAI-compatible gateways), and every touchpoint is covered: CLI provider menu, onboarding flow, `.env.example`, and the frontend Settings dropdown (reads `llm_providers.json` dynamically, so no separate frontend change needed). Verified locally: 19 provider/CLI tests green, both CI gates pass. Also propagated Requesty into the other 4 locale READMEs, which your diff didn't touch.

PR #479, opened 2026-07-12
fix(preflight): reset EnvConfig after dotenv load to pick up .env values

### LINKED ISSUE #477: `run` preflight ignores `~/.vibe-trading/.env` while `provider doctor` reads it correctly
## Environment
- vibe-trading-ai: 0.1.11
- Python: 3.11 (reproduced on Windows 11 and Linux)
- Install: `pip install vibe-trading-ai`

## Description
With a valid `~/.vibe-trading/.env`:

```
LANGCHAIN_PROVIDER=openai
LANGCHAIN_MODEL_NAME=claude-sonnet-5
OPENAI_BASE_URL=https://api.anthropic.com/v1
OPENAI_API_KEY=sk-...
```

`vibe-trading provider doctor` resolves everything correctly (provider `openai`, model `claude-sonnet-5`, key `set`). But `vibe-trading run -p "..."` fails preflight:

```
 N/A  │ LLM (openai)       │ LANGCHAIN_MODEL_NAME not set in .env (agent cannot function)
Critical check failed - agent cannot start without a working LLM provider.
```

Note the preflight reports provider `openai` even when the `.env` sets a different provider — it's the schema default, indicating the `.env` was never loaded on this path.

## Suspected cause
`src/preflight.py` calls `_ensure_dotenv()` then `get_env_config()`, but by that point the `EnvConfig` snapshot appears to already be cached from an earlier import-time call, built from bare `os.environ`. `_load_env_file` uses `override=False`/`setdefault`, and no `reset_env_config()` runs after the dotenv load, so the cached snapshot never picks up the file values.

## Workaround
Exporting the same variables as real environment variables (e.g. `setx` on Windows, `export` on Linux) makes `run` pass preflight.

## Expected
`run` and `provider doctor` should resolve the same configuration from `~/.vibe-trading/.env`.

ananaymital:
Fixes https://github.com/HKUDS/Vibe-Trading/issues/477

## Problem

Preflight always reports `LLM (openai)` with `LANGCHAIN_MODEL_NAME not set`, even when `agent/.env` correctly configures a different provider (e.g. `openai-codex`).

## Root Cause

`cli/theme.py:187` calls `_is_dark_terminal()` at **module import time**, which calls `get_env_config()` and caches the `EnvConfig` singleton **before** any `.env` file is loaded. When `_check_llm_provider()` later calls `_ensure_dotenv()`, the values are written to `os.environ`, but the subsequent `get_env_config()` returns the stale cached snapshot with defaults (`provider="openai"`, `model=""`).

## Fix

Call `reset_env_config()` after `_ensure_dotenv()` in `_check_llm_provider()` so the singleton is rebuilt from the now-populated `os.environ`.

## Verification

Before fix: `LLM (openai) — LANGCHAIN_MODEL_NAME not set in .env (agent cannot function)`
After fix: `LLM (openai-codex) — openai-codex/gpt-5.4 via ChatGPT OAuth (...)`

### DIFF
--- a/agent/src/preflight.py
+++ b/agent/src/preflight.py
@@ -15,7 +15,7 @@ from typing import List, Optional
 from rich.console import Console
 from rich.table import Table
 
-from src.config.accessor import get_env_config
+from src.config.accessor import get_env_config, reset_env_config
 
 
 @dataclass(frozen=True)
@@ -34,6 +34,11 @@ def _check_llm_provider() -> CheckResult:
     from src.providers.llm import _ensure_dotenv, _sync_provider_env, provider_diagnostics
 
     _ensure_dotenv()
+    # The EnvConfig singleton may have been cached before _ensure_dotenv()
+    # loaded the .env file (e.g. by theme.py's import-time _is_dark_terminal
+    # call). Reset it so get_env_config() rebuilds from the now-populated
+    # os.environ.
+    reset_env_config()
     _cfg = get_env_config()
     provider = _cfg.llm.langchain_provider.strip()
     model = _cfg.llm.langchain_model_name.strip()


Raoufboukyxz:
i have the same problem !

Raoufboukyxz:
fix it

ananaymital:
@Raoufboukyxz Please verify if this PR fixes the issue on your machine too. Helps the maintainer approve and merge it quickly

PR #483, opened 2026-07-12
fix(preflight): refresh config after loading dotenv

### LINKED ISSUE #477: `run` preflight ignores `~/.vibe-trading/.env` while `provider doctor` reads it correctly

Bortlesboat:
## Summary

- refresh the cached environment schema immediately after preflight loads dotenv
- add a regression test for a config snapshot created before dotenv loading

## Why

Closes #477.

`run` preflight could cache `EnvConfig` before `_ensure_dotenv()` populated `os.environ`. It then continued using that stale snapshot and reported the model as missing even though `~/.vibe-trading/.env` contained it. Provider setup already resets this cache on its own path; preflight now does the same before reading provider and model values.

## Changes

- call `reset_env_config()` between `_ensure_dotenv()` and `get_env_config()`
- reproduce the stale-cache ordering and assert the loaded provider/model are used

## Test Plan

- [x] New regression test added
- [x] `python -m pytest agent/tests/test_preflight.py agent/tests/test_dotenv_observability.py -q --basetemp .pytest-tmp` (12 passed)
- [x] `python -m ruff check agent/src/preflight.py agent/tests/test_preflight.py`
- [x] `python -m compileall -q agent/src/preflight.py`

## Checklist

- [x] No hardcoded credentials or local application paths
- [x] Code follows `CONTRIBUTING.md`, including DCO sign-off
- [x] Documentation is unchanged because this restores the documented `.env` behavior

## Risk and rollback

No broker, order, MCP, credential-write, network-listener, or deployment behavior changes. The reset only discards an in-memory config snapshot immediately after dotenv loading. Rollback is a direct revert of this commit.

### DIFF
--- a/agent/src/preflight.py
+++ b/agent/src/preflight.py
@@ -15,7 +15,7 @@ from typing import List, Optional
 from rich.console import Console
 from rich.table import Table
 
-from src.config.accessor import get_env_config
+from src.config.accessor import get_env_config, reset_env_config
 
 
 @dataclass(frozen=True)
@@ -34,6 +34,7 @@ def _check_llm_provider() -> CheckResult:
     from src.providers.llm import _ensure_dotenv, _sync_provider_env, provider_diagnostics
 
     _ensure_dotenv()
+    reset_env_config()
     _cfg = get_env_config()
     provider = _cfg.llm.langchain_provider.strip()
     model = _cfg.llm.langchain_model_name.strip()
--- a/agent/tests/test_preflight.py
+++ b/agent/tests/test_preflight.py
@@ -75,6 +75,40 @@ def test_llm_preflight_probe_reports_request_errors(monkeypatch) -> None:
     assert "Timeout: timed out" in result.message
 
 
+def test_llm_preflight_refreshes_config_after_loading_dotenv(monkeypatch) -> None:
+    """Preflight must not reuse a config snapshot created before dotenv loading."""
+    import src.providers.llm as llm
+    from src.config.accessor import get_env_config
+
+    for name in ("LANGCHAIN_PROVIDER", "LANGCHAIN_MODEL_NAME", "OPENAI_BASE_URL"):
+        monkeypatch.delenv(name, raising=False)
+    get_env_config()
+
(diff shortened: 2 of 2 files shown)


warren618:
Thanks for tracking down the same root cause — this is byte-for-byte the same fix as #479 (EnvConfig cached before dotenv loads, so preflight sees a stale snapshot). Merging #479 instead since it also carries a regression test covering this exact scenario. Appreciate you digging into #477 independently — closing this one as a duplicate, not a rejection of the fix itself.

PR #517, opened 2026-07-13
fix: filter unsupported hypothesis update fields

sTunnaSu:
## Summary

- Prevent `UpdateHypothesisTool` from forwarding unsupported arguments to `HypothesisRegistry.update()`
- Reject undeclared properties in the tool schema
- Filter update fields through an explicit allowlist
- Add a regression test for the unexpected `run_dir` argument

## Why

The agent could pass `run_dir` to `update_hypothesis`, causing:

`HypothesisRegistry.update() got an unexpected keyword argument 'run_dir'`

## Test Plan

- [x] Existing hypothesis registry tests pass
- [x] New regression test added
- [x] `python -m pytest agent/tests/test_hypothesis_registry.py -q`
- [x] 8 tests passed

### DIFF
--- a/agent/src/tools/hypothesis_tool.py
+++ b/agent/src/tools/hypothesis_tool.py
@@ -96,13 +96,30 @@ class UpdateHypothesisTool(BaseTool):
             "invalidation_notes": {"type": "string"},
         },
         "required": ["hypothesis_id"],
+        "additionalProperties": False,
     }
 
     def execute(self, **kwargs: Any) -> str:
         """Update a hypothesis and return it as JSON."""
         try:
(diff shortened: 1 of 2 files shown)


sTunnaSu (commented):
This fixes the unexpected `run_dir` argument crash and adds a regression test. All targeted tests pass.

warren618:
Thank you @sTunnaSu — this contribution has now landed on main via #534, with your original commit authorship preserved. We integrated the locally verified fix together with the full regression gate, so I’m closing this source PR as incorporated. Really appreciate the contribution!

PR #523, opened 2026-07-13
fix(tools): add Tushare fallbacks for flow tools

### LINKED ISSUE #516: Add fallback providers for single-source data tools (fund_flow / dragon_tiger / margin_trading / northbound)
## Context

Raised in discussion #486: `get_fund_flow` fails entirely when Eastmoney (or
the caller's own proxy en route to it) is unreachable, with no fallback
provider. Confirmed via code read: `fund_flow_tool.py`, `dragon_tiger_tool.py`,
`margin_trading_tool.py`, and `northbound_tool.py` are all single-source
Eastmoney wrappers — this is consistent across the whole tool family, not a
bug specific to one file.

## Open questions before implementing

- Does tushare (or another already-integrated source) actually carry
  equivalent bucketed net-inflow data (主力/超大单/大单/中单/小单) in the
  same shape, or would a fallback source need a different response schema /
  partial-feature degradation?
- If we add a fallback to one of these 4 tools, do we add it to all 4 for
  consistency, or is fund-flow specifically higher-priority?
- What's the fallback UX — silently switch source and label it in the
  response envelope's `source` field (already present), or surface a
  warning?

## Not urgent

Eastmoney is free / no-auth and generally reliable; the reporter's specific
failure looks environment-side (a broken `HTTPS_PROXY`), not a systemic
outage. This is a resilience improvement, not a live bug affecting most
users.

asahikiko:
## Summary
- add optional Tushare fallback adapters for the Eastmoney-backed fund-flow, dragon-tiger, northbound-flow, and margin-trading tools
- keep Eastmoney as the primary source; fall back only on provider failure or empty margin data, and label fallback responses with `source: tushare` plus warnings
- preserve existing error envelopes when Tushare is unavailable, including the fallback failure reason for easier configuration/debugging

Closes #516.

## Tests
- `python -m pytest agent/tests/test_fund_flow_tool.py agent/tests/test_dragon_tiger_tool.py agent/tests/test_northbound_tool.py agent/tests/test_margin_trading_tool.py agent/tests/test_tushare_fallbacks.py`
- `python -m ruff check agent/src/tools/tushare_fallbacks.py agent/src/tools/fund_flow_tool.py agent/src/tools/dragon_tiger_tool.py agent/src/tools/northbound_tool.py agent/src/tools/margin_trading_tool.py agent/tests/test_fund_flow_tool.py agent/tests/test_dragon_tiger_tool.py agent/tests/test_northbound_tool.py agent/tests/test_margin_trading_tool.py agent/tests/test_tushare_fallbacks.py`

### DIFF
--- a/agent/src/tools/dragon_tiger_tool.py
+++ b/agent/src/tools/dragon_tiger_tool.py
@@ -20,6 +20,7 @@ from typing import Any
 
 from backtest.loaders import eastmoney_client
 from src.agent.tools import BaseTool
+from src.tools import tushare_fallbacks
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


warren618:
Thank you @asahikiko — this contribution has now landed on main via #534, with your original commit authorship preserved. We verified the fallback field and unit contracts and integrated it through the full regression gate, so I’m closing this source PR as incorporated. Really appreciate the contribution!

PR #526, opened 2026-07-13
feat(mcp): add Streamable HTTP transport for modern clients (fixes 405 from QwenPaw etc.)

huangcheng:
## Goal

Add a `--transport http` option to `vibe-trading-mcp` that exposes the MCP **Streamable HTTP** transport (spec 2025-03-26+) as a single `POST`/`GET` `/mcp` endpoint, alongside the existing `stdio` and legacy `sse` transports.

Modern MCP clients (e.g. [QwenPaw](https://github.com/agentscope-ai/QwenPaw)) negotiate the transport by `POST`ing an `InitializeRequest` to the server URL (per the spec's recommended client flow). The currently-shipped **legacy SSE** transport uses two endpoints — `GET /sse` for the stream and `POST /messages/` for requests — so a client that `POST`s to `/sse` gets **`405 Method Not Allowed`**. Streamable HTTP accepts `POST` on its single endpoint, resolving this.

> The spec is explicit that the legacy HTTP+SSE transport is **deprecated**, and that clients should POST an `InitializeRequest` first to detect the transport.

## Affected areas

- `agent/mcp_server.py` — `main()`: added `"http"` to `--transport` choices and a new `elif args.transport == "http"` branch calling `mcp.run(transport="streamable-http", host=args.host, port=args.port)`. Added a `--host` option scoped to the `http` transport (default `127.0.0.1`). **The legacy `sse` branch is byte-for-byte unchanged** (`mcp.run(transport="sse", port=args.port)`).
- `agent/tests/test_mcp_server_http_transport.py` — new integration test: spawns the real server with `--transport http` and asserts `POST /mcp initialize` → `200` + `MCP-Session-Id` + `serverInfo`, and that the legacy `POST /sse` path returns `404` (not mounted by this transport).
- `README.md` + 4 translations (`README_zh/ja/ko/ar.md`) — document the three transport options and warn not to point HTTP clients at `/sse`.

## Out of scope

- No change to the legacy `sse` transport behavior, the `stdio` transport, any tool, the registry, broker/order/mandate surfaces, or credentials.
- No change to FastMCP version (`fastmcp>=2.14.0` already in `pyproject.toml`; resolves to `3.4.4`).
- No new dependencies.

## Test plan

Run with the project venv (`pip install -e ".[dev]"`):

```bash
python -m py_compile agent/mcp_server.py
pytest agent/tests/test_mcp_server_http_transport.py \
       agent/tests/test_mcp_server_smoke.py \
       agent/tests/test_mcp_sse_integration.py \
       agent/tests/test_mcp_streamable_http_integration.py -q
```

Results (Python 3.12, fastmcp 3.4.4):
- `test_mcp_server_http_transport.py` (new) — **PASS** (POST `/mcp` initialize → 200 + session + `Vibe-Trading` serverInfo; `POST /sse` → 404)
- `test_mcp_server_smoke.py` (stdio regression) — **PASS**
- `test_mcp_sse_integration.py` (legacy SSE, 3 tests) — **PASS**
- `test_mcp_streamable_http_integration.py` (client adapter, 4 tests) — **PASS**
- `git diff --check` — clean

**Not run:** the full suite (`pytest --ignore=agent/tests/e2e_backtest …`) and the frontend build, since this change is backend-MCP-only; the narrowest matching suite (all MCP tests) is green.

I also ran a Streamable HTTP conformance probe against the running endpoint: `POST initialize` → 200 + `MCP-Session-Id`; `notifications/initialized` → 202; `tools/list` → 200 (54 tools); wrong `Accept` → 406; wrong `Content-Type` → 400; missing session id on a post-init request → 400; `DELETE /mcp` with session id → 200 (session termination).

## Risk

- **MCP / network:** additive only — a new optional transport. Default transport remains `stdio`; the new `http` transport binds `127.0.0.1` by default (loopback only), so it does not start an externally reachable server unless the operator explicitly passes `--host 0.0.0.0`. Per the Agent Contributor Guide, starting an externally reachable MCP server requires operator approval; this change does **not** do that by default.
- **Security note (FastMCP default):** FastMCP 3.4.4 ships with Host/Origin (DNS-rebinding) validation **disabled by default**. For loopback use this is low-risk. Operators deploying `--transport http` beyond loopback should enable host-origin protection and front it with an authenticated proxy; this is worth a follow-up CLI flag but is out of scope here.
- **Spec nuance (inherited from FastMCP, not introduced here):** a sessionless `GET /mcp` returns `400 "Missing session ID"` rather than the strict `405`/SSE. A `GET` with a valid session id correctly returns `200 text/event-stream`. Real clients `POST initialize` first, so this does not affect the normal flow. This is FastMCP library behavior; our wrapper adds no protocol logic.
- **No broker, order, mandate, credential, or live-trading surfaces are touched.**

## Rollback / close path

Fully reversible with `git revert <sha>` (or close the PR without merge). The change is additive — reverting removes the `http` branch and `--host` option and restores the prior two-transport CLI, leaving `stdio` and legacy `sse` exactly as they were. No data migrations, no persistent state.

---

DCO: signed off on the commit (`Signed-off-by:` trailer present). Per `CONTRIBUTING.md`, no AI-assistant attribution trailers were added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -967,9 +967,17 @@ profile. `run_swarm` requires an LLM key.
 <summary><b>Cursor / Windsurf / other MCP clients</b></summary>
 
 ```bash
-vibe-trading-mcp                  # stdio (default)
-vibe-trading-mcp --transport sse  # SSE for web clients
+vibe-trading-mcp                   # stdio (default)
+vibe-trading-mcp --transport http  # Streamable HTTP (current MCP spec default) at http://127.0.0.1:8900/mcp
+vibe-trading-mcp --transport sse   # legacy SSE (deprecated) for older clients
 ```
+
+For HTTP clients (QwenPaw, and any client that negotiates by POSTing an
+`InitializeRequest`), use `--transport http` and point the client at the single
+`/mcp` endpoint — e.g. `http://127.0.0.1:8900/mcp`. Do **not** point an HTTP
+client at `/sse`; that path belongs to the deprecated two-endpoint SSE transport
+and will return `405 Method Not Allowed` on `POST`. Override the bind address
+with `--host` / `--port`.
 
 </details>
 
--- a/README_ar.md
+++ b/README_ar.md
@@ -934,8 +934,9 @@ skills:
 <summary><b>Cursor / Windsurf / عملاء MCP آخرون</b></summary>
 
 ```bash
-vibe-trading-mcp                  # stdio (default)
-vibe-trading-mcp --transport sse  # SSE for web clients
+vibe-trading-mcp                   # stdio (default)
+vibe-trading-mcp --transport http  # Streamable HTTP (spec default) at /mcp
+vibe-trading-mcp --transport sse   # legacy SSE (deprecated)
 ```
 
 </details>
--- a/README_ja.md
+++ b/README_ja.md
@@ -931,8 +931,9 @@ skills:
 <summary><b>Cursor / Windsurf / other MCP clients</b></summary>
 
 ```bash
-vibe-trading-mcp                  # stdio (default)
-vibe-trading-mcp --transport sse  # SSE for web clients
+vibe-trading-mcp                   # stdio (default)
+vibe-trading-mcp --transport http  # Streamable HTTP (spec default) at /mcp
+vibe-trading-mcp --transport sse   # legacy SSE (deprecated)
 ```
 
 </details>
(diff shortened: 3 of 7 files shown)


warren618:
Thank you @huangcheng — this contribution has now landed on main via #534, with your original commit authorship preserved. We verified the MCP transports and integrated it through the full regression gate, so I’m closing this source PR as incorporated. Really appreciate the contribution!

PR #527, opened 2026-07-13
fix(frontend): authenticate correlation requests

ShiroKSH:
## Summary

- Route Correlation requests through the shared authenticated API client.
- Preserve the page's existing query parameters and error handling.
- Add regression coverage for the stored API-key header.

## Root cause

The Correlation page used a local fetch helper, which omitted the stored Authorization header. Remote backends correctly rejected the request.

## Scope

- No changes to broker, live-runner, backtest, or server-side authorization behavior.

## Checks

- `npm run test:run` — 247 passed.
- `npm run build`.


warren618:
Thank you @ShiroKSH — this contribution has now landed on main via #534, with your original commit authorship preserved. We integrated the locally verified fix together with the full regression gate, so I’m closing this source PR as incorporated. Really appreciate the contribution!

PR #531, opened 2026-07-14
fix(backtest): account for terminal close costs

YZY0108:
## Summary

Include terminal liquidation slippage and exit commission in the final reported portfolio equity.

## Root cause

At the end of a backtest, remaining positions were closed at the raw final mark without applying the engine's exit slippage. Although `_close_position` deducted exit commission from cash, the equity series used by metrics and artifacts had already captured a pre-liquidation snapshot, so the terminal cost never reached reported performance.

## Changes

- apply symbol-aware, directionally adverse slippage to forced terminal exits
- replace the last equity snapshot with post-liquidation cash
- explicitly report zero open positions and zero unrealized P&L after liquidation
- add long/short regressions covering exit price, commission, cash, and final equity

## Validation

- `PYTHONPATH=agent pytest -q agent/tests/test_terminal_close_accounting.py agent/tests/test_base_engine.py agent/tests/test_crypto_engine.py agent/tests/test_forex_engine.py agent/tests/test_china_a_engine.py agent/tests/test_china_futures_engine.py agent/tests/test_global_futures_engine.py agent/tests/test_global_equity_engine.py agent/tests/test_india_equity_engine.py agent/tests/test_engine_robustness.py --tb=short`
  - 255 passed, 1 skipped
- `python -m compileall -q agent/backtest/engines/base.py agent/tests/test_terminal_close_accounting.py`
- `git diff --check`

The complete `agent/tests` collection is unavailable in this local environment because optional project dependencies such as `fastmcp`, `langchain_core`, `ccxt`, and `python-multipart` are not installed.

## Scope

This PR only addresses terminal liquidation accounting. Open-rebalance causality and realized-turnover accounting are intentionally submitted separately.

PR #532, opened 2026-07-14
fix(backtest): report turnover from executed trades

YZY0108:
## Summary

Report portfolio turnover from actual executed position sizes instead of target-weight changes.

## Root cause

The engine passed its target-position frame directly to the turnover metric. That frame is produced before execution, so it still contains orders rejected by market rules and ignores lot/contract rounding. As a result, reported turnover could describe intended trades that never occurred.

## Changes

- record actual entry and exit margin-equivalent values on completed trades
- aggregate filled entry/exit legs by execution timestamp and normalize them by portfolio equity
- preserve the existing `0.5 * gross traded weight` convention
- keep position-implied turnover as a backward-compatible fallback for external `calc_metrics` callers
- add regressions for rounded fills and a rejected China A-share short target

The realized series also counts modeled liquidations and end-of-backtest exits because those are recorded executions.

## Validation

- `PYTHONPATH=agent pytest -q agent/tests/test_realized_turnover.py agent/tests/test_metrics.py agent/tests/test_models.py agent/tests/test_validation.py agent/tests/test_base_engine.py agent/tests/test_crypto_engine.py agent/tests/test_forex_engine.py agent/tests/test_china_a_engine.py agent/tests/test_china_futures_engine.py agent/tests/test_global_futures_engine.py agent/tests/test_global_equity_engine.py agent/tests/test_india_equity_engine.py agent/tests/test_engine_robustness.py --tb=short`
  - 326 passed, 1 skipped
- `python -m compileall -q agent/backtest/engines/base.py agent/backtest/metrics.py agent/backtest/models.py agent/tests/test_realized_turnover.py`
- `git diff --check`

The complete `agent/tests` collection is unavailable in this local environment because optional project dependencies such as `fastmcp`, `langchain_core`, `ccxt`, and `python-multipart` are not installed.

## Scope

This PR only changes turnover measurement. Open-rebalance causality and terminal-close cost accounting are intentionally submitted separately.


warren618:
Merged on top of #531 after unifying turnover with the existing #478 position-frame metric. When fill accounting is available, turnover now comes from actual rounded entry/exit margin—including terminal liquidation, futures multipliers, composite routing, and full rotations—while the position-frame calculation remains the explicit fallback. The updated focused suite and full remote CI passed. Thank you for making the reported metric reflect real execution.

PR #533, opened 2026-07-14
fix(backtest): cap annual return after full loss

### LINKED ISSUE #524: Backtest metrics crash (TypeError) when final equity is negative
### Describe the bug

`backtest/metrics.py::calc_metrics` computes the annualised return as:

```python
ann_ret = float((1 + total_ret) ** (bpy / max(n, 1)) - 1)
```

A leveraged or short book can finish at or below zero equity, so `total_ret <= -1` and the base `(1 + total_ret)` is non-positive. Raising a negative base to a fractional power yields a Python `complex`, and the wrapping `float(...)` then raises `TypeError: float() argument must be a string or a real number, not 'complex'`. This crashes the whole metrics computation (and the backtest run that reports it) instead of returning a metric.

`_calc_equity` in `backtest/engines/base.py` returns `capital + sum(margin + pnl)` with no floor at zero, so a leveraged position that gaps past its maintenance margin (crypto-perp / futures engines) can produce a negative final equity that flows into `calc_metrics`.

### How to reproduce

```python
import numpy as np, pandas as pd
from backtest.metrics import calc_metrics

dates = pd.bdate_range("2025-01-01", periods=252)
eq = pd.Series(np.linspace(1_000_000, -500_000, 252), index=dates)  # book ends below zero
calc_metrics(eq, [], 1_000_000, 252)
# TypeError: float() argument must be a string or a real number, not 'complex'
```

A full wipeout to exactly zero equity (`total_ret == -1`) does not crash; only a strictly negative final equity does.

### Expected behavior

`calc_metrics` should never crash on a valid equity series. A total wipeout should report a -100% annualised return.

CooperSheroy:
## Summary

- Prevent `calc_metrics` from raising when final equity is below zero and annualization uses a fractional exponent.
- Cap `annual_return` at `-1.0` once total return reaches or exceeds a full loss.
- Add regression coverage for a negative final-equity curve.

## Why

Closes #524.

A leveraged or short strategy can produce a valid equity curve whose final value is below zero. In that case `1 + total_return` is negative, and fractional annualization can produce a complex number before `float(...)` raises. The metrics path should report a bounded loss metric instead of crashing the backtest report.

## Changes

- Guard annualized return calculation in `agent/backtest/metrics.py` for `total_ret <= -1`.
- Add `test_negative_final_equity_caps_annual_return` in `agent/tests/test_metrics.py`.

## Test Plan

- [x] Existing tests pass (`.venv/bin/python -m pytest agent/tests/test_metrics.py -q`)
- [x] New tests added

Manual check: reproduced the negative-final-equity case locally and confirmed it now returns `annual_return == -1.0` instead of raising.

## Checklist

## AI assistance disclosure

This PR was prepared with AI assistance from OpenAI Codex/Catalina for repository inspection, patch drafting, and validation. I reviewed the change and ran the focused tests locally.

### DIFF
--- a/agent/backtest/metrics.py
+++ b/agent/backtest/metrics.py
@@ -213,7 +213,7 @@ def calc_metrics(
     port_ret = equity_curve.pct_change().fillna(0.0)
 
     total_ret = float(equity_curve.iloc[-1] / initial_cash - 1)
-    ann_ret = float((1 + total_ret) ** (bpy / max(n, 1)) - 1)
+    ann_ret = -1.0 if total_ret <= -1 else float((1 + total_ret) ** (bpy / max(n, 1)) - 1)
     vol = float(port_ret.std())
     sharpe = float(port_ret.mean() / (vol + 1e-10) * np.sqrt(bpy))
 
--- a/agent/tests/test_metrics.py
+++ b/agent/tests/test_metrics.py
@@ -229,6 +229,15 @@ class TestCalcMetrics:
         m = calc_metrics(eq, [], 1_000_000, 252)
         assert m["total_return"] < 0
 
+    def test_negative_final_equity_caps_annual_return(self) -> None:
+        dates = pd.bdate_range("2025-01-01", periods=100)
+        eq = pd.Series(np.linspace(1_000_000, -500_000, 100), index=dates)
+
+        m = calc_metrics(eq, [], 1_000_000, 252)
+
+        assert m["total_return"] == pytest.approx(-1.5)
+        assert m["annual_return"] == pytest.approx(-1.0)
+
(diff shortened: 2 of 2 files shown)


warren618:
Thanks for catching this — the crash is real (uncapped annualized-return math going non-numeric after a total/leveraged loss).

Closing as already fixed: main independently shipped an equivalent fix in 75fb2dd (via #534), merged about 50 minutes after this PR was opened — same approach (cap at -1.0 when growth is non-positive) and equivalent regression tests for both the wipeout and negative-final-equity cases. That's also why this PR shows merge conflicts: main rewrote the exact line you touched.

No action needed on your end — appreciate the fix and the tests either way.

PR #535, opened 2026-07-14
fix: preserve run data source during chart reconstruction

xkam7ar:
## Summary

- Preserve the executed market-data source when reconstructing historical chart prices.
- Add a network-free regression covering an OKX-configured completed run.

## Why

`reconstruct_price_series()` looked for `source` in the run context, but that context does not contain the field. It therefore silently defaulted to `tushare`, even when the canonical `config.json` recorded `okx`.

This can make historical analysis query the wrong loader and return empty or mismatched chart data.

## Changes

- Read `source` from the existing run `config.json` inside reconstruction.
- Preserve the current `tushare` fallback for missing or legacy configuration.
- Test actual loader selection with isolated OKX and Tushare stubs.

## Test Plan

- [x] New tests added
- [x] Tested manually

Targeted validation:

- `test_ui_services.py`, `test_api_infrastructure.py`, and `test_run_card.py`: **45 passed**
- New test Ruff and Black checks: **passed**
- Direct offline reproduction selected Tushare before the fix and only OKX afterward
- `git diff --check`: **passed**

The full suite, live market data, LLM, browser, and E2E paths were not run.

## Risk and scope

No request schema, HTTP response, dependency, credential, network, broker, or configuration format changes. Missing or malformed legacy configuration retains the previous Tushare fallback. Rollback is a single commit revert.

## Checklist

- [x] Documentation update is not applicable to this internal reconstruction correction

### DIFF
--- a/agent/backtest/runner.py
+++ b/agent/backtest/runner.py
@@ -13,6 +13,7 @@ import inspect
 import json
 import logging
 import sys
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional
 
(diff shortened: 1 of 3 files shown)


warren618:
Merged after converging this onto the central loader registry. Chart reconstruction now preserves the original data source without hard-coding Tushare/YFinance/OKX, keeps auto/runtime fallback behavior centralized, and deep-copies loader config to avoid mutation leaks. Focused and combined regression suites passed locally. Thank you for the contribution.

PR #536, opened 2026-07-14
fix(backtest): keep Sharpe and information ratio finite for single-bar runs

santhreal:
### Problem

`calc_metrics` (`agent/backtest/metrics.py`) computes the Sortino denominator with a small-sample guard:

```python
downside_std = float(downside.std()) if len(downside) > 1 else 1e-10
```

but the **Sharpe** and **information-ratio** denominators use `Series.std()` unguarded:

```python
vol = float(port_ret.std())                 # Sharpe
active_std = float(active_ret.std())         # information ratio
```

`Series.std()` defaults to `ddof=1`, so a **single-observation** return series returns `NaN`. A one-bar backtest (`start_date == end_date`, or a data window with a single bar) produces a length-1 `equity_series` in `engines/base.py`, which flows straight into `calc_metrics` — only `len == 0` is guarded (the empty-equity early return), so `n == 1` falls through. The result is `sharpe = NaN` and `information_ratio = NaN`, while `sortino` is correctly `0.0`. The asymmetry is the giveaway: the small-sample pitfall was already handled for Sortino but missed for the other two.

The run-card writer sanitizes non-finite metrics to `null`, so today a legitimate one-bar run silently reports `sharpe: null` instead of `0.0`.

### Reproduction (against current `main`, pandas 3.0.3 / numpy 2.5.1)

```python
import pandas as pd, numpy as np, math
eq = pd.Series([1_000_000.0], index=pd.bdate_range("2025-01-01", periods=1))
port_ret = eq.pct_change().fillna(0.0)
vol = float(port_ret.std())                                   # -> nan  (ddof=1, single element)
sharpe = float(port_ret.mean() / (vol + 1e-10) * np.sqrt(252))
assert not math.isfinite(sharpe)   # NaN
```

### Fix

Apply the same `len > 1` guard `downside_std` already uses to `vol` and `active_std`, so all three risk metrics stay finite. A one-bar run now reports `0.0` instead of `NaN`.

### Tests

Added `TestCalcMetrics::test_single_bar_equity_metrics_finite` (a one-bar equity curve; asserts `sharpe`, `sortino`, `information_ratio`, `annual_return`, `max_drawdown`, `calmar` are all finite and Sharpe/IR are `0.0`). It fails on `main` (`isfinite(nan) → False`) and passes with this change.

`agent/tests/test_metrics.py`: **47 passed**. `ruff check` clean.

### DIFF
--- a/agent/backtest/metrics.py
+++ b/agent/backtest/metrics.py
@@ -222,7 +222,10 @@ def calc_metrics(
         ann_ret = -1.0
     else:
         ann_ret = float(growth ** (bpy / max(n, 1)) - 1)
-    vol = float(port_ret.std())
+    # ``Series.std()`` uses ddof=1, so a single-observation return series
+    # (e.g. a one-bar backtest) yields NaN and poisons the Sharpe ratio.
+    # Guard the small sample the same way ``downside_std`` is guarded below.
+    vol = float(port_ret.std()) if len(port_ret) > 1 else 0.0
     sharpe = float(port_ret.mean() / (vol + 1e-10) * np.sqrt(bpy))
 
     # Drawdown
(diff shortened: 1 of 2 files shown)


warren618:
Merged after focused audit. The single-bar regression now keeps Sharpe and information ratio finite without changing normal multi-bar behavior. Remote CI passed, and the combined convergence suite also passed locally. Thank you for the precise reliability fix.

PR #537, opened 2026-07-14
feat(providers): add iFlytek Spark as an OpenAI-compatible provider

FenjuFu:
Adds iFlytek Spark as a provider, following the wiring pattern of the other OpenAI-compatible entries (DashScope, Zhipu, MiniMax, Xiaomi MIMO). Spark's HTTP endpoint (`https://spark-api-open.xf-yun.com/v1`) is plain OpenAI-compatible — the APIPassword from the iFlytek console is sent as a Bearer token — so it rides the generic path with no special capability flags. `iflytek` is accepted as an alias, mirroring the existing `qwen`/`glm` alias pattern.

Changes:
- `agent/src/providers/llm_providers.json` — `spark` entry (+ `iflytek` alias) with default model `4.0Ultra` (Spark's strongest model on this endpoint; supports function calling) and default base URL `https://spark-api-open.xf-yun.com/v1`.
- `agent/src/providers/capabilities.py` — `_SPARK_CAPABILITIES` on the generic OpenAI-compatible path (no reasoning fields on the v1 chat path), registered under `spark` and `iflytek`.
- `agent/cli/_legacy.py` — `_provider_key_env` / `_provider_base_env` mappings, an `iFlytek Spark` entry in `_PROVIDER_CHOICES`, and `SPARK_API_KEY` / `SPARK_BASE_URL` in the `.env` render order.
- `agent/.env.example` — commented Spark block alongside the other Chinese providers.
- `README.md` — Spark added to the supported-provider list.
- Tests: registry/legacy default-model coverage for `spark`/`iflytek`, a capabilities test asserting the generic path and the alias, and the two `cmd_init` index-based prompts bumped for the new `_PROVIDER_CHOICES` entry.

Validation:
- `python -m py_compile` over the touched Python files and a JSON parse of `llm_providers.json` pass locally.
- `pytest agent/tests/test_llm_provider_defaults.py agent/tests/test_provider_diagnostics.py agent/tests/test_cli_init.py` not run locally (no full dev env on this machine); relying on CI. The changed tests only extend existing parametrized data and mirror the Requesty test shape from #474.

Not added to the interactive `cli/onboard.py` short list, consistent with the other regional providers (DashScope, Zhipu, Moonshot, MiniMax are also only in the legacy chooser).

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,7 +531,7 @@ vibe-trading-mcp               # start MCP server (stdio)
 - **Docker** for Path A
 - OpenAI Codex can also be used with ChatGPT OAuth: set `LANGCHAIN_PROVIDER=openai-codex`, then run `vibe-trading provider login openai-codex`. This does not use `OPENAI_API_KEY`.
 
-> **Supported LLM providers:** OpenRouter, Requesty, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, Z.ai, Ollama (local). See `.env.example` for config.
+> **Supported LLM providers:** OpenRouter, Requesty, OpenAI, DeepSeek, Gemini, Groq, DashScope/Qwen, Zhipu, Moonshot/Kimi, MiniMax, Xiaomi MIMO, iFlytek Spark, Z.ai, Ollama (local). See `.env.example` for config.
 
 > **Tip:** All markets work without any API keys thanks to automatic fallback. yfinance (HK/US), OKX (crypto), mootdx (A-shares, TCP-direct, no IP throttle), and AKShare (A-shares, US, HK, futures, forex) are all free. Tushare token is optional — mootdx is the preferred no-token A-share fallback, with AKShare as a broader backup.
 
--- a/agent/.env.example
+++ b/agent/.env.example
@@ -98,6 +98,13 @@ OPENROUTER_BASE_URL=https://openrouter.ai/api/v1
 # MIMO_API_KEY=xxx
 # MIMO_BASE_URL=https://api.xiaomimimo.com/v1
 
+# --- iFlytek Spark ---
+# APIPassword from the iFlytek console (https://console.xfyun.cn), sent as a Bearer token.
+# LANGCHAIN_PROVIDER=spark
+# LANGCHAIN_MODEL_NAME=4.0Ultra                # or generalv3.5 / pro-128k / lite
+# SPARK_API_KEY=xxx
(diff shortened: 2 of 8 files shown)


warren618:
This one needs a couple of fixes before it can merge, on top of the protected-core sign-off (`agent/src/providers/capabilities.py` + `llm_providers.json` need explicit review since they're under `agent/src/providers/`):

1. **Duplicate UI entry**: this PR adds both a `spark` and an `iflytek` block to `llm_providers.json` with the identical label "iFlytek Spark" — traced it through `settings_routes.py` into the actual Settings dropdown render, and it'll show up twice, indistinguishable to the user. This is the first duplicate label in that file's history. The NVIDIA-NIM PR this was modeled on avoided exactly this by keeping alias handling internal to `capabilities.py`/`_legacy.py` and only adding ONE entry to the enumerable JSON list — could you do the same here (drop one of the two JSON blocks; the alias will still resolve fine for anyone setting `LANGCHAIN_PROVIDER=iflytek` directly)?
2. **Heads up on merge order**: there's another open PR, #562 (AnyRouter), touching the same files. A test-merge of both together shows Git reports zero conflicts, but the merged test suite actually fails 2 tests in `test_cli_init.py` — both PRs independently bump the same menu index to different, individually-plausible-but-wrong values. Whoever's second to merge between this and #562 will need to manually recompute that index and rerun the suite; a clean git merge won't catch it. Just flagging so it doesn't surprise anyone.

FenjuFu:
Thanks for tracing that through the Settings render — you're right, two identical "iFlytek Spark" labels in the dropdown would be confusing. Fixed in the latest push:

- Dropped the `iflytek` block from `llm_providers.json`; only the `spark` entry remains in the enumerable list. The alias still resolves for `LANGCHAIN_PROVIDER=iflytek` via `capabilities.py` and the `_legacy.py` env maps, same as the NVIDIA-NIM `nvidia-nim` alias.
- Removed the corresponding `iflytek` row from `EXPECTED_PROVIDER_DEFAULTS` in `test_llm_provider_defaults.py` (it asserts against the JSON registry, which no longer has that key). Alias coverage stays in `test_provider_diagnostics.py`, which asserts against `get_provider_capabilities` directly.
- Also rebased onto current main to pick up the July 15 README news so the provider-list line applies cleanly.

On #562: understood on the index collision — if AnyRouter merges first I'll recompute the two `IntPrompt` indices in `test_cli_init.py` against the merged `_PROVIDER_CHOICES` and rerun that file before flagging this ready again.

warren618:
Thanks @FenjuFu — merged. iFlytek Spark lands as a clean OpenAI-compatible provider; menu indices reconciled with the SiliconFlow add on merge.

PR #538, opened 2026-07-14
fix(backtest): validate validation-config knobs instead of crashing

santhreal:
## Problem

The three validation tools in `agent/backtest/validation.py` read their count/seed knobs directly from user `config["validation"]` (via `run_validation` → the engine), but none of them validate the values. A single misconfigured knob either aborts the whole backtest with an opaque error, or silently returns NaN-laden garbage:

| Config | Result |
|---|---|
| `walk_forward: {n_windows: 0}` | `ZeroDivisionError` (`len // n_windows`) — aborts the run |
| `walk_forward: {n_windows: -2}` | **Silently** returns `{windows: [], return_mean: NaN, consistency_rate: -0.0}` — no error |
| `monte_carlo: {n_simulations: 0}` | `ZeroDivisionError` (`count / n_simulations`) |
| `monte_carlo`/`bootstrap: {seed: -1}` | `ValueError` from `np.random.default_rng` |
| `bootstrap: {n_bootstrap: 0}` | `IndexError` from `np.percentile` on an empty array |
| `bootstrap: {confidence: 1.5}` | `ValueError` from `np.percentile` (pct out of `[0,100]`) |

All reproduced deterministically. There is no upstream schema validation of the `validation` sub-config, so these values reach the functions unchecked.

## Fix

Guard each knob at function entry and return the same `{"error": ...}` shape these functions already use for insufficient data (e.g. `"need at least 3 trades"`). A bad knob now surfaces cleanly in the validation results instead of crashing the backtest or silently corrupting it:

- `n_simulations >= 1`, `n_bootstrap >= 1`, `n_windows >= 1`
- `0 < confidence < 1`
- `seed >= 0` (monte_carlo and bootstrap)

This is non-breaking — valid configs are unaffected, and the error dict renders fine through `run_card` / `validation.json`.

Out of scope: float-typed counts (`n_simulations=2.5` → `TypeError`) are a type-coercion concern better handled in the config schema than in these three functions.

## Tests

Added regression tests in `agent/tests/test_validation.py` covering non-positive `n_simulations`/`n_bootstrap`/`n_windows`, out-of-range `confidence`, and negative `seed`. Each asserts a clean error return rather than a raised exception. Full validation + metrics suites pass; `ruff` clean.

### DIFF
--- a/agent/backtest/validation.py
+++ b/agent/backtest/validation.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import json
 import math
+from numbers import Integral, Real
 from pathlib import Path
 from typing import Any, Dict, List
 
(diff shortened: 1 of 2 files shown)


warren618:
Merged after hardening all validation knobs to fail closed on booleans, strings, invalid fractions, and non-finite values. The focused validation suite and full remote CI passed. Thank you for turning runtime crashes into deterministic input errors.

PR #539, opened 2026-07-14
fix(backtest): keep backtest outputs finite and validation.json strict

santhreal:
Two related robustness gaps that let non-finite values into backtest outputs.

## 1. `initial_cash` is not validated → `inf` metrics

`BacktestConfigSchema` uses `extra="allow"` and never declares `initial_cash`, so a config with `initial_cash: 0` (or negative) passes validation untouched and reaches `calc_metrics`, where `equity[-1] / initial_cash - 1` produces `inf` `total_return` / `annual_return`:

```python
BacktestConfigSchema(codes=["AAPL.US"], start_date="2024-01-01",
                     end_date="2024-12-31", initial_cash=0)   # accepted today
```

Both production entry points (the CLI `python -m backtest.runner` and the MCP `backtest` tool, which runs the runner as a subprocess) go through this schema, so it is the right boundary to reject the value.

**Fix:** declare `initial_cash: float = Field(default=1_000_000, gt=0)` — non-positive values are now rejected at config validation with a clear error instead of running to produce non-finite metrics.

## 2. In-engine `validation.json` write bypassed strict JSON

`run_card` and the standalone `python -m backtest.validation` both sanitize validation results with `_json_safe` + `allow_nan=False`, but `BaseEngine.run_backtest` wrote the same artifact with `json.dumps(v_results, indent=2, ensure_ascii=False)` — default `allow_nan=True`. So when a validation metric is non-finite (e.g. a Sharpe from a path whose equity touches zero), the automatic in-engine `validation.json` contains bare `NaN` / `Infinity` tokens, which are invalid RFC-8259 JSON and rejected by strict parsers — while the standalone CLI writes `null` for the same run.

**Fix:** extract the strict write into a shared `write_validation_json` helper (sanitize + `allow_nan=False` + create the artifacts dir) and call it from both the engine and the standalone `main`, so every writer of `validation.json` emits the same valid JSON.

## Tests

- `initial_cash` of `0` / negative is rejected; default (`1_000_000`) and positive values accepted.
- `write_validation_json` serializes `inf` / `NaN` as `null` and its output parses under a strict `parse_constant` loader; it also creates a missing parent directory.

Relevant validation, metrics, and config-schema suites pass; `ruff` clean on the changed code.

Independent of #538 (branched off `main`); they touch different parts of `validation.py` and merge cleanly.

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -469,17 +469,17 @@ class BaseEngine(ABC):
 
         # 7. Validation (optional — triggered by config["validation"])
         if config.get("validation"):
-            from backtest.validation import run_validation
+            from backtest.validation import run_validation, write_validation_json
             v_results = run_validation(
                 config, equity_series, self.trades, self.initial_capital, bars_per_year,
             )
             m["validation"] = v_results
-            # Write validation.json artifact. The artifacts dir is normally
-            # created by _write_artifacts() below (step 8), so ensure it exists
-            # here to avoid a FileNotFoundError when run_dir/artifacts is absent.
-            v_path = run_dir / "artifacts" / "validation.json"
-            v_path.parent.mkdir(parents=True, exist_ok=True)
-            v_path.write_text(json.dumps(v_results, indent=2, ensure_ascii=False), encoding="utf-8")
+            # Write validation.json through the shared strict writer so a
+            # non-finite validation metric is serialized as null rather than an
+            # invalid bare NaN/Infinity token (matching the standalone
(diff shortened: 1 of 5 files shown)


warren618:
Merged after reconciling it with the validation-parameter guard and central loader changes. Initial cash and result metrics now reject or sanitize non-finite values, validation artifacts are strict RFC 8259 JSON, and NumPy scalars/arrays serialize deterministically. Focused regressions and the post-merge-base remote CI passed. Thank you for the reliability hardening.

PR #543, opened 2026-07-14
fix: MiniMax relay 401, dotenv override, shutdown hardening, Binance loader & OKX robustness

tyj147454413-cmd:
## Summary

This PR addresses several reliability and security issues encountered during real-world usage with MiniMax and other OpenAI-compatible providers.

### Key Changes

**1. MiniMax relay 401 fix — dotenv override**
- Shell-level exports (e.g. `export MINIMAX_API_KEY=...`) silently shadow the real keys in `.env` when `load_dotenv(override=False)`, causing 401 invalid api key errors against stale shell exports.
- Now force `override=True` so project `.env` beats shell env by default. Controlled by `VIBE_TRADING_DOTENV_OVERRIDE` flag (default `True`) for test compatibility.

**2. Shutdown endpoint hardening**
- Accidental or CSRF-style loopback shutdowns were possible with a bare POST.
- Now requires all of: same-site guard, explicit `confirm=true` query (or `X-Confirm-Shutdown` header), and either a valid `API_AUTH_KEY` or opt-in `VIBE_TRADING_ALLOW_LOCAL_SHUTDOWN=1` for lab use.

**3. Binance loader (new)**
- Dedicated `source=binance` loader as a crypto fallback alongside OKX — not a replacement.
- Public market data only; no API key required.

**4. OKX loader robustness**
- Deep-pagination support, rate-limit handling, and bounded history fetch.

**5. Thought-signature length guard**
- `payload["messages"]` and converted `messages` could diverge in length, causing IndexError and cross-pollinating thought signatures across turns. Added a defensive break.

**6. Stream retry**
- Configurable `VT_STREAM_MAX_RETRIES` (default 3) for transient LLM stream failures.

**7. Other fixes**
- Data directory migration (legacy `~/.vibe-trading` compatibility)
- fsync on state writes for crash safety
- Redaction and security test coverage improvements
- Frontend: i18n zh-CN updates, agent store fixes, layout improvements

### Test plan

- [x] All modified tests pass locally
- [x] MiniMax relay verified with `api.minimaxi.com` (domestic endpoint)
- [x] Shutdown endpoint rejects bare POST, requires confirm + auth
- [x] Binance loader fetches BTC-USDT spot and USD-M perp OHLCV
- [x] OKX loader deep-pagination fetches >1000 candles


warren618:
This PR bundles 7 distinct changes across 74 files, including 8 files inside protected core (`agent/src/agent/{context,loop,trace}.py`, `agent/src/session/{models,service,store}.py`, `agent/src/providers/{llm,openai_codex}.py`), none of which had prior discussion — bundling them with unrelated feature work makes atomic review impossible.

Could you split this into separate PRs? Roughly: (1) shutdown-endpoint hardening — looks close to mergeable on its own once isolated; (2) the dotenv `override=True` default flip — a real behavior change that needs its own discussion, and overlaps file-for-file with open PR #563 (same `providers/llm.py` base blob) so worth coordinating with that author too; (3) new Binance loader; (4) OKX loader robustness; (5) remaining protected-core pieces (thought-signature guard, stream retry, fsync-on-write) — each needs its own sign-off; (6) frontend i18n/layout fixes.

Happy to review each independently once split — the shutdown-hardening fix in particular looks solid and worth fast-tracking on its own.

warren618:
Closing this parent bundle. It's a large multi-feature branch that no longer applies cleanly, and it has been superseded by the focused per-feature splits you opened from it — several of which are already merged (Binance loader #643, fsync run state #645, and others). Reviewing and landing those individually is much cleaner. Thanks for breaking it up. 🙏

PR #544, opened 2026-07-14
fix(tools): mark core query tools as repeatable to prevent silent dedup

Marnie0415:
## Summary

- Mark 4 core query tools (`get_market_data`, `search_symbol`, `screen_market`, `get_fundamentals`) as `repeatable = True`
- Fixes silent dedup that drops second tool call when user queries multiple symbols

## Why

`loop.py` deduplicates tool calls by name: once a non-repeatable tool succeeds, subsequent calls with different arguments are silently skipped with `{"skipped": true}`. `BaseTool.repeatable` defaults to `False`, but four core query tools were missing the override.

This causes the most common multi-symbol workflow ("compare AAPL and MSFT") to silently drop the second call, returning stale or missing data with no error.

**Reproduction:**
1. Start a session and ask: "Compare AAPL and MSFT recent prices"
2. Agent calls `get_market_data(codes=["AAPL.US"])` → succeeds
3. Agent calls `get_market_data(codes=["MSFT.US"])` → silently skipped
4. User sees only AAPL data, MSFT data is missing

**Affected tools:**
| Tool | File | Issue |
|------|------|-------|
| `get_market_data` | `market_data_tool.py` | Missing `repeatable = True` |
| `search_symbol` | `symbol_search_tool.py` | Missing `repeatable = True` |
| `screen_market` | `market_screener_tool.py` | Missing `repeatable = True` |
| `get_fundamentals` | `get_fundamentals_tool.py` | Missing `repeatable = True` |

## Changes

Added `repeatable = True` to 4 tool classes:
- `agent/src/tools/market_data_tool.py`
- `agent/src/tools/symbol_search_tool.py`
- `agent/src/tools/market_screener_tool.py`
- `agent/src/tools/get_fundamentals_tool.py`

All 4 tools are idempotent read-only queries that users naturally call multiple times per session.

## Test Plan

- [x] Verified all 4 tools now have `repeatable = True` via Python import test
- [x] No existing tests broken (ran `test_agent_config.py`)
- [ ] Manual test: ask agent to compare two stocks and verify both return data

## Checklist

- [x] Documentation not needed (behavioral fix, no API change)

### DIFF
--- a/agent/src/tools/get_fundamentals_tool.py
+++ b/agent/src/tools/get_fundamentals_tool.py
@@ -114,6 +114,7 @@ class GetFundamentalsTool(BaseTool):
         },
         "required": ["symbols", "fields", "start", "end"],
     }
+    repeatable = True
 
     def execute(self, **kwargs: Any) -> str:
         """Validate inputs, call the loader, and return a JSON envelope.
--- a/agent/src/tools/market_data_tool.py
+++ b/agent/src/tools/market_data_tool.py
@@ -76,6 +76,7 @@ class MarketDataTool(BaseTool):
         },
         "required": ["codes", "start_date", "end_date"],
     }
+    repeatable = True
 
     def execute(self, **kwargs: Any) -> str:
         return fetch_market_data_json(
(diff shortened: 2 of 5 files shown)


warren618:
Merged with the Agent-loop regression expanded to cover all four repeatable market-query tools across iterations. The tests verify parameterized second calls are not silently deduplicated, and mutation checks confirm each repeatable flag is required. The protected Agent loop itself remains unchanged. Thank you for catching this subtle failure mode.

PR #552, opened 2026-07-15
fix(backtest): honor explicit local source for engine routing and benchmark

### LINKED ISSUE #550: [Bug] Explicit local data source is not honored for US equity engine routing and benchmark loading
### Description

**What happened:**

When a backtest explicitly uses the opt-in `local` data source for a frozen US-equity dataset, the data-source choice is not honored end to end:

1. `AAPL.US` is routed to `CryptoEngine` instead of `GlobalEquityEngine`.
2. An explicit benchmark such as `AAPL.US` is fetched through yfinance even when `source: local` is configured.

This gives a local US-equity backtest the wrong market execution/cost rules and prevents an offline, reproducible benchmark comparison.

**What I expected:**

Market engine selection should follow the instrument market, not the loader name. A local `AAPL.US` dataset should use `GlobalEquityEngine` with US-equity rules.

For an explicit source, benchmark loading should use that source's loader. `source: auto` may retain the existing yfinance default when no concrete source is selected.

### Steps to Reproduce

1. Configure a local data bridge containing daily OHLCV data for `AAPL.US`.

2. Run a backtest with the following settings:

```yaml
source: local
codes: [AAPL.US]
benchmark: AAPL.US
start_date: "2023-01-03"
end_date: "2024-12-31"
interval: 1D
```

3. Observe that `_create_market_engine("local", config, ["AAPL.US"])` routes the local US-equity dataset to `CryptoEngine` instead of `GlobalEquityEngine`.

4. Observe that `resolve_benchmark(..., source="local", explicit="AAPL.US")` calls `_fetch_benchmark()` without forwarding the configured source. `_fetch_benchmark()` then creates `YfinanceLoader()`.

5. The benchmark therefore makes a network request even though all required price data is available locally. If yfinance is unavailable or rate-limited, benchmark loading may fail or fall back, making the offline comparison non-reproducible.

### Error Logs

```shell
No exception or stack trace.

Observed behavior:
- local AAPL.US is routed to CryptoEngine instead of GlobalEquityEngine.
- An explicit local benchmark unexpectedly attempts to use yfinance.
```

### Interface

CLI (vibe-trading)

### LLM Provider

N/A (local backtest reproduction)

### Version

0.1.11

### Environment

macOS 15.7.7 / Python 3.12.13

xor-xe:
## Summary

- Route data sources without a dedicated engine branch (e.g. `local`) by detected instrument market, so a local `AAPL.US` dataset gets `GlobalEquityEngine` US-equity rules instead of the `CryptoEngine` default.
- Fetch explicit benchmarks through the configured source's loader first, keeping yfinance as the fallback, so `source: local` backtests stay offline and reproducible.

## Why

Engine routing in `_create_market_engine` fell through to the crypto default for any source without its own branch, and `resolve_benchmark` always created a `YfinanceLoader`, ignoring the configured source. Closes #550.

## Changes

- `agent/backtest/runner.py`: market-based routing (US/HK equity → `GlobalEquityEngine`) in the source fallback branch.
- `agent/backtest/benchmark.py`: `resolve_benchmark`/`_fetch_benchmark` accept the source's loader and try it before yfinance.
- `agent/backtest/engines/base.py`: pass the run's loader through; `source: auto` keeps the yfinance default.
- `agent/tests/test_local_source_routing.py`: routing + benchmark-forwarding regression tests.
- `CHANGELOG.md`: entry under Unreleased → Fixed.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`) — 5391 passed, 10 skipped
- [x] New tests added (`agent/tests/test_local_source_routing.py`)
- [x] Tested manually: verified `_create_market_engine("local", ..., ["AAPL.US"])` returns `GlobalEquityEngine` and benchmark resolution uses the local loader without constructing a yfinance loader

## Checklist


warren618:
Thanks for fixing the engine-routing half of #550. One blocking issue remains: when an explicit `source=local` loader returns no benchmark data or raises, `_fetch_benchmark()` silently falls back to yfinance. That contradicts the PR’s offline/reproducible contract and introduces an unexpected network request.

Please make explicit `source=local` fail closed for benchmark loading—return no external benchmark rather than falling back to yfinance—and replace `test_falls_back_to_yfinance_when_source_loader_has_no_data` with a regression proving `YfinanceLoader` is never constructed for explicit local runs. `source=auto` may retain the existing yfinance fallback.

xor-xe:
Done, in later commit local now fails closed for benchmark loading: no yfinance fallback when the local loader yields no data or raises (or when no loader is passed at all), so the run returns no external benchmark instead of making a network request. The replacement regression test monkeypatches YfinanceLoader to assert it is never constructed for explicit local runs, parametrized over the empty-loader, raising-loader, and missing-loader cases. source: auto and other explicit sources keep the existing yfinance fallback. Also merged main to resolve the CHANGELOG conflict.

warren618:
Holding this — the fix doesn't fully close the gap it's meant to close yet. Ran a focused test directly against this PR's code: when the existing (unrelated to this diff) empty-result fallback in `fetch_data_map()` silently swaps the loader for a network source (e.g. the `yahoo` fallback, which triggers whenever the primary local loader returns nothing for a requested symbol — an easy case to hit, not contrived), `config['source']` never gets updated to reflect that. `resolve_benchmark()` then calls `.fetch()` on whatever loader it's handed *before* checking `allow_fallback` — so a `source=local` run can still make a real network call for the benchmark. That's exactly the behavior the CHANGELOG/docstring claim this PR prevents, and the same class of problem flagged in the original review thread.

The new regression tests only assert `YfinanceLoader` is never constructed, which doesn't cover this path since the leak goes through a different loader class entirely.

Could you add a check in `resolve_benchmark`/`_fetch_benchmark` that respects `allow_fallback`/source *before* calling `.fetch()` on the handed-in loader, not just before constructing a fresh `YfinanceLoader`? Happy to merge once that's closed — the engine-routing half of this fix is solid.

xor-xe:
Done in 488fc2f. resolve_benchmark() now verifies the handed-in loader actually belongs to the configured source (via the registry name attribute) before calling .fetch() on it  so a loader silently swapped by fetch_data_map()'s empty-result fallback is dropped and the run fails closed, instead of the check only applying to a fresh YfinanceLoader.

The fail-closed regression is now parametrized over four cases for explicit source=local: empty local loader, raising local loader, a swapped-in network loader (whose fetch() asserts if ever invoked), and no loader at all,  with YfinanceLoader construction also asserting. That covers both leak paths: fetching through a foreign loader and constructing a fresh network loader.

warren618:
Thanks so much, @xor-xe — this is now merged into `main` as part of a reviewed batch integration, with your original commits and authorship preserved. Really appreciate the contribution!

PR #561, opened 2026-07-16
fix: support Web UI settings persistence on Windows

CRui5in:
## Summary

- Guard the POSIX-only `os.fchmod` call used while atomically persisting Web UI settings.
- Preserve POSIX permission hardening and atomic replacement.
- Make POSIX mode-bit assertions conditional on non-Windows platforms.
- Add a regression test for platforms without `os.fchmod`.

## Root cause

Saving Agent LLM settings on Windows raised `AttributeError: module 'os' has no attribute 'fchmod'`, causing `PUT /settings/llm` to return HTTP 500.

## Validation

- `python -m pytest agent/tests/test_settings_api.py agent/tests/test_api_infrastructure.py -q`
- Local `PUT /settings/llm` smoke test returns HTTP 200 on Windows.

### DIFF
--- a/agent/src/api/helpers.py
+++ b/agent/src/api/helpers.py
@@ -97,9 +97,19 @@ def _atomic_write_secret(path: Path, content: str) -> None:
         return
     try:
         os.write(fd, data)
-        os.fchmod(fd, 0o600)
+        # ``os.fchmod`` is unavailable on Windows.  Keep descriptor-level
+        # permission hardening on platforms that support it, then use the
+        # portable path-based best effort after the descriptor is closed.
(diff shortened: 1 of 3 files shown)


warren618:
Merged — thanks for the fix. Verified independently: reverted to the pre-fix code, reproduced the actual `AttributeError: module 'os' has no attribute 'fchmod'` crash, then confirmed your fix resolves it cleanly with no POSIX regression. One honest note for visibility: this repo's CI only runs on Linux, so there's no automated way to ever exercise the Windows-specific branch going forward — today's manual reproduction is the extent of verification this fix will get.

PR #564, opened 2026-07-16
fix(deps): regenerate runtime lock

### LINKED ISSUE #558: requirements-lock.txt has multiple broken pins causing --require-hashes to fail (caio, websockets, pydantic-core)
### Description

Building the Docker image from a fresh clone of `main` fails during
`pip install --no-cache-dir --require-hashes -r requirements-lock.txt`
with a `ResolutionImpossible` error. I found and manually patched three
separate broken pins so far — each one is a package pinned to a version
that conflicts with another package's declared dependency range, all in
the *same* lockfile. This looks like the lockfile has drifted out of
sync with the actual dependency graph, rather than three unrelated bugs.

### Environment

- Fresh `git clone` of `main`, no local modifications before hitting this
- `docker compose up --build` on Docker Desktop (Windows)
- Base image per Dockerfile: `python:3.11-slim`

### Conflicts found

1. **caio** — locked to `caio==0.10.2`, but `aiofile==3.11.1` (also
   pinned) requires `caio~=0.9.0`.

### Steps to Reproduce

create new container on a fresh system 

### Error Logs

```shell
ERROR: Cannot install -r requirements-lock.txt (line 17) and caio==0.10.2
because these package versions have conflicting dependencies.
The conflict is caused by:
The user requested caio==0.10.2
aiofile 3.11.1 depends on caio~=0.9.0


2. **websockets** — locked to `websockets==16.1`, but `langgraph-sdk==0.4.2`
   requires `websockets<16,>=14`.

3. **pydantic-core** — locked to `pydantic-core==2.47.0`, but
   `pydantic==2.13.4` requires `pydantic-core==2.46.4` exactly.
```

### Interface

Docker

### LLM Provider

ollama

### Version

_No response_

### Environment

_No response_

tianrking:
## Summary

- regenerate the Python 3.11 universal, hash-verified runtime lock
- resolve the incompatible `caio`, `pydantic-core`, and `websockets` pins reported in #558

## Root cause and impact

The previous lock selected versions outside the constraints required by `aiofile`, `pydantic`, and `langgraph-sdk`. That makes the Docker image's `pip install --require-hashes` step fail before the service starts.

## Validation

- `python -m uv pip compile agent/requirements.txt --generate-hashes --universal --python-version 3.11 -o requirements-lock.txt`
- `python -m uv pip install --dry-run --python-version 3.11 --python-platform linux --require-hashes -r requirements-lock.txt`
- `git diff --check`

## Scope and risk

This only refreshes dependency resolution; it does not change broker, credential, MCP, or trading code. Docker is unavailable on this host, so the image build was not run. Reverting this commit restores the previous lock.

Fixes #558


warren618:
Merged — thanks for tracking down the exact pins. Verified independently: reproduced the original `--require-hashes` failure against main's lockfile, then confirmed this PR's regenerated lockfile resolves cleanly (caio 0.9.25, websockets 15.0.1, pydantic-core 2.46.4 — matching what issue #558 needed). Appreciate the fix.

PR #563, opened 2026-07-16
refactor(providers): centralize credentials + fix gateway misroute (#549, #553)

shadowinlife:
## Summary

- Fix gateway provider inference (#549): Added regression tests confirming that explicit gateway providers (OpenRouter/Requesty) are never overridden by model-name inference. The existing `get_provider_capabilities()` explicit-provider check already protects gateway users — `provider="openrouter"` returns openrouter capabilities regardless of model name. No code change needed in the inference logic itself.
- Centralize provider credential resolution (#553): Extract duplicated `provider→env_name→os.getenv→credential` chain from `_sync_provider_env()` and `provider_diagnostics()` into a single `get_llm_credentials()` in `capabilities.py`.
- Add 10 regression tests (4 gateway protection cases + 1 backward-compat case + 5 credential resolution cases) covering gateway / inference / empty / explicit / ollama-fallback scenarios.

## Why

- Closes #549 — adds defense-in-depth regression tests ensuring gateway providers (OpenRouter/Requesty) are never silently misrouted when their model names contain direct-provider prefixes like `deepseek/`. The existing explicit-provider check (`if normalized and normalized != "openai"`) already handles this correctly; the tests prevent regression.
- Closes #553 — maintainability: ~40 lines of identical `os.getenv()` fallback logic were duplicated across two files.
- Both issues share the same fix surface; combining them reduces code churn and ensures every credential-consumption site benefits from the centralized resolver.

## Changes

- `agent/src/providers/capabilities.py`: add `import os` + new `get_llm_credentials()` function + updated docstring for `get_provider_capabilities()`.
- `agent/src/providers/llm.py`: refactor `_sync_provider_env()` and `provider_diagnostics()` to call `get_llm_credentials()`.
- `agent/tests/test_llm.py`: add `test_gateway_provider_not_inferred_from_model` (4 parametrized cases) + `test_default_openai_provider_with_glm_model_infers_zhipu` (backward compat) + `test_uninferable_model_with_empty_provider_falls_back_to_openai` + `TestGetLlmCredentials` class (5 tests).

**Out of scope (by design):**
- `_sync_runtime_env()` in `settings_routes.py` — WRITE operation (saves Web UI settings to env vars), fundamentally different pattern from the READ path.
- `_build_native_deepseek()` in `llm.py` — hard-codes `provider="deepseek"` (native adapter init, not generic credential read).
- `system_routes.py` readiness check — narrower probe (just checks key presence for `/health`), not full credential resolution.

## Test Plan

[...]

Issue #549, opened 2026-07-15
`_infer_from_model` silently misroutes gateway-provider users (OpenRouter/Requesty) when provider field is empty


shadowinlife:
## Description

`_infer_from_model()` in `agent/src/providers/capabilities.py` uses naive model-name prefix matching to reverse-infer the LLM provider, which breaks gateway-style providers (OpenRouter, Requesty) that host other providers' models.

## Problem

When a user configures `langchain_provider=openrouter` (or `requesty`) with `langchain_model_name=deepseek/deepseek-v4-pro`, the provider is correctly matched via `_PROVIDERS["openrouter"]` — but the `_infer_from_model` fallback is a ticking landmine:

```python
def _infer_from_model(model: str) -> str | None:
    lowered = model.strip().lower()
    if lowered.startswith("deepseek"):     return "deepseek"     # ❌ openrouter user!
    if lowered.startswith("gemini"):       return "gemini"       # ❌ openrouter user!
    if lowered.startswith("nvidia/"):      return "nvidia"       # ❌ openrouter user!
    if lowered.startswith("glm"):          return "zhipu"        # ❌ openrouter user!
    if "kimi" in lowered or "moonshot" in lowered: return "moonshot"
    return None
```

The guard in `get_provider_capabilities()` currently prevents this from firing:

```python
if normalized and normalized != "openai":
    return _PROVIDERS.get(normalized, _PROVIDERS["openai"])  # ← returns here for openrouter
inferred = _infer_from_model(model or "")                    # ← only reached for ""/"openai"
```

But the design invites bugs — any future caller that omits `provider` will silently get wrong credentials routed.

## Root cause

PR #247 (2026-06-16, @mvanhorn) introduced `_infer_from_model` to handle the narrow case of a user configuring `provider=openai` with a `glm-*` model name, needing zhipu reasoning capture. Subsequent PRs (#329 added Requesty, NVIDIA NIM patch added `nvidia/`) kept piling prefix checks on top instead of fixing the architecture.

## Reproduce

```python
from src.providers.capabilities import get_provider_capabilities

# User on OpenRouter using DeepSeek — should get OPENROUTER_API_KEY
caps = get_provider_capabilities(provider="openrouter", model="deepseek/deepseek-v4-pro")
assert caps.api_key_env == "OPENROUTER_API_KEY"  # ✅ passes

# Same user, but caller forgets to pass provider — gets wrong credentials
[...]

PR #568, opened 2026-07-16
refactor: centralize duplicated credential resolution in llm.py

### LINKED ISSUE #553: refactor: centralize duplicated credential resolution in llm.py
## Problem

`_sync_provider_env()` and `provider_diagnostics()` in `agent/src/providers/llm.py` independently call `provider_env_names()` → `os.getenv()` with the same fallback chain. The logic is identical but duplicated.

## Proposal

Extract `get_llm_credentials() → dict[str, str]` in `capabilities.py` that resolves `api_key` and `base_url` from `provider_env_names()` + `os.getenv()`. Both functions call this single method instead.

## Context

- Discovered during PR #548 (pluggable OCR engine) where a third copy would have been introduced
- CI env var gate (`tools/ci_env_var_gate.py`) requires `# noqa: env-gate` on all `os.getenv()` calls outside `src/config/`

zerafachris:
Closes #553

Extracts the duplicated credential resolution pattern into a single get_llm_credentials() function in capabilities.py.

Changes:
1. capabilities.py added get_llm_credentials(provider, model) returning dict with api_key and base_url
2. llm.py refactored _sync_provider_env(), provider_diagnostics(), _build_native_deepseek() to use it
3. system_routes.py readiness check uses it instead of independent resolution

The function respects the same fallback chain every call site used:
- api_key: provider-specific env var -> OPENAI_API_KEY -> ollama for keyless providers
- base_url: provider-specific env var -> OPENAI_BASE_URL -> OPENAI_API_BASE

Tests: all credential-related tests pass (26/33; 7 pre-existing failures from missing langchain-openai).
CI env var gate: 0 violations.

### DIFF
--- a/agent/src/api/system_routes.py
+++ b/agent/src/api/system_routes.py
@@ -121,7 +121,7 @@ def _provider_readiness() -> Tuple[bool, str]:
     """
     try:
         from src.config.accessor import get_env_config
-        from src.providers.capabilities import provider_env_names
+        from src.providers.capabilities import get_llm_credentials
         from src.providers.llm import _sync_provider_env
     except Exception as exc:  # noqa: BLE001 — degrade to not-ready, never crash the probe
         logger.warning("readiness: provider config import failed: %r", exc)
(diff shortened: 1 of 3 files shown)


warren618:
Thanks for this, @zerafachris — it's a clean refactor. I'm closing it as a duplicate of #563, which centralizes the same credential resolution into `get_llm_credentials()` and goes a little further: it also fixes the gateway misroute (#549) and adds test coverage (`test_llm.py`). Both PRs close #553.

One piece your PR has that #563 doesn't: routing the `system_routes.py` readiness check through the new helper. That's worth keeping — if you're up for it, a small follow-up on top of #563 adding just that call site would be very welcome and fully credited to you. Thanks again!

PR #576, opened 2026-07-16
Load every page of Futu price history

### LINKED ISSUE #575: Futu price history reads only the first page
## Description

The loader expects the wrong response shape and stops after one page. It should follow Futu's three-part response and reject incomplete history when a later page fails.

## Steps to reproduce

2. Return a page key from the first mocked Futu history response and a second data page from the next response. Current main does not load the full result correctly.

## Expected behavior

## Scope

This issue covers only finding A12. The accompanying pull request uses one signed commit.

## Interface

Python API

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Read Futu's complete response and keep requesting pages until no page key remains. Do not save a partial symbol when a later page fails.

## Why

Closes #575.

## Changes

- Implements only finding A12.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_futu_loader.py -q

## Risk and scope

## Checklist


warren618:

PR #578, opened 2026-07-16
Close only the requested number of option contracts

### LINKED ISSUE #577: Options close signals close the whole position
## Description

The options engine reads the requested close quantity but removes the whole matching position. It should close only the requested contracts and leave the rest open.

## Steps to reproduce

2. Open more than one matching option lot, then request a close smaller than the open quantity. Current main removes the complete matching position.

## Expected behavior

## Scope

This issue covers only finding A13. The accompanying pull request uses one signed commit.

## Interface

Python API

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Close the requested contracts in first-in, first-out order, preserve remaining contracts, and reject over-closes before changing account state.

## Why

Closes #577.

## Changes

- Implements only finding A13.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_options_portfolio_correctness.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -17,7 +17,7 @@ Artifacts: equity.csv, metrics.csv, trades.csv, greeks.csv.
 import json
 import sys
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for this — the FIFO partial-close accounting is correct, and closing only the requested quantity is exactly the right behavior. One thing to fix before I can merge: the new logic raises an uncaught ValueError on an over-close, a missing matching lot, or qty=0, and that exception propagates out of run_options_backtest and aborts the whole backtest. The old code intentionally no-op'd those cases (closing a position already removed by auto-expiry, a double close), so this is a behavioral regression.

- Make over-close / missing-lot / zero-qty degrade gracefully: clamp the close quantity to what's available and no-op when nothing matches, preserving the old semantics — or catch it in the driver's close branch and skip with a warning.
- Add an integration test that runs through run_options_backtest, not just a unit test on the helper.
- Heads-up: this and #580 both add agent/tests/test_options_portfolio_correctness.py with different content — please rebase onto #580 and merge the two test bodies so they don't clobber each other.

Once over-close degrades gracefully and there's an end-to-end test, this is good to go.

warren618:
Thanks! This targets #577, which is already fixed on current main via 387f8bc ("honor partial-close quantity") — it clamps over-closes with min(abs(requested), full_mag), scales cash/PnL to the closed fraction, and rebuilds the remainder as a new immutable OptionPosition. This PR conflicts with that fix, and its FIFO multi-lot variant mutates the position in place and raises on over-close (which aborts a vectorized run, where main degrades gracefully). Closing as superseded. If FIFO multi-lot closing across separately-opened lots is genuinely useful, a fresh small PR against current main — keeping the immutable-rebuild + clamp semantics — would be welcome. Thanks!

PR #580, opened 2026-07-16
Handle unusual options result values safely

### LINKED ISSUE #579: Options metrics crash or report the wrong final equity
## Description

A one-value result reports starting cash instead of the observed value. Negative or invalid final values can also produce numbers that JSON cannot store.

## Steps to reproduce

2. Calculate options metrics for an empty series, one value, zero final equity, negative final equity, and a non-finite value. Current main returns incorrect or unsafe values.

## Expected behavior

## Scope

This issue covers only finding A14. The accompanying pull request uses one signed commit.

## Interface

Python API

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Use the observed final value and return null with a clear warning when an annual or risk metric cannot be calculated safely.

## Why

Closes #579.

## Changes

- Implements only finding A14.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_options_portfolio_correctness.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -532,7 +532,7 @@ def run_options_backtest(
         warnings=config.get("content_filter_warnings") or None,
     )
 
-    print(json.dumps(metrics, indent=2))
+    print(json.dumps(metrics, indent=2, allow_nan=False))
     return metrics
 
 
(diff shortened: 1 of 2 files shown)


warren618:

Issue #583, opened 2026-07-16
Slack optional extras cannot be installed


xkam7ar:
## Description

The Slack install options require a slackify-markdown version that is not published. Use a published range that still provides the function imported by this project.

## Steps to reproduce

1. Check out the current main branch.
2. Ask pip to resolve the Slack optional dependency on current main. Dependency resolution fails because the requested version does not exist.
3. Observe the incorrect behavior described above.

## Expected behavior

The behavior should follow the focused correction described above without changing unrelated features.

## Scope

This issue covers only finding A16. The accompanying pull request uses one signed commit.

## Interface

CLI

## Version

Current main at 86f6012.

## Environment

Verified on macOS with Python 3.13.13. Frontend checks used Node 24.14.0 where relevant.

warren618:
Fixed in #584 (merged to main): slack/channels extras pinned an unpublishable slackify-markdown range.

PR #584, opened 2026-07-16
Fix reliability across packaging, web, scheduler, swarm, and CLI

xkam7ar:
## Summary

- Consolidates the thirteen single-purpose fixes into PR #584, as requested
- Keeps every fix as a separate DCO-signed commit and rebases the full batch onto current `main`.
- Fixes confirmed problems in packaging, the web app, scheduled research, background commands, swarm runs, and the CLI.

## Changes

1. PR #584 fixes the Slack optional dependency so the published package can be installed.
2. PR #586 sends frontend `/auth` requests through the Vite development proxy.
3. PR #588 prevents an old comparison request from replacing newer results.
4. PR #590 keeps run-detail data tied to the current route and ignores late responses from the previous run.
5. PR #592 clears correlation output when its inputs change and ignores stale responses.
6. PR #596 applies standard cron rules when both day-of-month and day-of-week are restricted.
7. PR #600 stops the full process tree after a background-command timeout while preserving exit-code reporting.
8. PR #604 finds nested swarm artifacts, returns stable run-relative paths, and excludes symlink escapes.
9. PR #606 removes private task fields and local paths from the public swarm run response.
10. PR #612 reads and writes continued CLI sessions in one trace directory.
11. PR #614 cleans up development-server child processes after startup failures, interrupts, and exits.
12. PR #618 returns a failure status when a channel start or stop command fails, including JSON output mode.
13. PR #642 ignores authenticated SSE connections that become stale while waiting for a ticket or retry.

## Integration notes

- PR #600 keeps the newer `main` behavior that reports background exit codes and treats a nonzero exit as an error.
- PR #604 keeps the newer swarm output contract tests while adding recursive, contained artifact collection.
- PR #606 keeps the newer swarm SSE behavior and the existing `worker_iterations` response field while adding task redaction.

## Scope and risk

- Affected areas: optional dependencies, the Vite development server, frontend request state, scheduled research, background commands, swarm artifacts and API responses, and CLI process handling.
- Deliberately out of scope: broker and order behavior, live trading, MCP providers, database schemas, deployment, and public documentation.
- Network and secret risk: the Vite change only proxies the existing local `/auth` endpoint. No credentials, tokens, local paths, or private run data were added.
- Rollback: each fix is a separate commit, so a maintainer can revert one fix without reverting the full batch.

## Test Plan

- [x] Full Python suite: 5,557 passed and 10 skipped with the CI exclusions and coverage enabled.
- [x] Focused and adjacent Python regressions: 112 passed.
- [x] Full frontend suite under Node 20.20.2: 284 passed.
- [x] TypeScript compilation and the Vite production build passed under Node 20.20.2.
- [x] Repository safety gates, environment gate, and Python compile checks passed.
- [x] New regression tests cover every fix.
- [x] `git diff --check` passed, and all thirteen commits have matching DCO sign-offs.
- [ ] Manual live testing was not run because these paths are covered by automated regressions and do not require a live broker, deployment, or external service.

## Checklist

- [x] No protected `src/agent/`, `src/session/`, or `src/providers/` area was changed.
- [x] No API keys, secrets, local machine paths, or private data were added.
- [x] Code follows `CONTRIBUTING.md` and `AGENT_CONTRIBUTOR_GUIDE.md`.
- [x] Documentation is not needed because these changes restore existing behavior and add regression coverage.

### DIFF
--- a/agent/cli/_legacy.py
+++ b/agent/cli/_legacy.py
@@ -5332,9 +5332,8 @@ def cmd_dev(
     )
     console.print("[dim]Press Ctrl+C to stop both servers.[/dim]\n")
 
-    backend = subprocess.Popen(backend_cmd, cwd=str(AGENT_DIR))
-    frontend = subprocess.Popen(frontend_cmd, cwd=str(frontend_dir))
-    children = [backend, frontend]
+    children: List[subprocess.Popen] = []
+    exit_code = EXIT_SUCCESS
 
     def _terminate_all() -> None:
         for child in children:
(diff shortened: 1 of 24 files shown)


warren618:
Hi @xkam7ar — thanks for this batch of fixes, they're genuinely useful.

You've opened a lot of small, single-purpose PRs (#584, #586, #588, #590, #592, #596, #600, #604, #606, #612, #614, #618, #642). Many of them touch the same files, so they conflict with one another and are slow to review one at a time.

Could you consolidate all thirteen into this single PR? Just fold the other commits onto this branch and we'll review the batch as one unit — it's much easier for us to land that way and avoids the cross-PR conflicts. We're closing the other twelve to keep the queue readable; nothing is lost, and none of this is a rejection of the work. Thanks again!

warren618:
Thanks @xkam7ar — merged. Consolidating these was exactly right; every folded fix landed with its test and reconciled cleanly against the just-merged connector work. Resolves #603, #605, #611, #613, #617, #639 plus the folded #583/#585/#587/#589/#591/#595/#599.

PR #596, opened 2026-07-16
Match the standard cron rules for days

### LINKED ISSUE #595: Cron requires both restricted day fields to match
## Description

Standard five-field cron runs when either restricted day field matches, but this scheduler requires both fields to match.

## Steps to reproduce

2. Use a schedule with a specific day of month and day of week, then check a date that matches only one field. Current main does not run it.

## Expected behavior

## Scope

This issue covers only finding A26. The accompanying pull request uses one signed commit.

## Interface

Python API

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Use the standard either-or rule when both day fields are restricted and preserve wildcard behavior.

## Why

Closes #595.

## Changes

- Implements only finding A26.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_scheduled_research_executor.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/src/scheduled_research/executor.py
+++ b/agent/src/scheduled_research/executor.py
@@ -111,12 +111,19 @@ def _parse_cron_field(part: str, low: int, high: int) -> set[int] | None:
 
 
 def _day_matches(dt: datetime, doms: set[int] | None, months: set[int] | None, dows: set[int] | None) -> bool:
+    if months is not None and dt.month not in months:
+        return False
+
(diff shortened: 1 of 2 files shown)


warren618:
Closing to consolidate this batch into a single PR — please fold this change into #584 along with your other small fixes (see the note there). The change itself looks fine; this is purely to combine the batch into one reviewable PR. Thanks @xkam7ar!

PR #608, opened 2026-07-16
Handle missing runs and reconnects in swarm event streams

### LINKED ISSUE #607: Swarm event streams hang for missing runs and repeat old events
## Description

A missing run opens a stream that does not end, and browser reconnects ignore the Last-Event-ID header.

## Steps to reproduce

2. Request events for a missing run, or reconnect after event 2 with Last-Event-ID set to 2. Current main hangs or repeats old events.

## Expected behavior

## Scope

This issue covers only finding B15. The accompanying pull request uses one signed commit.

## Interface

Web UI

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Return 404 before streaming a missing run, stop if a run disappears, and resume after the browser's last event ID.

## Why

Closes #607.

## Changes

- Implements only finding B15.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_swarm_route_contracts.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/src/api/swarm_routes.py
+++ b/agent/src/api/swarm_routes.py
@@ -9,7 +9,7 @@ import json
 from pathlib import Path
 from typing import Any, Awaitable, Callable
 
-from fastapi import Depends, FastAPI, HTTPException, Query, Request
+from fastapi import Depends, FastAPI, Header, HTTPException, Query, Request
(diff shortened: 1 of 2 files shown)


warren618:

PR #610, opened 2026-07-16
Send the required details when approving a mandate

### LINKED ISSUE #609: CLI mandate approval sends an invalid request
## Description

The approval command omits the broker required by the API and drops the optional account reference.

## Steps to reproduce

2. Approve a mandate proposal through the CLI and inspect the commit request. Current main sends no broker, so the API rejects it.

## Expected behavior

## Scope

This issue covers only finding B19. The accompanying pull request uses one signed commit.

## Interface

CLI

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Copy broker and account reference from the proposal, validate the complete request, and stop before the network call when required data is missing.

## Why

Closes #609.

## Changes

- Implements only finding B19.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_cli_live.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/cli/main.py
+++ b/agent/cli/main.py
@@ -1043,18 +1043,29 @@ def _commit_mandate(proposal: Dict[str, Any], selected_ordinal: int) -> Dict[str
     """
     import httpx
 
+    from src.api.live_routes import CommitMandateRequest
     from src.config.accessor import get_env_config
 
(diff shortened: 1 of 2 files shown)


warren618:

PR #622, opened 2026-07-16
Do not mark successful local data as missing

### LINKED ISSUE #621: Successful local data is marked as missing
## Description

The local loader removes the local prefix from its result, but the missing-data check looks only for the original request name.

## Steps to reproduce

2. Request a local-prefixed symbol and return valid data under the shorter symbol name. Current main also reports the request as unresolved.

## Expected behavior

## Scope

This issue covers only finding B27. The accompanying pull request uses one signed commit.

## Interface

Python API

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Match both the requested local-prefixed symbol and the shorter result name when checking for missing data.

## Why

Closes #621.

## Changes

- Implements only finding B27.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_market_data.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/src/market_data.py
+++ b/agent/src/market_data.py
@@ -89,6 +89,12 @@ def fetch_market_data(
 ) -> dict[str, Any]:
     """Fetch normalized OHLCV data through the repository loader layer."""
     results: dict[str, Any] = {}
+    result_aliases = {
+        code: code.split(":", 1)[1]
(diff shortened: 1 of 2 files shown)


warren618:

PR #626, opened 2026-07-16
Include the complete final month in journal filters

### LINKED ISSUE #625: Month-range filters omit most of the final month
## Description

A month-only end value adds one day, so the filter includes only the first day of the final month.

## Steps to reproduce

2. Filter a journal with 2024-01 to 2024-02 and include trades on February 29. Current main omits most of February.

## Expected behavior

## Scope

This issue covers only finding B31. The accompanying pull request uses one signed commit.

## Interface

Python API

## Version

Current main at 86f6012.

## Environment

xkam7ar:
## Summary

- Treat month-only end values as the end of the whole month while preserving full-date behavior.

## Why

Closes #625.

## Changes

- Implements only finding B31.

## Test Plan

- [x] Focused regression check passed: cd agent && pytest tests/test_trade_journal.py -q

## Risk and scope

## Checklist

### DIFF
--- a/agent/src/tools/trade_journal_tool.py
+++ b/agent/src/tools/trade_journal_tool.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import json
 import logging
+import re
 from collections import defaultdict, deque
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


warren618:

PR #636, opened 2026-07-16
fix(ibkr): thread-local pool + snapshot quotes to fix hangs under parallel agent execution

MikeCer:
## Problem

The local IBKR connector had two critical issues that caused failures under the agent's parallel tool executor (ThreadPoolExecutor, up to 8 workers):

**1. Client-ID collision:** Every tool call opened a fresh ib_async.IB() connection with the same hard-coded client ID. When the agent fired multiple trading tools in parallel, TWS/IB Gateway rejected duplicates with error 326: "Unable to connect as the client id is already in use."

**2. Thread-safety deadlock:** The initial connection pool shared a single socket across all threads, but ib_async is not thread-safe — its internal event loop is bound to one thread. Parallel 	rading_quote calls deadlocked.

**3. Quote timing & reliability:** 
eqMktData(..., snapshot=True) returns immediately but the ticker fields populate asynchronously. The connector originally read fields instantly (returning None/NaN), then had a fixed 2s sleep (unreliable under parallel load — only 3/10 quotes returned data), and finally a waitOnUpdate polling loop that exited prematurely because waitOnUpdate returns True for any socket message, not just the targeted ticker.

## Solution

**Thread-local connection pool** (_TwsPool):
- Each worker thread gets its own ib_async.IB socket with a unique client ID (atomic counter via itertools.count)
- Per-thread reference counting: connection persists across multiple tool calls within the same thread, disconnects only when refcount hits zero
- All five public functions use _pool.acquire(cfg) / _pool.release()

**Incremental ib.sleep() polling for quotes** (_wait_for_tick):
- Uses short ib.sleep(0.1) increments (NOT waitOnUpdate) to pump the event loop
- Checks _tick_has_data after each cycle — rejects both None and NaN fields
- Exits as soon as real market data arrives, with a 5s total timeout
- ib.sleep() reliably processes all pending socket messages per call, unlike waitOnUpdate which can return True for unrelated messages

**Contract qualification errors surfaced:** _qualify_contract now raises IBKRConnectionError instead of silently swallowing failures (e.g., Samsung KRX contracts that can't be qualified).

## Impact

- No more client-ID collision errors under parallel agent call patterns
- No more deadlocks — each thread has its own isolated IB connection
- Quotes return real data reliably (9/10 symbols in testing) instead of timeout/NaN
- No resource leak from orphaned connections
- 14 tests pass including concurrency tests (thread isolation, refcounting, idempotent release) and timing tests (delayed tick arrival, NaN→real data polling)


warren618:
Thanks for this, @MikeCer — the thread-local IBKR connection pool is a genuinely good improvement and the right pattern here. One correctness issue in the new reqMktData path to fix before merge: it reads the quote immediately without waiting for the async tick to arrive, whereas the previous code had a short wait. In production this will frequently read empty/None values. The current test hides it because its mock returns instantly, so the timing regression never surfaces.

- Restore an explicit wait for the market-data tick before reading, or switch to a snapshot/blocking request.
- Add a test that exercises the async-arrival timing (tick arriving after a delay) rather than an instantly-returning mock.
- A concurrency test for the new pool would be a great addition too.

Once the read waits for the tick and there's a test covering the timing, this is good to merge.

MikeCer:
Thanks @warren618  good catch about timing regression issue. 
The ib_async docs state: "The ticker will initially be empty and gradually (after a couple of seconds) be filled" even with snapshot=True. BTW, I decided to leverage of the waitOnUpdate to handle the event loop timing in 100ms increments until at least one of bid/ask/last is populated, with a 2 secs timeout. This improves the overall call performances. Regarding testing, now the test cases are also covering timing handling and thread pool concurrency scenarios.

MikeCer:
I performed extensive testing using a local IB Gateway instance and decided to revert the waitOnUpdate approach. It is designed for streaming subscriptions rather than snapshot mode, as it returns True when any update is received on the socket—not for a specific ticker.
A fixed sleep is also not a reliable solution, so I implemented a hybrid approach: incremental waits of 0.1 seconds combined with context-aware checks on the data structure.

warren618:
Merged to `main` — thanks @MikeCer. Thread-local pool + snapshot quotes for the IBKR connector fixes the hangs under parallel agent execution. Solid fix. 🙏

PR #637, opened 2026-07-17
fix(security): harden shutdown authorization

tyj147454413-cmd:
Split from #543 for independent review.\n\nRequires explicit shutdown confirmation plus either a valid `API_AUTH_KEY` or local-only `VIBE_TRADING_ALLOW_LOCAL_SHUTDOWN=1` opt-in.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_security_auth_api.py -q` (53 passed).

### DIFF
--- a/agent/src/api/security.py
+++ b/agent/src/api/security.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import hmac
 import ipaddress
 import logging
+import os
 import re
 import secrets
 import threading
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for tackling this — the shutdown-authorization hardening is correct and exactly the behavior we want. The gap is only in the tests: all three edited cases set `API_AUTH_KEY` and append `?confirm=true`, which the old code ignored too, so they pass identically against both implementations and exercise none of the new branches. No code change needed — just add regression tests that fail without the fix:

- api-key set + no `confirm` → 400
- no key + no `VIBE_TRADING_ALLOW_LOCAL_SHUTDOWN` → 403
- no key + `VIBE_TRADING_ALLOW_LOCAL_SHUTDOWN=1` + `confirm` + loopback → 200

Once those genuinely guard the new logic, this is good to merge.

warren618:
Thanks for the security focus here. The main threat this targets — a cross-site browser fetch() hitting the shutdown endpoint — is already mitigated on current main by _reject_cross_site_browser_request (rejects sec-fetch-site: cross-site and mismatched Origin, checked first). As written, this change regresses the default local-shutdown path (loopback + no api-key now 403s unless an undocumented VIBE_TRADING_ALLOW_LOCAL_SHUTDOWN env is set) and places the confirm gate before the api-key check, which breaks the authenticated shutdown contract. Closing for now. If you'd like defense-in-depth here, a reworked PR that applies only the confirm=true gate, leaves the authenticated path unchanged, and documents any new env var would be considered. Thanks!

PR #641, opened 2026-07-17
Honor cancellation before the first AgentLoop iteration

### LINKED ISSUE #638: Cancelling a queued session does not stop it from starting
## Description

A cancellation can be accepted after a session publishes its AgentLoop but before the executor starts it. `AgentLoop.run()` clears that cancellation and still calls the model.

## Steps to reproduce

1. Queue a session run.
2. Cancel it after the AgentLoop is available but before its first iteration.
3. Let the executor call `run()`.

The model is called even though the cancellation was accepted.

## Expected behavior

A request cancelled before its first iteration should finish as cancelled without calling the model or any tools.

## Proposed fix

Preserve the first cancellation signal. Reset it only when a completed AgentLoop is deliberately reused.

xkam7ar:
## Summary

- Preserve cancellation accepted while an AgentLoop is waiting to start.
- Reset cancellation only when a completed loop is deliberately reused.
- Add a regression proving the cancelled run makes zero model calls.

## Why

Closes #638.

A queued cancellation is currently cleared at the start of `run()`, so work that the user cancelled can still call the model.

## Changes

- Track whether the loop has run before.
- Keep the initial cancellation event intact.
- Preserve the existing reuse behavior after a completed run.

## Test Plan

- [x] New tests added
- [ ] Tested manually
- [x] 9 AgentLoop terminal-state tests passed
- [x] 12 adjacent stream, content-filter, and session tests passed
- [x] Ruff and diff checks passed

## Risk

This changes AgentLoop lifecycle state. Existing mid-stream cancellation behavior is unchanged. No real provider, broker, network server, or credential was used.

## Out of scope

This does not change cancellation after model or tool execution has started.

## Rollback

Revert this one commit to restore the previous cancellation reset behavior.

## Checklist

- [x] The protected agent area is covered by issue #638 and focused tests
- [x] No hardcoded API keys, file paths, or magic values

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -550,6 +550,7 @@ class AgentLoop:
         self._previous_summary: str = ""
         self._persistent_memory = persistent_memory
         self._run_iteration: int = 0
+        self._has_run = False
 
     def cancel(self) -> None:
         """Cancel the current loop.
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @xkam7ar — merged. Preserving a pre-first-iteration cancel closes the queued-cancel gap (#638).

PR #643, opened 2026-07-17
feat(loaders): add Binance crypto fallback

tyj147454413-cmd:
Split from #543 for independent review.\n\nAdds an explicit Binance OHLCV loader, registers `binance` as a valid source, and makes the crypto fallback chain `OKX → Binance → CCXT → Yahoo`.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_registry.py agent/tests/test_binance_fallback.py -q` (25 passed).

### DIFF
--- a/agent/backtest/loaders/binance_loader.py
+++ b/agent/backtest/loaders/binance_loader.py
@@ -1,0 +1,44 @@
+"""Binance spot / USD-M perpetual OHLCV loader (via CCXT).
+
+Dedicated source name ``binance`` so it can sit **alongside** ``okx`` in the
+crypto auto-fallback chain — not as a replacement. Public market data only;
+no API key required.
+
+Use explicitly with ``source="binance"``, or let ``source="auto"`` fall through
+to Binance when OKX is unavailable.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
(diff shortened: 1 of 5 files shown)


warren618:
Thanks for this — the Binance loader and its registration are clean and slot in nicely. One behavioral issue to fix before merge: the new fallback loop in `fetch_market_data` catches `NoAvailableSourceError` and walks to the next chain source even when the user pinned `source='local'` (or `local:` symbols). `local` is a no-network-fallback source on purpose, so this silently serves network data (e.g. tencent) for a request meant to stay offline, masking what should be a fail-loud config error — the registry deliberately raises there and the old path let it propagate.

- Don't let an unavailable no-network-fallback source (or a propagated `NoAvailableSourceError` for the explicitly requested source) fall through to network sources — keep it fail-loud / unresolved for `local` and `qveris`.
- Add a test covering the `local`-source case so this stays fixed.

With that guard in place it's ready to merge.

warren618:
Merged to `main` — thanks @tyj147454413-cmd. The Binance crypto fallback is now wired into the loader registry (22nd source). I added a small follow-up commit bumping the skill manifest count 21→22 so the manifest-consistency test stays green. 🙏

PR #644, opened 2026-07-17
fix(okx): bound history fetches and handle rate limits

tyj147454413-cmd:
Split from #543 for independent review.\n\nContains only OKX loader pagination, rate-limit handling, and bounded history-fetch changes.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_okx_loader_bounded.py -q` (12 passed).

### DIFF
--- a/agent/backtest/loaders/okx.py
+++ b/agent/backtest/loaders/okx.py
@@ -1,11 +1,28 @@
 """OKX spot candle loader (crypto).
 
 Uses OKX V5 public REST API (no auth).
-Supports 1m/5m/15m/30m/1H/4H/1D.
-Up to 300 bars per request; paginates with ``after`` for longer history.
+
+Endpoints
+---------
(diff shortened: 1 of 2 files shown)


warren618:
Nice work here — the dual-endpoint OKX fetch, bounded retry, and proxy / rate-limit handling are solid. Two things to change before merge:

- `is_available()` now does a live network probe (it used to return `True`). That adds a round-trip to every crypto resolution and, worse, breaks offline cached-data use: offline the probe fails, OKX is skipped, and its okx-namespaced cache never gets read. Please keep `is_available()` cheap (return `True` — `fetch` already bounds failures and the chain handles real outages), or make the probe non-fatal and cached.
- `_should_use_history` uses the deprecated `pd.Timestamp.utcnow()`, which warns on pandas 3 and raises on pandas <3. Replace it with `pd.Timestamp.now('UTC').tz_localize(None)`.

Everything else looks good.

PR #645, opened 2026-07-17
fix(state): fsync persisted run state

tyj147454413-cmd:
Split from #543 for independent review.\n\nFlushes and fsyncs `state.json` writes to avoid truncated state after a process crash.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_state_fsync.py -q` (2 passed).

### DIFF
--- a/agent/src/core/state.py
+++ b/agent/src/core/state.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import os
 import uuid
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for looking at run-state durability — this is a real concern worth fixing. As written, though, it doesn't yet deliver what the description promises: it opens the real `state.json` in `'wb'` mode, which truncates the file to zero in place before writing, so a crash mid-write still leaves an empty/truncated file — exactly the failure it's meant to prevent. `fsync` only guards against power loss after a completed write, not the process-crash case.

- Write to a same-directory temp file, `fsync` it, `os.replace()` it onto `state.json`, then `fsync` the parent directory.
- We already have this atomic-write pattern in `scheduled_research/store.py` and `backtest/engines/base.py` — please mirror one of those so it stays consistent with the rest of the codebase.

With that in place this is good to merge.

warren618:
Merged to `main` — thanks @tyj147454413-cmd. fsync on persisted run state closes the crash-durability gap. 🙏

PR #653, opened 2026-07-17
fix(providers): guard thought-signature payload alignment

tyj147454413-cmd:
Split from #543 for protected-core review.\n\nStops outbound payload reinjection when upstream serialized messages and converted messages diverge in length, preventing an `IndexError` and stale-message cross-pollination.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_kimi_reasoning_content.py -q` (36 passed).

### DIFF
--- a/agent/src/providers/llm.py
+++ b/agent/src/providers/llm.py
@@ -268,6 +268,8 @@ if ChatOpenAI is not None:
             for i, m in enumerate(payload["messages"]):
                 if m.get("role") != "assistant":
                     continue
+                if i >= len(messages):
+                    break
                 source_message = messages[i]
                 if caps.normalize_assistant_content and m.get("content") is None:
                     m["content"] = ""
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for the patch. Looking closely against current `main`, we don't think the guard covers a reachable path: in `_get_request_payload`, both `payload["messages"]` and `super()._convert_input(input_).to_messages()` come 1:1 from the same input via the same `_convert_input`, so the lengths are always equal in practice and `i` never reaches `len(messages)`. We also reproduced the added test on `main` without the guard and it passes unchanged — the stub makes `payload["messages"]` shorter (length 1) than the converted list (length 2), so the branch is never exercised and no IndexError occurs. Since the test doesn't demonstrate the described failure and the condition isn't reachable in production, we're closing this. If you can share a concrete repro where `payload["messages"]` is genuinely longer than the converted messages, we'll gladly reopen. Appreciate the effort.

PR #654, opened 2026-07-17
Use one prefetched dataset for every backtest source

### LINKED ISSUE #646: Explicit-source backtests fetch and use market data twice
## Description

A backtest with an explicit data source fetches and validates one dataset, then the engine fetches the same source again. The second response can differ from the run card and can also repeat a paid provider charge.

## Steps to reproduce

1. Use an explicit source such as `yahoo`.
2. Count calls made by that loader during one backtest.
3. Return different data from the second call.

The loader is called twice and the engine uses the second response.

## Expected behavior

The engine should use the one dataset that was already fetched, validated, sanitized, and recorded.

## Proposed fix

Wrap the prefetched data for every source, not only `auto`, and pass that snapshot to the engine.

xkam7ar:
## Summary

- Reuse the dataset already fetched and validated for explicit sources.
- Keep run-card provenance aligned with the data used by the engine.
- Add a regression proving the provider is called once.

## Why

Closes #646.

Explicit-source backtests currently fetch twice. The second call can change the data and repeat provider cost.

## Changes

- Use the existing prefetched loader adapter for every source.
- Keep the auto-source path unchanged.
- Verify the engine receives the first validated snapshot.

## Test Plan

- [x] New tests added
- [x] 12 relevant runner and service tests passed

## Risk

The engine API is unchanged. The input now comes from the validated snapshot already used for provenance. No provider credential or paid endpoint was used.

## Out of scope

This does not change source selection or fallback order.

## Rollback

Revert this one commit to restore the second loader call.

## Checklist

### DIFF
--- a/agent/backtest/runner.py
+++ b/agent/backtest/runner.py
@@ -919,9 +919,10 @@ def main(run_dir: Path) -> None:
     else:
         bars_per_year = calc_bars_per_year(interval, effective_source)
 
-    # Auto mode: wrap preloaded data in a dummy loader
-    if source == "auto":
-        loader = _AutoLoader(data_map)
+    # Every source has already been fetched, sanitized, and enriched above.
+    # Reuse that exact snapshot so provider costs and run-card provenance stay
+    # aligned with the data consumed by the engine.
+    loader = _AutoLoader(data_map)
 
     if engine_type == "options":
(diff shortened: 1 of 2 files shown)


warren618:

PR #655, opened 2026-07-17
Keep risk-parity portfolios on the long-only simplex

### LINKED ISSUE #647: Risk parity can produce short weights in a long-only backtest
## Description

The risk-parity refinement can produce negative weights for valid covariance matrices with adverse correlations. A long-only signal can therefore become a short position.

## Steps to reproduce

Run `RiskParityOptimizer` with a positive semidefinite covariance matrix whose correlations make one intermediate risk contribution negative.

## Expected behavior

Risk-parity weights should be finite, nonnegative, and sum to one.

## Proposed fix

Solve the equal-risk-contribution objective on the long-only simplex and keep the inverse-volatility seed as a safe fallback.

xkam7ar:
## Summary

- Constrain risk-parity weights to the long-only simplex.
- Fall back to finite inverse-volatility weights if optimization fails.
- Add an adverse-correlation regression.

## Why

Closes #647.

The previous multiplicative refinement can produce negative weights and turn long signals into shorts.

## Changes

- Solve equal risk contribution with nonnegative bounds and a sum-to-one constraint.
- Reject non-finite covariance input safely.
- Normalize the result before returning it.

## Test Plan

- [x] New tests added
- [x] Focused risk-parity tests passed
- [x] 30 adjacent optimizer tests passed

## Risk

This changes portfolio weights for cases where the old method was unstable. The fallback remains long-only and finite. Synthetic covariance matrices were used; no market account was contacted.

## Out of scope

This does not add short risk parity or change other optimizers.

## Rollback

Revert this one commit to restore the previous iterative calculation.

## Checklist

### DIFF
--- a/agent/backtest/optimizers/risk_parity.py
+++ b/agent/backtest/optimizers/risk_parity.py
@@ -1,7 +1,4 @@
-"""Risk parity: equalize marginal risk contributions.
-
-Iterative refinement so w_i * MRC_i is approximately equal across assets.
-"""
+"""Long-only risk parity: equalize marginal risk contributions."""
 
 from typing import Any, Dict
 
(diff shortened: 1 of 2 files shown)


warren618:

PR #657, opened 2026-07-17
Apply alpha_t_train during strict OOS confirmation

### LINKED ISSUE #649: Strict Alpha Bench can confirm a factor that fails in training
## Description

With an out-of-sample split, Strict Alpha Bench can mark a factor as confirmed when the full sample and test period pass even though the training period fails the configured alpha threshold.

## Expected behavior

A confirmed factor should pass the strict threshold in both training and test periods.

## Proposed fix

Require `alpha_t_train` as well as `alpha_t_test` to clear the configured threshold before returning `confirmed_alive`.

xkam7ar:
## Summary

- Require the training statistic during strict out-of-sample confirmation.
- Preserve behavior when no split is requested.
- Add a regression for a failing train period and passing test period.

## Why

Closes #649.

Strict Alpha Bench currently reads the full and test statistics but can ignore a failed training statistic.

## Changes

- Gate `confirmed_alive` on both split statistics.
- Return noise when training does not meet the configured threshold.
- Keep the existing test-period sign-flip rule.

## Test Plan

- [x] New tests added
- [x] Strict benchmark tests passed
- [x] Factor safety gates passed

## Risk

Some previously confirmed factors will be downgraded when their training evidence is weak. No factor formula or market data is changed.

## Out of scope

This does not change thresholds or no-split mode.

## Rollback

Revert this one commit to restore the previous classifier.

## Checklist

### DIFF
--- a/agent/src/factors/bench_runner_strict.py
+++ b/agent/src/factors/bench_runner_strict.py
@@ -253,7 +253,7 @@ def categorise_strict(
 
     Rules:
         - ``alpha_t_full <= -thr``                          → ``reversed_strict``
-        - ``alpha_t_full >=  thr`` and (no OOS or ``alpha_t_test >= thr``)
+        - ``alpha_t_full >=  thr`` and (no OOS or both split t-stats pass)
                                                             → ``confirmed_alive``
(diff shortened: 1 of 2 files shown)


warren618:

PR #658, opened 2026-07-17
Require enough observations in every strict OOS split

### LINKED ISSUE #650: Strict Alpha Bench ignores undersized train and test splits
## Description

Strict Alpha Bench checks the full-sample observation count but does not apply the same minimum to each out-of-sample split. A factor can be confirmed from a test period with only a few observations.

## Expected behavior

Training and test periods should each meet `min_ic_count` before a strict category is assigned.

## Proposed fix

Apply the existing observation floor independently to both split counts.

xkam7ar:
## Summary

- Apply the existing observation floor to both OOS splits.
- Reject missing or undersized split counts.
- Add an underpowered-test-period regression.

## Why

Closes #650.

A large combined sample can currently hide a training or test period with too few observations for strict confirmation.

## Changes

- Read the existing train and test counts during classification.
- Require each count to meet `min_ic_count`.
- Preserve no-split behavior.

## Test Plan

- [x] New tests added
- [x] Strict benchmark tests passed
- [x] Factor safety gates passed

## Risk

Some categories will become noise when either split is underpowered. The configured floor itself is unchanged.

## Out of scope

This does not change split dates or t-statistic calculation.

## Rollback

Revert this one commit to restore the aggregate-only count check.

## Checklist

### DIFF
--- a/agent/src/factors/bench_runner_strict.py
+++ b/agent/src/factors/bench_runner_strict.py
@@ -250,6 +250,8 @@ def categorise_strict(
         - ``alpha_t_train`` (Optional[float])  — None when no OOS split was run
         - ``alpha_t_test`` (Optional[float])   — None when no OOS split was run
         - ``ic_count`` (int)
+        - ``ic_count_train`` / ``ic_count_test`` (Optional[int]) — split
+          sample counts when OOS was run
 
     Rules:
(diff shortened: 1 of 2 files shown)


warren618:

PR #659, opened 2026-07-17
Keep unmatched Shadow Account markets at zero

### LINKED ISSUE #651: Shadow backtests apply rules to the wrong market
## Description

When no Shadow Account rule matches a symbol market, the generated engine falls back to the first rule. A US-only profile can therefore trade China, Hong Kong, and crypto symbols.

## Steps to reproduce

Generate an engine with one US rule and request signals for US, China, Hong Kong, and crypto symbols.

## Expected behavior

Unmatched markets should stay flat. An explicit `other` rule should remain the only wildcard.

## Proposed fix

Return no rule when the market loop has no match.

xkam7ar:
## Summary

- Remove the implicit first-rule fallback for unmatched markets.
- Preserve matching rules and the explicit `other` wildcard.
- Add generated-engine coverage across four markets.

## Why

Closes #651.

A market with no matching rule currently receives the first unrelated rule and can trade.

## Changes

- Return no rule after an unsuccessful market lookup.
- Verify a US-only profile stays flat in China, Hong Kong, and crypto.

## Test Plan

- [x] New tests added
- [x] The complete Shadow Account test suite passed

## Risk

Unmatched markets will now remain flat. Explicit cross-market `other` rules continue to work. Tests use synthetic prices only.

## Out of scope

This does not change symbol market detection.

## Rollback

Revert this one commit to restore the first-rule fallback.

## Checklist

### DIFF
--- a/agent/src/shadow_account/templates/signal_engine.py.j2
+++ b/agent/src/shadow_account/templates/signal_engine.py.j2
@@ -75,7 +75,7 @@ class SignalEngine:
         for rule in RULES:
             if rule["market"] == market or rule["market"] == "other":
                 return rule
-        return RULES[0] if RULES else None
+        return None
 
     @staticmethod
     def _conditional_entry(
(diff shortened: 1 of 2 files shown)


warren618:

PR #660, opened 2026-07-17
Keep Shadow entry features point-in-time safe

### LINKED ISSUE #652: Shadow entry features include the buy day's future close
## Description

For an intraday journal entry, Shadow feature extraction includes the eventual daily close for the buy date. That value was not available when the trade was opened, so learned rules and backtest results contain lookahead.

## Steps to reproduce

1. Build daily history through an intraday buy date.
2. Change only the close on the buy date.
3. Recompute the entry RSI and prior return.

Both features change even though that daily bar was incomplete at entry time.

## Expected behavior

Daily entry features should use only sessions completed before the buy date.

## Proposed fix

Exclude the normalized buy date from the as-of slice and add an extreme-close invariance test.

xkam7ar:
## Summary

- Exclude the incomplete buy-date daily bar from entry features.
- Use only completed prior sessions for RSI and prior return.
- Add an extreme-close invariance regression.

## Why

Closes #652.

The buy-day close is not available at an intraday entry, but it currently changes the learned Shadow features.

## Changes

- Slice daily history strictly before the normalized buy date.
- Update the availability comments.
- Prove an extreme buy-day close cannot change the features.

## Test Plan

- [x] New tests added
- [x] The complete Shadow Account test suite passed

## Risk

Entries recorded after market close also use the prior day because the journal does not provide a reliable per-market availability calendar. This is the conservative point-in-time rule. Tests use synthetic prices only.

## Out of scope

This does not add exchange calendars or intraday bars.

## Rollback

Revert this one commit to restore the inclusive daily slice.

## Checklist

### DIFF
--- a/agent/src/shadow_account/extractor.py
+++ b/agent/src/shadow_account/extractor.py
@@ -190,8 +190,9 @@ def _fetch_price_history(
             normalization in v1).
         market: Journal market label (e.g. ``"china_a"``).
         start: Inclusive fetch start (already buffered for indicator warmup).
-        end: Inclusive fetch end — never later than the roundtrip's buy_dt, so
-            look-ahead is structurally impossible.
+        end: Inclusive fetch end. Entry features still exclude the buy date
+            because this loader returns completed daily bars, not intraday
+            availability timestamps.
 
     Returns:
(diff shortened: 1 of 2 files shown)


warren618:

PR #661, opened 2026-07-17
fix(frontend): improve run detail and session UX

tyj147454413-cmd:
Split from #543 for independent frontend review.\n\nIncludes layout/session UX, run-detail empty-state handling, and the localized copy it requires. All locale files carry the new run-detail keys so the existing parity contract remains valid.\n\nVerification: `npm run test:run` (247 passed) and `npm run build`.


warren618:
Thanks — the RunDetail UX work is a nice improvement and the 7 new keys are correctly added to all five locales. One gap to close before merge: five other new user-facing strings go through `t()` but are missing from every locale file — `layout.sessionsLoadFailed`, `layout.deleteFailed`, `layout.renameFailed`, `agent.failedToLoadHistory`, `agent.failedToLoadSettings`. They render only via their inline English `defaultValue`, so non-English users get English error toasts, and the i18n parity test passes only because the keys are absent everywhere.

- Add those five keys to the locale files — at minimum `en.json` and `zh-CN.json`, ideally all five — so the error toasts localize like the rest of the new UI.

Once they're in, this is ready to merge.

warren618:
Thanks — the RunDetail empty-state + `failedReason` block is a genuine, correct improvement (`run.reason` is a real backend field, i18n usage matches the existing pattern), and the cache bump / toast handling are fine.

Two things before merge:

1. i18n (blocking): the 5 new toast strings — `layout.sessionsLoadFailed`, `layout.deleteFailed`, `layout.renameFailed`, `agent.failedToLoadHistory`, `agent.failedToLoadSettings` — are defaultValue-only and in no locale file, so zh-CN users see English toasts. You already did `runDetail.*` correctly across all 5 locales; please add these 5 keys to all locale files too (en + zh-CN mandatory).
2. The tool-id fix is half-wired: the backend SSE tool events don't emit `call_id` (loop.py's `tool_call`/`tool_result` sites only carry `tool`), so `d.call_id` is always undefined and it degrades to "most-recent-running-by-name" — which still can't disambiguate two concurrent same-name calls, the stated goal. Either add `call_id=tc.id` to the 3 backend emit sites so it works end-to-end, or drop the call_id framing and document it as a name+running heuristic.

Once the locale keys are in, we'll merge. Thanks!

warren618:
Thank you for splitting #543 into reviewable pieces; #662 and #675 from that
series have both been merged with your authorship.

This one I'm closing, because the frontend it targets has been through two
overhauls since the branch was cut and the merge would now regress behavior we
added deliberately. Specifically, the tool-call fallback here resolves an
unknown `call_id` to the newest running tool with the same name; the current
store refuses name-based fallback whenever a call ID is present, and there's a
test covering interleaved same-name calls that this would break. The
empty-report predicate is also now a shared helper that additionally accounts
for validation, price series, trade markers and the separately fetched code
object, so the narrower version here can hide the code tab when code did load.

Two things from this PR are still worth having and aren't covered upstream:
the swallowed session list/delete/rename errors in the layout, and the missing
`reason` field on the run type. I'll pick those up separately. If you'd like to
send them yourself against current main, I'll review promptly.

PR #662, opened 2026-07-17
fix(trace): harden artifact writes and persistence

tyj147454413-cmd:
Split from #543 for protected-core review.\n\nHardens trace artifact path handling and fsyncs trace writes for crash durability.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_trace_writer.py -q` (5 passed).


warren618:
Thanks — the `os.fsync` in `write()` is a fair point; the docstring claims crash-safety and we were only calling `flush()` (and `store.append_message` already fsyncs). Two asks before merging: (1) please drop the `try/except` around the bare `pass` in `__init__` — it wraps a no-op and reads as dead code; (2) `TraceWriter.write` runs per start/message/tool_result/LLM event, so an fsync on every call is a lot of blocking syscalls per run in the agent thread. Could you make it opt-in (env-gated) or fsync on close / periodically rather than every record? We want the durability without a per-event fsync tax on long runs. Thanks!

warren618:
Thanks — the fsync in `write()` is correct and worth keeping: it makes the "crash-safe" docstring actually true and matches what `store.py` already does for messages (flush + `os.fsync`).

One cleanup before merge: the `__init__` change wraps a `try: pass / except: close+raise` around nothing — `try: pass` can't raise, so the `except` is unreachable dead code (and against the house rule of not keeping useless code). Please strip that sentinel block and keep just the one-line fsync in `write()`. Then it's a clean merge. Thanks!

warren618:
Merged with your authorship, extended in one respect: the fsync on the JSONL
record needed the sidecar to be durable first. Large tool results and answers
were still written with a plain `write_text`, so a crash could persist a
fsynced record whose `result_path` pointed at a missing or truncated file. The
merged version writes sidecars temp -> fsync -> `os.replace` -> parent-directory
fsync before the referencing record is written, fsyncs new directory entries,
and degrades to flush-only with a single warning if fsync is unsupported.
Thanks — the durability gap you identified was real.

PR #663, opened 2026-07-17
fix(codex): classify stream HTTP failures

tyj147454413-cmd:
Split from #543 for protected-core review.\n\nAdds a typed Codex stream error with HTTP status so callers can distinguish deterministic 4xx failures from retryable transport failures.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_openai_codex.py -q` (10 passed).


warren618:
Thanks @tyj147454413-cmd — merged. Classifying codex deterministic 4xx as non-retryable avoids the wasted retry.

PR #669, opened 2026-07-17
Enforce exposure limits on signed post-trade positions

### LINKED ISSUE #664: Gross exposure limits can be bypassed by opening a short
## Description

The live mandate gate adds buys and subtracts sells from one unsigned portfolio total. A sell from a flat position therefore looks like lower exposure even when it opens a short.

For example, an account with no position and a USD 500 exposure limit can submit a USD 600 sell. The current calculation is negative and the order is allowed.

## Expected behavior

The gate should calculate each symbol's signed post-trade position, then sum the absolute values. A sell should reduce exposure only while it closes an existing long. Any excess should count as short exposure.

## Proposed fix

Apply the order to its symbol-level signed position and use the resulting gross exposure for the mandate and leverage checks.

xkam7ar:
## Summary

- Calculate gross exposure from symbol-level signed positions.
- Count opening and oversized sells as short exposure.
- Add opening-short, mixed-book, and reducing-sell regressions.

## Why

Closes #664.

The current unsigned total lets a sell reduce exposure below zero and bypass a mandate cap.

## Changes

- Resolve each current position's symbol and direction.
- Apply the proposed order to that symbol.
- Sum absolute post-trade values and fail closed on ambiguous rows.

## Test Plan

- [x] New tests added
- [x] 354 adjacent live-safety tests passed
- [x] Ruff, compile, CI safety gates, diff, DCO, and secret checks passed
- [x] All broker transports were mocked

## Risk

This changes the live mandate and leverage calculation. It can block sells that would open or enlarge a short, while still allowing risk-reducing sells. No real broker, account, credential, or order was used.

## Out of scope

This does not change broker sizing or short availability.

## Rollback

Revert this one commit to restore the previous exposure formula.

## Checklist

- [x] The live safety area is covered by issue #664 and focused tests

### DIFF
--- a/agent/src/live/enforcement.py
+++ b/agent/src/live/enforcement.py
@@ -306,6 +306,34 @@ def _positions_market_value(positions: object) -> float | None:
     return total
 
 
+def _post_trade_gross_exposure(
+    positions: object,
+    *,
+    symbol: str,
+    signed_order_notional: float,
+) -> float | None:
+    """Return gross exposure after applying one signed order to its symbol.
+
+    Gross exposure is computed from signed, symbol-level positions. This lets a
+    sell reduce an existing long only until it reaches zero; any excess becomes
+    short exposure instead of subtracting from the whole portfolio.
+    """
+    rows = _coerce_position_rows(positions)
+    if rows is None:
+        return None
+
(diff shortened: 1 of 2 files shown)


warren618:

PR #674, opened 2026-07-17
feat(agent): add stream retry configuration

tyj147454413-cmd:
Split from #543 as the parent for stream-retry consumers.\n\nAdds `VT_STREAM_MAX_RETRIES` and its bounded runtime accessor without changing retry-loop behavior. Follow-up PRs can independently review agent and swarm retry behavior.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_stream_retry_config.py -q` (1 passed).

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -101,6 +101,15 @@ def _stream_retry_delay_s() -> float:
     return get_env_config().agent_tuning.vt_stream_retry_delay_s
 
 
+def _stream_max_retries() -> int:
+    """Return the configured retry count for transient stream failures."""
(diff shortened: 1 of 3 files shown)


warren618:
Thanks @tyj147454413-cmd, but I'm closing this. The `VT_STREAM_MAX_RETRIES` knob and its accessor are added, but no code path actually consumes the value — so it's inert config with no behavioral effect. To matter, stream-retry backoff would need to be wired into the real streaming call sites (`loop.py` / the provider stream loop). Happy to review a PR that does that wiring. 🙏

PR #675, opened 2026-07-17
fix(redaction): preserve safe content and scrub sensitive payloads

tyj147454413-cmd:
Split from #543 for independent security review.\n\nRedacts sensitive payload fields and internal paths without treating generic report/chat `content` as a secret by name.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_redaction.py agent/tests/test_redaction_shared.py -q` (53 passed).


warren618:
Thanks for this — dropping `content`/`env` from the redaction key set is defensible, and the audit tests do show no leak. Two defects to resolve before merge:

- The new `redact_text` function is dead code: it's added with tests but called nowhere, so its stated benefit (scrubbing shell output / logs) isn't actually delivered. Either wire it into the real sink it's meant for (the shell/log path) or drop it.
- `redact_internal_paths` only matches a root followed by a separator, so a bare root — or a root at end-of-string — is no longer redacted. That's an under-redaction regression, and for a security scrubber a false-negative path leak is the worse direction. Please also redact a root that appears at a string boundary (end, or followed by a non-path char).

One more thing: since removing `content`/`env` loosens a control that feeds the audit ledger, please call that out explicitly in the PR description for a maintainer sign-off, even though the current audit tests show no leak.

warren618:
Thanks. Two issues you target are real (the `content`/`env` keys over-redact legitimate payloads; the substring path match can clobber `/Users/alicetest` from `/Users/alice`). But as written the PR net-loosens a security module and ships dead code:

1. `redact_text()` has zero call sites (grep-verified) — the entire "scrub sensitive payloads" half never runs, and it injects a regex into a module that advertises itself regex-free (zero-ReDoS). Wire it into the surfaces it's meant to protect, or drop it.
2. Removing `env` from the sensitive set regresses the live audit ledger: a non-marker child like `DATABASE_URL=postgres://user:pass@…`, or a bare `env` string value, now lands unredacted — please keep `env` (or scope the exception precisely).
3. Your new path fix only matches `root + separator`, so a bare root at end-of-string is no longer redacted — `redact_internal_paths("/Users/alice")` returns the raw username-bearing path (the comment claims it handles this, but it doesn't; existing tests all use trailing-separator paths so it passes silently). Please fix the bare-root case and add a test.

The `content` removal and the `@cache` change are defensible in isolation. Happy to re-review once the dead code is gone, `env` is kept, and bare-root redaction is fixed. Thanks!

warren618:
CI has run here for the first time (the workflow was unapproved, so "no checks" meant no signal). It fails on the redaction contract this PR is meant to strengthen:

```
agent/tests/test_swarm_worker_report.py:67: in test_preview_tool_arguments_redacts_sensitive_values
E   Differing items:
E   {'content': '# very long confidential report'} != {'content': '[redacted]'}
```

The existing invariant is that `content` is redacted out of swarm worker tool previews. After this change it passes through verbatim, so the confidential body reaches the preview surface. That is the opposite direction from the PR title.

I can see the intent — "preserve safe content" — and there is a real usability cost to blanket redaction. But `content` is exactly the field most likely to carry the sensitive payload, so if it should now pass through, that needs an explicit rationale and a rule that distinguishes safe from unsafe content, not a removal of the redaction plus a test that still says otherwise.

This also lines up with the "redaction regressions" note from the earlier review, which still stands.

warren618:
Merged with your authorship — the two problems you identified were both real:
plain-text tool results were persisted and streamed unredacted, and path
redaction over-matched shared prefixes.

Three maintainer changes:

- `redact_text` is now actually called. It arrives through a single
  `redact_tool_result` choke point that both the trace writer and the SSE
  preview read, and the same gap in the swarm worker's preview path is wired to
  it as well.
- `content` and `env` couldn't simply leave the sensitive set. Tool-call
  arguments are persisted through the same helper, and `write_file` takes a raw
  `content` argument, so entire user documents and generated credentials would
  have landed in traces. Classification is now sink-aware: `content` is
  released only in the explicit result sink (which is where your fix was
  correct — the trace was blanking the very payload it exists to explain) and
  stays redacted in the fail-closed default used by arguments and the audit
  ledger. `env` is never released. Results additionally get their string leaves
  pattern-scrubbed, because shell output arrives as a JSON envelope where
  key-based classification can't reach secrets inside `stdout`.
- Exact-root redaction is restored — the loop only matched `root + separator`,
  so a string equal to a root stopped being redacted — while keeping your
  actual fix of not redacting unrelated paths that merely share a prefix.

There's coverage now proving a planted secret in a plain-text result and in a
`write_file` content argument reaches neither the trace files nor the streamed
preview.

PR #676, opened 2026-07-17
fix(session): enforce lifecycle and concurrent-run invariants

tyj147454413-cmd:
Split from #543 for protected-core review.\n\nAdds per-session active-loop ownership, cancellation terminal state, metrics persistence, and safe partial JSONL reads.\n\nVerification: `PYTHONPATH=agent ~/.venvs/vibe-trading/bin/python -m pytest agent/tests/test_session_service_lifecycle.py -q` (9 passed).


warren618:
Thanks for the thorough pass. This bundles several unrelated changes into protected session code and a couple don't hold up, so we can't take it as one PR:

- The two genuinely useful pieces are the `CANCELLED` status (`mark_cancelled` instead of folding a user cancel into `mark_failed`) and mirroring `result['metrics']` onto `attempt.metrics`. Please split those into a small standalone PR — happy to take them.
- The `send_message` concurrency guard checks `self._active_loops`, but that dict is only populated later inside the background task (after an awaited `build_registry`), well after `send_message` returns — so two rapid sends both pass the guard and it doesn't actually prevent concurrent runs. The test passes only because it plants `_active_loops` by hand. Also `SessionBusyError` is never caught by the API (`/sessions/{id}/messages` maps only `ValueError`→404), so it would surface as a 500, not the 409 the docstring describes; and rejecting a second send is a user-facing behavior change that needs a frontend story. To pursue it, the guard has to be set synchronously in `send_message`, wired to a 409 in `sessions_routes.py`, and coordinated with the UI.
- The `store.py` partial-line retry guards a case that can't happen: `append_message` writes the record and its trailing newline in a single `write()`, then `flush()`+`fsync()` before returning, so a reader never sees a line without its newline. It also adds a blocking `sleep` to the read path — please drop it.

Let's land the cancelled-status + metrics fixes first.

warren618:
Thanks — two genuinely correct fixes are bundled here that we'd like to take: the cancelled-status mapping (a user cancel currently gets mis-reported as a failure via `mark_failed`) and the metrics-persistence one-liner (`attempt.metrics = result.get("metrics")`; on main computed metrics are silently dropped because `attempt.metrics` is never assigned).

But the PR can't merge as-is:

- The headline concurrency guard doesn't enforce the invariant — `_active_loops` is registered only deep inside `_run_with_agent` after an awaited call, so two rapid sends both pass the top-of-`send_message` check (the classic double-submit) and both spawn loops. It must register synchronously in `send_message` before `create_task`.
- The docstring says callers return 409, but `SessionBusyError` is a `RuntimeError` and `sessions_routes` only catches `ValueError` → it actually 500s, and the channel runtime surfaces it as a raw error string to IM users.
- The `store.py` sleep-retry and the message-history rewrite are speculative changes to protected core (`src/session`) — please drop them.

Could you split this into (a) a minimal, clean PR with just the cancelled-status + metrics fixes, and (b) if you want to pursue the concurrency guard, a separate PR that registers the guard synchronously and wires `SessionBusyError`→409 + graceful channel handling, with a test that actually drives the double-submit race (the current tests pre-plant `_active_loops`, so they validate the stub, not the race). Thanks!

warren618:
Merged with your authorship. Every defect you identified was real and still
present; the fixes are placed differently.

- The concurrency guard needed to move earlier. `_active_loops` is only
  populated after the registry is built, long after both concurrent sends have
  persisted a message and created an attempt, so checking it there left the race
  open. The session is now claimed synchronously before anything is written, and
  released in exactly one terminal path.
- The 409 needed the API side as well: the error subclassed RuntimeError while
  the endpoint converts only ValueError, so a busy rejection surfaced as a 500.
  The IM channel path shares send_message and now answers in plain language
  rather than leaking an exception name.
- attempt.cancelled needed a consumer. The backend event alone would have been
  correct only until reload, because the SSE hook subscribed to completed and
  failed only and history hydration reconstructed every non-failed status as
  done. Both now understand it.
- The partial-JSONL retry was dropped: it slept 10 ms and then returned the
  older text anyway, so the record was skipped regardless, and the store already
  writes with flush+fsync and validates record shapes on read.

Your other three splits: #662 and #675 are merged; #661 is closed, since the
frontend it targets has been overhauled twice and the tool-call fallback would
regress the current call-id matching. Details are on that PR.

PR #687, opened 2026-07-17
Derive China-futures leverage per contract

### LINKED ISSUE #679: China futures sizing depends on symbol order
## Description

All China futures contracts currently inherit leverage from the first configured symbol. Products with different margin rates therefore receive different sizes when the code order changes.

## Expected behavior

Each contract should derive leverage from its own margin rate. An explicit global margin override should still apply to every contract.

## Proposed fix

Add a symbol-aware leverage hook and delegate it through composite portfolios.

xkam7ar:
## Summary

- Derive China futures leverage from each contract's margin rate.
- Delegate symbol-aware leverage through composite portfolios.
- Preserve the explicit margin override.

## Why

Closes #679.

The first configured symbol currently controls sizing for every China futures contract.

## Changes

- Add a symbol-aware leverage hook to the base engine.
- Implement contract margin lookup in the China futures engine.
- Add order-invariance and composite regressions.

## Test Plan

- [x] New tests added
- [x] 100 relevant engine tests passed
- [x] Source Ruff, diff, DCO, and secret checks passed
- [x] Tests used synthetic prices only

## Risk

Multi-product China futures sizes can change to match each product's margin. The explicit override remains unchanged. No broker or exchange was contacted.

## Out of scope

This does not update the margin-rate table.

## Rollback

Revert this one commit to restore first-symbol leverage.

## Checklist

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -376,6 +376,11 @@ class BaseEngine(ABC):
         """Convert target notional exposure to number of units/contracts."""
         return target_notional / price
 
+    def _leverage_for_symbol(self, symbol: str) -> float:
+        """Return leverage used to size and margin one symbol."""
+        del symbol
+        return self.default_leverage
+
     # ── Main entry ──
 
     def run_backtest(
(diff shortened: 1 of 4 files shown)


warren618:

PR #688, opened 2026-07-17
Preserve requested crypto symbols after provider normalization

xkam7ar:
## Summary

- Track provider-normalized crypto aliases.
- Restore original request keys before merging data.
- Add a slash-form OKX regression.

## Why

Closes #680.

Auto routing can fetch `BTC-USDT` successfully but lose it because downstream code requests `BTC/USDT`.

## Changes

- Pair normalized codes with their original symbols.
- Restore keys for both primary and fallback results.
- Keep provider-facing formats unchanged.

## Test Plan

- [x] New tests added
- [x] 89 relevant routing tests passed
- [x] Ruff, diff, DCO, and secret checks passed
- [x] No provider was contacted

## Risk

Only returned map keys change, back to the caller's requested form.

## Out of scope

This does not change symbol detection or provider symbol formats.

## Rollback

Revert this one commit to restore normalized result keys.

## Checklist

- [x] No hardcoded credentials or local paths
- [x] Code follows `CONTRIBUTING.md`
- [x] No user-facing documentation change is needed

PR #689, opened 2026-07-17
Complete partial market-data results through fallback

### LINKED ISSUE #681: Partial loader results silently shrink the backtest universe
## Description

Any nonempty primary loader result prevents fallback, even when requested symbols are missing. The backtest can then run on a smaller universe without reporting it.

## Expected behavior

Fallback should request only missing symbols. If every source still leaves incomplete coverage, the run should stop with a clear error.

## Proposed fix

Merge successful primary data with per-symbol fallback results and raise `NoAvailableSourceError` for unresolved symbols.

xkam7ar:
## Summary

- Retry fallback providers only for missing symbols.
- Merge successful primary and fallback data.
- Stop explicitly when requested coverage remains incomplete.

## Why

Closes #681.

A nonempty partial response currently prevents fallback and silently shrinks the backtest universe.

## Changes

- Restore provider aliases to requested symbols.
- Track missing symbols after every source.
- Raise `NoAvailableSourceError` when any remain.
- Add auto, explicit, and unresolved-coverage regressions.

## Test Plan

- [x] New tests added
- [x] 95 relevant runner and loader tests passed
- [x] All data providers were fakes

## Risk

Runs that previously continued with an incomplete universe now fail with a clear error. This is intentional fail-closed behavior.

## Out of scope

This does not change fallback ordering.

## Rollback

Revert this one commit to restore all-or-nothing fallback.

## Checklist

### DIFF
--- a/agent/backtest/runner.py
+++ b/agent/backtest/runner.py
@@ -584,6 +584,16 @@ def _normalize_codes(codes: List[str], source: str) -> List[str]:
     return codes
 
 
+def _map_result_to_requested(
+    result: dict,
+    requested_codes: List[str],
+    provider_codes: List[str],
+) -> dict:
+    """Restore provider-normalized result keys to this fetch's request keys."""
+    aliases = dict(zip(provider_codes, requested_codes))
+    return {aliases.get(code, code): frame for code, frame in result.items()}
+
+
 def _columns_required_from_factor_spec(spec: Any) -> list[str]:
     """Extract ``columns_required`` from supported factor spec shapes.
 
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for tackling this — the auto-source fallback path is correct and does what it should. Two things to change before merge:

- Explicit-source path: the fallback fills the returned data_map, but the backtest still runs on the primary loader's partial set, so the headline goal isn't actually achieved for an explicit-source run. Please make the merged data_map reach the engine (wrap it the way auto mode does, or reassign the loader) so the fix holds end-to-end.
- The new blanket fail-closed raise fires uncaught in run(), so a 50-symbol basket with one delisted name now hard-crashes the whole backtest where it previously ran on the other 49. Either scope it (skip/annotate the missing symbols instead of aborting) or gate it behind an explicit option.
- Also please rebase on top of #688 — the two conflict on _fetch_auto.

With those sorted this is in good shape.

warren618:
Thanks for the update. A couple of things have moved since the last pass:

- #688 has now merged and reworked `_fetch_auto` (provider-normalized keys are restored via `_restore_original_codes`), so this branch conflicts on `runner.py`. Please rebase onto current `main` and fold the per-symbol fallback into that structure instead of the older `_map_result_to_requested` path.
- The blanket `raise NoAvailableSourceError` when symbols are still missing after fallback is still there, and it's uncaught in `run()`. As-is, a 50-symbol basket with one delisted name hard-crashes the whole backtest where it used to run on the other 49 — a regression against the goal of this PR. Please either skip/annotate the unresolved symbols and surface them in the run metadata, or gate the strict-abort behind an explicit opt-in flag; `test_fetch_stops_when_fallbacks_leave_symbols_missing` will need to move with that decision.
- Good news on the earlier explicit-source concern: `main` now rebuilds the loader from the validated snapshot before the engine runs, so once you rebase, the merged `data_map` reaches the engine on the explicit path automatically.

The per-symbol fallback itself is the right approach and the tests are solid; with the rebase and the abort behavior softened, this should be in good shape.

warren618:
Thanks — this fixes a real, still-present bug (#681): on main both the explicit-source and auto paths only fall back when the result is *completely* empty, so any partial response silently proceeds on a shrunk universe. Your fix (fetch only the missing symbols via the fallback chain over the same window, then fail closed with `NoAvailableSourceError` if any remain) is correct — no look-ahead, no date-frame mixing.

Before merge:

1. Rebase onto current main (it currently conflicts).
2. On rebase, drop the new `_map_result_to_requested` helper and reuse main's existing `_restore_original_codes` (added after you branched) — they're logic-identical.
3. Worth a release note: runs that previously proceeded silently on a partial universe will now hard-fail — intended fail-closed behavior per #681.

Once rebased and deduped, we'll merge. Thanks!

warren618:
Rebased onto current main and merged (as be77c94). I deduped as suggested — it now reuses the existing `_restore_original_codes` helper instead of the parallel `_map_result_to_requested`, applies the same missing-symbol completion to both `_fetch_auto` and the explicit-source path, and keeps your new tests. 263 tests across the fetch/fallback paths pass. Thanks @xkam7ar! (Fixes #681.)

PR #691, opened 2026-07-17
Verify CCXT coverage after bounded pagination

### LINKED ISSUE #683: CCXT backtests can silently end before the requested date
## Description

Bounded CCXT pagination can exhaust 200 full pages before reaching the requested end and still return success. A long intraday backtest then uses a shorter date range.

## Expected behavior

A full-cap result should be accepted only when it reaches the requested end within one interval.

## Proposed fix

Keep the page cap and raise a clear incomplete-history error when end coverage is missing.

xkam7ar:
## Summary

- Keep the existing 200-page safety cap.
- Reject a full-cap result that does not reach the requested end.
- Add a deterministic incomplete-range regression.

## Why

Closes #683.

CCXT pagination can hit its cap and return a truncated date range as success.

## Changes

- Track whether pagination stopped because of the cap.
- Compare the final bar with the requested end using interval tolerance.
- Raise a clear incomplete-history error when coverage is missing.

## Test Plan

- [x] New tests added
- [x] 40 relevant CCXT loader tests passed
- [x] The exchange client was a fake

## Risk

Long requests that exceed bounded coverage now fail so fallback or the user can respond. The cap and normal short responses are unchanged.

## Out of scope

This does not increase the pagination or wall-clock budget.

## Rollback

Revert this one commit to restore capped partial results.

## Checklist

### DIFF
--- a/agent/backtest/loaders/ccxt_loader.py
+++ b/agent/backtest/loaders/ccxt_loader.py
@@ -32,6 +32,16 @@ _INTERVAL_MAP = {
     "1H": "1h", "4H": "4h", "1D": "1d",
 }
 
+_TIMEFRAME_DELTA = {
+    "1m": pd.Timedelta(minutes=1),
+    "5m": pd.Timedelta(minutes=5),
+    "15m": pd.Timedelta(minutes=15),
+    "30m": pd.Timedelta(minutes=30),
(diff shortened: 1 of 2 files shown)


warren618:

PR #692, opened 2026-07-17
Reject mootdx history that does not reach the requested start

### LINKED ISSUE #684: mootdx returns incomplete history after reaching its page limit
## Description

When pagination cannot reach the requested start, mootdx logs a warning and returns a shorter dataset. Downstream research treats that partial range as complete.

## Expected behavior

Exhausting the page limit before the requested start should be a failure visible to fallback.

## Proposed fix

Raise an incomplete-history error instead of returning the capped result.

xkam7ar:
## Summary

- Keep the existing mootdx page cap.
- Raise an incomplete-history error when the cap cannot reach the requested start.
- Add a small fake-client regression.

## Why

Closes #684.

The loader currently warns but returns a shorter date range as success.

## Changes

- Convert page-cap exhaustion into a typed failure.
- Preserve successful behavior when a page reaches the start boundary.

## Test Plan

- [x] New tests added
- [x] 96 relevant mootdx loader tests passed
- [x] No TCP or provider access occurred

## Risk

Requests beyond the bounded history now fail visibly instead of running on partial data.

## Out of scope

This does not change page size or symbol support.

## Rollback

Revert this one commit to restore warning-only behavior.

## Checklist

### DIFF
--- a/agent/backtest/loaders/mootdx_loader.py
+++ b/agent/backtest/loaders/mootdx_loader.py
@@ -191,9 +191,10 @@ class DataLoader:
             if first_dt <= start_ts:
                 break
         else:
-            logger.warning(
-                "mootdx: %s %s pagination hit cap (%d pages) without reaching %s",
-                symbol, freq, _MAX_PAGES, start_date,
+            raise ValueError(
+                "incomplete mootdx history: "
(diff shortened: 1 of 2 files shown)


warren618:

PR #693, opened 2026-07-17
feat: add Korea equity (KRX: KOSPI/KOSDAQ) market support

JungHoonGhae:
## Summary

- Add first-class **Korean equity (KRX: KOSPI/KOSDAQ)** support across the research stack: data ingestion → dedicated backtest engine → `equity_kr` factor universe.
- New `KoreaEquityEngine` modeling KRX cash-equity rules: **same-day round trips allowed** (no T+1 — unlike China A-share/India delivery), the **±30% daily price limit** (blocks buys at limit-up / sells at limit-down), bilateral brokerage + **sell-only securities transaction tax** (config-driven rates), 1-share lots, and no shorting by default (`allow_short` opt-in).
- New **pykrx loader**: free, no-auth KRX EOD OHLCV behind an optional `[krx]` extra; `005930.KS` (KOSPI) / `247540.KQ` (KOSDAQ) symbol convention. The `kr_equity` fallback chain degrades to yahoo/yfinance (which carry `.KS`/`.KQ` natively) when pykrx isn't installed.

## Why

Korean symbols previously had **zero** research support: a `005930.KS` request matched no market pattern and fell through to China infrastructure (`a_share` default → tushare/`ChinaAEngine`), producing no data and wrong market rules. Korea is a top-15 world equity market with an active retail/quant community; this wires the same path India got in #305 so existing strategies, factors, and metrics work on KRX bars.

## Changes

- **Data ingestion:** new `backtest/loaders/pykrx_loader.py` (pykrx → canonical OHLCV via a pure `_normalize`, batch-safe per-symbol failure handling, shared loader cache); `kr_equity` fallback chain `["pykrx", "yahoo", "yfinance", "local"]` in `registry.py`; `pykrx` added to `VALID_SOURCES`; `.KS`/`.KQ` accepted by `yahoo_loader`/`yfinance_loader`.
- **Routing:** `.KS`/`.KQ` regex in `_market_hooks.py` (`kr_equity` market type), `market_data.py` source patterns, and `correlation.infer_market`; `runner.py` source map + single-market routing (precedes source-based routing for the same fall-through reason as India); `composite.py` cross-market sub-engine.
- **Engine:** new `backtest/engines/korea_equity.py` (see Summary for the rule set; module docstring documents the cost-stack defaults and their revision caveat).
- **Factors:** `equity_kr` added to the `Universe` literal and `Market` enum (vwap uses the typical-price branch — pykrx bars carry raw price/volume, no Tushare scaling); tagged all 101 Alpha101 and 154 QLib158 factors for `equity_kr`. GTJA191 left China-only (Tushare-scale math). Registry loads 461 factors, 0 failures; 255 listed for `equity_kr`.
- **Docs:** `SKILL.md` updated (9 engines, 22 sources, Korea symbol convention + engine rules); `pyproject.toml` gains the optional `[krx]` extra.
- Cost-stack rates and the price-limit band are config-driven defaults (documented), not magic constants.

Deliberately **not** included: live/paper broker wiring (no Korean broker connector exists in this codebase — unlike India's Shoonya/Dhan there is nothing to bridge), and no mandate/`AssetClass` changes.

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest`)
- [x] New tests added — `test_korea_equity_engine.py` (rules incl. the same-day-sell contrast with India/China), `test_korea_backtest_smoke.py` (in-memory E2E + Korea costs strictly poorer than the zero-commission US engine on identical bars), `test_pykrx_loader.py` (symbol mapping + `_normalize` on synthetic pykrx-shaped frames — no network, no pykrx needed), `tests/factors/test_korea_universe.py` (255 listed, GTJA191 excluded, representative alphas compute on a KRX-shape panel), plus Korea cases in `test_market_detection.py` and the `kr_equity`/`pykrx` updates to `test_registry.py`
- [x] Tested manually — targeted suites pass locally (95 Korea-specific tests; 1104 passed across `test_india_*`, `tests/factors/`, `test_registry.py`, `test_base_engine.py`, `test_engine_robustness.py`); real-data end-to-end backtest against live KRX via pykrx (005930.KS + 247540.KQ, 6 months, MA-crossover signal → 18 trades, full metrics + benchmark comparison); `tools/ci_grep_gates.sh` all gates pass; `ruff check` clean on all new files

## Checklist

- [x] No hardcoded values (API keys, file paths, magic numbers) — Korea cost/band defaults are config-overridable
- [x] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines — `ruff check` passes on the contribution footprint; public signatures type-annotated; Google-style docstrings; DCO sign-off on the commit
- [x] Documentation updated (if user-facing change) — `SKILL.md`

---

**Reviewer notes**
- The first Test Plan box is left unchecked for the same reason as #305: I verified targeted suites + ruff + grep gates locally but not the full suite (sandbox lacks some optional deps, e.g. `fastmcp`). Please let CI confirm.
- Korean securities transaction tax stepped down over 2023–2025 (KOSPI/KOSDAQ effectively 0.15% on sell as of 2025) and brokerage varies by firm — defaults are documented in the module docstring and exposed as `kr_brokerage` / `kr_tax_sell` config knobs; verify against a current schedule before relying on absolute figures.
- pykrx scrapes KRX public endpoints and rate-limits by IP; the loader spaces calls (0.5s) with its own throttle because pykrx owns its HTTP session and can't route through `_http.throttled_get`. Recent pykrx versions print a `KRX 로그인` warning for *some* endpoints (short-sell data); daily OHLCV — the only thing this loader uses — needs no login (verified live).
- Bare 6-digit codes stay routed to A-share (they're ambiguous with Shenzhen codes); Korea always requires the `.KS`/`.KQ` suffix, mirroring how HK/A-share disambiguation already works.


warren618:
Thanks, @JungHoonGhae — the Korea (KRX) equity engine is solid and follows the India engine pattern faithfully, which makes it easy to review. One blocker plus a heads-up:

- CI never ran on this — please let it run and confirm the suite is green before we merge.
- Heads-up (not introduced by your PR): it inherits a latent issue from the India engine where the circuit-band logic depends on a `pre_close` field that no loader populates, so those bands are effectively dormant on real data. If you'd like the KRX circuit bands to actually fire, a loader would need to populate `pre_close` — but we can track that separately and it shouldn't hold up this PR.

Get CI green and this is in good shape.

warren618:
Thanks — this is a solid, idiomatic addition. The KRX engine correctly models the real market rules (same-day round trips allowed unlike China T+1, ±30% price band, sell-side STT + bilateral brokerage, short blocked by default), pykrx is properly gated as an optional extra with a fallback chain, and the 255 `equity_kr` factor-universe tags follow the existing per-market gating pattern (same as India's `equity_in`). 116 tests pass.

Two things before merge:

1. Rebase onto current main — the branch is behind and the `equity_kr` universe edits now conflict with zoo files main re-touched since your base; please resolve (mechanical: re-add `'equity_kr'` to the universe list).
2. Minor: confirm the sell-side STT rate tracks the current-year KRX schedule (your in-code docstring already flags this).

Once rebased and green, we'll merge. Thanks for the careful market-rule work!

JungHoonGhae:
Rebased onto current `main` (was 293 commits behind) and force-pushed. Conflicts were in `equity_kr` universe/source-list touches to `SKILL.md`, `pyproject.toml`, `registry.py`, `market_data.py`, `yahoo_loader.py` — resolved by keeping both sides' additions (this PR's Korea entries alongside main's newer mt5/binance/futures-forex entries). All 119 targeted tests (Korea engine/loader/factors + market detection + registry) pass on the rebased branch.

warren618 (changes requested):
Thanks for rebasing. I reviewed exact head `669198166a52e4a8ec0d3f9a4e8f59f9bdc6a9d0` against current `main`. The merge is clean, and the six targeted Korea/registry suites pass (`119 passed`).

Two blockers remain:

1. The full non-E2E suite is not green: `1 failed, 6386 passed, 12 skipped`. `agent/tests/test_yahoo_loader.py::TestLoaderMetadata::test_name_and_markets` still expects the old market set. Please add `kr_equity` to that assertion.

2. `kr_tax_sell` still defaults to the 2025 rate (`0.0015`). For 2026, KOSPI charges 0.05% securities transaction tax plus 0.15% special rural-development tax, while KOSDAQ charges 0.20%, so the effective sell-side default for both boards is 0.20%. Please update the default to `0.0020`, update the module documentation, and add an explicit numeric regression test.

Sources:
- https://www.law.go.kr/LSW/lsRvsDocListP.do?chrClsCd=010202&lsId=008301&lsRvsGubun=all
- https://kind.krx.co.kr/external/2026/03/18/002312/20260318010780/11011.htm

The previously noted `pre_close` limitation can remain a separate issue. Once these two changes are pushed, we can approve and run CI on the corrected head.

warren618:
CI has now run on this branch for the first time — the workflow had been sitting unapproved, so the earlier "no checks" state was not a pass, it was no signal at all. I have approved the queue, so runs report normally from here.

It reproduces the first review point directly:

```
agent/tests/test_yahoo_loader.py:306: in test_name_and_markets
    assert loader.markets == {"us_equity", "hk_equity", "india_equity"}
E   AssertionError: Extra items in the left set
```

The loader advertises the new Korean market but `TestLoaderMetadata::test_name_and_markets` still asserts the old three-market set, so the registry expectation has to move with it.

The second point from the earlier review — the stale 2026 KRX sell-tax default — still stands and is not covered by CI.

No rush, just flagging that the red check is real rather than infrastructure noise.

warren618:
Merged with your authorship — thank you, the engine and loader follow the house
pattern closely and that made this straightforward to finish.

Maintainer changes worth knowing:

- The ±30% guard could not fire. It needed pct_chg or pre_close, but pykrx
  returns OHLCV only (by design) and Yahoo likewise, so it always received None
  and permitted orders at locked limits. Feeding it the current bar's close
  would have been lookahead instead — the engine fills at that same bar's open.
  The band now comes from a strictly historical base price and is compared
  against the prospective fill price.
- KRX tick quantization was missing, so limit prices were treated as a
  continuous percentage and fills never snapped to the grid. Note that KRX
  unified the KOSPI and KOSDAQ tables in its 2023-01-25 reform, so this is one
  table; several English references still print the older split bands.
- The loader accepted any interval and returned daily bars under the requested
  key, so a 1m or 4h backtest silently ran on daily data. Non-daily intervals
  are now refused so the runner falls through.
- The securities transaction tax default was the 2025 rate; from 2026-01-01 the
  aggregate sell-side rate is 0.20%.
- Costs were keyed to open/close rather than trade side. Since KRX covered-short
  and uptick rules are not representable on daily bars, the engine now refuses
  allow_short outright rather than silently mis-modelling shorts.
- Run cards could claim pykrx when Yahoo actually served the data.
- equity_kr reached factor metadata but every surface rejected or hid it; REST,
  the CLI, the get_market_data enum and the Web Alpha Zoo now accept it, with
  labels in all five locales.

One request for next time: the 255 one-line factor-zoo edits were mechanically
uniform and I kept them as written, but that is the last such sweep this repo
should take. Zoo-wide market capability belongs in a registry-level default with
per-factor exceptions, not in 255 modules per new market.

PR #701, opened 2026-07-17
fix(providers): force temperature=1 for the whole Kimi K-series

sambazhu:
kimi-k3 rejects any temperature other than 1 ("invalid temperature: only 1 is allowed for this model"), same as kimi-k2, but the forcing guard only matched the kimi-k2 prefix — so a default LANGCHAIN_TEMPERATURE=0.0 config fails on every request with kimi-k3.

Replace the hardcoded prefix tuple with a K-family regex (kimi-k<digits> + kimi-for-coding) so kimi-k3 and future K-series reasoning models are covered without another whack-a-mole patch.

Verified against the Kimi for Coding endpoint: both kimi-k3 and its k3 alias return the temperature error; moonshot-v1-* models are unaffected. Note: bare aliases like "k3" are intentionally out of scope for the name-based guard.

## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/src/providers/llm.py
+++ b/agent/src/providers/llm.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 from collections.abc import Sequence
 from importlib import import_module
 from importlib.metadata import PackageNotFoundError, version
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @sambazhu — nice future-proofing. Generalizing the `temperature=1` forcing from `kimi-k2` to the whole K-series (`kimi-k2`/`k3`/…/`for-coding`) via the `kimi-(k\d+|for-coding)` regex is the right call, and non-K Moonshot models (e.g. `moonshot-v1-8k`) correctly keep their configured temperature. Merged 🙏

PR #706, opened 2026-07-17
Add secure ICICI paper-order approval workflow

aryaveer20000-tech:
## Summary

Adds a secure, server-owned approval workflow for ICICI Breeze paper orders.

## Included

- Immutable, one-time paper-order proposals
- Explicit Approve and Reject actions
- Second confirmation dialog before simulated order placement
- Proposal expiry and duplicate-use protection
- Agent timeline integration through SSE
- Paper-only execution guard
- Risk checks and manual-confirmation recording
- CLI compatibility for string and dictionary order statuses

## Safety validation

- No live ICICI broker-write function is called
- Paper order successfully created and cancelled locally
- `broker_write_called: false`
- Proposal rejection and duplicate-use protection verified
- Backend tests passed
- Frontend: 27 test files and 243 tests passed
- Production frontend build passed
- Git working tree clean


warren618:
Thanks for the detailed work here — the ICICI Breeze connector itself looks well-structured: paper-locked, first-line hard-refuse on non-paper, local ledger with no broker-write, and clean risk controls.

Before we can merge, a few things need to change:

1. **Blocker:** adding `confirmed` to `service.place_order` and threading it into the shared `place_kwargs` breaks every other connector. Those kwargs are forwarded verbatim to `connector_module.place_order(...)` for both paper and (via `execute_live_order`) live orders, and no connector except ICICI accepts `confirmed`, so tiger/alpaca/okx/binance/futu/longbridge/dhan/shoonya/trading212 will raise `TypeError` on order placement. Please scope `confirmed` to the ICICI path only (e.g. via the connector's own config/overrides rather than the generic `place_kwargs`).
2. Please move the paper-order module out of `sessions_routes.py` into its own routes module, and drop the `>>> … >>>` sentinel blocks and the aliased mid-file imports — it should read like the rest of the codebase.
3. Add backend tests for the proposal create/approve/reject/expire endpoints, including the `paper`/`broker_write_called` safety invariants you assert in the approve route.

Happy to re-review once those are addressed.

warren618:
Thanks — the ICICI connector itself is red-line clean: it correctly follows the paper-capped pattern for a broker with no runtime paper/live discriminator (no `*-live-trade` profile; `place_order`/`cancel_order` hard-refuse any non-paper config at the first line; paper is a local JSON ledger with zero broker writes; not MCP-exposed). Good work mirroring the Dhan/Shoonya precedent.

One safety-critical blocker: the change to `service.place_order` injects `confirmed` into the shared `place_kwargs` dict forwarded to *every* connector. The other 9 connectors have keyword-only signatures with no `confirmed`/`**kwargs`, so this raises `TypeError` for all of them — breaking paper placement for every `*-paper-trade` profile and live placement for the runtime-discriminator brokers (it fails closed after the mandate returns ALLOW, so no unintended order, but the path is broken). The suite misses it because the gate test uses a fake connector that absorbs `**kwargs`.

Required changes:

1. Fix the `confirmed` plumbing so it doesn't hit the other 9 connectors — route it through `overrides`→`build_config` for ICICI only, or gate it to connectors that accept it; don't inject unconditionally into the shared dict.
2. Add a regression test that `service.place_order` binds cleanly for every registered order profile (this one test would have caught it).
3. Add backend unit tests for the ICICI connector (paper ledger buy/sell/cancel, the risk-deny paths, and the live-readonly hard-refuse).
4. Add `"icici_breeze"` to `test_no_discriminator_brokers_expose_no_live_trade_profile`.
5. Style: extract the 339-line proposal module out of `sessions_routes.py` into its own file, and drop the aliased-stdlib imports / comment fences.

The connector is a faithful paper-capped clone; it's the shared-code regression + missing tests that block. Happy to re-review once those are in. Thanks!

warren618:
Thank you for this — the safety architecture is the part that matters most and
you got it right: local paper simulation only, a live profile that is
read-only, and no Breeze order API call anywhere in the diff. That matches how
Dhan and Shoonya are capped, and it is why this is a changes-requested rather
than a decline.

Four things need to change before it can land.

1. `confirmed` cannot go into the shared place_kwargs. That dict is splatted
   straight into every connector's place_order, and none of the ten existing
   connectors accepts the parameter or **kwargs. As written, this makes every
   existing paper order raise TypeError and every mandate-approved live order
   fail closed inside the gate with "unexpected keyword argument". Pass it only
   for the ICICI connector, or give ICICI its own service entry point. Please
   add a test asserting the flag never reaches another connector.

2. The approval workflow is unreachable from the product. `createPaperOrderProposal`
   is declared in the frontend but nothing calls it, and the agent tool invokes
   service.place_order directly, so a `confirmation_required` result is never
   converted into a stored proposal and no SSE event is emitted — the card can
   only appear if someone POSTs the private endpoint by hand. This needs a real
   bridge, and the proposal state needs to live somewhere the trading layer can
   reach rather than in sessions_routes.

3. Proposal state is a module-level dict, so a page refresh, a server restart,
   or a second worker loses a card that is still valid, and terminal
   transitions emit no event so other tabs stay stale. `agent/src/scheduled_research/store.py`
   is the house pattern for crash-safe atomic persistence under the runtime
   root. Relatedly, the expiry handling is inconsistent: pending proposals are
   swept to `expired` before the approval checks run, so the explicit expiry
   branch is unreachable and approval returns 409 where 410 was intended, while
   the card keeps its initial state and still shows active buttons.

4. Please declare the Breeze SDK as an optional extra in pyproject.toml
   (`icici = ["breeze-connect>=..."]`). There is no dependency declaration at
   the moment, and both profiles need it — the paper one included, since it
   prices simulated fills off live Breeze quotes.

Two more notes. The branch needs a rebase: `agent/src/trading/profiles.py` and
all five locale JSONs have moved. And a design question worth settling before
you invest in item 2 — this project's stance is bounded autonomy, where the
agent trades inside a mandate without per-order confirmation, so a per-order
approval surface only makes sense framed explicitly as a paper rehearsal
feature. Please say in the PR description which it is meant to be.

The unrelated CLI fix for a string order status is a good catch; feel free to
send that as its own small PR and I will merge it quickly.

warren618:
Thanks for the ICICI contribution. The current patch adds a shared `confirmed` keyword that breaks existing connector implementations, stores approval proposals only in process memory, and does not connect the normal Agent tool path to proposal creation. Combined with the stale pre-rewrite history, this is not safe to merge, so we are closing the PR. A fresh ICICI read-only connector and local paper simulator can be proposed separately from current `main`, without changing every connector’s call contract. Please delete the old source branch.

PR #709, opened 2026-07-18
fix(factors): reject non-positive n_groups in factor analysis

santhreal:
compute_group_equity passed n_groups=-1 into pandas.qcut and raised IndexError. Require n_groups >= 1 and return a JSON error from the factor_analysis tool instead of crashing.


warren618:
Thanks @santhreal — merged to `main`. Rejecting non-positive `n_groups` before writing artifacts is the right guard, and the added test covers it. 🙏

PR #710, opened 2026-07-18
fix(alpha): reject inverted period ranges in _parse_period

santhreal:
_parse_period accepted inverted ranges like 2024-2020 and returned a reversed (start, end) pair. Reject start > end with ValueError, matching validate_date_range.


warren618:
Merged to `main` — thanks @santhreal. Inverted period ranges now fail fast in `_parse_period` instead of silently producing an empty window. 🙏

PR #711, opened 2026-07-18
fix(backtest): handle unnamed DatetimeIndex in correlation matrix

santhreal:
_rolling_correlation_matrix assumed a trade_date column when the index was not named trade_date, so standard OHLCV frames with an unnamed DatetimeIndex raised KeyError. Accept DatetimeIndex close series directly.


warren618:
Merged to `main` — thanks @santhreal. An unnamed `DatetimeIndex` in the correlation matrix is now handled cleanly. 🙏

PR #712, opened 2026-07-18
fix(pattern): reject non-positive detection window

santhreal:
pattern detectors passed window<=0 into peak scans and polyfit, causing IndexError, TypeError, or every-bar peak/valley garbage. Require window >= 1 and return a JSON error from run_pattern.


warren618:
Merged to `main` — thanks @santhreal. Non-positive detection windows are now rejected up front. 🙏

PR #713, opened 2026-07-18
fix(validation): accept equity.csv nav/value column aliases

santhreal:
validation CLI hardcoded df['equity'], so artifacts that store nav or value raised KeyError. Accept equity/nav/value (or a single numeric column).


warren618:
Merged to `main` — thanks @santhreal. `equity.csv` nav/value column aliases are now accepted, which makes validation more forgiving of real-world exports. 🙏

PR #714, opened 2026-07-18
fix(journal): reject empty A-share codes instead of 000000.SZ

santhreal:
Blank securities codes were zfill(6)-mapped to 000000.SZ. Reject empty codes in _qualify_a_share and skip blank-code rows when parsing Tonghuashun/Eastmoney exports.


warren618:
Merged to `main` — thanks @santhreal. Empty A-share codes are now rejected instead of being coerced to `000000.SZ`. Good catch. 🙏

PR #715, opened 2026-07-18
fix(mcp): align factor_analysis wrapper with the registered tool contract

### LINKED ISSUE #635: MCP `factor_analysis` always fails with `KeyError: 'factor_csv'` — wrapper and registered tool have mismatched contracts
## Summary

The MCP server's `factor_analysis` tool cannot succeed for any input: the
`@mcp.tool` wrapper in `agent/mcp_server.py` forwards `codes`/`factor_name`
arguments, but the only registered implementation
(`agent/src/tools/factor_analysis_tool.py`) requires pre-built CSV artifact
paths (`factor_csv`, `return_csv`, `output_dir`). Every MCP call therefore
dies on `KeyError: 'factor_csv'` before any analysis runs.

## Reproduction

Connect any MCP client to `vibe-trading-mcp` (tested over Streamable HTTP)
and call:

```json
{
  "tool": "factor_analysis",
  "arguments": {
    "codes": ["SBIN.NS", "HDFCBANK.NS", "ICICIBANK.NS"],
    "factor_name": "pb",
    "start_date": "2025-07-16",
    "end_date": "2026-07-16",
    "source": "yfinance"
  }
}
```

Result (same for any market/symbols, e.g. US tickers, and for `source:
"auto"` / `"tushare"`):

```text
factor_analysis error: 'factor_csv'
```

## Root cause

`agent/mcp_server.py` — the wrapper accepts a codes-based signature and
forwards exactly those keys:

```python
@mcp.tool
def factor_analysis(
    codes: list[str],
    factor_name: str,
    start_date: str,
    end_date: str,
    source: str = "auto",
    top_n: int = 10,
    bottom_n: int = 10,
) -> str:
    ...
    return registry.execute(
        "factor_analysis",
        {
            "codes": codes,
            "factor_name": factor_name,
            "start_date": start_date,
            "end_date": end_date,
            "source": source,
            "top_n": top_n,
            "bottom_n": bottom_n,
        },
    )
```

`agent/src/tools/factor_analysis_tool.py` — the registered tool is
artifact-based and reads different keys:

```python
"required": ["factor_csv", "return_csv", "output_dir"],
...
factor_csv=kwargs["factor_csv"],   # <- KeyError: 'factor_csv'
```

`FactorAnalysisTool` performs no data fetching (it only `pd.read_csv`s the
two matrices), and I could not find any other code path that builds a factor
matrix from `codes` + `factor_name` (the wrapper's docstring mentions
`daily_basic` fields like `pe_ttm`/`pb`, which suggests an intended
tushare-backed builder that isn't wired up).

Verified at commit `86f6012` and, by reading the current sources, still
present on `main` as of 2026-07-16.

## Environment

- vibe-trading-ai installed from git @ `86f6012` (pip, Python 3.12, macOS arm64)
- `vibe-trading-mcp --transport http` (Streamable HTTP), stock configuration
- No broker connections, no mandate; pure research usage

## Expected behavior (either would resolve it)

1. The MCP wrapper matches the registered tool's contract — i.e. it accepts
   `factor_csv`/`return_csv`/`output_dir` so agents can feed artifacts they
   built with `get_market_data` + `write_file` (the flow the
   `factor-research` skill documents); or
2. A codes→factor-matrix builder is added behind the current wrapper
   signature (e.g. tushare `daily_basic` for CN, with an honest error for
   markets lacking a fundamentals history source).

Happy to provide more detail or test a fix. Thanks for the project — the
MCP server design (research-only surface, no order tools) is excellent.

Robin1987China:
## Summary

The MCP `factor_analysis` tool could never succeed: the wrapper forwarded `codes`/`factor_name`/`start_date`/`end_date`/`source`/`top_n`/`bottom_n`, but the only registered implementation (`FactorAnalysisTool`) requires pre-built CSV artifact paths (`factor_csv`, `return_csv`, `output_dir`). Every MCP call died on `KeyError: 'factor_csv'` before any analysis ran.

This PR mirrors the registered tool's real contract in the wrapper and pins it with regression tests.

Closes #635

## Why

The wrapper advertised a fetch-by-codes capability that no registered tool implements, so the MCP schema was 100% broken — any well-formed call raised `KeyError` inside `execute()`. The registered tool's CSV contract is the source of truth (it is what the agent loop uses), so the minimal honest repair is to align the wrapper to it. Implementing fetch-by-codes would be a separate feature.

## Changes

- `agent/mcp_server.py`: `factor_analysis` wrapper now takes `(factor_csv, return_csv, output_dir, n_groups=5)` and forwards exactly those keys to the registry; docstring rewritten for the CSV-based contract.
- `agent/tests/test_mcp_factor_analysis_contract.py`: 3 regression tests — forwarded-keys pinning, wrapper↔`FactorAnalysisTool.parameters` drift guard, and an end-to-end call through the real registry on synthetic CSVs (the exact issue scenario).

## Test Plan

- [x] ruff check — clean
- [x] pytest (new regression tests) — 3 passed
- [x] pytest (mcp + factor related) — 1193 passed
- [x] pytest (full suite excl e2e) — 5539 passed, 9 skipped; 2 failures reproduce on a clean `main` worktree (`test_bottleneck_available` env check, `test_factor_gate_fund_prefix` runner test) and are unrelated to this change

## Checklist

- [x] No hardcoded values

### DIFF
--- a/agent/mcp_server.py
+++ b/agent/mcp_server.py
@@ -466,39 +466,30 @@ def backtest(run_dir: str) -> str:
 
 @mcp.tool
 def factor_analysis(
-    codes: list[str],
-    factor_name: str,
-    start_date: str,
-    end_date: str,
-    source: str = "auto",
-    top_n: int = 10,
-    bottom_n: int = 10,
+    factor_csv: str,
+    return_csv: str,
+    output_dir: str,
+    n_groups: int = 5,
 ) -> str:
-    """Compute factor IC/IR analysis and layered backtest for a cross-section of stocks.
+    """Compute factor IC/IR analysis and layered backtest from prepared CSVs.
 
     Analyzes factor predictive power using Spearman rank IC, IR (IC/std),
-    and top/bottom quintile return spreads.
+    and quantile group return spreads.
 
     Args:
-        codes: List of stock codes (e.g. ["000001.SZ", "600519.SH"]).
-        factor_name: Factor column name in daily_basic data (e.g. "pe_ttm", "pb", "turnover_rate").
-        start_date: Start date (YYYY-MM-DD).
-        end_date: End date (YYYY-MM-DD).
-        source: Data source ("tushare", "yfinance", "auto").
-        top_n: Number of top-ranked stocks per period.
-        bottom_n: Number of bottom-ranked stocks per period.
+        factor_csv: Path to factor values CSV (index=date, columns=codes).
+        return_csv: Path to returns CSV (same structure as factor_csv).
+        output_dir: Directory for output files (ic_series.csv, ic_summary.json, group_equity.csv).
+        n_groups: Number of quantile groups (default 5).
     """
     registry = _get_registry()
     return registry.execute(
         "factor_analysis",
         {
-            "codes": codes,
-            "factor_name": factor_name,
-            "start_date": start_date,
-            "end_date": end_date,
-            "source": source,
-            "top_n": top_n,
-            "bottom_n": bottom_n,
+            "factor_csv": factor_csv,
+            "return_csv": return_csv,
+            "output_dir": output_dir,
+            "n_groups": n_groups,
         },
     )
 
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @Robin1987China — this fixes the MCP `factor_analysis` contract (#635): the wrapper now mirrors the registered tool's real `factor_csv`/`return_csv`/`output_dir`/`n_groups` signature, so calls no longer die on `KeyError` before running. The signature-drift guard test is a great addition. Merged 🙏

PR #726, opened 2026-07-18
fix: Robinhood connector account_number wiring + remote-MCP display shape

nareshkps:
## Summary

Two compounding bugs that made the Robinhood remote MCP connector's read tools (`connector account` / `connector positions`) non-functional even on a fully authorized, healthy connection.

---

## Bug 1 — `account_number` silently dropped

**What was happening**

`get_account` / `get_positions` / `get_open_orders` in `service.py` always called `_call_remote(profile, op, {})` — a hardcoded empty dict, ignoring any `account` override passed in via the CLI `--account` flag or an agent tool call. `robinhood/mcp.py`'s `remote_arguments()` also returned `{}` for every operation except `quote`.

Robinhood's `get_portfolio` / `get_equity_positions` / `get_equity_orders` tools require `account_number` as a parameter, so every read call failed with:

```
invalid params: validating "arguments": validating root:
required: missing properties: ["account_number"]
```

**Fix**

- `remote_arguments()` now maps `account`/`account_number` into the field Robinhood's schema expects.
- `service.py` adds `_account_arg()` and wires it into all three read functions.
- `_call_remote` now also echoes `account_number` back into the result, so callers can display which account was actually queried (previously always showed as unknown).

---

## Bug 2 — CLI printer didn't understand the remote-MCP response shape

**What was happening**

Even after fixing Bug 1, `connector account` / `connector positions` still printed "No account summary returned" / "No positions returned" on a working connection. Two issues:

1. `fastmcp`'s auto-generated `Root` response type is a **dataclass** (built via `dataclasses.make_dataclass` from the tool's JSON Schema), not a Pydantic model — code checking `hasattr(value, "model_dump")` never matched, so the raw dataclass instance fell through unconverted.
2. Robinhood's response is **double-wrapped**:
   `result["data"] → Root(data=Root(<actual fields>), guide="<text>")`
   The real fields are one level deeper than `result["data"]`, alongside a sibling `guide` field containing the broker's own advisory text (e.g. "use `buying_power.buying_power`, not `cash`, as the authoritative spendable figure").

**Fix**

- `_normalize_mcp_value` now checks `dataclasses.is_dataclass()` and uses `dataclasses.asdict()` (which also recurses into nested dataclass fields like `buying_power`), falling back to `model_dump()` for any Pydantic-model shape.
- Both printers drill into the `data.data` / `data.guide` double-wrap when present, before falling back to the flat `broker_sdk` shape.
- The positions table now uses Robinhood's actual field names (`symbol`, `type`, `quantity`, `shares_available_for_sells`, `average_buy_price`) instead of the IBKR-style names it previously assumed.

---

## Verification

Reproduced and fixed against a live, read-only-authorized Robinhood Agentic Trading connector profile. Values below are illustrative placeholders, not real account data.

**Before:**
```
$ vibe-trading connector account --account XXXXXXXX
Session termination failed: 400
invalid params: validating "arguments": validating root:
required: missing properties: ["account_number"]
```

**After:**
```
$ vibe-trading connector account --account XXXXXXXX
Account: XXXXXXXX
                 Account Summary · robinhood-live-mcp
  Field                                    Value
  total_value                        12345.678901 USD
  equity_value                          2345.670000 USD
  crypto_value                          1000.000000 USD
  cash                                  9000.000000 USD
  buying_power.buying_power             9000.000000 USD
  ...
```

```
$ vibe-trading connector positions --account XXXXXXXX
Account: XXXXXXXX
              Positions · robinhood-live-mcp
  Symbol   Type        Qty   Avail. Sell   Avg Buy Price
  ABCD     long   10.000000     10.000000       42.500000
```

**No changes to any write/order-placement path** — both fixes are confined to read operations (`account`, `positions`, `orders`).

### DIFF
--- a/agent/cli/_legacy.py
+++ b/agent/cli/_legacy.py
@@ -3908,13 +3908,89 @@ def cmd_connector_check(
     return EXIT_SUCCESS
 
 
+def _normalize_mcp_value(value: Any) -> Any:
+    """Unwrap a value from a remote MCP call into plain JSON-safe data.
+
+    Remote connectors (e.g. Robinhood) return their payload as an instance of
+    ``fastmcp``'s auto-generated ``Root`` type — a *dataclass* built at runtime
+    via ``dataclasses.make_dataclass`` from the tool's JSON Schema, not a
+    Pydantic model. ``dataclasses.asdict()`` is the correct unwrap (it also
+    recurses into nested dataclass fields, e.g. ``buying_power``); a stray
+    Pydantic model elsewhere falls back to ``model_dump()``. Anything else
+    (already a dict/list/scalar) is returned as-is.
+    """
+    import dataclasses
+
+    if dataclasses.is_dataclass(value) and not isinstance(value, type):
+        return dataclasses.asdict(value)
+    if hasattr(value, "model_dump"):
+        return value.model_dump()
+    return value
+
+
+def _flatten_account_fields(data: dict[str, Any], prefix: str = "") -> list[tuple[str, str]]:
+    """Flatten a remote-MCP account payload into (tag, value) rows.
+
+    Nested one level (e.g. ``buying_power.buying_power``) rather than
+    recursing arbitrarily deep, since broker account payloads are shallow.
+    Skips ``None`` and zero-valued numeric-looking fields, matching the
+    guidance the broker's own response typically includes (zero means no
+    holdings in that asset class — noise to omit, not a real 0.00 balance).
+    """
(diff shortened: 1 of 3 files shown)


warren618:
Thanks @nareshkps — merged. Reconciled the remote-MCP account/positions rendering with the Longbridge schema-tolerant renderers (#735): account now renders balances (broker_sdk) → remote-MCP nested (Robinhood) → IBKR summary, and positions read both the remote-MCP shape and the key aliases.

PR #727, opened 2026-07-19
fix(channels): avoid hang in split_message when max_len <= 0

santhreal:
When max_len is 0 or negative, the cut pointer never advances and split_message loops forever. Return the content unsplit so outbound IM adapters cannot wedge.

Repro: split_message("hello world", max_len=0).


warren618:
Thanks @santhreal — clean catch. `split_message` genuinely spins forever when `max_len <= 0` (the cut pointer can't advance past 0), and returning the content unsplit is a sane fallback. The regression test is appreciated. Merged 🙏

PR #728, opened 2026-07-19
fix(tools): reject inverted PDF page ranges in doc_reader

santhreal:
`doc_reader` page specs like `10-5` produced an empty index list, so the PDF path returned success with no text. Ascending ranges are the documented grammar, and `alpha_bench` already rejects inverted periods.

`_parse_pages` now raises `ValueError` for inverted ranges.

Repro: `_parse_pages("10-5", 20)` returned `[]` on main and now raises.


warren618:
Thanks @santhreal — good hardening. Inverted page ranges like `10-5` silently read nothing before; now they fail fast, and the raise surfaces cleanly through `read_document`'s error envelope. Merged 🙏

PR #729, opened 2026-07-19
fix(tools): reject inverted trade-journal date filters

santhreal:
Trade journal `filter_expr` values like `2026-03 to 2026-01` parsed cleanly but selected zero rows, so analysis ran on an empty frame with no error. Ascending ranges are the documented examples, and alpha_bench already rejects inverted periods.

`_apply_filter` now raises `ValueError` when the lower bound is not before the upper bound.

Repro: `_apply_filter(df, "2026-03 to 2026-01")` returned an empty frame on main and now raises.

### DIFF
--- a/agent/src/tools/trade_journal_tool.py
+++ b/agent/src/tools/trade_journal_tool.py
@@ -389,10 +389,15 @@ def _apply_filter(df: pd.DataFrame, expr: str) -> pd.DataFrame:
                 if hi_month
                 else hi_base + pd.Timedelta(days=1)
             )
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @santhreal, merged 🙏 The fail-fast direction is right — silently returning an empty frame on an inverted range can read as "no trades in that window." One small follow-up landed on top (`0538db1`): `analyze_trade_journal` was calling `_apply_filter` unwrapped, so the new `ValueError` escaped as a raw traceback — it now returns the standard `{status: error}` envelope, and a pre-existing test that asserted the old silent-empty behavior was updated to expect the raise.

PR #730, opened 2026-07-19
fix(tools): return real Yahoo news articles for US/HK stock news

yxhuang:
## Goal

`get_stock_news` currently cannot return news for US/HK codes: the module itself documents that `yahoo_client.search()` only exposes the search endpoint's instrument `quotes`, so the tool returns related-instrument *matches* instead of articles. Yahoo's same public v1 search endpoint also carries a `news` array — this PR exposes it and makes US/HK return real headlines.

## Affected areas

- `agent/backtest/loaders/yahoo_client.py`: new `search_news(query, count)` helper, same throttled path (`throttled_get_json` + `HOST_KEY` + `_min_interval()`) as the existing `search()`; no existing function changed.
- `agent/src/tools/stock_news_tool.py`: US/HK route now returns `articles` in the exact shape the Eastmoney path already uses (`title/url/source/published/snippet`); `providerPublishTime` (epoch seconds) is formatted as UTC. The stopgap `matches` contract (`result_type='matches'`) is removed — repo grep shows no other consumer.
- `agent/tests/test_stock_news_tool.py`: mocked coverage for US/HK articles, empty `news`, limit clamping, and the error envelope; no network in tests.

## Out of scope

`scope='global'` behavior (Eastmoney), other loaders/tools, no new dependencies.

## Test plan

- `pytest agent/tests/test_stock_news_tool.py -q` → 17 passed
- `pytest agent/tests -k 'news or yahoo' -q` → 84 passed
- `black --check` clean on the three files; `ruff check` reports only two pre-existing F401s already present on `main` (left untouched to keep the diff focused).

## Risk notes

Read-only public endpoint, no auth, no broker/MCP/secret surface. Every outbound call stays behind the frozen throttled client. Failure mode unchanged: any fetch error returns the tool's error envelope, never raises.

## Rollback

Single revertable commit; `git revert` restores the previous matches behavior.

---
Agent-assisted contribution per AGENT_CONTRIBUTOR_GUIDE.md; reviewed and tested by the submitting maintainer of this fork.

### DIFF
--- a/agent/backtest/loaders/yahoo_client.py
+++ b/agent/backtest/loaders/yahoo_client.py
@@ -387,3 +387,28 @@ def search(query: str) -> List[Dict[str, Any]]:
     )
     quotes = (payload or {}).get("quotes") or []
     return [quote for quote in quotes if isinstance(quote, dict)]
+
+
+def search_news(query: str, count: int) -> List[Dict[str, Any]]:
+    """Look up matching news via the v1 search endpoint.
+
+    Args:
+        query: Free-text query (ticker fragment or company name).
+        count: Maximum number of news items requested from Yahoo.
+
+    Returns:
+        The ``news`` list (each a dict with ``title``, ``publisher``, ``link``,
+        ``providerPublishTime``, ``relatedTickers``, ...), or an empty list when
+        none match.
+
+    Raises:
+        requests.RequestException: On a network/HTTP failure.
+    """
+    payload = throttled_get_json(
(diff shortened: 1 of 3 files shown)


warren618:
Thanks @yxhuang — verified against the live endpoint: Yahoo's `v1/finance/search` does carry a `news` array (title/publisher/link/providerPublishTime), so US/HK now return real articles instead of instrument matches, and it stays on the frozen IP-throttled client. Merged 🙏

PR #734, opened 2026-07-19
fix(composite): fall back to first available sub-engine for unknown symbols

Marnie0415:
## Summary

`CompositeEngine._rule_for()` defaults to `"a_share"` for any symbol whose market type is not in `_symbol_market`. When `_rule_engines` does not contain an `"a_share"` key (because no A-share codes were in the original request), the method raises `KeyError: 'a_share'` and the backtest crashes.

## Reproduction

```python
from backtest.engines.composite import CompositeEngine

config = {"initial_cash": 100_000}
engine = CompositeEngine(config, ["BTC-USDT", "ETH-USDT"])

# Any method that dispatches through _rule_for crashes:
engine.round_size(1.0, 100.0)  # KeyError: 'a_share'
```

## Root cause

`_rule_for()` uses `self._symbol_market.get(symbol, "a_share")` which maps unknown symbols to `"a_share"`, then indexes `self._rule_engines[market]` without checking if that key exists. When the engine was constructed only for crypto+forex codes, no `"a_share"` sub-engine was built.

## Fix

Replace the hard index `self._rule_engines[market]` with `.get(market)` and fall back to the first available sub-engine when the resolved market has no dedicated rule engine. This keeps the current behavior for all known markets while preventing a crash on unrecognised symbols.

## Testing

Verified in Docker environment:
- `round_size` for an unknown symbol returns a valid float instead of crashing
- `calc_commission` for an unknown symbol returns a valid float
- `_leverage_for_symbol` for an unknown symbol returns a positive float
- All existing engine imports and behaviors unchanged (no regressions)

### DIFF
--- a/agent/backtest/engines/composite.py
+++ b/agent/backtest/engines/composite.py
@@ -92,6 +92,8 @@ class CompositeEngine(BaseEngine):
             market = "china_futures" if _is_china_futures(symbol) else "global_futures"
         engine = self._rule_engines.get(market)
         if engine is None:
+            if not self._rule_engines:
+                raise ValueError("No sub-engines available for composite backtest")
             engine = next(iter(self._rule_engines.values()))
         return engine
 
--- a/agent/backtest/engines/composite.py
+++ b/agent/backtest/engines/composite.py
@@ -90,7 +90,10 @@ class CompositeEngine(BaseEngine):
         market = self._symbol_market.get(symbol, "a_share")
         if market == "futures":
             market = "china_futures" if _is_china_futures(symbol) else "global_futures"
-        return self._rule_engines[market]
+        engine = self._rule_engines.get(market)
+        if engine is None:
+            engine = next(iter(self._rule_engines.values()))
+        return engine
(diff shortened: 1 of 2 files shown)


Marnie0415:
Updated to address review feedback:

- Added `tests/test_composite_engine_fallback.py` with 5 pytest tests covering:
  - `test_unknown_symbol_round_size` — unknown symbol does not crash
  - `test_unknown_symbol_calc_commission` — unknown symbol returns valid fee
  - `test_unknown_symbol_leverage` — unknown symbol returns positive leverage
  - `test_known_symbols_still_use_correct_engine` — known symbols route correctly
  - `test_empty_rule_engines_raises` — empty sub-engines raises ValueError

- Added defensive guard for empty `_rule_engines` dict (raises `ValueError` instead of `StopIteration`)

warren618:
Thanks @Marnie0415 — merged. Falling back to the first available sub-engine is the right guard for a cross-market composite with no A-share leg.

PR #736, opened 2026-07-20
fix(channels): avoid hang splitting telegram fenced markdown chunks

santhreal:
Long unclosed fenced bodies (and mid-size fence-at-start cuts) could land the split on the fence line and reopen the same chunk forever.

Advance past the fence header, bail when reopen does not shrink the buffer, and reject non-positive `max_len` like `split_message`.


warren618:
Thanks @santhreal — merged. Good catch on the fence-split reopen loop; the signal.alarm hang-regression tests are a nice touch.

PR #737, opened 2026-07-20
fix(ui): skip non-positive MA periods in indicator overlays

santhreal:
Run-detail MA overlays infer periods from planner/design JSON. A zero `ma_window` made `sum(window)/period` raise `ZeroDivisionError`.

Skip non-positive periods when inferring and when building the series.


warren618:
Thanks @santhreal — merged. A zero/negative MA period should never reach the overlay math.

PR #738, opened 2026-07-20
fix(api): skip commented dotenv keys when upserting settings

santhreal:
Web UI settings writes matched `# KEY=...` before the active `KEY=` line, so the update could be overwritten on read and silently discarded when `.env` had commented examples.

Skip commented lines when matching keys; append missing keys as before.


warren618:
Thanks @santhreal! Folding this into #742, which rewrites the same `_write_env_values` loop — it already skips commented `# KEY=` lines while also fixing the last-duplicate-active-key upsert. Your commented-key regression case is preserved as a test on `main`, so the coverage isn't lost. Closing as superseded by #742.

PR #739, opened 2026-07-20
fix(metrics): avoid OverflowError annualizing explosive equity paths

santhreal:
calc_metrics annualised with float(growth ** (bars_per_year / n) - 1). Short crypto 1m windows (bars_per_year=525600) overflow that power on CPython for ordinary gains like 2x over a few hundred bars, crashing the metrics step after an otherwise successful backtest.

Catch OverflowError and treat annual_return as +inf (run_card already sanitizes non-finite values to null).

Repro: calc_metrics(pd.Series([1.0, 1_000_000.0]), [], 1.0, 252) raised OverflowError on main and now returns annual_return=inf.

### DIFF
--- a/agent/backtest/metrics.py
+++ b/agent/backtest/metrics.py
@@ -267,7 +267,14 @@ def calc_metrics(
     if growth <= 0:
         ann_ret = -1.0
     else:
-        ann_ret = float(growth ** (bpy / max(n, 1)) - 1)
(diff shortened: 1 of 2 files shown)


warren618:
Thanks @santhreal — merged. OverflowError on explosive equity paths is exactly what short minute-bar windows hit; treating annual_return as non-finite is the right call.

PR #740, opened 2026-07-20
fix(options): catch OverflowError in options metrics annualization

santhreal:
Minute-bar options equity paths annualize with bars_per_year=525600. Modest growth then overflows float(growth ** factor) before the isfinite guard. Catch OverflowError and report annual return as undefined with a warning.

Repro: `_calc_options_metrics(equity 1e5→1.2e5 over 100 bars, bars_per_year=525_600)` raises OverflowError on main.


warren618:
Thanks @santhreal — merged. Same annualization-overflow guard for the options metrics path.

PR #741, opened 2026-07-20
fix(pattern): reject window=1 for trend_line_slope

santhreal:
trend_line_slope accepted window=1, then polyfit on a single point raised LinAlgError. Require window>=2 (and return a JSON error from run_pattern for trend_slope).

Repro: trend_line_slope(pd.Series(range(30), dtype=float), window=1)


warren618:
Thanks @santhreal — merged. window>=2 is the correct floor for a polyfit slope.

PR #742, opened 2026-07-20
fix(api): upsert last duplicate active dotenv key

santhreal:
When `.env` has duplicate active `KEY=` lines, read is last-wins but upsert updated the first line, so settings writes could stick on a stale value. Upsert the last active line instead.


warren618:
Thanks @santhreal — merged. Upserting the last active key (and skipping commented ones) fixes the stale-value write.

PR #743, opened 2026-07-20
fix(api): keep hash characters inside quoted dotenv values

santhreal:
Settings quotes dotenv values that contain `#`. On read, `_strip_env_value` stripped ` #` comments before handling quotes, so values like `secret # still-part-of-value` were truncated on reload.

Parse a leading quote through to its closing quote (skip escapes), then strip unquoted ` #` comments only.

Repro: `_strip_env_value('"secret # still-part-of-value"')` returned `"secret` on main.


warren618:
Thanks @santhreal — merged. Quoted dotenv values containing '#' should survive a read/write round-trip.

PR #744, opened 2026-07-20
fix(options): treat non-positive spot/strike as intrinsic in BS helpers

santhreal:
bs_price/bs_greeks raise ZeroDivisionError when a signal leg uses strike 0, because S/K is undefined before log. Fall back to intrinsic value the same way the expired/zero-vol path already does; iv_smile_adjustment already soft-guards non-positive S/K.

Repro: bs_price(100.0, 0, 1.0, 0.05, 0.2, "call")


warren618:
Thanks @santhreal — merged. Non-positive S/K falling back to intrinsic keeps the BS helpers safe on degenerate legs.

PR #755, opened 2026-07-21
fix(providers): pass reasoning_effort through to OpenAI-compatible providers

1anter:
## Goal

Make OpenAI's gpt-5.6 family usable as the agent LLM. These models reject function tools on `/v1/chat/completions` unless the request carries an explicit `reasoning_effort` — including the literal `'none'`:

```
provider_stream_error provider=openai model=gpt-5.6-sol: BadRequestError: Error code: 400 -
"Function tools with reasoning_effort are not supported for gpt-5.6-sol in /v1/chat/completions.
 To use function tools, use /v1/responses or set reasoning_effort to 'none'."
```

Today `LANGCHAIN_REASONING_EFFORT` only feeds the OpenRouter `extra_body` opt-in and never reaches `ChatOpenAI` for direct providers, and both the settings API allowlist and the Settings UI reject the value `none` — so there is no way to satisfy the API's requirement and every agent run on a gpt-5.6-* model fails on its first tool-calling iteration.

## What changed

- `agent/src/providers/llm.py`: forward the configured effort to `ChatOpenAI` for non-relay providers. Unset (`""`) still passes `None`, which langchain omits from the request — no behavior change for existing configs. The OpenRouter `extra_body` path is untouched (relay providers still get `reasoning_effort=None` to avoid double-sending).
- `agent/src/api/settings_routes.py`: accept `"none"` in `LLM_REASONING_EFFORTS`.
- `frontend/src/pages/Settings.tsx`: add the `none (explicit)` option to the reasoning-effort dropdown, distinct from `Off` (which omits the field entirely).
- `agent/tests/test_llm_reasoning_effort.py`: regression tests — effort `'none'` is forwarded for `openai`, unset stays absent, OpenRouter keeps its `extra_body` opt-in with `reasoning_effort=None`, and the settings allowlist admits `'none'`.

## Affected areas

Provider factory (`build_llm` kwargs), settings API validation, Settings UI dropdown. No changes to the agent loop, tools, trading, or MCP surfaces.

## Deliberately out of scope

- Auto-detecting gpt-5.6-* models and defaulting them to `'none'` (kept explicit and operator-controlled).
- Migrating the OpenAI provider to `/v1/responses` (the API's alternative suggestion).
- Plumbing reasoning effort per-model rather than globally.

## Test plan

- `pytest agent/tests/test_llm_reasoning_effort.py agent/tests/test_llm.py agent/tests/test_settings_api.py -q` → 60 passed.
- Full suite collection clean (5,649 tests).
- Verified live: with `LANGCHAIN_PROVIDER=openai`, `LANGCHAIN_MODEL_NAME=gpt-5.6-sol`, `LANGCHAIN_REASONING_EFFORT=none`, a natural-language backtest request (ETH-USDT SMA crossover via OKX data) previously failed with the 400 above on the first tool call and now completes end-to-end with tool calls, backtest execution, and final report.

## Risk

- LLM/provider risk only: the new kwarg is `None` unless an operator sets the env var/setting, so existing deployments see byte-identical requests. Relay providers (OpenRouter) are explicitly excluded from the pass-through.
- No live-trading, broker, MCP, secret, or network-surface changes.

## Rollback

Single revertable commit; reverting restores prior behavior (gpt-5.6-* models unusable, all other providers unaffected). The regression tests pin the new contract if a refactor touches `build_llm`.


warren618:
Thanks — the change is correct and well-targeted: `reasoning_effort` is a native ChatOpenAI field so it reaches the request as a top-level param (not one of the dropped message-conversion boundaries), the relay-vs-direct split is right, and adding `'none'` unblocks the gpt-5.6 function-tools case. Tests cover forward/unset/relay.

Two mechanical blockers: (1) it's still a draft — mark it ready when you're set; (2) it conflicts — main reformatted the adjacent `extra_body` block to multi-line after you branched, so please rebase (trivial, no logic change).

Once ready + rebased, we'll merge. Thanks!

warren618:
Merged with your authorship. The core observation was right, and I verified it
independently: `reasoning_effort` does reach the payload as a top-level field,
including the literal "none".

The gating changed. Keying on "every provider whose `openrouter_reasoning_body`
is false" would have sent the field to DeepSeek, Gemini, Groq, DashScope,
Zhipu, NVIDIA, Spark and MiniMax, any of which may reject an unknown field —
and it contradicted an existing test asserting DeepSeek must not receive it. The
merged version uses a positive allowlist covering direct OpenAI only, checking
both the resolved capability name and the configured provider name, because
either alone leaks: model inference maps `provider=openai` plus a DeepSeek
model onto DeepSeek capabilities, and unknown providers fall back to OpenAI
capabilities, so a flag on the capability record alone would start sending the
field to any unrecognized gateway. OpenRouter/Requesty keep `extra_body`.

Also added: "none" in the settings allowlist with a corrected rejection
message, the option through `t()` with labels in all five locales, and the
resolved capability in `provider_diagnostics` so `doctor` can explain why an
effort is or isn't sent.

PR #777, opened 2026-07-22
feat(backtest): portfolio risk x-ray core + agent tool (Portfolio Studio slice)

he-yufeng:
## Summary

- First slice of the Portfolio Studio roadmap item (risk x-ray; the turnover-aware optimizer already shipped, the rest is still planned): a pure-computation risk x-ray for weighted baskets.
- A new `portfolio_risk_xray` agent tool that fetches daily closes through the standard loader fallback chain and returns a strict-JSON report.

## Why

Portfolio Studio has the optimizer but nothing that answers the question that comes first: how risky is this basket right now. This PR is the diagnostic half of the studio. Concentration, volatility, drawdown, tail risk, and co-movement, in one call the agent can make during portfolio review or pre-rebalance research.

## Changes

- `agent/backtest/risk_xray.py`: pure core, no I/O or network. Given a close panel and weights, it computes HHI / effective N / top-1 / top-3 concentration, annualized and downside volatility, max drawdown with start and trough dates, historical VaR and expected shortfall at 95/99, diversification ratio, and pairwise correlation plus beta against an equal-weight proxy. All statistics are trailing-window only, so look-ahead is impossible by construction. Weights are validated and renormalized with a recorded warning. Thin symbols fall out through a min-history filter, with renormalization over the survivors. Every emitted float is finite-guarded, so the payload always survives `json.dumps(allow_nan=False)`. v1 is long-only: negative weights are rejected explicitly instead of mis-computing concentration, and a short-leg slice can define its own semantics later.
- `agent/src/tools/portfolio_risk_tool.py`: `BaseTool` wrapper. Validates arguments, shapes the fetch envelope into a close panel (records with missing or mixed timestamps fall back to loader order), and returns a strict-JSON envelope with `unresolved_symbols` / `skipped` bookkeeping. The fetcher is injectable so the tool is testable without network.
- `agent/tests/test_risk_xray.py`: 17 tests. Hand-computed math (HHI, drawdown on a known curve, ES on a known tail, beta=1 against the equal-weight proxy, unit diversification ratio for identical series), the edge matrix (renormalization, negative/unknown weights, thin symbols, single asset, zero-variance panel, mixed timestamps), and the tool end to end.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q`): **6037 passed, 12 skipped**
- [x] New tests added: 17, all passing
- [x] Tested manually: the tool runs end to end against a stub fetcher in tests. The live path goes through the same `fetch_market_data` fallback chain as the other market tools. Not run: a live-network fetch (no market data committed, per repo rules).

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`); new files only under `agent/backtest/` and `agent/src/tools/`
- [x] Code follows CONTRIBUTING.md guidelines (DCO signed)
- [ ] Documentation updated: deliberately deferred. Wiki docs for the new tool read better as their own slice once the shape is reviewed.

Scope notes: backend + agent tool only. Natural next slices are the constraints model and rebalance notes from the same roadmap row, and a UI card once the report shape settles.

### DIFF
--- a/agent/backtest/risk_xray.py
+++ b/agent/backtest/risk_xray.py
@@ -1,0 +1,277 @@
+"""Portfolio risk x-ray: concentration, volatility, drawdown, tail risk, and
+co-movement for a weighted basket, computed from close-price panels.
+
+This is the Portfolio Studio risk-x-ray backend (roadmap slice). Everything
+here is a pure function of prices and weights — no I/O, no network, no
+loader imports — so the same computation serves the agent tool, the API,
+and tests. Data fetching lives in the caller.
+
+Conventions:
+- Long-only v1: negative weights are rejected up front rather than silently
+  mis-computing concentration (a short leg is a different feature).
+- Non-finite inputs are never trusted and non-finite outputs are never
+  emitted: every reported float passes through ``_finite`` so the result
+  survives ``json.dumps(..., allow_nan=False)``.
+- Look-ahead is structurally impossible: all statistics are computed over
+  trailing returns only.
+"""
+
+from __future__ import annotations
+
+import math
+from typing import Any, Mapping, Sequence
+
+import numpy as np
+import pandas as pd
+
+MIN_HISTORY_DAYS = 30
+PERIODS_PER_YEAR = 252
+VAR_LEVELS = (0.95, 0.99)
+
+
+def _finite(value: float | None) -> float | None:
+    """Return ``value`` when finite, else ``None`` (strict-JSON safe)."""
+    if value is None:
+        return None
+    try:
+        out = float(value)
+    except (TypeError, ValueError):
+        return None
+    return out if math.isfinite(out) else None
+
+
+def _validate_weights(
+    closes: pd.DataFrame, weights: Mapping[str, float]
+) -> tuple[dict[str, float], list[str]]:
+    """Normalize weights to sum 1; reject unknown symbols and bad values.
+
+    Returns the cleaned weight map (restricted to available columns) and any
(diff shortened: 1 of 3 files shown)


warren618:
Merged — thanks @he-yufeng, this is a clean, well-scoped slice. 🎯

What stood out on review:

- **Pure core, right boundaries.** `risk_xray.py` is a pure function of prices + weights — no I/O, no loader imports — so the same computation serves the tool, the API, and tests. Only `backtest/` and `src/tools/` touched; protected areas untouched.
- **Look-ahead is impossible by construction** (trailing-window returns only), and every emitted float passes through `_finite`, so the payload always survives `json.dumps(allow_nan=False)`.
- **The edge matrix is real, not decorative** — renormalization, negative/unknown weights, thin-symbol filtering, single asset, zero variance, and mixed timestamps are all covered. 17/17 green locally, and the tool auto-registers through `BaseTool.__subclasses__()` (verified: it shows up in the built registry).
- Nice call rejecting negative weights explicitly instead of silently mis-computing concentration, and leaving a short-leg slice for later.

One non-blocking follow-up: the live `fetch_market_data` path is only exercised through the stub in tests (correct — no market data in the repo). I checked the envelope shape against what `_closes_frame` expects and it lines up, so the live path should hold. Docs deferral is fine as its own slice once the report shape settles.

Really solid work. We'd love to have you on as a maintainer — sending an invite now. Welcome aboard 🙌

PR #799, opened 2026-07-23
fix(channels): keep empty edge columns in Telegram md tables

santhreal:
Markdown tables with empty leading or trailing cells lost those columns in Telegram output, shifting later fields.


warren618:
Merged, thanks @santhreal! Third of the channel table fixes — empty leading/trailing columns now survive Telegram conversion too. Thanks for the thorough, consistent series!

PR #800, opened 2026-07-23
fix(channels): keep empty edge columns in Signal md tables

santhreal:
Markdown tables with empty leading or trailing cells lost those columns in Signal output, shifting later fields.


warren618:
Merged, thanks @santhreal! Same edge-column preservation as the Slack fix, applied consistently to Signal — keeping the channel formatters aligned in behavior is much appreciated.

PR #801, opened 2026-07-23
fix(channels): keep empty edge columns in Slack md tables

santhreal:
Markdown tables with empty leading or trailing cells lost those columns in Slack output, shifting later fields.


warren618:
Merged, thanks @santhreal! `strip("|")` eating empty edge cells was a subtle one — peeling only the bounding pipes preserves column alignment. The stubbed-deps test keeps it hermetic; nice.

PR #802, opened 2026-07-23
fix(connectors): map Futu 1H/4H periods to hour bars

santhreal:
Project-style `1H`/`4H` periods fell through to daily bars on Futu.

Map them to hour bars like the other connectors.


warren618:
Merged, thanks @santhreal! Mapping 4h/4H to `K_240M` instead of silently reusing `K_60M` is a real correctness fix — a 4H request getting 1H bars would quietly poison timeframe-sensitive backtests.

PR #803, opened 2026-07-23
fix(connectors): reject unsupported Dhan history periods

santhreal:
Unsupported Dhan history periods silently became daily bars.

Reject them with a clear error, matching Shoonya.


warren618:
Merged, thanks @santhreal! Failing closed with a clear "unsupported period" error beats silently serving daily bars for a 1H request — silent data substitution is the worst kind of bug to debug downstream. Clean tests too.

PR #804, opened 2026-07-23
fix(loaders): keep lowercase 1h/4h/1w as ccxt hour/week bars

santhreal:
Fleet-style lowercase `1h`/`4h`/`1w` fell through to daily on the ccxt loader.

Alias them to the hour/week bars (same class as the MT5 loader aliases).


warren618:
Merged, thanks @santhreal! Keeping lowercase 1h/4h/1w as hour/week bars closes the case-sensitivity gap symmetrically with the uppercase aliases. Small fix, exactly the right size.

PR #805, opened 2026-07-23
fix(connectors): map Binance 1H/4H periods to ccxt timeframes

santhreal:
Project-style `1H`/`4H` were passed raw to ccxt, which expects lowercase unified timeframes.

Map them before the history request.


warren618:
Merged, thanks @santhreal! Passing raw user period tokens straight into ccxt was fragile — the explicit map with a safe `1d` default is much better, and the monkeypatched-exchange tests cover it without touching the network.

PR #806, opened 2026-07-23
fix(connectors): map MT5 1H/4H/1D periods to hour/day bars

santhreal:
Project-style `1H`/`4H`/`1D` periods fell through to daily on the MT5 connector.

Map them to hour/day bars.


warren618:
Merged, thanks @santhreal! Project-style 1H/4H/1D/1W aliases now resolve to the right MT5 timeframes, and the comment calling out 1m-vs-1M case sensitivity will save the next reader a double-take.

PR #807, opened 2026-07-23
fix(tools): keep empty cells when parsing report_audit tables

santhreal:
Empty mid-row cells in report_audit markdown tables dropped out and shifted later columns.

Keep empty cells when splitting rows.


warren618:
Merged, thanks @santhreal! Keeping empty cells so values stay under the right headers is the correct fix — filtering them silently shifted quarters, exactly as your tests demonstrate. The `_split_md_table_row` docstring explains the "why" well.

PR #808, opened 2026-07-23
fix(journal): normalize stringified Excel serials in Tonghuashun dates

santhreal:
Tonghuashun exports that stringify Excel serials (for example `45321.5`) kept the raw number instead of a datetime.

Normalize stringified serials the same way as numeric ones.


warren618:
Merged, thanks @santhreal! Good catch on the dtype=str load path turning Excel serials into strings, and the `< 100_000` guard keeps YYYYMMDD ints safe. The xlsx round-trip test is the right coverage.

PR #809, opened 2026-07-23
fix(journal): normalize stringified Excel serials in Futu dates

santhreal:
Futu exports that stringify Excel Date/Time serials kept the raw number instead of a datetime.

Normalize stringified serials the same way as numeric ones.


warren618:
Merged, thanks @santhreal! Handling both the serial date *and* the fractional-day time column (0.375 → 09:00) is exactly right — most fixes would have missed the time half.

PR #810, opened 2026-07-23
fix(journal): normalize Excel serials in generic datetime columns

santhreal:
Generic journal datetime columns that arrive as stringified Excel serials kept the raw number.

Normalize those serials before formatting.


warren618:
Merged, thanks @santhreal! Completing the trilogy — the generic parser now survives dtype=str Excel serials too, with the same careful numeric guards. Consistent, well-tested work across all three parsers.

PR #811, opened 2026-07-23
fix(journal): normalize Excel day-count serials in Eastmoney dates

santhreal:
Eastmoney exports that store `成交日期` as a General-format day serial (for example `44941`) kept the raw number after `dtype=str`, so the journal row became `NaT`.

Normalize those day-count serials the same way as the Tonghuashun/Futu path.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `9178cda`. Appreciate the contribution.

PR #812, opened 2026-07-23
fix(loaders): map Yahoo 4H to 1h like yfinance

santhreal:
Project `4H` fell through to unsupported Yahoo chart interval `4h`, so symbols were omitted or empty.

Map `4H` to `1h` like the yfinance loader.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `0891fcf`. Appreciate the contribution.

PR #813, opened 2026-07-23
fix(loaders): reject unsupported BaoStock history intervals

santhreal:
BaoStock is daily-only, but `1H`/`4H` still logged in and returned daily bars cached under the requested timeframe.

Reject unsupported intervals with an empty map, matching tencent/fmp.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `cf33899`. Appreciate the contribution.

PR #814, opened 2026-07-23
fix(loaders): reject unsupported Alpha Vantage history intervals

santhreal:
Alpha Vantage always called `TIME_SERIES_DAILY` even for `1H`/`4H`, so intraday runs silently got day bars and stopped the fallback chain.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `38a994e`. Appreciate the contribution.

PR #816, opened 2026-07-23
feat(backtest): opt-in support for non-positive prices (closes #571)

### LINKED ISSUE #571: BaseEngine rejects opening fills at price <= 0 — blocks EU power backtests (negative prices are valid data)
## Problem

`BaseEngine._plan_open_order` returns `None` when `open_price <= 0`, so a backtest can never OPEN a position on a bar with a zero/negative price. For equities that's a data-sanity guard; for European power it rejects valid market states: day-ahead zone prices clear negative routinely (our ENTSO-E-sourced 2019–2025 panel has 243 zone-days with a *negative daily average*, min −85.27 EUR/MWh — individual hours go much lower). `validate_ohlc` in `backtest/loaders/base.py` has the same assumption (drops bars with any `price <= 0`).

Positions already held mark and close correctly through negative prints — only opening fills are blocked, which silently biases entries in energy backtests (we measured 19 blocked DELU days in a 7-year day-ahead spread study).

## Context

We build an external energy extension ([vibe-trading-energy](https://github.com/darkknight4563/vibe-trading-energy)) on the 0.1.11 loader/engine interfaces — market engines subclass `BaseEngine` following the `IndiaEquityEngine` pattern, and everything else about the contract handles negative prices fine (our engine uses absolute EUR/MWh slippage and `abs(price)`-based margin with a floor, both well-defined through zero).

## Suggested direction (happy to PR whichever shape you prefer)

1. An engine-level opt-in, e.g. class attribute `allow_nonpositive_prices = False` on `BaseEngine` that subclasses (or config) can flip — the `open_price <= 0` guard becomes `open_price <= 0 and not self.allow_nonpositive_prices`, and `validate_ohlc` gains an `allow_nonpositive=False` parameter for loaders that opt in.
2. Alternatively, move the guard into `can_execute` so market engines own the rule entirely (equity engines keep rejecting, commodity/power engines allow).

Option 1 is the smaller diff and keeps current behavior as the default everywhere. If either direction sounds right, I'll follow up with a PR including tests (negative-price open/close round trip, sizing/margin sanity at negative prints).

Related: #570 (user presets directory) from the same extension work.

🤖 AI-assisted issue (Claude Code), per AGENT_CONTRIBUTOR_GUIDE.md; submitted by the account owner.

darkknight4563:
# feat(backtest): opt-in support for non-positive prices (closes #571)

## Summary

Adds `allow_nonpositive_prices` (default `False`) to `BaseEngine` and `validate_ohlc`, so subclasses that trade markets where non-positive prices are valid — European day-ahead power, spreads — can ingest and trade those bars. Default behaviour is unchanged for every existing engine.

Per @warren618's guidance in #571: engine-level opt-in, subclasses set it `True`, the same flag gates `validate_ohlc` in the loader.

## Why this is a data-loss bug, not only a missing feature

While preparing this change I measured what the current guards do to a real European power dataset (DE-LU and NO2 day-ahead, 165 calendar days each, loaded via a local CSV bridge):

| | source days | days reaching the engine | dropped |
|---|---|---|---|
| NO2 (Norway, hydro) | 165 | 157 | 8 |
| DE-LU (Germany, solar-heavy) | 165 | 128 | 37 |

Every dropped bar has `low <= 0` in the source; every surviving bar has `low > 0` — 45/45 correlation, no exceptions. The asymmetry matches the physics: negative clears cluster on solar-heavy DE-LU, so it loses ~4.5× more days than hydro-dominated NO2.

The bars are dropped **silently** — no warning, no counter — and the survivors are simply re-indexed, so any rolling statistic computed downstream (z-scores, correlations, volatility) is calculated over a series with holes in exactly its most volatile days. Nothing in the run output indicates this happened.

Full reproduction and per-day breakdown in the issue comment on #571.

## What changed

| File | Change |
|---|---|
| `agent/backtest/loaders/base.py` | `validate_ohlc` takes `allow_nonpositive_prices` (default `False`). Structural invariants (`high >= low`, OHLC ordering, NaN rejection) are always enforced. When the flag is set, **negative** prices are treated as valid data rather than dropped; an exactly-`0` price is still dropped (undefined for the downstream return math — see below). |
| `agent/backtest/engines/base.py` | `BaseEngine.allow_nonpositive_prices = False` class attribute. `_plan_open_order` always rejects `open_price == 0`; it rejects negative prices only when the flag is off. `_calc_raw_size` and `_calc_margin` use `abs(price)`. |
| `agent/tests/test_nonpositive_prices.py` | New. |
| `agent/tests/fixtures/negative_close_bars.csv` | New — real non-positive-close bars (see provenance below). |

## Negative prices are kept; exactly zero is not

The flag admits **negative** prices and continues to reject exactly `0`:

- **Negative prices** are real published auction outcomes and divide cleanly — the engine sizes on `abs(price)` (below), so a negative entry is well-defined. These flow through both layers.
- **Exactly `0`** is undefined for the return math this codebase runs downstream. Both `benchmark.py` and `engines/base.py` compute per-bar returns as `close.pct_change().fillna(0.0)` over a *raw* close series; a surviving `0.00` close makes the *next* bar divide by zero and produce `inf`, which `fillna(0.0)` does **not** neutralize (it fills NaN, not inf) — the benchmark's `(1 + r).prod()` then collapses to `nan`. So `0` is dropped at the loader, and `_plan_open_order` rejects `price == 0` at the engine (`target_notional / 0` is undefined) regardless of the flag.

Making `0` a valid, tradable-through bar would require an inf-guarded return path at those two call sites; that is a follow-up (see *Out of scope*), not part of this change. The fixture carries the real `0.00` clear (DE-LU 2025-10-26) so the tests pin that it stays dropped, alongside DE-LU 2025-10-04 which *opened* at `-0.01` and is kept.

## Relaxing the guard alone is not sufficient

@Robin1987China called this correctly in #571: the two guards are not the only obstruction. With a negative open price, `target_notional` (always positive) divided by a negative price yields a negative size, which then trips the existing `size <= 0` check two lines below — so the order is still rejected, just further downstream.

`abs(price)` in `_calc_raw_size` and `_calc_margin` resolves it: size and margin are magnitudes, and the position's direction is already carried by the side. `_calc_pnl` is a price *difference* and is well-defined through zero, so it is unchanged.

## Backward compatibility

Default behaviour is byte-identical for positive-price markets:

- `abs(price)` is a no-op when `price > 0`.
- With the flag off, the engine gate reduces to the previous `open_price <= 0` rejection.
- With the flag off, `validate_ohlc` behaviour is unchanged.

No existing engine sets the flag; nothing opts in by accident.

## Testing

New tests (`agent/tests/test_nonpositive_prices.py`):

- a subclass with the flag set opens a position on a negative-price bar
- size, margin, and PnL are well-defined and correctly signed through zero
- `price == 0` is still rejected for opening, with the flag on
- with the flag off (default), negative prices are still rejected
- `validate_ohlc` keeps **negative** bars when the flag is set and still drops an exactly-`0` bar, with a regression test pinning *why* — a surviving `0.00` close yields `inf` at the `close.pct_change()` sites in `benchmark.py` / `engines/base.py` — drops all non-positive bars when the flag is off, and enforces structural invariants either way

Results:

- 15 passed — new tests plus the existing OHLC-validation tests
- 125 passed — existing engine suites (base, china_a, china_futures, forex), zero regressions
- ruff: no new lint findings introduced by this change — the changed files and the new test were verified against pristine `upstream/main` (the one pre-existing `E402` in `engines/base.py` is on a line this PR does not touch)

## Fixture provenance

`agent/tests/fixtures/negative_close_bars.csv` contains five real day-ahead bars with non-positive closes from German (DE-LU) and Norwegian (NO2) markets. Source: Bundesnetzagentur | SMARD.de and ENTSO-E via Energy-Charts (Fraunhofer ISE), licensed CC BY 4.0. Attribution is in the file header. No values are synthetic.

## Out of scope

- Extending this to intraday or other resolutions.
- Any change to benchmark or metrics computation.

Separately: the return path assumes strictly positive prices in at least two places — `benchmark.py:92` and `engines/base.py:221` both call `close.pct_change().fillna(0.0)` on a raw price series. This compounds unstably as prices approach zero and produces `inf` at zero, which is why this PR stops at negatives. I'll open a separate issue with a reproduction.

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -360,6 +360,14 @@ class BaseEngine(ABC):
         self.config = config
         self.initial_capital: float = config.get("initial_cash", 1_000_000)
         self.default_leverage: float = config.get("leverage", 1.0)
+        # Markets that clear at or below zero (e.g. EU day-ahead power) opt in
+        # to opening on negative-price bars. Default False preserves the legacy
+        # "reject any open_price <= 0" behavior. An exactly-zero open is always
+        # rejected (undefined size = notional / price); negatives are handled by
+        # abs()-based sizing and margin below.
+        self.allow_nonpositive_prices: bool = bool(
+            config.get("allow_nonpositive_prices", False)
+        )
         self.capital: float = self.initial_capital
         self.positions: Dict[str, Position] = {}
         self.trades: List[TradeRecord] = []
@@ -439,14 +447,24 @@ class BaseEngine(ABC):
     def _calc_margin(
         self, symbol: str, size: float, price: float, leverage: float,
     ) -> float:
-        """Margin (collateral) required for a position."""
-        return size * price / leverage
+        """Margin (collateral) required for a position.
+
+        ``abs(price)`` so collateral stays positive when the entry price is
+        negative; for the usual positive price this is unchanged.
+        """
+        return size * abs(price) / leverage
 
     def _calc_raw_size(
         self, symbol: str, target_notional: float, price: float,
     ) -> float:
-        """Convert target notional exposure to number of units/contracts."""
-        return target_notional / price
+        """Convert target notional exposure to number of units/contracts.
+
+        Size is a positive magnitude — direction carries the sign elsewhere —
+        so divide by ``abs(price)``; a negative entry price must not flip the
+        size negative (which the ``size <= 0`` guard would then reject). For a
+        positive price this is unchanged.
+        """
+        return target_notional / abs(price)
 
     def _leverage_for_symbol(self, symbol: str) -> float:
         """Return leverage used to size and margin one symbol."""
@@ -949,7 +967,10 @@ class BaseEngine(ABC):
         if not self.can_execute(symbol, direction, bar):
             return None
         open_price = float(bar.get("open", bar.get("close", 0)))
-        if open_price <= 0:
+        # Zero is always rejected (size = notional / price is undefined);
+        # negatives are rejected unless this engine opted into non-positive
+        # prices, in which case abs()-based sizing/margin below handle them.
+        if open_price == 0 or (open_price < 0 and not self.allow_nonpositive_prices):
             return None
         price = self.apply_slippage(open_price, direction)
         leverage = self._leverage_for_symbol(symbol)
--- a/agent/backtest/loaders/base.py
+++ b/agent/backtest/loaders/base.py
@@ -47,7 +47,12 @@ def validate_date_range(start_date: str, end_date: str) -> None:
         raise ValueError(f"start_date ({start_date}) > end_date ({end_date})")
 
 
-def validate_ohlc(frame: pd.DataFrame, *, strategy: str = "drop") -> pd.DataFrame:
+def validate_ohlc(
+    frame: pd.DataFrame,
+    *,
+    strategy: str = "drop",
+    allow_nonpositive_prices: bool = False,
+) -> pd.DataFrame:
     """Drop, flag, or reject bars that violate OHLC invariants.
 
     Loaders only drop NaN rows, so structurally dirty bars — ``high < low``,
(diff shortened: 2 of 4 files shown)


warren618:
Merged — thanks @darkknight4563! Now on `main` via squash commit `62aea55`. Appreciate the contribution.

PR #817, opened 2026-07-23
feat(openbb): OpenBB Workspace agent support

shugaoye:
## Summary

- Add an OpenBB Workspace integration layer (`openbb_bridge`) that exposes Vibe-Trading as a custom AI Agent via `GET /agents.json` (agent discovery) and `POST /v1/query` (streaming query).
- Translate Vibe-Trading's `AgentLoop` execution events into OpenBB-standard SSE (`text_delta`→`message_chunk`, `reasoning_delta`→`reasoning_step`, `tool_call`/`tool_result`→`reasoning_step` INFO) and inject Workspace widget/dashboard context into agent prompts.
- Allow OpenBB Workspace (`https://pro.openbb.co`) cross-origin access by adding it to the default CORS origins, so the bridge works regardless of `.env` load timing.

## Why

- Lets Vibe-Trading be used as a conversational finance agent directly inside OpenBB Workspace, extending it beyond the standalone API/CLI/MCP/frontend.
- Without the CORS entry, Workspace's browser preflight to `/agents.json` is rejected (`400 Disallowed CORS origin`), blocking agent discovery entirely. (The env-based `CORS_ORIGINS` is captured at import time before `load_dotenv` populates it, so the default origins had to be updated directly.)

## Changes

- **New `agent/src/openbb_bridge/` package**: request adapter (`conversation_id` ↔ `session_id` mapping), SSE event mapper, widget/dashboard context injector, agent manifest + query routes (`routes.py`), and optional AI-service routes (`ai_service_routes.py`).
- **`agent/api_server.py`**: register `openbb_bridge` routes behind a `try/except` so the core API still starts if `openbb_ai` is not installed (logs `OpenBB Workspace bridge not registered` warning instead of crashing).
- **`agent/requirements.txt`**: add the `openbb_ai` dependency.
- **`agent/src/api/security.py`**: add `https://pro.openbb.co` to `_DEFAULT_CORS_ORIGINS`.
- **Tests**: new `agent/tests/test_openbb_bridge/` covering adapter, context injector, event mapper, and routes.
- **Docs** (`docs/vibe-trading/openbb-agent.md`): added a "部署与运维" section documenting startup (`vibe-trading serve` from `agent/`), `uv pip install` of `openbb_ai` into the uv-created `.venv`, and the CORS gotcha.

## Test Plan

- [x] New tests added (`agent/tests/test_openbb_bridge`: `test_adapter`, `test_context_injector`, `test_event_mapper`, `test_routes`)
- [x] Tested manually:
  - Started the server via `cd Vibe-Trading/agent && vibe-trading serve --port 8000`.
  - `GET /agents.json` returns `200` with the `vibe_trading_agent` manifest.
  - `OPTIONS /agents.json` with `Origin: https://pro.openbb.co` returns `200` + `access-control-allow-origin: https://pro.openbb.co`, resolving the Workspace preflight failure.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) — only `agent/src/api/security.py` and the new `agent/src/openbb_bridge/` were touched.
- [x] No hardcoded values (API keys, file paths, magic numbers) — the added CORS entry is the documented OpenBB Workspace host origin.
- [x] Documentation updated (user-facing: OpenBB Workspace setup/CORS in `docs/vibe-trading/openbb-agent.md`)


warren618:
CI has run on this branch for the first time — the workflow had been sitting unapproved, so the previous "no checks" was no signal rather than a pass. I've approved the queue.

It fails on a security invariant rather than anything incidental:

```
agent/tests/test_security_auth_api.py:482: in test_default_cors_origins_are_loopback_only
E   assert False
```

That test exists to keep the server's *default* CORS origins loopback-only, so a default install cannot be reached cross-origin from an arbitrary site. Whatever this PR adds for the Workspace integration widens that default.

If OpenBB Workspace genuinely needs a non-loopback origin, that's a legitimate need — but it has to be opt-in configuration the operator sets, not a change to the shipped default, and the test should keep asserting the default is loopback-only.

The broader review of this integration is still pending separately; this one is worth fixing regardless of how that lands.

warren618:
Merged with your authorship — thank you for building this against the SDK
properly.

Five security changes were needed first:

- `pro.openbb.co` could not be a default CORS origin. CORS is installed
  globally with credentials and keyless loopback mode admits safe GETs by peer
  address, so scripts on that origin could have read /sessions and /runs. Extra
  origins are now an additive opt-in through VIBE_TRADING_EXTRA_CORS_ORIGINS,
  and the loopback defaults are untouched.
- POST /v1/query now requires auth. It reaches the full agent registry, which
  auto-registers the order tools, so an unauthenticated caller could have driven
  actions inside an already-active mandate.
- The three LLM routes (chat title, dashboard title, prompt enhancement) were
  removed rather than authenticated: they were absent from the manifest and
  duplicated an existing authenticated path.
- Conversation identity could not hash the first message — any two conversations
  opening with the same text mapped onto one persistent session and leaked
  history between them, and later requests' full history was ignored. openbb-ai
  2.1.0 exposes no conversation id and the contract is stateless, so each
  request now runs an ephemeral session replaying the supplied history.
- The SDK is now installed by the dev extra and verified in CI. It landed only
  in agent/requirements.txt while CI installs the extras, so every bridge test
  silently skipped — the feature could look green while being uninstallable.

Workspace context is ingested under a bounded budget now instead of being
dropped, and widget-dashboard-search is advertised as false because the widget
groups carry metadata only and the value round-trip is not implemented.

One operational note for the README/description: a browser POST from a remote
Workspace origin is still refused by the pre-existing cross-site guard,
independently of CORS and of a valid token. Self-hosted same-origin Workspace
and server-side callers work. I pinned that as a test so it is documented
behaviour rather than a surprise.

Issue #456, opened 2026-07-10
[Feature] Portfolio Studio: turnover-aware optimizer + constraints (planning)


Robin1987China:
## Problem or Motivation

The roadmap lists **Portfolio Studio** (Risk x-ray, constraints, turnover-aware optimizer, rebalance notes) as *Planned*.

The backtest engine already has a clean optimizer extension point — `config.optimizer` → `backtest.optimizers.<name>.optimize(ret, pos, dates, **optimizer_params)` (`agent/backtest/engines/base.py`) — with four weighting schemes today: mean-variance, risk parity, equal-volatility, max-diversification.

But none of them account for **turnover cost**. Because weights are recomputed on a rolling window, an optimizer can churn positions every rebalance, so an optimized backtest can overstate net returns once real transaction costs are considered. There is also no way to cap per-name or per-group exposure, and no rebalance/turnover reporting.

## Proposed Solution

Deliver Portfolio Studio bottom-up as small, independently reviewable PRs, backend-first:

1. **Turnover-aware optimizer** — a new `backtest/optimizers/turnover_aware.py` that penalizes weight changes between rebalances (`turnover_penalty` param), reusing the existing `BaseOptimizer` contract. Report realized turnover in the result so cost-adjusted performance is visible.
2. **Constraint layer** — composable weight constraints (per-name cap, group/sector exposure caps, min weight) that can be applied on top of any optimizer.
3. **Rebalance notes + risk x-ray** — per-rebalance turnover / weight-drift detail and a risk-contribution breakdown surfaced in the run output.

Scope notes for maintainers:
- Step 1 follows the existing `optimizer` / `optimizer_params` config contract — no engine API changes, and it stays off by default (current behavior unchanged).
- Backend-only through Step 2; any frontend is optional and confined to Step 3.
- Is this direction welcome, and would you prefer Step 1 to land as its own PR first (same incremental style as the earlier shadow-account steps)?

## Area

Backtest (engines, loaders, metrics)

## Alternatives Considered

- Handling turnover purely as a post-hoc transaction-cost deduction in metrics — but that only measures the drag after the fact; it does not let the optimizer trade off expected return against turnover, so churn is never reduced.
- Adding turnover control inside each existing optimizer — rejected to avoid touching four working modules; a dedicated module keeps the change isolated and opt-in.

## Contribution

I'm willing to implement this.

warren618:
[...]

PR #818, opened 2026-07-23
feat(backtest): composable weight constraints for any optimizer

he-yufeng:
## Summary

- Adds `backtest/constraints.py`: per-name cap, per-name floor, and per-group exposure caps that compose on top of any optimizer, configured via `constraints` in config.json.
- Wired through `_load_optimizer`, so all five weighting schemes can run under caps, not just turnover_aware's built-in bounds. Off by default.

## Why

Portfolio Studio step 2 (#456) asks for a constraint layer that works with any optimizer. Today only turnover_aware has caps, baked into its SLSQP solve; mean-variance, risk parity, equal-volatility, and max-diversification have none, and there is no min-weight floor anywhere.

## Changes

- New module with MaxWeight (clip plus pro-rata redistribution of the excess), MinWeight (lift funded by the names above the floor), and GroupExposure (per-group scale-down). All three preserve signs, never touch inactive names, and validate config the same way turnover_aware does.
- `_load_optimizer` wraps the loaded optimizer with the layer; `constraints` set without `optimizer` warns and skips.
- Infeasible books degrade deterministically: an unsatisfiable cap shrinks gross exposure, an unsatisfiable floor falls back to equal weights for that date.

## Test Plan

- [x] New tests added (23, covering redistribution, infeasible books, sign preservation, config validation, and engine wiring)

## Checklist

- [x] Documentation updated (if user-facing change): the module docstring carries the config schema with an example

### DIFF
--- a/agent/backtest/constraints.py
+++ b/agent/backtest/constraints.py
@@ -1,0 +1,205 @@
+"""Composable weight constraints applied on top of any optimizer.
+
+Portfolio Studio step 2 (#456): per-name cap, per-name floor, and per-group
+exposure caps that work with every optimizer, not only the bounds baked into
+``turnover_aware``. Configured via ``constraints`` in config.json::
+
+    "constraints": [
+        {"type": "max_weight", "cap": 0.25},
+        {"type": "min_weight", "floor": 0.02},
+        {"type": "group_exposure",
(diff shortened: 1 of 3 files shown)


warren618:
Merged — thanks @he-yufeng! Now on `main` via squash commit `cc2c479`. Appreciate the contribution.

PR #716, opened 2026-07-18
feat(data): add historical USD-M funding settlements + maintenance brackets

honginp:
## Stack note

Second slice of the Binance USD-M data-contract stack (#462). Cut fresh from
`upstream/main` as `pr/usdm-funding-brackets` — not rebased from the old
`codex/binance-usdm-data` branch, since that branch's slice-1 commits already
merged in a different shape as #470. Only the funding-settlement idea from
that branch's third commit carries forward here as reference; the code is
re-applied against the post-#470 file. The maintenance-bracket half is new
work not present on that old branch.

## Summary

- Fetch historical funding-rate settlements for `*-USDT-PERP` symbols and
  align them to the required 00:00/08:00/16:00 UTC settlement bars.
- Fetch the current per-symbol maintenance-margin bracket schedule
  (`notional_cap`, `maintenance_rate`, `cumulative_maintenance_amount` per
  bracket) and stamp it with a content-hash version.
- Fail closed when funding settlements are missing/duplicated, or the bracket
  schedule is missing, incomplete, or not strictly increasing by notional cap.

## Why

Follow-up to merged #470 per maintainer guidance: "Still open for a follow-up
PR (1b): historical funding-rate settlement and maintenance-margin brackets.
Same constraints apply — opt-in, fail-closed, spot untouched." This PR only
extends the data contract established by #470; it does not wire funding or
brackets into `CryptoEngine`'s accounting or liquidation path (that's PR 2).

## Changes

`agent/backtest/loaders/ccxt_loader.py`:

- `_fetch_perpetual` now also calls `_fetch_funding_history` and
  `_fetch_maintenance_brackets` for `-PERP` symbols only; spot fetches are
  untouched.
- `_fetch_funding_history`: paginated `fetch_funding_rate_history`, bounded by
  the existing retry/budget helpers. Adds `funding_rate` (0.0 outside
  settlement bars) and `funding_settlement_time` (`NaT` outside settlement
  bars). Fails closed on a missing required settlement or a duplicate
  settlement timestamp.
- `_fetch_maintenance_brackets`: calls `fetch_leverage_tiers([symbol])`,
  reads `notionalCap`/`maintMarginRatio`/`cum`/`bracket` from each tier's raw
  `info` payload. Adds `maintenance_brackets` (JSON-encoded list) and
  `maintenance_bracket_version` (16-hex SHA-256 of the canonical bracket
  JSON) — constant across every row, since CCXT/Binance expose only the
[...]

Issue #462, opened 2026-07-10
[Feature] Exchange-grounded Binance USD-M perpetual backtesting (planning)


honginp:
## Problem or Motivation

Vibe-Trading already has a `CryptoEngine` described as a perpetual-contract engine, but its data and risk boundaries are not exchange-grounded enough for reproducible Binance USD-M research:

- CCXT accepts ambiguous `BTC-USDT` symbols and defaults to spot;
- execution prices and mark prices are not separated;
- funding is a fixed scalar rather than historical settlements;
- liquidation is position-local, close-price based, and uses a simplified OKX-like tier table;
- same-direction target changes do not resize open positions.

The result is a useful simulation skeleton, but not an honest isolated/cross-margin account model.

## Proposed Solution

Deliver three historical-only, independently reviewable PRs:

1. **Binance USD-M data contract**
   - explicit `BTC-USDT-PERP` symbols mapped to `BTC/USDT:USDT`;
   - separate `execution_open` and mark OHLC;
   - historical funding rate + settlement timestamp;
   - versioned per-symbol maintenance brackets;
   - strict timestamp synchronization and cache separation from spot.

2. **Pure margin risk model**
   - reusable `CrossMarginRiskModel` inputs: account state, positions, synchronized mark frames, funding, and bracket configuration;
   - isolated liquidation closes only the breached position and records `position_liquidation`;
   - cross liquidation evaluates shared USDT collateral at account level, closes all positions, records `account_liquidation`, and terminates the run;
   - strict runs emit explicit fidelity flags and never silently fall back to spot close or fixed funding.

3. **Position scaling integration**
   - opt-in same-direction target resizing with weighted entry, proportional realized P&L, and taker fees on delta quantity;
   - supports generic scale-in/scale-out and cascade research without a martingale-specific core abstraction;
   - persists risk snapshots, fee attribution, liquidation events, and fidelity metadata.

Would maintainers welcome PR 1 as the first data-only slice? It can be narrowed further to explicit perpetual symbol routing + execution/mark OHLC if historical funding/bracket ingestion is too large for one review.

## Accounting Boundary

This is a single-asset USDT model with no open-order margin.

Cross mode:

```
margin_balance = wallet_balance + total_unrealized_pnl
initial_margin = sum(position_initial_margin)
maintenance_margin = sum(maintenance_margin_per_position)
available_balance = margin_balance - initial_margin
```

Cross account liquidation triggers when `margin_balance <= maintenance_margin`.

[...]

PR #819, opened 2026-07-23
fix(backtest): consume historical USD-M funding + round settlement jitter

g0rdonL:
## What

Two related fixes that make the v0.1.12 historical USD-M funding feature actually flow into backtest PnL. Follows up on #716 / #462.

### 1. Loader: settlement-timestamp millisecond jitter breaks alignment

Binance returns funding settlement timestamps with millisecond jitter — real examples from `binanceusdm.fetch_funding_rate_history('BTC/USDT:USDT')`:

```
2026-05-05T08:00:00.011Z  -4.626e-05
2026-05-06T00:00:00.010Z  -9.153e-05
2026-05-06T16:00:00.002Z  -3.756e-05
```

`_fetch_funding_history` builds its index from the raw timestamps, so most settlements never intersect the bar index and the missing-settlement guard raises `funding settlement data is missing for BTC-USDT-PERP: <hundreds of slots>` on real fetches. Fix: `.round("s")` on the settlement index before aligning.

### 2. Engine: fetched funding data is never consumed

The loader writes `funding_rate` / `funding_settlement_time` columns into the frame, but `calc_crypto_funding_fee` only ever charges the fixed config rate (default `0.0001`) — the historical data is fetched, validated, aligned, and then ignored. Fix: prefer the bar's historical `funding_rate` when present and finite; fall back to the fixed config rate otherwise, so spot-proxy runs and the daily-fallback path (NaN on non-settlement bars) keep their existing behaviour.

If you'd rather gate the engine change behind an explicit config flag (e.g. `funding_source: historical`), happy to rework — the loader fix stands alone either way.

## Verification

- Live Binance USD-M data, BTC/ETH `-USDT-PERP`, 4H bars, Jan–Jul 2026: before fix 1, the fetch hard-errors; after both fixes, an always-short strategy returns +39.1% with historical funding vs +44.5% with the fixed rate — the ~5pp drag matches 2026's predominantly negative funding regime (shorts pay).
- `pytest tests/test_ccxt_perpetual_loader.py tests/test_crypto_engine.py tests/test_ccxt_loader_bounded.py` → 66 passed
- `pytest -k 'composite or market_hooks or engine'` → 353 passed, 2 skipped
- Not run: full suite, e2e harness, frontend build (unrelated to these paths).

## Tests added

- `test_perpetual_fetch_rounds_funding_settlement_jitter` — jittered settlement aligns to the bar and carries its rate
- `TestHistoricalFundingRate` (3 tests) — historical rate overrides fixed, negative funding pays longs, NaN falls back to fixed

AI-assisted contribution (Claude, supervised); verified against live exchange data as described above. DCO signed.

### DIFF
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -164,7 +164,8 @@ def calc_crypto_funding_fee(
         bar: Current bar data.
         timestamp: Bar timestamp.
         positions: Shared positions dict.
-        funding_rate: Fixed rate per settlement.
+        funding_rate: Fallback fixed rate per settlement, used when the bar
+            carries no historical ``funding_rate`` column.
         applied_set: (symbol, date, hour) dedup set — mutated.
         daily_done_set: (symbol, date) dedup set — mutated.
 
@@ -194,6 +195,12 @@ def calc_crypto_funding_fee(
 
     mark_price = float(bar.get("close", pos.entry_price))
     notional = pos.size * mark_price
+    # Prefer the bar's historical funding rate when the loader supplied one
+    # (USD-M perpetual data via BASE-USDT-PERP); fall back to the fixed
+    # config rate otherwise so spot-proxy runs keep their behaviour.
+    hist = bar.get("funding_rate")
+    if hist is not None and pd.notna(hist):
+        funding_rate = float(hist)
     return notional * funding_rate * pos.direction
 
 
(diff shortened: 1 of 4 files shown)


warren618:
Merged, thanks @g0rdonL! Two real bugs, two surgical fixes. The millisecond-jitter walkthrough with live `fetch_funding_rate_history` examples made the alignment failure obvious, and `.round("s")` is the minimal correct fix. On the engine side, preferring the bar's historical `funding_rate` with a clean NaN fallback keeps spot-proxy behavior untouched — and the $30-vs-$6 charge-difference tests make the intent unmistakable. Fits the #462 stack perfectly; historical funding is now actually consumed instead of fetched-and-ignored.

PR #823, opened 2026-07-24
fix(loaders): reject unsupported Finnhub history intervals

santhreal:
Finnhub always used resolution=D for unsupported intervals like 1H, so intraday runs silently got day bars.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `e8764b7`. Appreciate the contribution.

PR #824, opened 2026-07-24
fix(loaders): reject unsupported Tiingo history intervals

santhreal:
Tiingo always fetched EOD daily bars for unsupported intervals like 1H, so intraday runs silently got day bars.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `fe31272`. Appreciate the contribution.

PR #825, opened 2026-07-24
fix(loaders): reject unsupported Stooq history intervals

santhreal:
Stooq always returned daily CSV bars for unsupported intervals like 1H, so intraday runs silently got day bars.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `c18eb06`. Appreciate the contribution.

PR #826, opened 2026-07-24
fix(loaders): map OKX lowercase 1h/4h instead of silent daily

santhreal:
OKX mapped lowercase 1h/4h to daily instead of hour bars, so intraday runs silently got day bars.

Map 1h/4h like 1H/4H.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `d717911`. Appreciate the contribution.

PR #827, opened 2026-07-24
fix(loaders): treat QVeris 1m/5m/15m/30m as intraday granularity

santhreal:
QVeris left 1m/5m/15m/30m with empty granularity tokens, so the daily path won.

Treat those minute intervals as intraday tokens.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `bd785c9`. Appreciate the contribution.

PR #828, opened 2026-07-24
fix(loaders): reject non-daily intervals for AKShare US/HK/ETF/forex

santhreal:
A-share already raised on unsupported intervals, but US/HK/ETF/forex always called the daily endpoints, so interval=1H returned day bars under a 1H cache key and stopped the fallback chain.

Reject non-daily intervals for those markets (daily aliases like 1d still work).


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `c510432`. Appreciate the contribution.

PR #829, opened 2026-07-24
fix(loaders): accept lowercase 1h/1d in India broker interval map

santhreal:
The India broker interval map listed 1H/1D but not 1h/1d, so lowercase hours fell through.

Add the lowercase aliases next to the existing uppercase keys.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `caed1a2`. Appreciate the contribution.

PR #830, opened 2026-07-24
fix(loaders): accept lowercase 1h/4h/1d in Futu interval map

santhreal:
The Futu loader interval map accepted 1H/4H/1D but not the lowercase forms used elsewhere in the project.

Map 1h/4h/1d like their uppercase siblings.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `afba322`. Appreciate the contribution.

PR #831, opened 2026-07-24
fix(loaders): accept daily aliases in FMP interval gate

santhreal:
The FMP daily gate missed aliases such as 1d/day that other loaders treat as daily.

Accept the same daily aliases before rejecting non-daily intervals.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `2245184`. Appreciate the contribution.

PR #832, opened 2026-07-24
fix(loaders): accept lowercase 1h/4h in local resample rules

santhreal:
Local resample rules keyed 1H/4H only, so lowercase 1h/4h did not resample.

Accept the lowercase hour tokens alongside the uppercase ones.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `59eef3c`. Appreciate the contribution.

PR #833, opened 2026-07-24
fix(loaders): map yfinance lowercase 4h/1w like 4H/1W

santhreal:
Lowercase 4h/1w fell through to .lower() as invalid Yahoo tokens while 4H/1W already mapped to 1h/1wk.

Map 4h→1h and 1w→1wk like the uppercase keys.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `b4b047a`. Appreciate the contribution.

PR #834, opened 2026-07-24
fix(loaders): accept lowercase 1d/1w in Longbridge interval map

santhreal:
The Longbridge loader interval map missed lowercase 1d/1w that the rest of the project uses.

Add those aliases next to the existing daily/week keys.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `1c332e6`. Appreciate the contribution.

PR #835, opened 2026-07-24
fix(loaders): map MT5 1W/1M and reject unknown intervals

santhreal:
Unknown MT5 intervals silently fell back to daily bars, and week/month tokens were missing while the connector already exposes W1/MN1.

Map 1W/1w/1M and return empty instead of rewriting unknowns to daily.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `c8f0ea4`. Appreciate the contribution.

PR #837, opened 2026-07-24
fix(loaders): treat lowercase 1d as Tushare daily bars

santhreal:
Tushare treated lowercase 1d as non-daily while 1D already selected the daily path.

Treat 1d as daily like 1D.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `a1f8997`. Appreciate the contribution.

PR #838, opened 2026-07-24
fix(loaders): accept lowercase 1d/1w in Eastmoney KLT map

santhreal:
The Eastmoney KLT map missed lowercase 1d/1w, so those intervals did not resolve.

Add the lowercase aliases next to the existing keys.


warren618:
Merged — thanks @santhreal! Now on `main` via squash commit `482248d`. Appreciate the contribution.

PR #840, opened 2026-07-24
docs: fix broken Contributing nav anchor in README

dvirarad:
The header navigation links "Contributing" to `#-contributing`, but the `## Contributing` heading has no leading emoji, so GitHub generates its anchor as `#contributing` (no leading hyphen). As a result, clicking "Contributing" in the top nav doesn't scroll to the section.

Every other nav entry points at an emoji heading (e.g. `## 📰 News` → `#-news`), so their leading-hyphen anchors are correct. `README_ja.md` already uses the correct `#contributing`; this aligns `README.md` and `README_zh.md` with it.

Docs-only, no functional changes.


warren618:
Merged — thanks @dvirarad! Now on `main` via squash commit `4103b79`. Appreciate the contribution.

PR #841, opened 2026-07-24
fix(cli): forward universe meta into alpha bench JSON/HTML output

### LINKED ISSUE #797: CLI JSON envelope and HTML report drop _meta.survivorship_bias set by the sp500 loader
**Summary:** the sp500 loader sets `_meta.survivorship_bias` precisely so downstream surfaces can warn users that current-constituent backtests overstate performance — but the two surfaces most users actually read (CLI JSON envelope, HTML report) both drop it. Split out from #773 at maintainer uptake, per my comment there (verified on the installed 0.1.12 package and current `main` as of 2026-07-23).

**The wiring today:**

- `src/data/sp500_loader.py` sets `_meta.survivorship_bias` on the frame; `run_bench` forwards it as `meta` in its result dict — this part works.
- `src/factors/cli_handlers.py`: the `alpha bench` JSON envelope enumerates its output keys explicitly and `meta` is not among them — the flag dies here for CLI users.
- The HTML report context omits it as well, so the report renders a leaderboard with no survivorship warning.
- Credit where due: the api_server SSE path (`_result_for_wire`) **does** keep `meta`, so frontend users can see the warning — it's specifically the CLI and report surfaces that lose it.

**Why it matters:** survivorship bias is not a cosmetic caveat on sp500 backtests — it's often the whole edge. A user running `alpha bench` from the CLI today gets confident-looking IC numbers with the one attached health warning silently stripped between computation and display.

**Suggested minimal fix:** include `meta` in the CLI envelope (additive, no key removals, so existing consumers are unaffected) and surface `survivorship_bias` in the report header when present. Happy to open a PR if useful — it should compose cleanly with the new `--strict` envelope from #796.

AmirF194:
## Summary

- `vibe-trading alpha bench` silently drops the sp500 loader's `survivorship_bias`
  disclosure from both its JSON envelope and its HTML report, even though the
  web UI's SSE path already keeps it.

## Why

`bench_runner.run_bench` forwards the sp500 loader's universe metadata as
`result["meta"]` (`bench_runner.py:319-347`), and `alpha_routes._result_for_wire`
explicitly keeps `"meta"` in the keys it forwards to the frontend
(`alpha_routes.py:757-777`, comment: "so users see degraded runs clearly"). But
`cmd_alpha_bench` builds its own JSON envelope and HTML report context from
hand-enumerated dict literals (`cli_handlers.py:676-701`), and `"meta"` was not
among the keys in either one. Anyone running `vibe-trading alpha bench --universe
sp500` from the CLI, or reading the generated HTML report, gets no
survivorship-bias signal at all, while the same run through the web UI does.

Fixes #797.

## Changes

- `cli_handlers.py`: read `result.get("meta")` once and forward it into both the
  HTML report context and the JSON envelope when present, the same opaque
  passthrough the wire path already uses.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`,
      6299 passed, 13 skipped; the one unrelated pre-existing failure,
      `test_redaction.py::test_no_over_redaction_of_external_or_caller_paths`,
      reproduces identically on unmodified `main` and is untouched by this diff)
- [x] New tests added: `agent/tests/test_alpha_bench_meta_survivorship.py`, 3 cases
      (envelope carries `meta` when present, HTML context carries `meta` when
      present, both stay free of the key when the loader supplies none). All three
      fail on `main` (`KeyError: 'meta'` for the first two) and pass on this branch.
- [x] Tested manually: drove `cmd_alpha_bench` directly (the real registered
      handler) in a clean `python:3.11-slim` Docker container against fresh
      `origin/main` HEAD, with a fake `run_bench` returning a `meta` payload;
      confirmed the key was missing pre-fix and present post-fix in both outputs.
- [x] `bash tools/ci_grep_gates.sh`, `pytest tools/test_ci_env_var_gate.py`, and
      the `compileall`/`py_compile` syntax checks all pass.
- [x] Not touched: frontend build and `vitest run` (35 files, 310 tests) pass
      unmodified; this diff has no frontend-visible surface, verified only as a
      completeness check since neither file changed.

## Checklist

- [ ] Documentation updated (not user-facing beyond the bug itself; no README/CLI
      help text describes this field today)

### DIFF
--- a/agent/src/factors/cli_handlers.py
+++ b/agent/src/factors/cli_handlers.py
@@ -659,6 +659,10 @@ def cmd_alpha_bench(args: argparse.Namespace) -> int:
 
         top_rows = [_normalise_row(r) for r in top_rows_raw]
         failures_for_report = [_normalise_skipped(s) for s in skipped[:10]]
+        # Universe metadata (e.g. the sp500 loader's survivorship_bias flag),
+        # forwarded by bench_runner.run_bench as result["meta"] and already
+        # kept by alpha_routes._result_for_wire for the SSE/frontend path.
+        universe_meta = result.get("meta")
 
         report_path: Path | None = None
         try:
@@ -684,6 +688,8 @@ def cmd_alpha_bench(args: argparse.Namespace) -> int:
                 "top": top_rows,
                 "failures": failures_for_report,
             }
+            if universe_meta:
+                context["meta"] = universe_meta
             report_path.write_text(_render_html(context), encoding="utf-8")
         except Exception as exc:  # noqa: BLE001 — report is nice-to-have
             _err(f"warning: could not write HTML report: {exc}")
@@ -706,6 +712,8 @@ def cmd_alpha_bench(args: argparse.Namespace) -> int:
                     envelope[key] = result[key]
             if result.get("oos_split") is not None:
                 envelope["oos_split"] = result["oos_split"]
+        if universe_meta:
+            envelope["meta"] = universe_meta
         if report_path is not None:
             envelope["report_path"] = str(report_path)
         print(json.dumps(envelope, indent=2, default=str))
--- a/agent/tests/test_alpha_bench_meta_survivorship.py
+++ b/agent/tests/test_alpha_bench_meta_survivorship.py
@@ -1,0 +1,138 @@
+"""``alpha bench`` must forward result["meta"] (issue #797).
+
+bench_runner.run_bench forwards the sp500 loader's survivorship_bias flag as
+result["meta"], and alpha_routes._result_for_wire already keeps it for the
+SSE/frontend path. cmd_alpha_bench's own JSON envelope and HTML report
+context were built from hand-enumerated dict literals that never included
+that key, so the CLI/HTML path silently dropped the disclosure.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+
+import pytest
+
+from src.factors import cli_handlers
+
+
+def _ns(**overrides):
+    base = dict(
+        zoo="alpha101",
+        universe="sp500",
+        period="2020-2025",
+        top=20,
+        yes=True,
+        strict=False,
+        oos_split=None,
+        random_seeds=5,
+    )
+    base.update(overrides)
+    return argparse.Namespace(**base)
+
+
+def _result(**overrides):
+    base = {
+        "status": "ok",
+        "alive": 1,
+        "rows": [
+            {
+                "id": "alpha001",
+                "ic_mean": 0.05,
+                "ic_std": 0.01,
+                "ir": 0.9,
+                "ic_positive_ratio": 0.7,
+                "ic_count": 100,
+                "theme": ["momentum"],
+                "formula_latex": "x",
+                "_category": "alive",
+            }
+        ],
+        "skipped": [],
+        "wall_seconds": 1.0,
+    }
+    base.update(overrides)
+    return base
+
+
+_UNIVERSE_META = {
+    "universe": "sp500",
+    "survivorship_bias": True,
+    "constituent_source": "current wiki S&P 500 list",
+    "constituent_source_date": "2026-01-01",
+    "constituent_count": 500,
+}
+
+
+class _FakeReg:
+    def list(self, zoo=None):
+        return ["alpha001"]
+
+    def get(self, aid):
+        class _Entry:
+            zoo = "alpha101"
+            meta = {"theme": ["momentum"], "formula_latex": "x"}
+
+        return _Entry()
+
+
+def _envelope(out: str) -> dict:
+    start = out.find("\n{")
+    if start < 0:
+        start = out.find("{") if out.lstrip().startswith("{") else -1
+        if start < 0:
+            raise AssertionError(f"no JSON envelope in stdout: {out[:200]!r}")
+    else:
+        start += 1
+    return json.loads(out[start:])
+
+
+@pytest.fixture()
+def _reg(monkeypatch):
+    monkeypatch.setattr(cli_handlers, "Registry", _FakeReg)
+
+
+@pytest.fixture()
+def _captured_html_context(monkeypatch):
+    """Capture the context dict handed to the HTML renderer, write nothing."""
+    import src.tools.alpha_bench_tool as tool
+
+    captured = {}
+
+    def fake_render_html(context):
+        captured.update(context)
+        return "<html></html>"
+
+    monkeypatch.setattr(tool, "_render_html", fake_render_html)
+    return captured
+
+
+def _run(monkeypatch, capsys, args, result):
+    import src.factors.bench_runner as legacy_mod
+
+    monkeypatch.setattr(legacy_mod, "run_bench", lambda **kw: result)
+    rc = cli_handlers.cmd_alpha_bench(args)
+    return rc, capsys.readouterr()
+
+
+def test_envelope_includes_universe_meta_when_present(capsys, monkeypatch, _reg, _captured_html_context):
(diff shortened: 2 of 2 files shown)


AmirF194:
Noticed @bonyohana's comment on #797 only after opening this, not before, so flagging the overlap rather than leaving it silent.

The root cause here is the same one traced there (independently, same file and line ranges), and this PR is the narrower half of the split proposed in that comment: forward `result["meta"]` into both the JSON envelope and the HTML report context, without touching `alpha_bench_tool.py`'s render templates (its own docstring in `cmd_alpha_bench` says that module is intentionally left unmodified, which is why the HTML banner itself is out of scope here too).

Happy to close this in favor of a PR from @bonyohana if that is preferred; posting so the two do not collide.

bonyohana:
Confirming from the #797 side: this matches the trace exactly (independently reached, same file/line ranges — `bench_runner.run_bench` builds `result["meta"]`, `_result_for_wire` keeps it for SSE, `cmd_alpha_bench`'s hand-enumerated envelope and `context` literals drop it). Two notes, both supportive:

1. The JSON-envelope half of this PR fully closes the machine-readable gap — CLI consumers get the same `meta.survivorship_bias` disclosure API consumers already see. That's the higher-stakes half, since scripted pipelines are the ones most likely to silently ingest inflated alpha.
2. On the HTML half: forwarding `context["meta"]` is the right precursor, but as noted in #797 the report template doesn't currently render a `meta` block, so the human-visible warning stays absent until the template gains one. Keeping that out of scope here is reasonable (it touches the module you keep unmodified) — just flagging so the issue isn't closed as fully-fixed on merge; the template touch can ride a follow-up.

Nice test coverage — the envelope and context assertions pin both halves independently. Happy to see this land; it supersedes the PR I'd offered in #797.

Robin1987China:
Good work — independently traced the same root cause. Closed our #844 in favor of this PR (yours arrived first, with tests).

One gap bonyohana flagged: the HTML report context has `meta` but the template `_render_html_manual` in `alpha_bench_tool.py` doesn't render a warning. Since that falls outside the scope you intentionally kept, here's the minimal diff to add it:

```patch
--- a/agent/src/tools/alpha_bench_tool.py
+++ b/agent/src/tools/alpha_bench_tool.py
@@ -693,6 +693,13 @@
         f" &middot; {int(ctx['n_alphas_tested'])} tested, {int(ctx['n_skipped'])} skipped",
         "</div>",
     ]
+    meta = ctx.get("meta") or {}
+    if meta.get("survivorship_bias"):
+        parts.append(
+            "<div class=\"warning\">&#9888; Survivorship-biased universe: "
+            "constituents are the current index members, not the point-in-time "
+            "roster. IC statistics may be overstated."
+            "</div>"
+        )
```

Happy to submit this as a follow-up PR after this lands, or feel free to fold it in directly. Either way, the machine-readable half (JSON envelope) is the critical one and this PR nails it.

bonyohana:
@Robin1987China thanks for closing #844 in favour of this — and for picking up the template half rather than letting it drift. I checked your patch against `main` @ `8643fcd` and rendered both paths locally before replying; it needs two corrections, one of which is the same failure mode as the bug this PR fixes.

**1. It patches the fallback renderer, not the live one.**

`_render_html` (`agent/src/tools/alpha_bench_tool.py:663`) is:

```python
try:
    from jinja2 import Environment, select_autoescape
    env = Environment(autoescape=select_autoescape(["html", "xml"]))
    return env.from_string(_JINJA_TEMPLATE).render(**context)
except ImportError:
    return _render_html_manual(context)
```

`jinja2` is a hard declared dependency — `pyproject.toml:57` (`"jinja2>=3.1.0"`), pinned at `jinja2==3.1.6` in `requirements-lock.txt`. So in any environment installed per the project's own metadata, `_render_html_manual` is unreachable and `_JINJA_TEMPLATE` (line 610) is what actually renders. Patching only the manual path leaves the warning invisible in normal operation — structurally the same bug as the one being fixed here: a second, hand-maintained rendering path drifting out of sync with the canonical one. Both need the block, or the divergence just moves down a layer.

**2. There is no `.warning` CSS class.** `_REPORT_CSS` (lines 598–608) defines exactly `body`, `h1, h2`, `table`, `th, td`, `th`, `.meta`, `.formula`, `.skipped`. A `<div class="warning">` renders as unstyled body text — for a disclosure whose entire purpose is human visibility, that's most of the value gone. `.skipped { color: #a33; font-size: .9em; }` is the in-house precedent for an alert style.

**3. Minor: the hunk won't apply as posted.** At HEAD the `parts` literal doesn't close after `"</div>",` — lines 693–695 add the `<h2>` and table header before the `]` at 696, so appending after the literal would emit the div after `<table>`'s header row.

Version I actually rendered and checked (warning present when flagged; suppressed when `meta` is `{}` *or* absent entirely; autoescape holds against a hostile `constituent_source_date`):

```diff
--- a/agent/src/tools/alpha_bench_tool.py
+++ b/agent/src/tools/alpha_bench_tool.py
@@ _REPORT_CSS
 .skipped { color: #a33; font-size: .9em; }
+.warning { color: #8a4b00; background: #fff6e5; border-left: 3px solid #d98200;
+           padding: .6em .8em; margin-bottom: 1.5em; font-size: .95em; }
@@ _JINJA_TEMPLATE — after the meta div closes
 </div>
+{% if meta and meta.survivorship_bias %}
+<div class="warning">&#9888; Survivorship-biased universe: constituents are the
+current index members{% if meta.constituent_source_date %} (sampled
+{{ meta.constituent_source_date }}){% endif %}, not the point-in-time roster.
+IC statistics may be overstated.</div>
+{% endif %}
@@ _render_html_manual — after the "</div>", entry, before the <h2> entry
         "</div>",
+        *_survivorship_block(ctx),
         f"<h2>Top {len(ctx['top'])} by IR</h2><table>",
```

with the shared helper:

```python
def _survivorship_block(ctx: dict[str, Any]) -> list[str]:
    meta = ctx.get("meta") or {}
    if not meta.get("survivorship_bias"):
        return []
    when = meta.get("constituent_source_date")
    return [
        "<div class=\"warning\">&#9888; Survivorship-biased universe: "
        "constituents are the current index members"
        + (f" (sampled {_esc(when)})" if when else "")
        + ", not the point-in-time roster. IC statistics may be overstated."
        "</div>"
    ]
```

Using `constituent_source_date` is worth it — the meta blob already carries it (and `constituent_count`), so the warning can name the actual sample date rather than generic prose, which is the difference between a reader shrugging and a reader checking their window.

**Separately, a larger hole in the same family — filing as its own issue rather than widening this PR:** `_load_csi300_panel` (line 303) never sets `panel["_meta"]` at all, and it has a documented degradation path — when `pro.index_weight` fails it falls back to `_CSI300_FALLBACK_CODES`, a **30-name** static blue-chip list standing in for a 300-name index (line 342), announced only by `logger.warning(... "degraded run")` in the server log. That is a strictly more severe survivorship situation than the sp500 case, and it is disclosed through *no* surface — not the API, not the CLI JSON envelope this PR fixes, not the HTML report. Out of scope here; noting it so survivorship disclosure isn't treated as done once this merges.

The JSON-envelope half remains the critical one and this PR still nails it — none of the above blocks the merge as scoped.

warren618:
Merged, thanks @AmirF194! Faithful implementation of the #797 fix — forwarding `result["meta"]` into both the JSON envelope and the HTML report context, mirroring the API's `_result_for_wire` contract. The tests pinning both the "meta forwarded when present" and "no stray key when the loader has none" (csi300) cases are exactly the right coverage. This also closed #797 automatically — thanks for the careful work!

AmirF194:
Thanks for the careful merge, and for taking the time on that thread with bonyohana and Robin1987China. Good to see it converge cleanly with #797.

PR #842, opened 2026-07-25
fix(mt5): convert numpy scalars to native types in historical bars

### LINKED ISSUE #774: MT5 connector: trading_history fails on every symbol with "Object of type int64 is not JSON serializable"
# MT5 connector: `trading_history` fails on every symbol with "Object of type int64 is not JSON serializable"

## Environment

- Installed from `main` branch, editable install (`pip install -e ".[mt5,anthropic]"`), commit checked out ~2026-07-22.
- OS: Windows 10 (Git Bash / MINGW64).
- Connector profile: `mt5-paper-sdk` (paper/demo).
- Broker: OANDA TMS (server `OANDATMS-MT5`), MT5 demo account, `trade_mode=0` confirmed via `get_account`.

## Summary

`trading_history` on the `mt5` connector fails for **every** symbol/period/limit combination tried, including `EURUSD` (a universally listed instrument), not just broker-specific symbols. The account/quote/position reads all work correctly — only historical bar retrieval is affected.

## Steps to reproduce

1. Configure `~/.vibe-trading/mt5.json` with valid `login`/`password`/`server` for an MT5 demo account, `"profile": "paper"`.
2. `vibe-trading connector use mt5-paper-sdk`
3. Confirm the connection works:
   ```
   python -c "from src.trading.service import get_account; import json; print(json.dumps(get_account('mt5-paper-sdk'), indent=2, default=str))"
   ```
   → returns `"status": "ok"` with correct account data.
4. Ask the agent (or call the underlying tool directly) to pull historical bars for `EURUSD` via `trading_history` on the `mt5-paper-sdk` profile, using either the IBKR-style `duration`/`bar_size` params or the SDK-style `period`/`limit` params.

## Expected behavior

Historical OHLCV bars are returned as JSON, same as `get_account` / `get_positions` / `get_quote` already do successfully.

## Actual behavior

Every call fails with:
```
Object of type int64 is not JSON serializable
```
This happens **connector-wide**, regardless of which symbol, period, or limit combination is used — including `EURUSD`, ruling out a symbol-resolution or data-availability problem. The likely root cause is a `numpy.int64` field (almost certainly tick volume) returned by the `MetaTrader5` SDK's `copy_rates_*` call not being cast to a native Python `int`/`float` before the response is JSON-serialized — the same class of bug already fixed for other connectors per the changelog (e.g. "non-finite values ... normalized before writing artifacts/validation.json").

## Suggested fix

Cast numpy scalar types (`int64`, `float64`, etc.) to native Python types when building the bar payload in `agent/src/trading/connectors/mt5/reads.py`, the same pattern already used elsewhere in the codebase for JSON-safety.

## Secondary observation (less certain, included for awareness)

While working around the above bug by requesting a two-instrument comparison (WTI crude + platinum proxies via yfinance/stooq) in the same chat session, the agent entered what looked like a repetitive tool-calling loop — repeatedly re-fetching the same run card / trade log data for both instruments without visible forward progress ("Now let me get the WTI trades... Now let me get the platinum trades..." repeated many times) until the run was manually cancelled. I don't have a clean minimal repro for this one, but flagging it in case it's a known failure mode (possibly related to the "reached max iterations masking empty model responses" class of issues mentioned in past changelog entries) worth a look if others report similar stalls on multi-instrument comparison prompts.

Robin1987China:
MT5 SDK copy_rates_from_pos returns numpy int64/float64 values (tick_volume, time, etc.) that are not JSON-serializable, causing every trading_history call to fail with "Object of type int64 is not JSON serializable". Add a `_native` helper that unwraps numpy scalars via `.item()` before building the bars payload.

Closes #774

### DIFF
--- a/agent/src/trading/connectors/mt5/reads.py
+++ b/agent/src/trading/connectors/mt5/reads.py
@@ -259,6 +259,16 @@ def get_historical_bars(
     )
 
 
+def _native(value: Any) -> Any:
+    """Unwrap a numpy scalar to a native Python type, else return as-is."""
+    if hasattr(value, "item"):
+        try:
+            return value.item()
+        except (TypeError, ValueError):
+            pass
+    return value
+
+
 def _rate_to_dict(rate: Any) -> dict[str, Any]:
     """Map one rates row (numpy structured row, mapping, or object) defensively."""
 
@@ -268,14 +278,14 @@ def _rate_to_dict(rate: Any) -> dict[str, Any]:
         except (TypeError, KeyError, IndexError, ValueError):
             return getattr(rate, key, None)
 
-    volume = _get("tick_volume")
+    volume = _native(_get("tick_volume"))
     return {
-        "time": _get("time"),
-        "open": _get("open"),
-        "high": _get("high"),
-        "low": _get("low"),
-        "close": _get("close"),
+        "time": _native(_get("time")),
+        "open": _native(_get("open")),
+        "high": _native(_get("high")),
+        "low": _native(_get("low")),
+        "close": _native(_get("close")),
         "volume": volume,
-        "spread": _get("spread"),
-        "real_volume": _get("real_volume"),
+        "spread": _native(_get("spread")),
+        "real_volume": _native(_get("real_volume")),
     }


Robin1987China:
Already fixed upstream by shadowinlife in commit 76d8867 (2026-07-22) — _int_or_none / _float_or_none helpers achieve the same numpy→native conversion. Closing as redundant. Closed #774.

Issue #735, opened 2026-07-20
Longbridge connector: 'connector positions' renders empty Qty/Avg Cost and 'connector account' shows no summary — CLI renderers read IBKR-style keys


Wenhui-King:
## Summary

With the Longbridge broker-SDK connector, `vibe-trading connector positions` and `vibe-trading connector account` render mostly-empty output even though the connector's data layer returns complete data. The CLI table renderers read IBKR-style dict keys (`position`, `avg_cost`, `sec_type`, `accounts`/`summary`), while the Longbridge connector returns its own schema (`quantity`, `cost_price`, `balances`, …). The service layer passes connector dicts through unadapted, so every non-matching key silently renders as an empty cell.

## Environment

- vibe-trading-ai 0.1.11 (pip install); same code is present on current `main` (67a393e)
- longbridge SDK 4.3.3
- Python 3.13, macOS
- Connector profile: `longbridge-paper-trade` (Longbridge paper trading account, `broker_sdk` transport)

## Reproduction

1. Configure a Longbridge connector profile (`~/.vibe-trading/longbridge.json`) and select it via `vibe-trading connector use`.
2. Hold at least one position in the account.

**Bug 1 — positions table renders Qty / Avg Cost (and Account / Type) as empty:**

```
$ vibe-trading connector positions

           Positions · longbridge-paper-trade
  Account   Symbol     Type   Qty   Avg Cost   Currency
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
            AAPL.US                            USD
            XXXX.HK                            HKD
            ...
```

But the data layer is fine — calling the connector directly returns complete rows:

```python
from src.trading.service import get_positions
r = get_positions()
# r["positions"][0] == {
#   "symbol": "AAPL.US", "symbol_name": "...",
#   "quantity": 20.0, "available_quantity": 20.0,
#   "cost_price": 3XX.XX, "currency": "USD", "market": "US",
# }
```

**Bug 2 — account summary claims nothing was returned:**

```
$ vibe-trading connector account
Accounts: (none)
No account summary returned.
```

But `get_account_snapshot()` returns full balances:

```python
from src.trading.service import get_account
r = get_account()
# r == {"status": "ok", "profile_id": "longbridge-paper-trade", ...,
#       "balances": [{"currency": "USD", "total_cash": ..., "net_assets": ...,
[...]

PR #843, opened 2026-07-25
fix(cli): show a meaningful account label in the broker_sdk balances view

### LINKED ISSUE #846: CLI shows "Accounts: (none)" in broker_sdk balances view
### Observed

When a broker_sdk connector (Longbridge, etc.) returns an account snapshot without an `accounts` key — only `balances` — the CLI renderer's broker_sdk fast-path prints:

```
Accounts: (none)
```

Expected: show the profile_id or a meaningful label instead.

### Root cause

`_print_connector_account` builds the account label from `result.get("accounts", [])` at line 4019. The broker_sdk fast-path added in #726 (for #735) renders the `balances` table correctly, but the header label was never wired to draw from `profile_id` — it falls through to the `"(none)""` default.

### Fix plan

When hitting the `balances` fast-path with an empty accounts list, fall back to `result.get("profile_id")` — which `_with_profile()` (service.py:426) always injects into every account result. One line change in `_legacy.py`.

Robin1987China:
### Problem

The `_print_connector_account` CLI renderer has a broker_sdk fast-path for connectors
that return `balances` instead of the IBKR-style `summary` rows (introduced in #726).
When that path is taken, the header always prints

```
Accounts: (none)
```

because `accounts` is derived from `result.get("accounts", [])` — a key that
Longbridge and other broker_sdk connectors do not include in their response.

### Fix

When `accounts` is empty and the renderer falls into the `balances` fast-path, fall
back to `result.get("profile_id")` — which `_with_profile()` (service.py:426)
always injects into every account result — so the label is informative:

```
Accounts: longbridge-paper-trade
```

### Why a separate follow-up (not part of #735)

#726 fixed the core rendering bug (#735): broker_sdk schemas were not decoded at
all. This is a cosmetic gap left behind: the header label that was always `(none)`
in the only path the fix introduced.

### Manual verification

- Ran `agent/tests/test_cli_connector_renderers.py` — 4/4 pass, including the
  balances-table test whose code path is the one being changed.
- Traced the full call chain: `get_account_snapshot()` → `_with_profile()` →
  `_print_connector_account()` — `profile_id` is guaranteed to be present.

Closes #846

### DIFF
--- a/agent/cli/_legacy.py
+++ b/agent/cli/_legacy.py
@@ -4021,6 +4021,8 @@ def _print_connector_account(result: dict[str, Any]) -> int:
     # broker_sdk connectors (Longbridge, …) return a ``balances`` list instead of
     # IBKR-style ``summary`` tag/value rows; render that when present (#735).
     if not rows and result.get("balances"):
+        label = accounts if accounts != "(none)" else result.get("profile_id", result.get("profile", "unknown"))
+        console.print(f"Accounts: [cyan]{rich_escape(str(label))}[/cyan]")
         return _print_connector_balances(result)
     if not rows:
         # Not the broker_sdk flat shape — try the remote-MCP nested shape.


warren618:
Merged, thanks @Robin1987China! Small change, real UX difference — showing the profile id instead of a bare `(none)` makes the broker_sdk balances view much clearer. This also closed #846 automatically. Appreciate the fix!

PR #844, opened 2026-07-25
fix(cli): forward survivorship_bias meta to JSON envelope and HTML report

### LINKED ISSUE #797: CLI JSON envelope and HTML report drop _meta.survivorship_bias set by the sp500 loader

Robin1987China:
The sp500 loader sets `_meta.survivorship_bias` on the panel so downstream surfaces can warn users about inflated IC stats, but the CLI JSON envelope and HTML report both dropped it — only the API SSE path preserved it. Include meta in both the envelope and report context, and render a warning banner in the HTML when survivorship_bias is true.

Closes #797

### DIFF
--- a/agent/src/factors/cli_handlers.py
+++ b/agent/src/factors/cli_handlers.py
@@ -662,6 +662,7 @@ def cmd_alpha_bench(args: argparse.Namespace) -> int:
                 "n_skipped": len(skipped),
                 "top": top_rows,
                 "failures": failures_for_report,
+                "meta": result.get("meta"),
             }
             report_path.write_text(_render_html(context), encoding="utf-8")
         except Exception as exc:  # noqa: BLE001 — report is nice-to-have
@@ -676,6 +677,7 @@ def cmd_alpha_bench(args: argparse.Namespace) -> int:
             "n_alphas_tested": len(rows),
             "n_skipped": len(skipped),
             "top": top_rows,
+            "meta": result.get("meta"),
             "wall_seconds": result.get("wall_seconds"),
         }
         if report_path is not None:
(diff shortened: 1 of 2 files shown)


Robin1987China:
Closing in favor of #841 which arrived first — same root fix for the JSON envelope and HTML context. Our diff also touched alpha_bench_tool.py to render a survivorship-bias warning banner in the HTML report; offering that as a standalone supplement in a comment on #841.

PR #839, opened 2026-07-24
feat: add Phase 01 Taiwan offline quant data foundation

zyllzl:
## Summary

- Add offline Taiwan DuckDB, CSV, Parquet, and snapshot data foundation
- Add TWSE/TPEX symbol parsing
- Add snapshot-only Taiwan loader routing
- Add data integrity verification and tests

## Scope

- Phase 01 only
- No FinMind, FinLab, MCP, provider integration, or live trading
- Taiwan execution engine remains intentionally unavailable

## Validation

Phase 01 and related tests: 197 passed
src/tw_quant coverage: 92%
Ruff: passed
Python syntax check: passed

PR #848, opened 2026-07-25
feat(tools): add read-only Taiwan stock snapshot integration

TSENGCHIENFENG:
## Summary

- restore a Docker-compatible Python dependency lock
- add a read-only Taiwan stock snapshot tool
- mount published Taiwan stock snapshots into the service as read-only
- support status, lookup, latest, history, and universe queries
- add unit tests for the Taiwan stock data tool

## Validation

- Docker image build passed
- dependency lock dry-run validation passed
- Taiwan stock tool unit tests: 7 passed
- Docker Compose service is healthy
- snapshot database integrity check: ok
- tool auto-registration verified
- published snapshot is mounted read-only


warren618:
CI has run on this branch for the first time — the workflow had been sitting unapproved, so "no checks" was no signal rather than a pass. I've approved the queue.

It fails the repository safety gates, not the test suite:

```
[gate e] no raw os.getenv / os.environ reads outside config layer ... FAIL
  agent/src/tools/taiwan_stock_data_tool.py:135: os.getenv() — use src.config accessor
  agent/src/tools/taiwan_stock_data_tool.py:145: os.getenv() — use src.config accessor
```

Environment reads go through the accessor in `agent/src/config/` rather than `os.getenv` directly, so that configuration has one place it can be read, validated and overridden — the Web Settings path and the `~/.vibe-trading/.env` migration both depend on that. Routing those two reads through the accessor should clear it.

Note this is separate from the architecture question on this PR: #839 covers overlapping Taiwan ground, and I still need to settle which approach to take before either can merge.

TSENGCHIENFENG:
Thanks for approving the workflow and for identifying the gate failure.

I’ll route the two environment reads in taiwan_stock_data_tool.py through the existing config accessor, keep the change minimal, and rerun the repository checks.

I understand that the architectural overlap with #839 is a separate decision, so I won’t expand the scope while that direction is being evaluated.

取得 Android 版 Outlook<https://aka.ms/AAb9ysg>
________________________________

warren618:
Merged with your authorship — thank you for the read-only design and the
`immutable=1` connection.

Four maintainer changes on top, all worth knowing about:

- Bulk responses are budgeted to 9,500 serialized characters. The agent loop
  truncates tool results at 10,000 characters, and at the documented maximum
  (200 rows x 50 stocks) the response was about 5.4 MB, so the prefix cut
  produced invalid JSON. Rows are now retained newest-first, round-robin across
  stocks, and responses carry `total_rows` / `returned_rows` / `truncated`.
- Indicators are no longer mis-dated: the newest price row and newest feature
  row were selected independently but reported under a single date, so
  07-23 indicators could be served as 07-24. Both dates are now returned with
  a `stale_features` flag.
- The Compose default host path moved outside the checkout. `./data/tw-stock`
  is inside the working tree and wasn't gitignored, and no market data may land
  in the repo; the default is now under `$HOME/.vibe-trading/`, overridable.
- The snapshot schema is validated on open, so an unrelated or stale SQLite
  file reports an actionable error instead of registering successfully and
  failing mid-query.

The `requirements-lock.txt` rewrite was dropped — it carried 21 unrelated pin
changes, including FastAPI and OpenAI downgrades, and the tool is stdlib-only.

PR #857, opened 2026-07-26
fix(providers): omit temperature for Claude Opus 4.7

### LINKED ISSUE #856: [Bug] Anthropic models after sonnet 4.6 are not accepted
### Description

What happened:
I tried to change my LANGCHAIN_MODEL_NAME to Opus 4.7, however I was informed via Claude that , Opus 4.7+ stays blocked by the package's hardcoded temperature parameter — a crisp upstream issue ("Anthropic provider sends temperature, rejected by claude-opus-4-7+") would likely get the same treatment their MiniMax temperature bug got, and then the model becomes a one-line env change.

LANGCHAIN_MODEL_NAME=claude-opus-4-7, any swarm → every worker fails at iteration 0 with BadRequestError 400: 'temperature' is deprecated for this model (request IDs available, e.g. req_011CdPoYamzMjp4HjU8sJXSn).
Cause: _build_anthropic (src/providers/llm.py:451) passes temperature=temperature unconditionally; LANGCHAIN_TEMPERATURE is float (env_schema.py:129) so None/omission is unreachable by config.
Suggested fix: same per-provider pattern as the MiniMax clamp (llm.py:727) and Kimi force (llm.py:730): pass temperature=None (omitting it) for Anthropic models that deprecate the parameter.
Workaround: claude-sonnet-4-6 works with identical config.

What I expect::
I want to use the most up-to-date anthropic models to run_swarm and other commands but I am relegated to Sonnet 4.6 at the most. 

### Steps to Reproduce

Run run_swarm on JNJ in Claude
Authenticate my Claude account
Receive error that the command can not be completed due to the hard-coded temperature requirement for the model configuration

### Error Logs
The bug, precisely located. src/providers/llm.py:451 — _build_anthropic() — passes temperature unconditionally:

```shell

python
return chat_anthropic(
    model=model,
    max_tokens=...,
    temperature=temperature,   # ← always sent; opus-4-7+ rejects the parameter
    ...
)

```
The underlying ChatAnthropic class can omit the parameter (it does when given None), but the package's float-typed schema can never produce None — so every Anthropic call carries an explicit temperature, and newer Claude models 400 it.

### Interface

MCP Plugin

### LLM Provider

anthropic

### Version

0.1.12

### Environment

Ubuntu/3.12.13

Solaris-star:
## Summary

- Omit the `temperature` request key for Claude Opus 4.7 model IDs.
- Preserve existing temperature behavior for Sonnet, older Opus releases, and unreported future Opus versions.
- Add provider-factory regressions for canonical, dated, and provider-qualified model IDs.

## Why

Closes #856.

`_build_anthropic()` currently forwards `temperature` unconditionally. Claude Opus 4.7 rejects requests containing this deprecated parameter, causing every worker to fail before its first iteration.

This draft is intentionally narrow because `src/providers` is protected by the contribution guide. It is open for maintainer confirmation of the model matching boundary before being marked ready.

## Changes

- Build Anthropic kwargs before constructing `ChatAnthropic`.
- Leave out `temperature` only when the model ID identifies Claude Opus 4.7.
- Keep the original parameter for existing unaffected model families.

## Test Plan

- [x] New tests added
- [ ] Tested manually against the live Anthropic API

Validation performed:

- Full `agent/tests` suite: `6355 passed, 13 skipped`.
- Focused provider/config audit tests: `41 passed`.
- Ruff passes for both changed files.
- Black passes for the new test file. The existing `llm.py` already fails repository Black formatting on the unchanged base revision, so this PR avoids unrelated whole-file formatting churn.
- The environment-variable gate reports only its pre-existing `os.environ.pop()` warning; the credential reads touched by this change retain their required suppressions.

## Checklist

- [ ] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — protected provider change; proposal posted on #856 and PR remains draft pending maintainer direction
- [x] Documentation updated (if user-facing change) — no documentation surface changed

## Contributor review note

I reviewed the final diff, verified the model-ID boundaries explicitly, ran the full agent test suite, and kept the change limited to the reported provider failure. No live API credential was used during testing.

### DIFF
--- a/agent/src/providers/llm.py
+++ b/agent/src/providers/llm.py
@@ -464,20 +464,26 @@ def _build_anthropic(
             'extra: pip install "vibe-trading-ai[anthropic]" (or pip install langchain-anthropic).'
         ) from exc
 
-    return chat_anthropic(
-        model=model,
-        max_tokens=get_env_config().llm.anthropic_max_tokens,
-        temperature=temperature,
-        timeout=get_env_config().llm.timeout_seconds,
-        max_retries=get_env_config().llm.max_retries,
-        callbacks=callbacks,
-        api_key=os.getenv("ANTHROPIC_API_KEY") or None,  # noqa: env-gate — native provider credential
-        base_url=(
+    kwargs: dict[str, Any] = {
+        "model": model,
+        "max_tokens": get_env_config().llm.anthropic_max_tokens,
+        "timeout": get_env_config().llm.timeout_seconds,
+        "max_retries": get_env_config().llm.max_retries,
+        "callbacks": callbacks,
+        "api_key": os.getenv("ANTHROPIC_API_KEY") or None,  # noqa: env-gate — native provider credential
+        "base_url": (
             os.getenv("ANTHROPIC_BASE_URL")  # noqa: env-gate — native provider endpoint
             or os.getenv("ANTHROPIC_API_URL")  # noqa: env-gate — SDK-compatible alias
             or None
         ),
+    }
+    is_opus_4_7 = re.search(
+        r"(?:^|[.:/])claude-opus-4-7(?:-|$)", model.lower()
     )
+    if not is_opus_4_7:
+        kwargs["temperature"] = temperature
+
+    return chat_anthropic(**kwargs)
 
 
 def _load_env_file(path: Path) -> None:
--- a/agent/tests/test_anthropic_provider.py
+++ b/agent/tests/test_anthropic_provider.py
@@ -1,0 +1,75 @@
+"""Regression coverage for Anthropic model-specific request parameters."""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
+
+from src.providers import llm
+
+
+def _capture_anthropic_kwargs(monkeypatch, model: str, temperature: float) -> dict:
+    captured: dict = {}
+
+    def fake_chat_anthropic(**kwargs):
+        captured.update(kwargs)
+        return object()
+
+    fake_module = SimpleNamespace(ChatAnthropic=fake_chat_anthropic)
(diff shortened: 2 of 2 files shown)


he-yufeng (commented):
Read through the diff and the regex behavior. The approach looks right to me:

- The boundary is drawn where it should be. `(?:^|[.:/])claude-opus-4-7(?:-|$)` catches the canonical id, dated suffixes, and provider-qualified forms, and deliberately leaves future Opus versions on the current path. A substring like `my-claude-opus-4-7-finetune` does not match, which is the conservative call for a request-shaping rule.
- Moving the unconditional `temperature=temperature` into conditional kwargs keeps every other model byte-identical, so the blast radius is exactly the one family that errors.
- The factory-level regression with a fake ChatAnthropic is the right altitude for this: it pins the request shape without needing a live key.

One non-blocking suggestion: when the parameter gets dropped, a one-line `logger.info` (or debug) saying temperature was omitted for this model would save someone a confusing afternoon when they set `LANGCHAIN_TEMPERATURE` and see no effect. The silence is the only part of the behavior a user cannot see from outside.

Leaving this as a comment rather than an approval since it touches the provider request path; worth another pair of eyes before it lands.

warren618:
Thanks for the careful writeup, and for opening this as a draft rather than treating a protected-area change as pre-approved — that was the right call.

I'm going with #890, which solves the same bug generically: it drops `temperature` when the API reports it unsupported, retries once, and remembers the model, so opus-5 / sonnet-5 / opus-4-8 are covered without a code change per model release. Your regex is correct for opus-4-7 specifically, but the affected set has already grown past it.

Your reproduction and boundary tests were useful in deciding the shape. Closing in favour of #890 — not because the work was wrong.

PR #861, opened 2026-07-26
fix(skills): parse frontmatter when closing fence is at EOF

santhreal:
SKILL.md files that end at the closing --- with no trailing newline were treated as plain text, so name and description never loaded. Empty --- / --- blocks had the same miss. Accept EOF and empty meta between fences, with tests.

### DIFF
--- a/agent/src/agent/frontmatter.py
+++ b/agent/src/agent/frontmatter.py
@@ -8,7 +8,7 @@ from typing import Any, Dict
 # Opening ---, optional meta lines, closing ---. The closing fence may be at
 # EOF (no trailing newline) or followed by a body; empty meta (---\n---) is ok.
 _FRONTMATTER_RE = re.compile(
-    r"^---[ \t]*\r?\n(.*?)(?:\r?\n)?---[ \t]*(?:\r?\n(.*))?$",
+    r"^---[ \t]*\r?\n(?:(.*?)\r?\n)?---[ \t]*(?:\r?\n(.*))?$",
     re.DOTALL,
 )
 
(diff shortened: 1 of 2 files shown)


warren618 (changes requested):
Thanks — the bug you're fixing is real: `main` can't parse frontmatter whose closing fence sits at EOF, so `"---\nname: eof-skill\n---"` returns `({}, ...)` instead of the parsed meta.

The new regex is a little too permissive, though. Making the newline before the closing `---` optional means the fence no longer has to stand alone:

```python
parse_frontmatter("---\nname: foo---")
# PR:   ({'name': 'foo'}, '')
# main: ({}, '---\nname: foo---')
```

Here `---` is just the tail of a value, not a fence, so this is a false positive that `main` correctly rejects.

Keeping the fence standalone while still allowing EOF termination — only the *trailing* newline should be optional, and the meta block itself may be empty:

```python
_FRONTMATTER_RE = re.compile(
    r"^---[ \t]*\r?\n(?:(.*?)\r?\n)?---[ \t]*(?:\r?\n(.*))?$",
    re.DOTALL,
)
```

That rejects the `name: foo---` case and still accepts everything you care about: EOF fence, empty block (`---\n---\nBody`), normal, CRLF, and a fence with trailing spaces.

One follow-on: with this form `group(1)` is `None` for the empty-block case, so it needs the same guard you already apply to `group(2)`:

```python
raw_meta = match.group(1) or ""
```

Could you also add `"---\nname: foo---"` as a negative test? That's what would have caught this.

santhreal:
Thanks. Updated _FRONTMATTER_RE in agent/src/agent/frontmatter.py to require a newline before the closing --- fence so it stands on its own line, while still allowing the fence to terminate at EOF without a trailing newline. Added test_rejects_inline_fence_tail_false_positive to test_skills.py.

warren618 (approved):
The updated regex matches the suggested form exactly, the group(1) guard is in place, and the inline-fence false positive now has a locking test. Verified the edge set locally (EOF fence, empty block, CRLF, trailing-dash value). Thanks for the quick turnaround — merging.

PR #867, opened 2026-07-26
fix(channels): preserve leading indent when splitting messages

santhreal:
After a newline cut, lstrip() removed indentation on the next chunk, so code blocks lost leading spaces across Discord/Slack/Telegram splits. Drop only the separator character, with a test.

### DIFF
--- a/agent/src/channels/signal.py
+++ b/agent/src/channels/signal.py
@@ -260,13 +260,13 @@ def _partition_styles(
         return [[] for _ in chunks]
 
     # Mirror split_message: after chunk 0, skip one separator char if present
-    # and not already part of the next chunk (so indented lines stay aligned).
+    # (split_message unconditionally drops at most one leading \n or space).
     chunk_ranges: list[tuple[int, int]] = []
     cursor = 0  # Python codepoint cursor in plain_text
     for i, chunk in enumerate(chunks):
         if i > 0 and cursor < len(plain_text):
             ch = plain_text[cursor]
-            if (ch == "\n" or ch == " ") and not chunk.startswith(ch):
+            if ch == "\n" or ch == " ":
                 cursor += 1
         utf16_start = _utf16_len(plain_text[:cursor])
         utf16_end = utf16_start + _utf16_len(chunk)
--- a/agent/src/channels/signal.py
+++ b/agent/src/channels/signal.py
@@ -248,27 +248,25 @@ def _partition_styles(
 ) -> list[list[str]]:
     """Partition Signal textStyle ranges across message chunks.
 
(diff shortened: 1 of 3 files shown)


warren618 (changes requested):
Thanks for this — the underlying bug is real. On `main`, `split_message` blanket-`lstrip()`s at each cut, so a wrapped code block loses its indentation (`'    def foo():'` → `'def foo():'`). The `utils.py` half of this patch is the right fix.

The `signal.py` half has a regression, though. `_partition_styles` mirrors the separator drop with:

```python
if (ch == "\n" or ch == " ") and not chunk.startswith(ch):
    cursor += 1
```

but `split_message` drops the separator **unconditionally** when it is `\n` or a space — it never looks at what the next chunk starts with. When the dropped separator and the next chunk both begin with a space, the cursor fails to advance and every style range in that chunk shifts.

Repro:

```python
text = "😀" * 5 + "  bold"
chunks = split_message(text, max_len=5)   # ['😀😀😀😀😀', ' bold']
styles = ["12:4:BOLD"]                    # 'bold' in full-text UTF-16 offsets
_partition_styles(text, chunks, styles)   # -> [[], ['2:3:BOLD']]
```

Expected `1:4:BOLD`. The offset is off by one *and* the length is truncated 4 → 3, so the bold run both moves and loses its last character. `main` returns the correct `0:4:BOLD` today, so this is a regression rather than pre-existing.

A randomized property check (one styled run, whitespace-heavy random text, random `max_len`) gives 840 wrong out of 3458 valid trials — about 24%.

CI stayed green because the new `test_signal_partition_styles_keeps_indent_chunk_aligned` only asserts `any("BOLD" in entry ...)` — it never checks offsets, so it can't catch this.

Dropping the guard fixes it:

```python
if ch == "\n" or ch == " ":
    cursor += 1
```

With that change the same property check gives 0 mismatches in 4000 trials, and the channel suite stays green (156 passed, 2 skipped).

Could you apply that and strengthen the new test to assert the exact `offset:length:STYLE` string rather than mere presence?

santhreal:
Thanks. Updated _partition_styles in agent/src/channels/signal.py to unconditionally advance cursor when encountering a space/newline separator, matching split_message's separator drop regardless of whether the next chunk is indented. Added a test covering space-separated indented chunks in test_split_message_nonpositive.py.

warren618 (approved):
The cursor now advances unconditionally on a \n/space separator, which is an exact mirror of split_message's drop rule — the emoji + double-space repro from the previous review now rebases correctly, and the new test locks the space-separator + indented-next-chunk case. Approving — merging.

PR #870, opened 2026-07-26
feat(trading): add vn.py RPC bridge connector

y85998607:
## Goal

Give Vibe-Trading a real-time, two-way link to a running VeighNa (vn.py)
instance, closing the loop with the existing `vnpy-export` skill: research
and backtest in Vibe-Trading -> export a `CtaTemplate` -> run it live in
vn.py against its professional-grade gateways (CTP, XTP, Tora, IB, ...) ->
Vibe-Trading reads positions/orders/quotes in real time and can place orders
back through the same connector.

## Affected areas

New:
- `agent/src/trading/connectors/vnpy/sdk.py`
- `agent/src/trading/connectors/vnpy/profiles.py`
- `agent/src/trading/connectors/vnpy/__init__.py`
- `agent/tests/test_vnpy_connector.py`

Modified (both one-line/one-entry additions):
- `agent/src/trading/service.py` — registered `"vnpy"` in `_SDK_CONNECTOR_MODULES`
- `agent/src/trading/profiles.py` — registered `VNPY_PROFILES` in `BUILTIN_PROFILES`
- `pyproject.toml` — new optional `vnpy` extra

## How it works

vn.py core ships `vnpy.rpc` (a ZeroMQ wrapper) and the separate
`vnpy_rpcservice` app exposes a running `MainEngine` over it: a REQ/REP
socket for remote calls (`send_order`, `cancel_order`, `subscribe`,
`query_history`, the `get_*`/`get_all_*` readers) and a PUB/SUB socket that
republishes every engine event (ticks, trades, orders, positions, accounts,
logs) in real time. This is vn.py's own documented mechanism for distributed
deployments, not a bolted-on hack.

`sdk.py` implements the same function contract every other `broker_sdk`
connector already implements (modeled directly on
`connectors/futu/sdk.py`): `check_status`, `get_account_snapshot`,
`get_positions`, `get_open_orders`, `get_quote`, `get_historical_bars`,
`place_order`, `cancel_order`. Operationally: the operator connects a real
gateway (CTP/XTP/Futu/IB/...) inside their own vn.py Trader, starts the RPC
Service app there, and points this connector's `req_address`/`sub_address`
at it.

## Safety note (please read — this is the important part)

Unlike Futu's OpenD, which tags every account row with an SDK-verifiable
`trd_env` (`SIMULATE`/`REAL`) that the existing Futu connector checks against
the profile, **a vn.py RPC endpoint has no runtime-discoverable paper/live
discriminator**. That boundary is entirely a property of which broker
account the operator logged the remote vn.py instance into — this connector
cannot verify it from the client side the way the other connectors verify
their own paper/live guard.

Given that, `place_order`/`cancel_order` fail closed unless the caller
explicitly sets `assume_environment_verified=true` in the connector config.
This is a manual acknowledgement, not a structural guard, and is documented
as weaker isolation than the other live-capable connectors in both the
module docstring and every profile's `notes` field. Flagging this explicitly
for review since broker/order-gate code is called out as safety-critical in
`AGENT_CONTRIBUTOR_GUIDE.md` even when a change looks small.

## Out of scope

- Not adding a `vnpy-rpc` MCP tool wrapper or CLI flags beyond what
  `connector configure/use/check` already provide generically for
  `broker_sdk` profiles — this PR is the connector module itself.
- Not changing the mandate/kill-switch plumbing itself, only plugging into it
  the same way every other `broker_sdk` connector does.
- Not touching `vnpy-export` (see the separate PR #869 for that skill's
  unrelated import-path fix).

## Test plan

- `pytest agent/tests/test_vnpy_connector.py -q` — 10 passed. Pure unit
  tests (config resolution, `vt_symbol` parsing, fail-closed order gate); no
  live vn.py process required.
- `pytest agent/tests/test_sdk_order_gate.py agent/tests/test_mandate_enforcement.py agent/tests/test_killswitch_blocks_orders.py agent/tests/test_readonly_default.py -q` — 69 passed (the targeted live/order-safety suite from `AGENT_CONTRIBUTOR_GUIDE.md`).
- `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` — 6366 passed, 12 skipped (full suite, pre-existing skips/warnings unrelated to this change).
- **Manual live round-trip** (not part of CI, no real broker/credentials involved): started a headless `vnpy.trader.engine.MainEngine` + `RpcEngine` in a separate process (`MainEngine.add_engine(RpcEngine)`, no Qt UI, no gateway connected), then called `check_status` / `get_account_snapshot` / `get_positions` from this connector in a completely separate Python process/venv and confirmed real cross-process RPC responses. Also confirmed `place_order` refuses to submit without `assume_environment_verified` set. Empty account/position data is expected — the demo instance has no broker gateway attached; the point was proving the RPC round-trip itself.

## Risk

- No secrets, credentials, or network-exposed servers are touched by this
  PR itself — it's a client that connects outward to an address the
  operator configures.
- The order-placement path is new surface touching the broker-write
  contract; mitigated by the fail-closed `assume_environment_verified` gate
  described above, and by reusing the exact same `_SDK_CONNECTOR_MODULES` /
  `TradingProfile` plumbing (mandate gate, kill switch, audit ledger) every
  other connector already goes through.

## Rollback

Single self-contained commit; `git revert` cleanly removes the new module
and the two registration entries with no other code depending on them.


warren618 (changes requested):
Thanks for building this, and in particular for being straight about the transport's limits in the profile docstrings rather than papering over them. That honesty is what makes the problem easy to state.

**1. The red line is structural, and this crosses it.**

`vnpy-live-trade` ships with `environment="live"`, `readonly=False` and `orders.place.requires_mandate`, gated only on `cfg.assume_environment_verified`. Your own profile notes say it plainly: *"there is no structural paper/live discriminator on this transport."*

That is exactly the case the contract covers. Paper-vs-live has to be a runtime guard the broker itself reports — account-id format, host separation, a demo flag, a trade environment — never a config value. `assume_environment_verified` is a config value, so a mistake or a stale config becomes a real order on a real account with no second line of defence.

The required shape for a connector with no discriminator is: no live order placement, no `*-live-trade` profile, and `place_order` / `cancel_order` refusing at the first line. Longbridge, Dhan and Shoonya are already capped that way.

**2. For this transport specifically, I think it has to go further than those three.**

Longbridge/Dhan/Shoonya can at least identify a paper session structurally, so paper order placement is safe there. The RPC bridge cannot: all four profiles reach the same `RpcService` endpoint, so `vnpy-paper-trade` has no more evidence of being paper than `vnpy-live-trade` does. A "paper" order here could land on a live gateway.

So the shape I can take is **fully read-only** — reads only, `place_order` / `cancel_order` hard-refusing regardless of configured environment, and only the two read profiles. Trading212 is the precedent. If vn.py's RpcService can report the remote Trader's gateway environment over the wire, that changes the analysis completely and I would rather have that than the cap — but it has to come from the remote, not from our config.

**3. CI is red, independently of the above.**

`ModuleNotFoundError: No module named 'vnpy'`. The connector imports an optional dependency at module load, so the whole suite fails wherever vn.py is not installed. It needs the same lazy/gated import the other optional-SDK connectors use.

(CI on this PR had never actually run — the workflow was sitting unapproved, so the "no checks" state was hiding this. I have approved the queue, so runs now report normally.)

Happy to keep going on the read-only version.

warren618:
Thanks for the vn.py bridge contribution. The project requires paper/live identity to be verified structurally at runtime; `assume_environment_verified` cannot authorize live order placement. Because this PR exposes a live-trade profile without that discriminator and also retains stale pre-rewrite history, we are closing it. A fresh read-only vn.py RPC PR from current `main`, with all place/cancel and live-trade surfaces removed, would be welcome. Please delete the old source branch.

PR #875, opened 2026-07-27
docs: architecture analysis handoff under docs/analysis

Tenant-farmer:
## Summary
- Add `docs/analysis/` as an AI/contributor handoff: narrative architecture report (Parts 1–28) plus catalogs for skills, alphas, swarm presets, channels, and tests
- Point `docs/README.md` at the analysis entry, and allow `docs/analysis/**` (plus `docs/README.md`) via a narrow `.gitignore` exception so shared docs are trackable while other internal `docs/` stay ignored

## Test plan
- [ ] Open `docs/analysis/README.md` and confirm navigation links resolve
- [ ] Spot-check catalog counts against repo (skills / alphas / swarm presets / channels / tests)
- [ ] Confirm `graphify-out/` and `.graphify_*.json` are not included in the PR
- [ ] Confirm unrelated private files under `docs/` remain untracked

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,8 +50,12 @@ data/.duckdb_tmp/
 data/parquet/
 data/backup/
 
-# Internal docs (plans, specs)
-docs/
+# Internal docs (plans, specs) — keep private by default;
+# shared architecture handoff lives under docs/analysis/
+docs/*
+!docs/README.md
+!docs/analysis/
+!docs/analysis/**
 !wiki/docs/
 !wiki/docs/**
 docs/screenshots/agent-run.png
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,7 @@
+# Moved
+
+분석 자료는 **`docs/analysis/`** 로 모았습니다.
+
+- AI 진입: [`docs/analysis/README.md`](./analysis/README.md)
+- 보고서: [`docs/analysis/REPORT.md`](./analysis/REPORT.md)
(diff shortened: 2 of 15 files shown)


warren618:
Apologies — I closed this without a word, which was not fair to a 15,000-line contribution. Let me explain the reasoning properly, because the effort here deserved that at the time.

`docs/` is gitignored on purpose (`.gitignore:54`). It is the internal planning area — specs, audits, half-finished investigations — and the rule that it never reaches the public tree is deliberate rather than an oversight. This PR adds a narrow exception to publish `docs/analysis/**`, which is a policy change rather than a docs addition, and that is really what I was reacting to.

The substantive worry: once a narrative architecture report is committed, readers treat it as authoritative. This repo moves quickly — the tool count, loader count, skill count, and engine list have all changed in the last few weeks — so a 28-part report describing the system would need someone to verify it against the code before it went in, and then keep verifying it. I could not commit to that, and publishing an unverified description of the architecture is worse than having none.

To be fair to your design, you did anticipate part of this: `_build_inventories.py` and `_build_channels_tests.py` mean the catalogs are regenerable rather than hand-maintained, and that is the right instinct.

If you would still like this to be useful to others, the cleanest route is to keep it in your own repository and link it from a Discussion — I am happy to point people at it as a community resource. And if there are specific pieces you believe belong upstream, small targeted PRs against tracked paths would get a proper review from me.

Sorry again for the silence. Whatever you decide, thank you for the work.

PR #798, opened 2026-07-23
feat(backtest): define perpetual margin state contracts

honginp:
## Stack note

First slice of the PR 2 (margin-risk model) work approved in `#462`. Cut fresh from `upstream/main`, not rebased or cherry-picked from the old `codex/perp-engine-fidelity` branch — that branch predates `#757` and still carries the pre-`#470` loader shape, so applying it directly would reintroduce already-fixed problems (see below). The *intent* of its first commit is reapplied here against the current, merged data contract.

## Summary

Pure, immutable state and risk contracts for historical USD-M perpetual backtests — no engine wiring yet:

- `MaintenanceBracket` / `MaintenanceSchedule` — versioned maintenance-margin brackets.
- `PositionState` — exchange-independent open-position accounting (signed quantity, entry price, leverage, fees; never stores a mark price).
- `AccountState` — single-asset USDT account state without open-order margin.
- `ExecutionFrame` / `MarketRiskFrame` — separate contracts for normal market fills vs. exchange mark/funding/bracket data, matching `#470`'s execution/mark price separation.

## Why

Part of `#462`'s three-PR plan (data contract → margin-risk model → engine integration). Sliced small per the maintainer's review-budget guidance, same pattern as `#470`/`#716`/`#757`.

## What changed vs. the old branch's version

The old `codex/perp-engine-fidelity` branch has the right shape but was written before `#757` (the bracket-artifact contract) existed:

- `MaintenanceBracket` gains `bracket_tier` and optional `notional_coefficient`, matching `#757`'s validated artifact tier fields exactly.
- `MaintenanceSchedule` drops its own independent SHA-256 hash computation in favor of a single caller-supplied `version` string — `#757`'s loader already validates content integrity once, at the artifact boundary (`_validate_bracket_artifact`); re-hashing here would just be a second, incompatible version scheme. This class re-checks structural invariants (tier and notional-cap ordering) as defense in depth, not content hashing.
- `MaintenanceSchedule.from_loader_columns(symbol, maintenance_brackets, maintenance_bracket_version)` builds a schedule directly from the loader's DataFrame columns.
- `MarketRiskFrame.schedule` is `Optional`: a `-PERP` fetch without a supplied bracket artifact (the default, zero-credential path) still produces a usable execution/mark/funding frame.

## Test Plan

- [x] `pytest agent/tests/test_perpetual_risk.py -q` — 21 passed.
[...]

PR #889, opened 2026-07-27
feat(backtest): model isolated and cross margin liquidation risk

honginp:
## Stack note

Second slice of the PR 2 margin-risk work approved in #462. It builds directly on the merged immutable state contracts from #798 and remains pure historical backtest logic: no engine wiring or live API surface.

## Summary

Add deterministic isolated- and cross-margin liquidation evaluation for single-asset Binance USD-M research:

- calculate maintenance margin from the versioned per-symbol bracket schedule;
- value notional, unrealized P&L, and initial margin from the selected mark price;
- liquidate only breached positions in isolated mode (`position_liquidation`);
- liquidate every open position when the cross-account threshold is breached (`account_liquidation`);
- use adverse `mark_low` for longs and `mark_high` for shorts by default;
- emit `conservative_intrabar_assumption` when multiple symbols use same-bar adverse extremes.

## Why

#798 established the state, execution, mark-risk, and bracket contracts but intentionally left liquidation logic out of scope. This slice provides a pure, reusable risk boundary before event ordering and engine integration are introduced in later PR 2c/2d slices.

## Accounting boundary

This remains the single-asset USDT model from #462, without open-order margin:

```text
margin_balance = wallet_balance + total_unrealized_pnl
initial_margin = sum(position_initial_margin)
maintenance_margin = sum(maintenance_margin_per_position)
available_balance = margin_balance - initial_margin
```

Cross liquidation uses the exact predicate `margin_balance <= maintenance_margin`. Per-position maintenance margin is:

```text
abs(notional) * maintenance_rate - cumulative_maintenance_amount
```

`notional_coefficient` remains schedule metadata and is not reapplied after the loader has produced authoritative bracket caps.

## Changes

- Add frozen `PositionRisk` and `RiskSnapshot` outputs.
- Add bracket-aware `maintenance_margin()` with symbol and cap validation.
- Add `evaluate_isolated()` for position-local liquidation decisions.
- Add `CrossMarginRiskModel.evaluate()` for account-level liquidation decisions.
- Reject missing schedules, symbol mismatches, unsynchronized frames, invalid margin-mode state, and notionals above the final bracket cap.
- Cover exact-threshold behavior, mark-price selection, cross-position P&L offsets, synchronized timestamps, frozen outputs, and conservative intrabar fidelity evidence.

## Test plan

- [x] `pytest agent/tests/test_perpetual_risk.py agent/tests/test_ccxt_perpetual_loader.py agent/tests/test_crypto_engine.py agent/tests/test_base_engine.py -q` — 123 passed.
- [x] `ruff check agent/backtest/perpetual_risk.py agent/tests/test_perpetual_risk.py` — passed.
- [x] `git diff --check upstream/main...HEAD` — passed.
- [x] PowerShell-equivalent checks for `tools/ci_grep_gates.sh` gates a-e — passed; the Git Bash wrapper itself timed out in this Windows environment, so CI remains authoritative for the canonical script.

## Safety and network risk

- No network calls, Binance credentials, broker writes, OAuth, paper trading, or live execution.
- Tests use synthetic deterministic state and do not contact an exchange.
- No changes to connectors, API routes, or order placement.

## Fidelity notes

This is a conservative historical backtest model, not an identical reproduction of Binance liquidation sequencing. Same-bar adverse extrema across symbols may not occur at the same millisecond; the resulting snapshot carries `conservative_intrabar_assumption`. Exact exchange liquidation waterfall, partial liquidation, insurance-fund behavior, ADL, and order-book replay remain unmodeled.

## Out of scope

- Applying funding or trading/liquidation fees in the engine event loop.
- Closing positions or terminating a run after a risk decision.
- Engine event ordering, risk artifact persistence, and degraded-fidelity reporting integration (PR 2c/2d).
- Open orders, hedge mode, non-USDT collateral, multi-assets mode, portfolio margin, Shadow Account API reconciliation, or live execution.

## Rollback

Revert this PR. The new evaluators are not wired into the engine execution path in this slice, so rollback does not require data migration or runtime cleanup.


he-yufeng:
Read the liquidation math against how the exchanges actually do it, and the core of it holds:

- Adverse price direction is right (longs liquidate at mark_low, shorts at mark_high), the per-position isolated equity (`isolated_margin + unrealized_pnl <= maintenance`) and the cross account-level aggregate (all positions targeted once wallet + uPnL drops under total maintenance) match the standard models.
- Requiring one shared timestamp across frames and flagging `conservative_intrabar_assumption` for multi-position adverse evaluation are both honest touches; simultaneous adverse prices in one bar really is the conservative bound, and it says so.

Two things worth a thought before merge:

1. `maintenance_margin` raises when notional exceeds the final bracket cap. With a real ccxt bracket ladder that is nearly unreachable, but with hand-built or partial schedules (like the two-bracket ones in the tests) a legitimately large position crashes the run instead of just paying the top rate. Is the raise intended as "schedule must be complete", or should the top bracket apply open-endedly?
2. The bracket walk assumes ascending `notional_cap` order. Nothing enforces it; an unsorted schedule silently prices maintenance in the wrong bracket. A one-line sort at schedule construction (or an assert) would make the assumption explicit.

warren618 (approved):
Re-derived the numbers in every scenario test by hand before merging — isolated (247.5 vs 100 breach, 259.54 aggregate), the exact-threshold cross case (259.5 <= 259.5), the PnL-offset case (367.1), and the bracket-boundary continuity identity (50000×0.004 == 50000×0.005−50) — all hold, and the adverse-direction and account-level aggregation match the standard exchange models.

On @he-yufeng's two points:

1. **The raise above the final cap is intended, and I'd keep it.** On the real venue the last bracket's notional cap *is* the maximum position size the exchange permits — a position above it cannot exist, so modelling it as payable-at-top-rate would price something the exchange would have rejected. For hand-built or partial schedules, crashing loudly is consistent with the fail-closed contract this stack committed to in #462 (strict consumers hard-fail on an unusable artifact rather than silently degrading). So: "schedule must be complete" is the ruling, and the error message says which invariant broke.

2. **Ascending order is already enforced upstream** — `MaintenanceSchedule.__post_init__` (landed with #798) rejects non-increasing `bracket_tier` *and* non-increasing `notional_cap` ("notional caps must be strictly increasing"), and `maintenance_margin` only accepts a `MaintenanceSchedule`, so an unsorted ladder is unconstructible rather than silently mispriced. The assumption in the walk is real but the type guarantees it.

Pure historical logic, no engine wiring, no live surface — red lines clean. Merging; looking forward to 2c/2d.

PR #890, opened 2026-07-27
fix(anthropic): self-heal models that deprecate the temperature field

yagnikpipaliya:
Problem
-------
Next-gen Claude models (e.g. claude-opus-5, claude-opus-4-8, claude-sonnet-5) reject the "temperature" request field with HTTP 400 "temperature is deprecated for this model." for ANY value, while older models (claude-opus-4-5, claude-sonnet-4-5, ...) still accept it. _build_anthropic unconditionally forwarded the global default temperature=0.0, so every run on a next-gen model failed before producing output. Model names are not predictable (opus-4-5 accepts it, opus-4-8 does not) and the rejection is value-independent, so neither a fixed model allowlist nor an "omit when 0.0" shortcut is a correct fix.

Solution
--------
Add a temperature-safe ChatAnthropic subclass (_make_temperature_safe_anthropic) that wraps all four generation entry points (_generate, _agenerate, _stream, _astream). When the API reports temperature as unsupported it drops the field and retries once, then records the model in a process-wide set so subsequent requests omit it up front (no repeated failed request). Models that accept temperature keep their configured deterministic value unchanged. The subclass is built from the lazily-imported class and cached per base class, preserving the optional langchain-anthropic dependency boundary and existing test doubles.

Correct for any temperature value, current and future model names, and the streaming path the Web UI relies on.

Tests
-----
Add regression tests in agent/tests/test_llm.py: self-heal retry drops temperature and remembers the model, temperature is preserved for supporting models, and the error matcher only triggers on temperature-unsupported errors. Full test_llm.py suite: 58 passed.

## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/src/providers/llm.py
+++ b/agent/src/providers/llm.py
@@ -448,13 +448,138 @@ def _build_native_deepseek(
     )
 
 
+# Anthropic model names discovered at runtime to reject the `temperature`
+# request field. Next-gen Claude models (e.g. claude-opus-5, claude-opus-4-8,
+# claude-sonnet-5) return HTTP 400 "`temperature` is deprecated for this model."
+# for ANY temperature value, while older models (claude-opus-4-5,
+# claude-sonnet-4-5, …) still honor it. Model names are not reliably
+# predictable, so membership is populated on first failure and then reused
+# process-wide to skip the redundant failed request on subsequent calls.
+_ANTHROPIC_TEMPERATURE_UNSUPPORTED: set[str] = set()
+
+# Cache of base ChatAnthropic class -> temperature-safe subclass, so the dynamic
+# subclass is built once per resolved base class (keyed to support test doubles).
+_TEMPERATURE_SAFE_ANTHROPIC_CACHE: dict[type, type] = {}
+
+
+def _is_anthropic_temperature_unsupported_error(exc: BaseException) -> bool:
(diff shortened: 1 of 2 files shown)


warren618:
Thanks — taking this one.

I compared it against #857, which fixes the same issue with a `claude-opus-4-7` allowlist. Yours is the right shape: the rejection is model-specific and not predictable from the name (opus-4-5 accepts the field, opus-4-8 does not), so an allowlist needs a patch per model release.

I checked the assumptions against `langchain-anthropic` 1.5.2 rather than the test double, since a fake base class would hide a mismatch: `_get_request_payload`, `_generate`, `_agenerate`, `_stream` and `_astream` all exist on `ChatAnthropic`, and `model` is a real field (not `model_name`), so `self.model` resolves.

Two non-blocking follow-ups, here or later:

- `_stream` retries the whole stream and relies on a comment for the claim that the 400 arrives before the first chunk. That holds for request validation, but setting a `produced` flag before the first `yield` would make duplicate output structurally impossible rather than merely unlikely.
- Seeding the known-rejecting model names would avoid one failed round-trip per model per process, with the self-heal still covering names we don't know yet.

`src/providers/` is protected, so this needed a maintainer decision — that's done. Approving the workflow run; it needs to be green before merge.

PR #891, opened 2026-07-28
fix(metrics): add missing annualisation entries for 15 data sources

### LINKED ISSUE #884: calc_bars_per_year silently returns 252 for 13 of 24 VALID_SOURCES, at every interval
13 of the 24 entries in `VALID_SOURCES` have no annualisation entry, so `calc_bars_per_year` silently returns `252` for them at **every** interval.

`agent/backtest/metrics.py` · `agent/backtest/loaders/registry.py`

`calc_bars_per_year` looks the source up in two tables and defaults on a miss:

```python
trading_days = _TRADING_DAYS.get(source_key, 252)
bars_per_day = _BARS_PER_DAY.get(interval_key, {}).get(source_key, 1)
return trading_days * bars_per_day
```

`_TRADING_DAYS` covers 8 sources (`akshare, ccxt, futu, mootdx, mt5, okx, tushare, yfinance`) plus 2 via `_SOURCE_ALIASES` (`yahoo`, `binance`). `VALID_SOURCES` has 24. Set difference, excluding `auto` (already resolved by `_detect_primary_source` before the call):

```
alphavantage  baostock  eastmoney  finnhub  fmp  india_broker
local  longbridge  qveris  sina  stooq  tencent  tiingo
```

For any of these, **both** lookups miss and the function returns `252 * 1 = 252` — for `1D`, and identically for `1m`, `5m`, `15m`, `30m`, `1H`, `4H`. Nothing warns.

**Why it matters.** `runner.py:964` feeds this into the engines:

```python
bars_per_year = calc_bars_per_year(interval, effective_source)
...
run_options_backtest(..., bars_per_year=bars_per_year)
market_engine.run_backtest(..., bars_per_year=bars_per_year)
```

and it lands in `validation.py:_sharpe`. Sharpe and annualised volatility scale as `sqrt(bars_per_year)`, so a wrong factor is a wrong *reported number*, not a crash. For an intraday run the gap is large — the `1m` entries in `_BARS_PER_DAY` range from 240 to 1440, so an affected source reports Sharpe roughly `sqrt(240*252/252)` ≈ **15x** too small at `1m`, ≈ **5x** at `15m`. And even at `1D`, a 24/7 venue among the 13 should be `365`, not `252` — about **20%** on annualised figures.

**Scope, stated rather than assumed.** I have *not* checked which of the 13 support sub-daily intervals; for a daily-only source the `252` may be right by coincidence, and for a daily-only 24/7 source it is still ~20% off. The impact is therefore "wrong for those that do", not "wrong for all 13" — but the silence is the same in every case, and the set is large enough that the default looks like a fallback for typos rather than for a majority of registered sources.

**Suggested fix.** Either give the 13 explicit entries, or make the miss loud rather than silent — e.g. keep the 252 fallback but emit a warning naming the source, so a run card records that annualisation was defaulted. A regression test analogous to `test_valid_sources_covers_all_registered_loaders` would keep the two tables from drifting apart again:

```python
def test_annualisation_covers_all_valid_sources():
    covered = set(_TRADING_DAYS) | set(_SOURCE_ALIASES)
    assert not (VALID_SOURCES - {"auto"}) - covered
```

---

Two smaller things found alongside, both currently harmless, reported because they are one edit away from mattering.

**1. `source` normalisation is applied to the alias lookup only.**

```python
source_key = _SOURCE_ALIASES.get(str(source or "").strip().lower(), source)
```

The lowercased/stripped string is the *alias key*, but the fallback is the **original** `source`. So a non-alias source that is not already exactly lowercase misses both tables:

```
calc_bars_per_year("1m", "okx")  -> 525600
calc_bars_per_year("1m", "OKX")  -> 252        # 2086x
calc_bars_per_year("1H", "okx ") -> 252        # trailing space
```

Aliases are unaffected (`"Yahoo"`, `"Binance"` both resolve correctly) because that lookup *does* get the normalised key — which is what shows the case-insensitivity is intended and applied to one of the two lookups. This is **not reachable today**: `valid_source` rejects anything outside the all-lowercase `VALID_SOURCES` before it gets here. It becomes live the moment the validator coerces instead of rejecting, or a source name with an uppercase letter is registered. `_normalize_interval` already does this properly for the interval half; the same treatment for `source_key` would close it.

**2. `BacktestConfigSchema` is constructed and discarded.**

`runner.py:903`:

```python
try:
    BacktestConfigSchema(**raw_config)
except Exception as exc:
    ...
    sys.exit(1)

config = raw_config
```

The validated model is never bound, so the **raw** dict is what flows onward. Validators that *raise* still work, which is why (1) is currently unreachable — but any validator that normalises or coerces would be silently inert, and the file already has `.strip()`-style logic in several validators whose returned values go nowhere. `config = BacktestConfigSchema(**raw_config).model_dump()` would make the validated values the ones actually used. Flagging it because (1) is exactly the bug that appears if someone later "fixes" `valid_source` to lowercase instead of reject.

Found while reading `metrics.py`; no behaviour was run against live data (this box has no pandas), so everything above is from the source plus direct evaluation of the lookup tables. Happy to open a PR for the coverage test and the `source_key` normalisation if useful.

Robin1987China:
### Problem

`calc_bars_per_year` in `metrics.py` has lookup tables (`_TRADING_DAYS` and `_BARS_PER_DAY`) that only cover 8 of 24 `VALID_SOURCES` entries. For the remaining 15 sources, the function returns 252 for **every** interval — including intraday — silently mis-annualising Sharpe/vol metrics.

### Fix

Add all 15 missing sources with appropriate trading-day and bars-per-day counts:

| Category | Sources | Trading Days | Intraday Pattern |
|---|---|---|---|
| Crypto | `binance` | 365 | 1440 bars/day (same as `ccxt`/`okx`) |
| A-share | `baostock`, `tencent`, `eastmoney`, `sina` | 252 | 240-min session (same as `tushare`) |
| US/Intl | `yahoo`, `finnhub`, `alphavantage`, `tiingo`, `fmp`, `stooq`, `longbridge` | 252 | 390-min session (same as `yfinance`) |
| Indian | `india_broker` | 252 | 375-min session (6.25h) |
| Daily-only | `qveris`, `local` | 252 | 1 bar/day for all intervals |

### Tests

24 new test assertions across 16 tests, covering every added source at both daily and intraday. Includes a cross-`VALID_SOURCES` sweep that verifies every source returns >0 at both 1D and 1m — so this regression cannot return silently.

Closes #884


he-yufeng:
Checked the new constants against the actual sessions and they hold up: US equity 390 (6.5h), A-share 240 (4h), NSE 375 (9:15-15:30), crypto 1440x365, mt5 1440x260 for a 24/5 book, and the per-interval rows divide down correctly (78/26/13/7/2 for US, 48/16/8/4/1 for A-share). qveris/local falling back to 1 bar per day is the honest choice for sources whose session shape is unknowable.

One thing worth a comment line in the table rather than leaving implicit: longbridge sits in the US 390 bucket but the broker also serves HK (330-minute session), the same mixed-calendar approximation futu already documents at 240. Fine as a deliberate choice, but writing it down stops a future reader from "correcting" it in either direction.

Tests assert concrete per-source values rather than row counts, which is the right shape for a constants table. Looks good; leaving the final pass to @warren618.

warren618 (changes requested):
Thanks — the direction is right and this closes a real gap. Three things before merge.

**1. `local` is pinned to daily-only, and a test locks the wrong value in.**

The PR gives `local` 1 bar/day at every interval, but the local loader resamples to whatever interval is requested — `_resample_to_interval` at `agent/backtest/loaders/local_loader.py:83`, whose docstring notes that "local files can hold any native granularity", and the fetch signature at `:264`: "Bar size (all intervals supported if data contains them)."

So a 1m `local` backtest would annualise at 252 instead of 252×390, understating Sharpe and annualised vol by about √390 ≈ 19.7×. `test_daily_only_local` then asserts that as correct, which makes it harder to fix later than it is today.

Same question for `qveris`: `qveris_loader.py:291` only says selection is "optimized for daily" while still accepting an interval — what did you base the daily-only classification on?

**2. The normalisation guard was removed rather than extended.**

```diff
-    source_key = _SOURCE_ALIASES.get(str(source or "").strip().lower(), source)
+    source_key = _SOURCE_ALIASES.get(source, source)
```

#884 asked for normalisation to be applied to *both* lookups; this drops it from the one place it existed. It is latent today — the only production caller is `backtest/runner.py:964`, fed by `_detect_primary_source`, which returns the exact pydantic-validated value from the `valid_source` validator at `runner.py:121` — but `calc_bars_per_year` is a public helper and the guard is cheap to keep.

**3. The drift guard cannot fail.**

The new coverage test asserts only `calc_bars_per_year(...) > 0`. The 252 fallback makes that true for every possible input, so it would pass unchanged with all 13 sources still missing. #884 proposed the form that actually bites:

```python
covered = set(_TRADING_DAYS) | set(_SOURCE_ALIASES)
assert not (VALID_SOURCES - {"auto"}) - covered
```

That fails the moment a loader is registered without an annualisation entry — which is the drift this PR exists to stop.

Happy to merge once those are addressed.

Robin1987China (commented):
Thanks for the thorough review. All three points addressed in `008069b`:

**1. local / qveris reclassified (daily-only → US equity session)**

Verified both loaders:
- `local_loader.py:83` `_resample_to_interval()` handles 1m–1D and the
  fetch docstring says "all intervals supported if data contains them."
- `qveris_loader.py:416` `_granularity_tokens()` explicitly distinguishes
  intraday/minute vs daily vs monthly.

Both now use 252d × 390-min (the yfinance paradigm) in all 7 interval
layers. The `_TRADING_DAYS` comment reads "resampling sources" to make
the rationale searchable.

**2. source_key normalisation**

This was fixed in the prior commit `d6770bd` — the guard is
`str(source or "").strip().lower()` on line 148, followed by
`_SOURCE_ALIASES.get(source_key, source_key)` on 149.
CI stays green; no change needed here.

**3. drift guard hardened**

`test_no_intraday_source_falls_to_default` now computes
`covered = set(_TRADING_DAYS) ∩ set(∪_BARS_PER_DAY)` and asserts
`not (VALID_SOURCES - {"auto"}) - covered`. That fails the moment a
loader is registered without both tables being updated — exactly the
form #884 proposed.

warren618 (approved):
All three points addressed: local/qveris are now modeled as resampling sources with the wrong-value test replaced, normalisation now applies to both the alias lookup and the fallback (better than the original guard — the case-insensitivity test locks it), and the coverage test now asserts the covered set so it fails the moment a loader lands without annualisation entries. Constants independently re-checked against session lengths. Merging — thanks for the careful revision.

PR #898, opened 2026-07-28
feat(loaders): add tickerall hosted MetaTrader 5 data source

### LINKED ISSUE #897: [Feature] tickerall data source: hosted MT5 forex/metals feed, no local terminal (any OS)
### Problem or Motivation

Backtesting forex/metals today relies on the `mt5` loader, which needs a local, logged-in MetaTrader 5 terminal on Windows. That rules out Linux, macOS, and headless/CI environments: there is no way to pull a broker's own MT5 candle feed without a running terminal.

### Proposed Solution

Add an optional `tickerall` data source: a forex/metals OHLCV loader that reads a broker account's candle history through a hosted MetaTrader 5 API (TickerAll, https://tickerall.com) over HTTP, so backtests can use the same broker feed as `mt5` from any OS with no local terminal.

It slots into the existing loader registry like every other source:
- A new `backtest/loaders/tickerall_loader.py`, an HTTP+key loader mirroring the `fmp` loader, with forex symbol resolution mirroring the `mt5` loader (broker suffixes like Exness `EURUSDm` resolved from the account's own symbol list).
- `is_available()` is False unless `TICKERALL_API_KEY` and `TICKERALL_ACCOUNT_ID` are set, so the forex fallback chain degrades cleanly (`mt5` > `tickerall` > `akshare`/`yfinance`/`local`) when it is not configured. Fully additive: nothing changes for anyone who does not set the keys.

### Area

Data Sources (Backtest loaders)

### Alternatives Considered

- The existing `mt5` loader: same broker feed, but it needs a local Windows terminal, so it cannot run on Linux/macOS/headless. `tickerall` is the no-terminal complement, not a replacement.
- Generic providers (yfinance/akshare): already in the forex chain as fallbacks, but they do not carry the user's own broker symbols/session times.

### Disclosure

Full disclosure: I build and operate TickerAll, the hosted MT5 API this loader uses, so I have a commercial interest here. The integration is optional and only active when a user supplies their own key; it does not affect the existing loaders. I have implemented it (loader + registry + config + docs + a mocked-HTTP test suite) and opened a PR alongside this issue.

miguelangelo78:
## Summary

- Adds an opt-in `tickerall` forex/metals data source: a broker's MetaTrader 5 candle feed over a hosted HTTP API (TickerAll, https://tickerall.com), with no local, logged-in MT5 terminal, so forex/metals backtests run on any OS.
- Purely additive: `is_available()` is False unless `TICKERALL_API_KEY` + `TICKERALL_ACCOUNT_ID` are set, so the forex fallback chain degrades to `mt5`/`akshare`/`yfinance`/`local` when unconfigured.

## Why

Backtesting forex/metals needs the `mt5` loader today, which requires a running Windows MT5 terminal and so cannot run on Linux/macOS/headless. This adds the same broker feed with no terminal. Closes #897.

## Changes

- `backtest/loaders/tickerall_loader.py` (new): an HTTP+key loader mirroring `fmp_loader`, with forex symbol resolution mirroring `mt5_loader` (broker suffixes like Exness `EURUSDm` resolved from the account's own symbol list, memoized). Requests route through `_http.throttled_get_json` (shared throttle/session), Bearer auth. The `[start, end]` window maps to the endpoint's relative `hours` lookback and is trimmed back to the window.
- `backtest/loaders/registry.py`: register `tickerall` in `VALID_SOURCES` + the loader-module list; insert into the forex fallback chain after `mt5`.
- `src/config/env_schema.py`: `TICKERALL_API_KEY` / `TICKERALL_ACCOUNT_ID` / `TICKERALL_BASE_URL`.
- Docs: the data-routing SKILL (source table + forex routing), README (source table + forex routing + loader-tree), the agent SKILL manifest (source count + list), and `.env.example`.
- Tests: `test_tickerall_loader.py` (39 cases); the forex-chain assertion in `test_registry.py` updated for the intentional addition.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py -q`): **6472 passed, 13 skipped**. The single failure is `test_signal_alignment_perf.py::TestAlignPerformance::test_5000bars_50symbols_under_35ms`, a hardware-tuned perf-regression threshold (median ~62 ms on my machine vs the target); this PR touches no signal-alignment code (the perf test and its code path are byte-identical to the base commit), so it is pre-existing and environment-specific, not from this change. Any collection skips are optional broker deps not installed in my env (baostock/mootdx/futu/etc.), also unrelated.
- [x] New tests added: `agent/tests/test_tickerall_loader.py` (39 cases: registration + `VALID_SOURCES`/forex-chain membership, auth gating, symbol resolution + broker-suffix + fallback, candle parsing, window trimming, interval->timeframe mapping, `_symbol_names` shapes, batch resilience).
- [x] Tested manually: a live read-only fetch through the loader returned real EURUSD + XAUUSD daily bars from a demo account (no local terminal, on Linux). Per `AGENT_CONTRIBUTOR_GUIDE.md`, no live-trading or broker-write flows were run; this loader is a read-only data provider (a data-source surface, not the order/mandate/broker-write surface).

## Checklist

- [x] No hardcoded values (keys/account/base-url come from config/env; base URL defaults with an env override)
- [x] Code follows CONTRIBUTING.md (DCO `Signed-off-by`; conventional commit; mirrors the `fmp`/`mt5` loader patterns)
- [x] Documentation updated (data-routing SKILL, README, agent SKILL manifest, `.env.example`)

## Disclosure

Full disclosure: I build and operate TickerAll, the hosted MT5 API this loader uses, so I have a commercial interest here. It is optional and only active when a user supplies their own key; the existing loaders are untouched.


warren618:
Hosted MT5 data remains valuable and we want to keep this capability, but this branch must be rebuilt from current `main` rather than rebased or merged with it. The rewrite should make TickerAll explicit-only instead of an automatic fallback, require a read-only scoped credential, include account ID and base URL in cache identity, and guarantee the requested historical window or fail clearly instead of silently truncating it. Please also replace or delete the old source branch so the pre-rewrite history is no longer exposed.

warren618:
Closing this in favour of the rebuild described above. The branch still carries pre-rewrite
repository history, so GitHub's diff and merge for it aren't reliable and the review can't be
anchored to anything stable.

The capability stays wanted and is tracked in #897. A fresh PR from current `main` lands it, under
the contract from my earlier comment: TickerAll is explicit-only and never enters an automatic
fallback chain, credentials are read-only and scoped, any cache identity includes the account and
base URL, and a partial history is an error rather than a silently short series.

PR #899, opened 2026-07-28
feat(providers): add GitHub Copilot as an LLM provider

sykuang:
## Summary

- Adds **GitHub Copilot** as an LLM provider, usable with a Copilot subscription instead of an API key.
- Authenticates through the user's existing GitHub credential, with a stdlib-only OAuth device code login as fallback — no new dependency, and the GitHub CLI is not required.
- Fixes a latent bug in the settings API where provider dispatch keyed on `auth_type == "oauth"` and then hardcoded the Codex implementation behind it.

## Why

Copilot exposes an OpenAI-compatible `/chat/completions` endpoint, so users with a Copilot subscription can already reach a broad set of frontier models (Claude Opus/Sonnet, GPT-5.4, Gemini) that they are paying for. Wiring it up as a provider avoids asking those users for a second, separately-billed API key.

The credential story is deliberately simple: Copilot accepts the long-lived `gho_`/`ghu_` GitHub token directly as the Bearer credential. The `copilot_internal/v2/token` JWT exchange used by some editor integrations is **not** used here — it returns 403 for individual accounts, and skipping it means there is no ~30-minute token and therefore no refresh loop to get wrong.

## Changes

**Provider**

- `llm_providers.json` / `capabilities.py`: register `copilot` (alias `github-copilot`). Copilot rejects requests without editor-identification headers, so `Editor-Version` / `Copilot-Integration-Id` / `Openai-Intent` / `x-initiator` are declared through the existing `ProviderCapabilities.default_headers` hook rather than a bespoke client.
- `copilot_auth.py` (new): credential resolution walks `COPILOT_GITHUB_TOKEN` → `gh auth token` → `~/.config/github-copilot/apps.json` (written by the VS Code extension). When none hit, `vibe-trading provider login copilot` runs GitHub's OAuth device code flow using only the standard library. Classic PATs (`ghp_`) are rejected up front because the Copilot API does not accept them.
- `_legacy.py`: `cmd_provider_login` becomes a dispatch instead of a single-provider check. The login is idempotent — it reports the existing credential and its source rather than forcing a redundant browser round trip. The token is persisted via the existing atomic dotenv writer and never echoed.

**Settings API bug fix**

Two call sites keyed behaviour on `auth_type == "oauth"` and hardcoded Codex behind it, so any second OAuth-ish provider would have been reported using Codex's login status. Dispatch is now per `auth_type`:

- `_build_llm_settings_response` reports Copilot as configured when any local source resolves, and names the source (never the secret) in `api_key_hint`. Previously a `gh`-CLI-only install displayed as unconfigured because `COPILOT_GITHUB_TOKEN` was unset.
- The runtime env sync no longer pops `OPENAI_API_KEY` for Copilot when the env var is empty; it falls back to the resolved token. This is what made the Web UI path unusable without the env var.

**Model default**

Default is `claude-sonnet-5`. Two constraints were verified live and are pinned by tests:

- Copilot serves `gpt-5.5` / `gpt-5.6-*` / `gpt-5.3-codex` / `gpt-5.4-mini` / `gemini-2.5-pro` **only** on `/responses`; they return HTTP 400 through the OpenAI-compatible chat path this provider uses.
- `claude-sonnet-4.6` answers a tool-bound **non-streaming** request with `finish_reason="tool_calls"` but an empty `tool_calls` array (streaming is unaffected). Since `chat.py`'s non-streaming path binds tools, defaulting to it would be a trap.

Models verified working on `/chat/completions` with tool calls: `claude-opus-4.6/4.7/4.8/5`, `claude-sonnet-5`, `claude-haiku-4.5`, `gpt-5.4`, `gpt-5-mini`, `gemini-3.6-flash`.

## Out of scope

- A `/responses` adapter for the `gpt-5.5` / `gpt-5.6-*` family. It is a separate transport, not a provider-registration concern; the README documents the limitation and the tests pin the default away from it.
- Copilot Enterprise / GHE proxy endpoints. `COPILOT_BASE_URL` is honoured, but only github.com was tested.

## Test Plan

- [x] Existing tests pass — `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` → **6451 passed, 12 skipped**
- [x] New tests added — `agent/tests/test_copilot_provider.py`, 17 tests covering catalog registration, headers, the credential precedence chain, resolution without the `gh` CLI, device-flow success/denial (HTTP mocked, no network), the two model constraints, and the settings-API regressions.
- [x] Negative control: reverting the `gh_cli` dispatch branch turns the settings regression tests red, confirming they detect the bug rather than passing vacuously.
- [x] `git diff --check`, `compileall agent/cli`, `py_compile agent/api_server.py agent/mcp_server.py` all clean.
- [x] Tested manually: live round trip against `api.githubcopilot.com` through `build_llm()`; `vibe-trading provider doctor` reports the correct provider/base_url with env vars only and no `.env` file; `vibe-trading provider login copilot` correctly short-circuits on an already-authenticated machine.

Not run: the frontend build (no frontend files changed) and `e2e_backtest` (excluded per the guide).

## Risk notes

- **Secrets:** no credential is committed, logged, or echoed. The settings API exposes only the credential *source* string, never the token. `.env.example` uses a placeholder.
- **Network:** the device flow contacts `github.com/login/device/code` and `/login/oauth/access_token` only, and only when the user explicitly runs `provider login copilot`. Normal operation makes no auth calls beyond the chat request itself.
- **Blast radius:** additive. All existing providers keep their exact prior behaviour — the two settings changes are new `auth_type` branches evaluated before the pre-existing ones, and no shared code path changes shape.
- **Live/broker:** none. This PR does not touch order gates, mandates, or connectors.

## Rollback

Self-contained: `git revert` the four commits, or drop the `copilot` entry from `llm_providers.json` to disable the provider while leaving the settings-API fix in place. No migrations, no persisted state, no config schema change.

## Checklist

- [x] `src/providers/` is a protected area — flagging it explicitly. The change is additive (one new provider entry, one new module); no existing provider's resolution logic was modified. Happy to split the settings-API fix into its own PR if maintainers prefer.
- [x] No hardcoded values. The OAuth client ID is GitHub's public VS Code Copilot Chat app ID (device flow uses no client secret) and is shared by every editor integration.
- [x] Code follows CONTRIBUTING.md — all four commits carry `Signed-off-by:` (DCO).
- [x] Documentation updated — README provider list, prerequisites, env-var table, and `agent/.env.example`.


warren618:
Thanks for building this out — the provider entry, capabilities wiring, CLI/settings surface, and
302 lines of tests are all in good shape, and Copilot as a provider is something people have asked
for.

I can't take it in this form. `copilot_auth.py` hardcodes an OAuth client ID that isn't ours
(`Iv1.b507a08c87ecfe98`) and drives GitHub's device-code flow with it, then calls
`api.githubcopilot.com` with `Editor-Version: vscode/1.104.1` and a `Copilot-Integration-Id` header.
That authenticates as an application we don't own and presents as an editor we aren't. Past the
terms question, it makes the integration silently dependent on someone else's registered client,
which can be revoked or rate-limited without notice — and GitHub's own guidance is explicit that a
public client ID is not a secret and shouldn't be reused this way.

There is a supported boundary: the Copilot SDK accepts an explicit `COPILOT_GITHUB_TOKEN`, stored
Copilot CLI credentials, or `gh` credentials. A version built on that — no borrowed client ID, no
editor-impersonation headers — is welcome, and most of the work in this PR carries over unchanged.

- https://docs.github.com/en/copilot/how-tos/copilot-sdk/auth/authenticate
- https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/best-practices-for-creating-an-oauth-app

PR #900, opened 2026-07-28
feat(backtest): emit risk x-ray artifacts in portfolio runs (Portfolio Studio slice)

he-yufeng:
The risk x-ray core and the `portfolio_risk_xray` agent tool landed earlier, but a backtest run still produced no x-ray of its own: the only way to get one was to call the tool by hand with a separate basket. This wires the computation into the engine so every portfolio run emits `risk_xray.json` and `risk_xray.md` next to the rebalance notes, and surfaces the headline numbers in metrics.

The basket is derived from the target position frame as time-averaged weights (`average_invested_weights`), because the final row alone is the wrong object for strategies that end flat. The average invested fraction is reported alongside. Derivation and computation both raise `ValueError` on unusable input (short runs under the 30-bar history floor, never-invested strategies), and the engine treats that as "no x-ray for this run" rather than a failure, matching how the artifact set already varies with config.

Sample of what a run now reports in metrics: `risk_xray_hhi`, `risk_xray_effective_n`, `risk_xray_annualized_vol`, `risk_xray_max_drawdown`, `risk_xray_avg_invested`.

## Verification

- New tests: weight derivation (including the flat-strategy and empty-frame rejections), strict-JSON write plus Markdown render, and a full `run_backtest` integration test with a stub engine/loader that asserts both artifacts land and the metrics fields are set.
- `pytest agent/tests/test_risk_xray.py agent/tests/test_rebalance_notes.py`: 26/26 green.
- Engine neighborhood regression: `test_equity_regression`, `test_execution_causality`, `test_realized_turnover`, `test_terminal_close_accounting`, `test_nonpositive_prices`, `test_constraints`, `test_options_portfolio_correctness`: 73/73 green.
- Not run: the full suite and the e2e harnesses (excluded per the agent contributor guide); frontend untouched.

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -623,6 +623,33 @@ class BaseEngine(ABC):
         m["rebalance_turnover_mean"] = rebalance_notes["summary"]["turnover_mean"]
         m["rebalance_turnover_max"] = rebalance_notes["summary"]["turnover_max"]
 
+        # Portfolio Studio: risk x-ray over the strategy's average basket.
+        # Short runs and never-invested strategies raise ValueError in the
+        # derivation and simply get no x-ray artifact.
+        from backtest.risk_xray import (
+            average_invested_weights,
+            compute_risk_xray,
+            render_risk_xray_markdown,
+            write_risk_xray,
+        )
+        try:
+            basket_weights, avg_invested = average_invested_weights(target_pos)
+            risk_xray = compute_risk_xray(
+                close_df, basket_weights, periods_per_year=bars_per_year,
+            )
+        except ValueError:
+            pass
+        else:
+            write_risk_xray(run_dir / "artifacts" / "risk_xray.json", risk_xray)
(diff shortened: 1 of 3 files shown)


warren618 (approved):
Wiring reviewed: the lazy import matches the validation-slice style already in run_backtest, the ValueError-to-no-artifact path means degenerate runs stay unaffected, and write_risk_xray reuses the strict-JSON contract. Verified the engine test's numbers (even two-name basket → hhi 0.5, avg_invested 39/40 from the one-bar signal shift).

One semantic gap fixed in a supplement commit rather than a round-trip: compute_risk_xray is long-only by contract, but average_invested_weights filtered negative mean weights out *before* that contract could fire — a long-short run would have emitted an x-ray describing only its long half while presenting it as the whole strategy. Net short average exposure now raises and takes the same no-artifact path as flat and short-history runs, with a test. Extending the x-ray to long-short books can be its own slice later. Merging.

Issue #541, opened 2026-07-14
Stale EnvConfig singleton: _ensure_dotenv() loads .env but callers still read bare-environ defaults (welcome panel, /settings, diagnostic log)


marichu99:
## Summary

`_ensure_dotenv()` loads `.env` into `os.environ` but does not refresh the cached `EnvConfig` singleton, so any caller that doesn't manually pair the load with `reset_env_config()` reads bare-environ defaults (`provider=openai`, empty model) instead of the values in `~/.vibe-trading/.env`.

PR #479 (commit `1b5c873`) fixed one symptom of this — preflight's `_check_llm_provider` — by adding a `reset_env_config()` at that call site. The root cause is still present and other call sites remain broken.

## Root cause

`agent/cli/theme.py` calls `get_env_config()` at **module import time** (`_is_dark_terminal`, invoked from module scope at `theme.py:187`) to read the theme override. This builds and caches the `EnvConfig` singleton from a bare `os.environ` before any `.env` file has been loaded. `_ensure_dotenv()` later populates `os.environ` but leaves the stale singleton in place — correctness then depends on every caller remembering to call `reset_env_config()` afterward.

## Still affected on current `main`

- **Interactive welcome panel** (`_build_welcome_panel` → `_ensure_cli_env`, `agent/cli/_legacy.py`): shows `provider=openai / model=(not set)` and credential state `MISSING` despite a valid `~/.vibe-trading/.env`. (It races the async preflight thread, which does reset — so the display is intermittently correct, which makes it extra confusing.)
- **`/settings` view** (`_show_settings` → `_ensure_cli_env`, `agent/cli/_legacy.py`): same stale provider/model rendering.
- **`agent/src/ui_services.py`**: `_ensure_dotenv()` before `DataLoader` construction, no reset.
- **`_ensure_dotenv`'s own P08 R1 diagnostic log line** reads `get_env_config()` immediately after loading and therefore logs the *stale* provider/model — the exact "stale .env costing hours" confusion the diagnostic was added to prevent.

## Reproduction (v0.1.11, where preflight was also affected)

With a valid `~/.vibe-trading/.env` (`LANGCHAIN_PROVIDER=deepseek`, `LANGCHAIN_MODEL_NAME=deepseek-chat`, key + base URL set):

```
$ vibe-trading run -p "..."
Preflight Check
 N/A  │ LLM (openai)       │ LANGCHAIN_MODEL_NAME not set in .env (agent cannot function)
```

[...]

PR #902, opened 2026-07-29
fix(cli): load ~/.vibe-trading/.env on the connector command path

### LINKED ISSUE #901: [Bug] Connector CLI path never calls _ensure_dotenv() — env-sourced broker credentials always read as unset
## Environment

- Version: `main` @ `e52b835` (also reproduces on the released 0.1.12)
- Python: 3.11.15, macOS (darwin 24.6.0)
- Install: uv venv, `uv pip install "longbridge>=0.2.5"` for the optional SDK

## Summary

The `vibe-trading connector ...` subcommand group never loads `~/.vibe-trading/.env`. Every connector that resolves credentials through `get_env_config()` therefore reports them as missing, even though the same file is read correctly by other subcommands of the same CLI.

Longbridge is the visible victim (it is the only connector whose *broker credentials* come from env vars rather than a JSON config file), but `FUTU_TRADE_PWD_MD5` is affected the same way.

### Steps to Reproduce

With a valid `~/.vibe-trading/.env` containing `LONGBRIDGE_APP_KEY`, `LONGBRIDGE_APP_SECRET`, `LONGBRIDGE_ACCESS_TOKEN`, and no `~/.vibe-trading/longbridge.json`:

```
$ vibe-trading connector check longbridge-live-sdk-readonly

  Field          Value
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━
  Connector      longbridge
  Environment    live
  Transport      broker_sdk
  Configured     no

Longbridge connector not configured: missing app_key, app_secret, access_token.
```

Sourcing the exact same file into the shell first makes it pass:

```
$ set -a && . ~/.vibe-trading/.env && set +a
$ vibe-trading connector check longbridge-live-sdk-readonly
...
  Configured     yes
Connector profile is ready.
```

## Contrast within the same CLI

In a clean shell (nothing sourced), `provider doctor` resolves the same `.env` correctly:

```
$ vibe-trading provider doctor
{
  "provider": "openai-codex",
  "env": { "LANGCHAIN_PROVIDER": "set", "LANGCHAIN_MODEL_NAME": "set", ... }
}
```

So one subcommand reads `~/.vibe-trading/.env` and the other does not.

### Error Logs

```shell

```

### Interface

CLI (vibe-trading)

### LLM Provider

_No response_

### Version

0.1.12

### Environment

Python: 3.11.15, macOS (darwin 24.6.0)

MuggleJinx:
## Summary

`main()` routes the `connector` subcommand group straight to `_dispatch_connector`, which — unlike the interactive TUI — never runs preflight, so nothing populated `os.environ` from `~/.vibe-trading/.env`. `EnvConfig` reads only `os.environ`, so every connector resolving credentials through `get_env_config()` reported them as missing despite a valid file on disk: Longbridge's app key / secret / access token (the only broker credential set sourced from env vars rather than a JSON config file), and Futu's `FUTU_TRADE_PWD_MD5` live-unlock hash.

Call the existing `_ensure_cli_env()` helper at the dispatcher entry. `_ensure_dotenv()` is idempotent and loads with `override=False`, so a real environment variable still outranks the file, and connectors that read JSON config are unaffected.

Distinct from #477 / #541, which were the stale-`EnvConfig`-singleton variant: there `_ensure_dotenv()` ran but no `reset_env_config()` followed. Here it never ran at all, so that fix could not reach this path.

## Why

Closes #901

## Changes

- update `_dispatch_connector`

## Test Plan

## Checklist

### DIFF
--- a/agent/cli/_legacy.py
+++ b/agent/cli/_legacy.py
@@ -387,7 +387,7 @@ def _terminal_width() -> int:
 
 
 def _ensure_cli_env() -> None:
-    """Load dotenv values before rendering CLI-only settings."""
+    """Load dotenv values before a CLI path reads configuration."""
     try:
         from src.providers.llm import _ensure_dotenv
 
@@ -4513,6 +4513,7 @@ def cmd_connector_revoke(profile_id: Optional[str]) -> int:
 
 def _dispatch_connector(args: argparse.Namespace) -> int:
     """Route parsed ``connector`` subcommands."""
+    _ensure_cli_env()
     sub = getattr(args, "connector_command", None)
     if sub == "list":
         return cmd_connector_list()
--- a/agent/tests/test_cli_connector_env.py
+++ b/agent/tests/test_cli_connector_env.py
@@ -1,0 +1,117 @@
+"""Regression (#901): the ``connector`` CLI path must load ``~/.vibe-trading/.env``.
+
+``EnvConfig`` reads only ``os.environ``; ``_ensure_dotenv()`` is what populates it
+from ``~/.vibe-trading/.env``. ``main()`` routes the ``connector`` group straight to
+``_dispatch_connector`` without the interactive TUI's preflight, so nothing loaded
+the file — and every connector resolving credentials through ``get_env_config()``
+(Longbridge's app key / secret / access token, Futu's ``FUTU_TRADE_PWD_MD5``)
+reported them as missing despite a valid file on disk.
+"""
+
+from __future__ import annotations
+
+import argparse
+
+import pytest
+
+from cli import _legacy
+
+pytestmark = pytest.mark.unit
+
+_CREDENTIAL_ENV_VARS = (
+    "LONGBRIDGE_APP_KEY",
+    "LONGBRIDGE_APP_SECRET",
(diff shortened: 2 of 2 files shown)


he-yufeng:
Checked the two things this one-line change has to get right, both fine:

- `_ensure_dotenv` is idempotent (the `_dotenv_loaded` guard), so calling `_ensure_cli_env()` at dispatch does not double-load or reorder anything when a caller already went through preflight.
- The loader uses `override=False` / setdefault, so shell-exported credentials still win over the file, which is the precedence an operator expects on a credential path.

The ordering test pins the real regression (dispatch used to reach the subcommand with an unpopulated environ). Nothing to block from me.

warren618 (approved):
Verified the diagnosis: main() routes the connector group straight to _dispatch_connector with no preflight, so EnvConfig only ever saw the live environment — and the fix correctly reuses _ensure_cli_env() at the dispatcher entry, keeping override=False semantics (the shell-outranks-file test locks that). Clean, minimal, and the ordering test pins the load-before-read contract. Merging — thanks for the sharp report and fix.

PR #907, opened 2026-07-30
fix(options-engine): safely handle trade records with null or non-numeric pnl

santhreal:
In _calc_options_metrics, trade records with "pnl": None raise TypeError because t.get("pnl", 0) returns None (and None != 0 evaluates to True, calling float(None)). Coerce pnl safely inside a try-except block.


warren618:
Merged with your authorship, with one addition: because `trade_count` still
reports every record, silently dropping malformed PnL meant `win_rate` and
`profit_loss_ratio` could be computed from a subset without saying so — in your
own test, four trades reporting a 50% win rate derived from two usable values.
The merged version counts the excluded records and appends a warning naming the
count, using the function's existing warnings collection. Thanks for finding
the crash.

PR #908, opened 2026-07-30
fix(regime): handle zero asset pairs in compute_edge_density

santhreal:
compute_edge_density raises ZeroDivisionError when called on a returns DataFrame with fewer than two assets (n_pairs = 0). Return NaN density series when n_pairs is 0.


warren618:
Merged with your authorship. This is the focused version of the guard, so I've
closed #912 (which bundled it with the `GroupExposure` change) as a duplicate
and #914 separately — reasoning is on those issues.

PR #909, opened 2026-07-30
fix(risk-xray): handle zero total surviving weight in compute_risk_xray

santhreal:
compute_risk_xray raises ZeroDivisionError when history filtering leaves surviving symbols with zero total weight. Raise ValueError with a clear message when surviving weight is zero.


warren618:
Merged with your authorship — thank you. Raising was the right call here: the
alternatives (equal-weighting the zero-weight survivors, or returning an
all-zero x-ray) would either invent holdings the user doesn't hold or
contradict the fully-invested-basket contract.

PR #910, opened 2026-07-30
fix(options-pricing): tolerate risk_free_rate=None in OptionsPricingTool

santhreal:
OptionsPricingTool.execute raises TypeError when risk_free_rate=None is supplied in kwargs. Fall back to default rate when risk_free_rate is None.


warren618:
Superseded rather than rejected: your `risk_free_rate=None` handling is
correct and it has landed as part of the reworked #913, which rewrites the
same conversion block (and additionally catches `OverflowError`, which was
still escaping the error envelope). Your authorship is retained on that
commit.

Closing this one to avoid landing the same behavior twice — thank you.

PR #912, opened 2026-07-30
fix(constraints): prevent ZeroDivisionError when group total is 0 in GroupExposure

santhreal:
GroupExposure.apply raises ZeroDivisionError when total group weight is 0 and total > cap + TOL. Guard division by checking if total group weight is positive.


warren618:
Closing as a duplicate bundle: the `GroupExposure` hunk here is byte-identical
to #914 (and shares the same unreachable-branch problem — details on that
issue), while the edge-density guard duplicates #908.

I've merged #908, which is the focused version of the `compute_edge_density`
fix, so that half of your work is in with your authorship retained. Thanks for
catching it.

PR #913, opened 2026-07-30
fix(tools): handle null and non-finite arguments in tool helpers and execute entrypoints

santhreal:
Tool execution methods in options_pricing, web_search, session_search, qveris, report_audit, and shadow_account raise TypeError or ValueError when passed null or non-finite parameters. Use safe argument coercion before type operations.


warren618:
Merged with your authorship — both PRs landed as one commit since they
overlap.

The direction changed where an action or a charge is involved. Coercing a
malformed value to None and continuing is a reasonable fallback for a read-only
helper, but not for these three:

- trading_connector_tool: `quantity="bad"` together with `notional=100` became
  a valid $100 notional order, and an invalid port or client_id silently fell
  back to the profile default. Malformed and non-finite values are now rejected
  before place_order or cancel_order is reached, and the numeric helpers reject
  NaN and Infinity, which the original still accepted.
- qveris_tool: a missing required `parameters` became `{}`, so a schema-invalid
  request could reach client.execute and spend credits. Validation now runs
  before the quote call, the budget reservation and execute.
- report_audit_tool (#911): mapping a missing reported_value to 0.0 made it
  compare equal to a fetched 0 and count as a PASS, so unverifiable evidence
  could be certified. It is now an explicit failing item with a reason, while a
  genuine reported 0 stays valid.

options_pricing_tool also catches OverflowError now, which was still escaping
the envelope for very large integer inputs. That block subsumes your #910, so I
closed it as superseded.

One thing worth knowing for future PRs: the test file imported through
`from tools.X import ...`, which does not resolve under the repo's pytest
pythonpath — the whole file would have failed collection. The rest of the suite
imports via `src.tools.*`. The tests were also reaching the network and the real
session index; they are mocked now.

The GroupExposure hunk was dropped: caps go through `_bounded_fraction`, which
constrains them to (0, 1], and the frame path filters zero-weight rows, so that
zero-division branch is unreachable for validated input.

PR #914, opened 2026-07-30
fix(backtest): handle zero total weight in GroupExposure apply

santhreal:
When total group weight is zero during portfolio exposure cap recalculation, GroupExposure.apply raised ZeroDivisionError.

Check for zero total weight before scaling group weights.


warren618:
Thanks for the sweep through the constraint code — the surrounding hardening
in this series has been genuinely useful.

I'm closing this one, though, because the guard can't be reached by a
validated configuration. `GroupExposure` caps go through `_bounded_fraction`,
which constrains them to `(0, 1]`, so for any accepted cap the
`total > cap + _TOL` branch can never be entered with `total == 0` — reaching
it needs a negative cap, which is exactly what the added test constructs by
calling the constructor directly. Returning zeros for a negative cap also
doesn't satisfy the constraint: zero group exposure still exceeds a negative
upper bound, so the fallback would mask invalid input rather than handle it.

If you'd like to harden this area, the useful change is validating caps at the
constructor boundary (so direct construction can't create an impossible
state), rather than a fallback inside `apply`. Happy to look at that.

The edge-density half of your #912 has landed as #908 with credit.

PR #915, opened 2026-07-30
fix(stock-profile): handle non-iterable sections in StockProfileTool

santhreal:
StockProfileTool.execute raises TypeError when passed non-iterable sections arguments like integers. Validate sections is iterable before processing.


santhreal:
Closing this one. sections=None already returns all sections on main; a non-list sections value is outside the tool schema, so the TypeError path is not worth a dedicated change.

PR #916, opened 2026-07-30
fix(web-search): handle null query and max_results in WebSearchTool

santhreal:
WebSearchTool.execute raises TypeError when max_results or query is None. Coerce max_results safely and handle empty queries gracefully.


santhreal:
Closing this one. Main already rejects a missing/null query and defaults max_results=None to 5, so the new tests do not cover a live bug.

PR #917, opened 2026-07-30
fix(cli): handle None input in slash_router match_commands

santhreal:
match_commands and find_exact in slash_router raise AttributeError when passed input_str=None. Return empty matches when input_str is None.


santhreal:
Closing this one. slash_router only sees user text from the CLI, so None is outside the input contract.

PR #919, opened 2026-07-30
fix(lockup-expiry): handle integer symbol code in get_lockup_expiry

santhreal:
When LockupExpiryTool or get_lockup_expiry receives an integer stock code (e.g., code=600519, common when LLMs pass unquoted numbers in tool call JSON), calling code.strip() raises AttributeError: 'int' object has no attribute 'strip'.

This change coerces code to a string before stripping, allowing numeric stock codes to be normalized and queried correctly.


warren618:
Merged, thanks @santhreal! Defensive `str()` coercion before `.strip()` is the right call here — tool arguments arriving from JSON can legitimately be ints, and the regression test with `code=600519` locks the behavior in.

PR #892, opened 2026-07-28
feat(desktop): add Electron shell lifecycle

QCYTSN:
## Summary

Adds a source-first Electron desktop shell for the existing Vibe-Trading FastAPI and React application.

This PR is intentionally limited to the desktop host and backend lifecycle. It does not add Windows packaging, credential storage, provider/model UX changes, messaging adapters, an updater, or release infrastructure.

## Why

The existing local web interface requires users to start the backend from a terminal and then open a browser manually. This shell manages that local lifecycle while preserving the existing FastAPI and React architecture.

## Changes

- adds an Electron host under `desktop/electron/`;
- enforces a single desktop application instance;
- selects a random loopback port for every launch;
- generates a 256-bit per-launch API authentication secret;
- injects authentication only for requests to the active backend origin;
- starts the existing Vibe-Trading backend and waits for its health check;
- captures backend output in the Electron log directory;
- provides startup diagnostics, retry, and log-folder actions;
- requests authenticated graceful shutdown before applying Windows process-tree cleanup as a fallback;
- restricts renderer navigation, permissions, Node.js access, and external-window handling.

## Security boundary

The authentication secret remains in the Electron main process and the owned Python child environment. It is not exposed directly to renderer JavaScript or written to configuration and log files.

The renderer uses:

- `contextIsolation: true`
- `nodeIntegration: false`
- Chromium sandboxing
- deny-by-default permission handlers
- navigation restricted to the active loopback backend origin

A compromised renderer could still exercise authenticated application routes through the Electron session, so renderer integrity remains security-significant.

The complete process-boundary analysis is documented in `desktop/electron/THREAT_MODEL.md`.

## Validation

Validated on Windows against the current upstream `main`:

- `npm ci`
- `npm run build`
- TypeScript strict compilation
- `npm run smoke:lifecycle`
- authenticated health and session requests succeed
- unauthenticated protected requests return HTTP 401
- graceful shutdown stops the listener
- no owned Python process remains after shutdown
- `npm audit` reports zero vulnerabilities
- CycloneDX SBOM generation succeeds

A clean-Windows source startup has not been rerun for this exact branch and remains recorded as an outstanding review item.

The file inventory, dependency/license summary, and validation ledger are in `desktop/electron/REVIEW_NOTES.md`.

## Scope exclusions

This PR deliberately contains no changes to:

- the agent loop;
- providers or model configuration;
- sessions;
- frontend model UX;
- messaging or IM channels;
- Python runtime packaging;
- credential storage;
[...]

PR #923, opened 2026-07-31
feat(desktop): add Electron shell lifecycle

QCYTSN:
## Rebuild note

This is the clean replacement for #892 after the upstream history rewrite invalidated that PR's merge base. It is rebuilt directly on upstream `main` at `261f007c410f7a6ff015a17f6830c8f809cd7413` and contains only DCO-signed commits.

The current diff is 19 files (`.gitignore`, the Windows CI job, and the source-only shell under `desktop/electron/`), with no source overlay and no changes to protected core paths.

## Summary

Adds a source-first Electron desktop shell for the existing Vibe-Trading FastAPI and React application.

This PR is intentionally limited to the desktop host and backend lifecycle. It does not add Windows packaging, credential storage, provider/model UX changes, messaging adapters, an updater, or release infrastructure.

## Why

The existing local web interface requires users to start the backend from a terminal and then open a browser manually. This shell manages that local lifecycle while preserving the existing FastAPI and React architecture.

## Changes

- adds an Electron host under `desktop/electron/`;
- enforces a single desktop application instance;
- selects a random loopback port for every launch;
- generates a 256-bit per-launch API authentication secret;
- injects authentication only for requests to the active backend origin;
- starts the existing Vibe-Trading backend and waits for its health check;
- captures backend output in the Electron log directory;
- provides localized startup diagnostics, retry, and log-folder actions in en/zh-CN/ja/ko/ar, including Arabic RTL layout;
- requests authenticated graceful shutdown before applying Windows process-tree cleanup;
- adds a separate parent-death watchdog so the Python tree is removed even when only the Electron main PID is force-terminated;
- constrains backend executable resolution to an explicit override, exact application/resource paths, a marker-anchored source virtual environment, and finally `PATH`;
- adds a clean `windows-2025` source-lifecycle CI job and targeted lifecycle/security regression tests;
- restricts renderer navigation, permissions, Node.js access, and external-window handling.

## Security boundary

The authentication secret remains in the Electron main process and the owned watchdog/Python child environment. It is not exposed directly to renderer JavaScript or written to configuration and log files.

Backend discovery never evaluates generic executable candidates while walking ancestors. Packaged mode checks only fixed application/resource paths. Source-mode `.venv` discovery requires a `pyproject.toml` whose `[project].name` is `vibe-trading-ai`; the final fallback uses non-empty `PATH` entries.

The renderer uses:

- `contextIsolation: true`
- `nodeIntegration: false`
- Chromium sandboxing
- deny-by-default permission handlers
- navigation restricted to the active loopback backend origin

A compromised renderer could still exercise authenticated application routes through the Electron session, so renderer integrity remains security-significant.

The complete process-boundary and executable-resolution analysis is documented in `desktop/electron/THREAT_MODEL.md`.

## Validation

Validated on Windows against the current PR source:

- `npm ci`
- `npm run build`
- TypeScript strict compilation
- `npm run smoke:lifecycle`
- five-locale key, placeholder, fallback, and RTL parity
- authenticated health and session requests succeed
- unauthenticated protected requests return HTTP 401
- graceful shutdown stops the listener and leaves no owned Python process
- a second Electron launch does not create a second backend
- force-terminating only the Electron main PID leaves no owned Python process or listening port
- explicit, packaged, marker-anchored source, and `PATH` backend resolution order
- planted unmarked-ancestor executable, wrong project marker, and packaged source-discovery regressions
- missing-backend startup produces an actionable diagnostic
- `npm audit` reports zero vulnerabilities
- CycloneDX SBOM generation succeeds

The clean-Windows job installs Python 3.12 and Node 22 on a fresh `windows-2025` runner, installs Vibe-Trading from the checked-out source, then runs `npm ci`, `npm run build`, and `npm run smoke:lifecycle`.

The file inventory, dependency/license summary, and validation ledger are in `desktop/electron/REVIEW_NOTES.md`.

## Scope exclusions

This PR deliberately contains no changes to:

- `agent/src/agent/`;
- `agent/src/session/`;
- `agent/src/providers/`;
- the live-trading mandate gate, kill switch, or audit ledger;
- frontend model UX;
- messaging or IM channels;
- Python runtime packaging;
- credential storage;
- application updates or release feeds.

## Unsigned-build limitations

This is a source-only desktop shell. It does not produce or publish an installer and does not claim an official HKUDS desktop release channel. Packaging, code signing, credential storage, and release ownership are reserved for separate review.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,3 +114,41 @@ jobs:
           agent/tests/test_background_tools.py
           agent/tests/test_bash_tool.py
           -q
+
+  desktop-electron-windows:
+    name: Windows desktop source lifecycle
+    runs-on: windows-2025
+    timeout-minutes: 30
+
+    steps:
+      # actions/checkout v7.0.1
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+
+      # actions/setup-python v7.0.0
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97
+        with:
+          python-version: "3.12"
+          architecture: x64
+          cache: pip
+
+      # actions/setup-node v7.0.0
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020
+        with:
+          node-version: "22"
+          cache: npm
+          cache-dependency-path: desktop/electron/package-lock.json
+
+      - name: Install Vibe-Trading source backend
+        run: python -m pip install -e .
+
+      - name: Install desktop dependencies
+        working-directory: desktop/electron
+        run: npm ci
+
+      - name: Build desktop shell
+        working-directory: desktop/electron
+        run: npm run build
+
+      - name: Run desktop lifecycle and parent-death smoke tests
+        working-directory: desktop/electron
+        run: npm run smoke:lifecycle
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,10 @@ AGENTS.md
 
 .DS_Store
 
+# Desktop shell build outputs
+desktop/electron/node_modules/
+desktop/electron/dist/
+
 # Alpha Zoo wiki artifacts (generated, not source)
 wiki/alpha-library/manifest.json
 wiki/alpha-library/content/
--- a/desktop/electron/README.md
+++ b/desktop/electron/README.md
@@ -1,0 +1,66 @@
+# Vibe-Trading Desktop (Unofficial Community Build)
+
+This directory contains a source-only Electron host for the existing
+Vibe-Trading FastAPI and React application. It does not contain an installer,
+an embedded Python runtime, credential storage, an updater, or changes to the
+agent, provider, session, frontend model UX, or messaging-channel code.
+
+## What the shell does
+
+- Enforces a single desktop application instance.
+- Starts one owned `vibe-trading serve` process.
+- Selects a free loopback port and binds the backend to `127.0.0.1`.
+- Generates a 256-bit authentication secret for each desktop process.
+- Adds the secret to same-origin renderer requests without exposing its value
+  to page JavaScript.
+- Waits for the authenticated `/health` endpoint before loading the UI.
+- Captures backend output in Electron's per-user log directory.
+- Reports startup failures and exposes retry and log-folder actions.
+- Requests authenticated graceful shutdown before terminating the owned
+  Windows process tree as a fallback.
+- Prevents in-window navigation away from the local backend origin.
+
+See [THREAT_MODEL.md](THREAT_MODEL.md) for the trust boundary and residual
+risks. See [REVIEW_NOTES.md](REVIEW_NOTES.md) for the file inventory and
+validation ledger.
+
+## Run from source
+
+From a complete Vibe-Trading checkout:
+
+```powershell
+python -m venv .venv
+.\.venv\Scripts\python.exe -m pip install -e .
+
+cd frontend
+npm ci
+npm run build
+
+cd ..\desktop\electron
+npm ci
+npm start
+```
+
+The shell searches for a repository `.venv\Scripts\vibe-trading.exe` before
+falling back to `vibe-trading.exe` on `PATH`. To select an explicit development
+backend:
+
+```powershell
+$env:VIBE_TRADING_EXECUTABLE = "C:\path\to\vibe-trading.exe"
+npm start
+```
+
+## Review scope
+
+This first change intentionally excludes:
+
+- Electron Builder, NSIS, release workflows, and embedded Python;
+- `safeStorage`, credential migration, and settings UI changes;
+- update checks or release-feed configuration;
+- provider/model discovery and response metadata;
+- optional IM adapters and personal WeChat pairing;
+- changes outside `desktop/`, except the two generated-output entries in the
+  repository `.gitignore`.
+
+The desktop shell remains unofficial. No release or distribution ownership is
+implied by this source directory.
--- a/desktop/electron/README.md
+++ b/desktop/electron/README.md
@@ -57,9 +57,13 @@ It verifies localized message parity, graceful shutdown, authentication, and
 the parent-death path by force-terminating only the Electron main PID before
 checking that the Python process and loopback listener are gone.
 
-The shell searches for a repository `.venv\Scripts\vibe-trading.exe` before
-falling back to `vibe-trading.exe` on `PATH`. To select an explicit development
-backend:
+Backend resolution is intentionally narrow. It checks an explicit
+`VIBE_TRADING_EXECUTABLE` override first, then exact application/resource
+locations used by packaged builds. In source mode only, it may use
+`.venv\Scripts\vibe-trading.exe` from an ancestor containing this project's
+`pyproject.toml` (`[project].name = "vibe-trading-ai"`). Its final fallback is
+`vibe-trading.exe` on `PATH`; arbitrary ancestor directories and the drive root
+are never searched for executables. To select an explicit development backend:
 
 ```powershell
 $env:VIBE_TRADING_EXECUTABLE = "C:\path\to\vibe-trading.exe"
--- a/desktop/electron/README.md
+++ b/desktop/electron/README.md
@@ -8,7 +8,8 @@ agent, provider, session, frontend model UX, or messaging-channel code.
 ## What the shell does
 
 - Enforces a single desktop application instance.
-- Starts one owned `vibe-trading serve` process.
+- Starts one owned `vibe-trading serve` process through a parent-death
+  watchdog.
 - Selects a free loopback port and binds the backend to `127.0.0.1`.
 - Generates a 256-bit authentication secret for each desktop process.
 - Adds the secret to same-origin renderer requests without exposing its value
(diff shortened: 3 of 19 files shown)


warren618 (changes requested):
The loopback binding, per-launch key, exact-origin header injection, sandbox, navigation restrictions, and graceful shutdown path are a solid baseline. Three blockers remain:

1. The current repository CI does not run `npm ci`, `npm run build`, or `npm run smoke:lifecycle` in `desktop/electron`, so the green checks do not validate this new package. Please add a Windows CI job and complete the clean-Windows source-start validation still left unchecked in REVIEW_NOTES.md.

2. The loading page, runtime status/errors, dialogs, and application menu are hardcoded in Chinese (`lang="zh-CN"`). Vibe-Trading defaults to English and requires en/zh-CN/ja/ko/ar parity, including RTL behavior where applicable. Please route all user-facing desktop strings through a locale layer with all five locales.

3. The claimed force-termination guarantee is not implemented structurally. `taskkill /T /F` is called only inside `BackendManager.stop()`, which runs during normal JavaScript shutdown; hard-killing only the Electron main process bypasses that cleanup. Please use a Windows Job Object with kill-on-close or a backend parent-death watchdog, and add an automated test that kills only the Electron main PID (not the whole process tree) before asserting that the Python child and listener are gone.

QCYTSN:
Addressed all three blockers in `8883a59a4ff74080ed6786592dddf0c29a624ba9`.

1. **Windows desktop CI and clean source start**
   - Added the `Windows desktop source lifecycle` job on a fresh `windows-2025` runner with Python 3.12 and Node 22.
   - The job runs `npm ci`, `npm run build`, and `npm run smoke:lifecycle` in `desktop/electron` after installing Vibe-Trading from the checked-out source.
   - The new job passed, as did the existing repository test and Windows background jobs: https://github.com/HKUDS/Vibe-Trading/actions/runs/30820654542
   - Updated `REVIEW_NOTES.md` to record this reproducible clean-Windows source validation.

2. **Five-locale desktop i18n**
   - Routed loading-page, runtime status/error, dialog-title, and application-menu strings through one typed locale layer.
   - English is the fallback/default; `en`, `zh-CN`, `ja`, `ko`, and `ar` have identical key and placeholder coverage.
   - Arabic selects `dir=rtl`, with RTL action ordering on the loading page.
   - Added an automated locale parity, fallback, placeholder, and direction check to the lifecycle suite.

3. **Structural parent-death cleanup**
   - Added a backend watchdog that verifies the Electron parent is alive before spawning Python, retains an IPC connection, and also polls the exact parent PID.
   - Parent IPC disconnect or death terminates the owned Python process tree independently of Electron's JavaScript shutdown callbacks.
   - Added an automated Windows test that launches the actual Electron main process, runs `taskkill /PID <main-pid> /F` without `/T`, and then separately asserts that the Python PID and loopback listener are gone.
   - This test passed locally and in the new clean-Windows CI job.

Local validation also passed: TypeScript build, locale parity, authenticated lifecycle/graceful shutdown, single-PID parent-death cleanup, `npm audit` (0 vulnerabilities), YAML parsing, and diff checks.

warren618 (changes requested):
This is a really thorough response — thank you for taking all three seriously rather than arguing them down. I checked each against the code rather than reading the summary:

1. `Windows desktop source lifecycle` runs `npm ci` / `npm run build` / `npm run smoke:lifecycle` on a clean `windows-2025` runner against a source install, and passes at 3m43s alongside the three existing jobs.
2. `locales.ts` carries 28 keys in each of en/zh-CN/ja/ko/ar with identical key sets and identical `{placeholder}` sets per key, `ar` resolves to `dir="rtl"`, and no Chinese literals remain under `desktop/electron/src/`.
3. `backend-watchdog.ts` is a genuine structural guarantee — a separate process holding the IPC channel plus a parent-PID poll, so the Python tree still goes away when Electron's JavaScript shutdown path never runs. And `smoke-parent-death.mjs` kills only the main PID (`/PID <pid> /F`, no `/T`) before asserting that both the backend PID and the loopback listener are gone. That's exactly the shape I was hoping for, and it's a nicer solution than the Job Object I suggested.

There's one more thing I'd like to sort out before this merges. It's in code that was already in your first commit, so it isn't a regression from this round — I simply missed it last time, and I'm sorry for surfacing it only now.

**`BackendManager.resolveBackend()` could end up starting a binary that someone else placed.** It seeds `roots` with `appPath`, `resourcesPath`, `__dirname`, and `process.cwd()`, then walks up to 10 ancestors of each — which on Windows reaches the drive root — and for every root tries `backend\python.exe`, `runtime\backend\python.exe`, `backend\Scripts\vibe-trading.exe`, `.venv\Scripts\vibe-trading.exe` and so on, taking the first hit. The `PATH` scan only happens after all of that.

Under default Windows ACLs a standard user can create a folder at the drive root, which would make `C:\backend\python.exe` plantable without elevation. It matters most in exactly the flow this PR serves: someone who ran `pip install -e .` into a global or conda environment has no `<repo>\.venv`, so no legitimate candidate matches, and the planted path would be reached before the `PATH` entry holding the real `vibe-trading.exe`. The shell would then start it and pass it `API_AUTH_KEY`.

`THREAT_MODEL.md` already names "the integrity of the executable selected as the backend" as a protected asset, but the text only covers the explicit `VIBE_TRADING_EXECUTABLE` override — so the implicit search just isn't described yet.

What I'd suggest: resolve from `appPath` / `resourcesPath` (and their packaged `resources` subtree), plus `VIBE_TRADING_EXECUTABLE`, plus a `PATH` lookup, without the ancestor walk toward the drive root — and extend `THREAT_MODEL.md` to cover implicit resolution. If you'd like to keep a source-tree convenience path for development, anchoring it on a marker file (a `pyproject.toml` naming this project) rather than on any directory that happens to contain `backend\python.exe` would work well.

One caveat so you can weigh this yourself rather than take my word for it: I read the resolution order out of the code and confirmed the `THREAT_MODEL.md` wording, but I haven't verified the default drive-root ACL on an actual Windows machine. If it turns out that planting there needs elevation on a default install, tell me and I'll happily re-rank it.

Much smaller, and genuinely optional: `VIBE_TRADING_DESKTOP_FAST_START=1` is set in the backend child environment, but nothing in this repository reads it — might be worth wiring up or dropping.

QCYTSN:
Addressed the backend executable-resolution blocker in `3815c1bb893cbc1539919597b83e7d54f69f998c`.

### Resolution boundary

- Removed the generic ancestor candidate walk and all `process.cwd()`-based discovery.
- Resolution order is now:
  1. an existing explicit `VIBE_TRADING_EXECUTABLE` override;
  2. fixed candidates anchored to the exact Electron application/resources paths (including fixed packaged `app` / `resources` subtrees);
  3. in source mode only, `<marked-root>\.venv\Scripts\vibe-trading.exe`, where the ancestor `pyproject.toml` must contain `[project].name = vibe-trading-ai`;
  4. non-empty `PATH` entries.
- Packaged mode disables source-ancestor discovery entirely, and the filesystem root is never evaluated as a source marker or executable root.
- This removes the unsafe candidate regardless of the machine's drive-root ACL, so the fix does not depend on resolving that caveat.

### Regression coverage and documentation

- Added `test-backend-resolution.mjs` covering explicit override precedence, exact packaged resources, marker-anchored source discovery, rejection of an executable in an unmarked ancestor, rejection of a wrong project marker, and disabled source discovery in packaged mode.
- Added that test to `npm run smoke:lifecycle`, so the clean `windows-2025` job exercises it.
- Extended `THREAT_MODEL.md` with implicit resolution controls and the residual trust placed in an explicit override and `PATH`.
- Updated the README, review inventory, validation ledger, and PR description.

I also dropped the unused `VIBE_TRADING_DESKTOP_FAST_START` environment variable. During final validation, npm's current advisory database flagged the transitive `undici@7.28.0`; the lock file now resolves `7.29.0`, and `npm audit` is back to zero findings.

Local Windows validation passed after a clean `npm ci`: TypeScript build, locale parity, all backend-resolution regressions, authenticated lifecycle/graceful shutdown, single-main-PID parent-death cleanup, and `npm audit` (0 vulnerabilities). CI has been triggered for the updated head.

QCYTSN:
CI is now fully green on the updated head: all four jobs passed, including the clean Windows desktop source lifecycle with the new executable-resolution regressions (3m5s). Run: https://github.com/HKUDS/Vibe-Trading/actions/runs/30868677189

warren618 (approved):
Reviewed the current head (`3815c1bb`) against current `main`. The earlier backend-selection and parent-death concerns are addressed: backend resolution is now explicit override → packaged roots → source `.venv` only under a matching `pyproject.toml`, and the watchdog terminates the Python child when the Electron parent exits. Local Electron build, locale-parity test, backend-resolution test, and `npm audit` all pass; the current GitHub checks, including Windows, are green. No blocking findings.

QCYTSN:
Thank you for the careful review and for merging the lifecycle boundary. I have now rebuilt the requested Windows packaging layer on current main and opened draft PR #1015. It keeps the updater, release feed, optional IM bundles, personal WeChat pairing, and broker extras out of scope; the unsigned review installer is not uploaded. The new Windows job also launches the real packaged application and verifies backend cleanup after killing only the Electron main PID. Signing identity and release ownership remain explicit gates before any user-facing installer.

PR #924, opened 2026-07-31
feat(llm): add model discovery and runtime identity UX

QCYTSN:
## Summary

- Add authenticated provider model discovery and an editable model picker that always exposes the full loaded model list.
- Show the configured provider, model, and reasoning setting in the chat UI, and show response duration on each completed assistant reply.
- Record the provider-returned model identifier in reply metadata, and make reply copying reliable with a clipboard fallback.

## Why

This is the independently useful UX work requested as **PR 3** in [Discussion #760](https://github.com/HKUDS/Vibe-Trading/discussions/760).

Model identity should come from runtime configuration and provider response metadata, rather than from asking a model to describe itself. Users also need to see all models returned by a provider without first editing or clearing the current model name.

This branch was rebuilt from current `main` (`261f007c410f7a6ff015a17f6830c8f809cd7413`) without the earlier 0.1.11 overlay. It contains four DCO-signed commits.

## Changes

- Add `POST /settings/llm/models` for authenticated OpenAI-compatible model discovery.
  - Only HTTP(S) endpoints without URL-embedded credentials are accepted.
  - A saved API key is sent only to the matching saved/default/known provider endpoint.
  - An arbitrary unsaved custom endpoint does not receive the stored key.
  - Discovery falls back to the configured/default model when a provider does not expose a usable model list.
- Add an accessible editable model combobox.
  - Loaded models remain visible even when the current typed value does not match them.
  - Keyboard navigation, custom model names, refresh, loading, empty, and error states are covered.
- Show runtime identity as provider + configured/returned model + reasoning setting.
  - Explicit `none` is distinguished from provider default.
  - Stale identity is cleared when switching to a session without runtime metadata.
- Persist reply runtime metadata and include it in the terminal SSE event:
  - provider
  - configured model
  - provider-returned model, when available
  - model source
  - reasoning effort
  - elapsed milliseconds
- Display response duration on each completed assistant reply. Token usage is intentionally not added to the UI.
- Fix reply copying by using the Clipboard API first and a textarea/`execCommand` fallback when necessary.
- Update the supported frontend locales and add focused backend/frontend tests.

### Protected-core disclosure

This PR touches:

- `agent/src/agent/loop.py`
- `agent/src/providers/chat.py`
- `agent/src/session/service.py`

These changes are limited to capturing and propagating runtime response metadata approved for PR 3 in [Discussion #760](https://github.com/HKUDS/Vibe-Trading/discussions/760). They do **not** change the system prompt, provider selection policy, tool permissions, agent planning behavior, live-trading mandate gate, kill switch, or audit ledger.

### Explicitly out of scope

- Electron shell and desktop lifecycle
- Windows packaging or credential storage
- Updater behavior
- IM adapters or personal WeChat pairing
- Token-usage display
- Agent identity/system-prompt changes

## Test Plan

Completed locally on Windows:

- Frontend: **46 files / 379 tests passed**
- Frontend production build: passed (existing chunk-size warning only)
- Targeted backend tests: **52 passed**
- Ruff on changed Python files: passed (two pre-existing malformed `noqa` warnings in `settings_routes.py`)
- Full backend suite: **6785 passed, 24 skipped, 15 failed, 9 errors**
  - A clean worktree of the exact upstream `main` reproduced **14 failures and the same 9 errors** under this Windows environment (symlink privilege, POSIX mode, `SIGALRM`, Windows file locking/path behavior, and related platform assumptions).
  - The one additional branch result was a 50 ms performance-threshold test at 52.4 ms; its immediate isolated rerun passed.

Focused tests cover response-model extraction, reply metadata persistence/SSE propagation, discovery authorization and credential boundaries, complete model-list display, keyboard behavior, stale-session metadata clearing, explicit reasoning `none`, runtime display, and copy fallback behavior.

## Checklist

- [x] Changes to protected areas were discussed and explicitly requested in [Discussion #760](https://github.com/HKUDS/Vibe-Trading/discussions/760)
- [x] User-facing strings updated across supported locales


warren618 (changes requested):
Thanks for the thorough implementation. I validated the exact current-main synthetic merge: 41 targeted backend tests passed, all 46 frontend test files / 379 tests passed, and the production build completed successfully.

Before merge, please address three correctness/i18n blockers:

1. agent/src/agent/loop.py:1046-1054 reads mutable global settings after the run completes. These values may not describe the ChatLLM instance that actually served the attempt when an explicit model override is used or settings change concurrently. Snapshot the effective provider, configured model, and reasoning effort when ChatLLM is constructed, then emit that immutable snapshot.

2. ModelRuntimeBar combines persisted historical provider/model metadata with the current settings reasoning effort. Agent.tsx also leaves the previous session identity visible until the next REST load completes, or indefinitely if that load fails. Persist and consume the runtime reasoning value, and clear runtime identity synchronously on every session transition and load failure.

3. Model-discovery warning prose is returned in English by the backend and rendered directly, while ModelPicker appends a hard-coded " options" suffix to ARIA labels. Replace backend prose with stable warning codes and translate all user-facing text through the five supported locales.

Please add regressions covering configuration mutation after ChatLLM construction, explicit model overrides, historical reasoning metadata, and failed/slow session transitions.

QCYTSN:
Addressed the requested changes in `11af7292d08533b275d1ae90276fb81645330f56`:

- Added an immutable `ChatLLM` runtime snapshot for the effective provider, configured model (including explicit overrides), and reasoning effort; `AgentLoop` now emits that snapshot instead of reading mutable settings after the run.
- Persisted and consumed historical reasoning metadata, keyed runtime identity to the active session, and synchronously cleared it during session transitions and on load failure.
- Replaced backend discovery prose with stable warning codes, translated those warnings across all five supported locales, and removed the hard-coded English ARIA suffix from `ModelPicker`.
- Added regressions for post-construction configuration mutation, explicit model overrides, historical reasoning metadata, and slow/failed session transitions.

Validation after the fix:

- Focused backend regressions: **28 passed**
- Expanded affected backend suite: **53 passed**
- Focused frontend regressions: **4 files / 13 tests passed**
- Full frontend suite: **47 files / 384 tests passed**
- Frontend production build: passed (existing chunk-size warning only)
- Ruff on changed Python files: passed

The latest workflow is currently waiting for maintainer approval to run for the new fork commit.

warren618 (approved):
Re-reviewed `11af7292` as a synthetic merge with current `main` (`4da304fe`). The three prior blockers are resolved:

- runtime provider/model/reasoning metadata is snapshotted immutably, including explicit model overrides;
- historical reasoning identity is persisted and session-scoped, with synchronous clearing on slow/failed transitions;
- model-discovery warnings use stable codes, and all warning/ARIA text is translated across the five locales.

There are no unresolved review threads.

Validation on the current-main merge: 7,029 backend tests passed (14 skipped), all 47 frontend files / 384 tests passed, production build passed, and Ruff, Python compilation, and diff checks passed. GitHub Actions also passed on the unchanged PR head. Approving.

PR #925, opened 2026-07-31
fix: store session and run history under the user runtime root, not the code

### LINKED ISSUE #904: Session and run history is stored relative to the code, not the user
## Environment

- Version: `main` @ `0e004d7`; the path expressions are identical in the released 0.1.12 wheel
- Python 3.11.15, macOS

## Summary

Session and run history is stored **relative to where the code lives**, not under the user-level runtime root. Two consequences:

1. On a normal `pip install vibe-trading-ai`, `sessions/`, `runs/`, and `.swarm/runs/` are created **inside `site-packages/` itself**.
2. In a dev checkout, history is bound to whichever clone the editable install points at, so moving or re-pointing the install silently orphans it.

This is inconsistent with the project's own convention: `~/.vibe-trading/` already holds `sessions.db`, `memory/`, `reports/`, `live/`, `shadow_accounts/`, and `workspace/`.

## The paths

```python
# agent/cli/_legacy.py:56-59
AGENT_DIR    = Path(__file__).resolve().parents[1]
RUNS_DIR     = AGENT_DIR / "runs"
SESSIONS_DIR = AGENT_DIR / "sessions"
SWARM_DIR    = AGENT_DIR / ".swarm" / "runs"

# agent/src/agent/loop.py:50
RUNS_DIR = Path(__file__).resolve().parents[2] / "runs"
```

In the repo, `parents[1]` from `agent/cli/` is `agent/` — the intended layout. But `[tool.setuptools] package-dir = {"" = "agent"}` publishes `cli`, `src`, and `backtest` as **top-level** packages, so there is no `agent/` directory in an installed distribution and the same expression resolves one level higher.

## Reproduction

Probing the published wheel without touching a real environment:

```console
$ uv pip install --no-deps --target /tmp/probe vibe-trading-ai
 + vibe-trading-ai==0.1.12
$ ls /tmp/probe
api_server.py  backtest  bin  cli  mcp_server.py  src  vibe_trading_ai-0.1.12.dist-info
```

Evaluating the shipped constants against that layout:

```
AGENT_DIR    = /tmp/probe                 # i.e. site-packages
SESSIONS_DIR = /tmp/probe/sessions
RUNS_DIR     = /tmp/probe/runs
SWARM_DIR    = /tmp/probe/.swarm/runs
```

So an end user's chat sessions, generated strategy code, backtest metrics, and execution traces are written into `site-packages/`.

In a dev checkout the same mechanism binds history to the code location rather than the user. Running `vibe-trading` from `$HOME` with an editable install still resolves to the checkout:

```
$ cd ~ && python -c "from cli._legacy import SESSIONS_DIR; print(SESSIONS_DIR)"
/Users/me/Repo/Vibe-Trading/agent/sessions
```

## Why this matters

- **Upgrades and reinstalls.** `pip install -U` / `uv tool upgrade` replaces the package directory. Anything the tool wrote there is at best carried along by accident and at worst removed; `pip uninstall` leaves the directory behind as untracked residue because these files are not in the RECORD.
- **Split-brain with the search index.** `~/.vibe-trading/sessions.db` (`agent/src/session/search.py:23`) is user-level and stores message text, so it survives a reinstall — while the canonical session records under `SESSIONS_DIR` and every artifact under `RUNS_DIR` do not. Search can therefore surface a session whose `/show`, `/code`, and `/trace` no longer resolve.
- **Multiple checkouts.** Contributors with more than one clone get history silently partitioned by which one the editable install points at, with no indication in the UI which store is active.
- **`git clean -xdf`** in a checkout deletes session and run history, since both directories are gitignored.

## Suggested direction

Route all three through the existing `get_runtime_root()` (`agent/src/config/paths.py:10`), which already returns `~/.vibe-trading`, so they join `sessions.db`, `memory/`, and `reports/` under one user-level root.

Open questions for the maintainers, since this is a behavior change rather than a bug fix:

- **Migration.** One-time move of an existing `<install>/sessions`, `<install>/runs`, and `<install>/.swarm/runs` into the runtime root on first run, or leave existing data in place and only write new records to the new location?
- **Override.** Worth an env var (e.g. `VIBE_TRADING_HOME` or a `VIBE_RUNS_DIR`) so multi-project users can keep separate stores deliberately? `get_runtime_root()` currently has no env override at all.
- **Scope.** `SWARM_DIR` and the `api/state.py` host-attr indirection (`_host_attr("SESSIONS_DIR", ...)`) suggest some callers already expect to override the location — I have not traced whether anything depends on the current placement.

Happy to open a PR once there's a preferred answer on migration and the override knob. Related in kind (packaged install vs. repo layout) but not in cause: #55.

MuggleJinx:
Fixes #904.

## Problem

`sessions/`, `runs/`, `.swarm/runs/`, and `uploads/` were resolved relative to the installed code. On a `pip install`, that is `site-packages/` itself — chat history, generated strategies, and backtest artifacts were written there, invisible to `~/.vibe-trading/sessions.db` (split-brain with search) and effectively lost on upgrade. In dev checkouts, history was silently partitioned per clone.

## What this does

**Single source of truth** — `src/config/paths.py` gains `get_sessions_dir()` / `get_runs_dir()` / `get_swarm_runs_dir()` / `get_uploads_dir()`, all under `get_runtime_root()` (`~/.vibe-trading`). Every code-relative definition site now derives from them — the audit found **eight** in total, six more than the issue listed: `cli/_legacy.py`, `src/agent/loop.py`, `src/api/helpers.py`, `src/api/uploads_routes.py`, `src/api/swarm_routes.py`, `src/swarm/store.py` (`swarm_runs_root()`), an independent re-derivation in `src/tools/swarm_tool.py`, and the `TraceWriter.find_trace_dir` defaults. `.swarm/runs` becomes `swarm/runs` — hiding a directory inside an already-hidden root serves no purpose. The `sessions.db` FTS index and the goal-store default also derive from the runtime root now, so the index always lives beside the sessions it indexes.

**Answers to the issue's open questions:**

- **Migration: one-time, child-level, crash-safe.** New `src/config/migrate.py`, wired into all three process entry points (CLI, `api_server` startup, `vibe-trading-mcp`). Each session/run/upload entry moves individually via a `.migrating-*` staging name plus an atomic rename — an interrupted move is recovered on the next run (redone while the source survives, promoted once it is gone) and never surfaces half-copied. Only a same-named entry already present at the destination is skipped (ids are timestamped/random, so a real collision means "already migrated") — necessary because autopilot runs and channel media already populate `~/.vibe-trading/{runs,uploads}` before any migration, and a whole-directory skip would strand those users' legacy history. A directory containing `__init__.py` is never touched (`site-packages/runs` may be someone else's distribution), one failing directory never blocks the rest, and failures are logged, never fatal to startup.
- **Override: `VIBE_TRADING_HOME`.** Relocates sessions, runs, swarm runs, uploads, and `sessions.db` (explicit `config_path` still wins); UNC paths are rejected, matching the hardening on the other root-widening env vars. Documented in `.env.example` as shell-env-only, since module-level constants bind at process start, before `.env` is read.
- **Scope: `_host_attr` traced.** The `_host_attr("SESSIONS_DIR", ...)` indirection exists purely so tests can monkeypatch `api_server` attributes; nothing depends on per-install placement. Constants stay module-level (only their initialization changed), so every existing monkeypatch-based test passes unmodified.

**Follow-through the paths change forces:**

- Browser-upload handles (`uploads/<name>`) resolve to the user-level uploads dir, and the file/write/run sandbox allow-lists gain the runtime-root variants while keeping every legacy root (including `agent/.swarm/runs`) — pure additions, so un-migrated data stays reachable and nothing existing is revoked.
- The welcome banner's *Workspace* row now reports the runtime root — the UI shows which store is active, which the issue called out as missing.

Docs: `README.md` + zh/ja/ko/ar localizations and `.env.example` describe the new layout and the override.

## Review

The branch went through an adversarial multi-agent review (two independent reviewers + verification of every finding) before the second commit; that pass caught the `swarm_routes.py` site, the destination-already-populated migration gap, and the UNC/env-split issues fixed in `56a82ea`.

## Testing

- 40+ dedicated tests across `test_user_state_paths.py`, `test_state_migration.py`, `test_state_migration_wiring.py` (written test-first): env-override precedence and UNC rejection, helper resolution, migration semantics (merge, collision, package guard, error isolation, staging recovery in both crash windows, symlinked destination, idempotency, legacy-root correctness), entry-point wiring for all three processes, upload-handle resolution, sandbox allow-list follow, API swarm-store root, `sessions.db` override follow, banner display.
- Full suite: **6843 passed, 0 failed**, 20 skipped. `ruff check` clean vs. baseline.


warren618:
Merged, thanks @MuggleJinx — this closes #904. The migration design is careful in all the ways that matter here: child-level merge so partially-populated destinations don't block, `.migrating-*` staging with atomic rename and recovery of interrupted moves, skip-if-it-looks-like-a-package protection for `site-packages`, and startup that can never be blocked by a failed move. `VIBE_TRADING_HOME` fills the zero-override gap on `get_runtime_root()` too. Really solid work.

PR #926, opened 2026-07-31
fix(pattern-tool): use abs(avg) when evaluating double top peak difference

santhreal:
When prices or return series carry negative values, double_top_bottom evaluated abs(v1 - v2) / avg < 0.03 with a negative average, which yielded a negative ratio and falsely flagged divergent negative peaks as a double top pattern. This updates the peak comparison in pattern_tool.py to use abs(avg) (matching the double-bottom valley logic) and adds a regression test.


warren618:
Merged, thanks @santhreal! Nice catch — the peaks loop now matches the valleys loop's `abs(avg)`, and the negative-price regression test proves the false double-top case is gone.

PR #927, opened 2026-07-31
fix(metrics): guard calendar_days calculation when equity_curve has non-datetime index

santhreal:
When bars_per_year=None (cross-market auto-detect mode) is passed to calc_metrics(), (last - first).days failed with AttributeError when equity_curve has an integer or non-datetime index. This guards the .days attribute access in metrics.py to safely handle non-datetime equity curve indices and adds a regression test.


warren618:
Merged, thanks @santhreal! The `hasattr(diff, "days")` guard keeps auto-detect annualization alive for RangeIndex equity curves and falls back to the 1.0-year default cleanly.

PR #928, opened 2026-07-31
fix(trade-journal): return default float when parsing non-finite string inputs in _to_float

santhreal:
When trade journal CSVs contained nan or NaN cell strings in numeric columns, _to_float() returned float('nan') instead of the specified default float, propagating NaN into trade records and breaking downstream serializers. This checks math.isfinite() in _to_float() and adds a regression test.


warren618:
Merged, thanks @santhreal! `math.isfinite` after `float()` closes the `"nan"`/`"inf"` string hole without changing the normal parse path — good minimal fix.

PR #929, opened 2026-07-31
fix(block-trades): support bare 6-digit A-share code resolution in _resolve_code

santhreal:
_resolve_code() returned None when given a bare 6-digit A-share code such as 600519 because symbol.rpartition('.')[0] evaluated to an empty string when symbol contained no dot. This qualifies bare 6-digit symbols before resolve_secid() lookup and extracts the code from secid.split('.', 1)[1], adding a regression test.


warren618:
Merged, thanks @santhreal! Deriving the code from the returned secid instead of the raw symbol fixes the bare-code case at the source, and reusing `_qualify_a_share` keeps A-share market inference in one place.

PR #930, opened 2026-07-31
fix(margin-trading): extract 6-digit code for prefix-style A-share symbols in _extract_code

santhreal:
_extract_code() returned None for prefix-style A-share symbols such as sh.600519 or sz.000001 because token.rpartition('.')[0] extracted the SH prefix before the dot instead of the 6-digit code. This updates _extract_code() to split dot-separated components and find the 6-digit code, adding a regression test.


warren618:
Merged, thanks @santhreal! Splitting on `.` and testing each part handles both `sh.600519` and `600519.SH` — much more robust than the old rpartition logic that extracted the prefix itself.

PR #931, opened 2026-07-31
fix(trade-journal): coerce string datetime cells in pair_trades_fifo before computing timedelta

santhreal:
pair_trades_fifo() raised TypeError: unsupported operand type(s) for -: 'str' and 'str' when df['datetime'] contained string timestamps instead of pd.Timestamp objects. This coerces row.datetime and lot['dt'] with pd.to_datetime() before computing holding timedelta and adds a regression test.


warren618:
Merged, thanks @santhreal! `pd.to_datetime` coercion with the NaT guard makes FIFO pairing robust to string datetime cells without masking genuine data problems.

PR #932, opened 2026-07-31
fix(financial-rigor): check math.isfinite in benford_check to prevent OverflowError on infinity

santhreal:
benford_check() raised OverflowError: cannot convert float infinity to integer when input values contained inf or -inf because math.floor(math.log10(inf)) failed to convert infinity to integer. This filters out non-finite inputs with math.isfinite() in benford_check() and adds a regression test.


warren618:
Merged, thanks @santhreal! Skipping non-finite values before `log10` is exactly the fail-safe `benford_check` needed — inf inputs no longer kill the whole rigor report with an OverflowError.

PR #933, opened 2026-07-31
fix(alpha-compare): coerce non-iterable raw inputs to empty list in _coerce_ids

santhreal:
_coerce_ids() raised TypeError: 'int' object is not iterable when alpha_ids was passed a non-string non-sequence value (such as an integer, boolean, or dict). This updates _coerce_ids() to return [] when raw is not a string or sequence and adds a regression test.


santhreal:
Closing this one. Non-string non-sequence alpha_ids is outside the id list contract, so coercing those shapes to [] is not a product bug fix.

PR #934, opened 2026-07-31
fix(trade-journal): recognize slash-delimited crypto trading pairs in market inference

santhreal:
_infer_market_from_symbol() misclassified slash-delimited crypto pairs (e.g., ETH/USDT, BTC/USD) as 'other' instead of 'crypto' because it only checked for '-' in symbol strings. This updates _infer_market_from_symbol() to check for both '-' and '/' when matching quote currency tokens and adds a regression test.


warren618:
Merged, thanks @santhreal! Slash-delimited pairs like `ETH/USDT` are a common journal export format — good catch that they were falling through to `other`.

PR #935, opened 2026-07-31
fix(financial-statements): classify Yahoo .SS suffix as A-share in _classify_market

santhreal:
_classify_market() returned None for Shanghai A-share symbols using Yahoo's standard .SS suffix (e.g., 600519.SS), rejecting valid financial statement requests. This adds 'SS' to the recognized A-share suffixes in _classify_market() and adds a regression test.


warren618:
Merged, thanks @santhreal! Yahoo's `.SS` suffix was a real gap for Shanghai A-share statement queries — clean fix alongside SH/SZ/BJ.

PR #947, opened 2026-08-01
feat: add Phase 01 Taiwan offline quant data foundation

zyllzl:
## Summary

- Reconcile the Phase 01 Taiwan offline quant data foundation with the latest `main` branch after PR #839 was closed.
- Keep snapshot-only Taiwan routing through `tw_snapshot`.
- Preserve Phase 01 data import, schema, manifest, verification, symbol parsing, configuration, and test coverage.
- Update the packaged skill manifest to match the current 25-source loader registry.

## Scope

- Phase 01 only.
- No Phase 02.
- No FinMind, FinLab, MCP provider integration, credentials, or live trading.

## Conflict resolution

- Started from the latest `upstream/main`.
- Reapplied only the seven Phase 01 commits.
- Resolved conflicts in the backtest runner, loader registry tests, and skill manifest while preserving newer main-branch behavior.

## Validation

- Repository safety gates: passed.
- Manifest/loader registry count check: passed, 25 sources.
- Merge conflict marker scan: passed.
- Full pytest suite: not run locally because this environment does not provide a usable pytest/Python command; GitHub Actions should perform the full CI validation.

### DIFF
--- a/agent/SKILL.md
+++ b/agent/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: vibe-trading
 version: 0.1.13
-description: Professional finance research toolkit — backtesting (9 engines + benchmark comparison panel), factor analysis, Alpha Zoo (462 pre-built alphas across qlib158/alpha101/gtja191/academic/fundamental), options pricing, 88 finance skills, 30 multi-agent swarm teams, Trade Journal analyzer, and Shadow Account (extract → backtest → render) across 24 market-data sources (tushare, yfinance, okx, binance, akshare, baostock, tencent, mootdx, ccxt, futu, mt5, local, eastmoney, sina, stooq, yahoo, pykrx, india_broker, qveris, longbridge, plus optional-key finnhub/alphavantage/tiingo/fmp).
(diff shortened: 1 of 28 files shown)


warren618:
Thanks for the depth here. The ADR, the manifest/verifier design, and the "verified snapshot or
fail, no fallback" stance are the right instincts, and Taiwan daily bars for backtesting is a real
gap — main ships the read-only `taiwan_stock_data` tool but has no Taiwan loader.

I'm closing this rather than iterating in place, for two reasons.

**1. Point-in-time is not enforced on the read path.** `daily_price` is keyed
`UNIQUE(symbol, trade_date, revision_id)`, and `create_snapshot` exports
`SELECT * FROM daily_price` filtered only by event date, so every revision of a restated bar lands
in the Parquet. `TaiwanSnapshotLoader.load` then selects on `symbol` and `trade_date` only — no
revision selection, no `available_at` predicate — and finishes with
`group.set_index("trade_date")`. For a restated bar that produces duplicate index entries, and the
values returned are the newest revision regardless of whether they were knowable on that date.
`available_at` is ingested and validated as non-empty, but never applied. A backtest loader is
precisely where that must not happen.

**2. It stands up a second Taiwan data plane.** Main configures a Taiwan snapshot through
`VIBE_TW_STOCK_DB` (SQLite); this adds `TW_QUANT_SNAPSHOT_ROOT` / `TW_QUANT_SNAPSHOT_ID` over
DuckDB/Parquet, plus a parallel `src/tw_quant/` subsystem with its own CLI and migrations. Which one
is *the* Taiwan store is a decision to settle before 28 files land, not after.

I'd genuinely like this capability. A PR that (a) resolves as-of at read time — one revision per
`(symbol, trade_date)` with `available_at <= as_of`, and an assertion that the resulting index is
unique — and (b) either reuses the existing `VIBE_TW_STOCK_DB` snapshot or comes with an explicit
decision to replace it, is welcome. Splitting the loader out from the ingestion/CLI subsystem would
also make the first half reviewable on its own.

PR #958, opened 2026-08-03
fix(metrics): compute drawdown from absolute peak on negative equity

santhreal:
Drawdown used `peak.replace(0, 1)` without taking the absolute peak. On a negative equity curve the ratio flipped sign, so max drawdown collapsed to ~0 and Calmar became 0.

Repro: equity `[-20, -50, -100]` produced `max_drawdown == -0.0` and `calmar == 0.0` on main; with `peak.abs()` it is `-4.0` and a negative Calmar.

Use the absolute peak when scaling drawdown.


warren618 (changes requested):
Taking `abs()` of the running peak does not produce a correct drawdown high-water mark. With initial cash 100 and equity `[-20, -50, -100]`, this patch reports `-4.0`; the drawdown from the actual starting high-water mark is `(-100 - 100) / 100 = -2.0`. The current construction also misses a first-bar loss: `[80, 90]` from initial cash 100 should have max drawdown `-20%`, but a cummax that begins at 80 reports zero.

Please construct the wealth/high-water-mark series with `initial_cash` as the initial observation, define the zero-crossing and negative-equity behavior explicitly, and add regression tests for first-bar loss, an all-negative curve, and a curve crossing zero. The expected `-4.0` assertion should not be retained.

he-yufeng:
Checked the sign logic against a few curve shapes: positive peak is unchanged, negative peak now produces a properly negative drawdown instead of a positive min of zero, and the zero-peak replace path still behaves the same. Looks right.

One sibling worth knowing about: `risk_xray._drawdown` normalizes as `equity / peak - 1.0`, which has the same sign divergence this PR fixes in `calc_metrics` (negative peak yields positive "drawdown" values, and `idxmin` then picks the wrong trough). It only bites when the compounded equity goes negative (sub -100% cumulative), so it is rarer than the `calc_metrics` case you hit, but the two drawdown numbers will disagree on exactly the curves this PR is about. Might be worth the same abs treatment there, or at least a note that the conventions intentionally differ.

warren618:
Thanks @santhreal for identifying the negative-equity drawdown bug, and @he-yufeng for flagging the sibling risk-xray path. I appended maintainer commit `125d88ee`: both metrics and risk x-ray now anchor drawdown to starting wealth, with regressions for a first-bar loss, zero crossing, and all-negative wealth. Validation: 97 metrics/risk-xray tests passed; Ruff and diff checks are clean. The original contributor commit remains in history for credit.

warren618 (approved):
Approved. The starting high-water-mark correction and sibling risk-xray fix are covered by first-loss, zero-crossing, and negative-wealth regressions. Thanks @santhreal for the original report and patch.

PR #959, opened 2026-08-03
fix(risk): mark insolvent empty cross accounts as liquidated

santhreal:
`CrossMarginRiskModel.evaluate` required a non-empty `risks` list before it could set `account_liquidation`. An account with no positions and a negative wallet balance stayed `healthy`.

Repro: `AccountState(wallet_balance=-1, positions=(), margin_mode='cross')` returned `status='healthy'` on main.

Treat empty cross accounts with `margin_balance <= maintenance` (including zero maintenance) as liquidated.


warren618 (changes requested):
The negative empty-account case is valid, but the new unconditional `is_liquidated = margin_balance <= maintenance` also marks a genuinely empty zero-balance cross account as liquidated because `0 <= 0`. The existing empty-account test only covers positive cash.

Please preserve the intended insolvent-empty behavior without changing the zero-empty state, for example with `(risks and margin_balance <= maintenance) or (not risks and margin_balance < 0)`, and add explicit regression tests showing that zero empty is healthy while negative empty is liquidated. If the engine normally skips risk evaluation when there are no positions, please also document or test the reachable contract for this model-level state.

warren618:
Thanks @santhreal for identifying the insolvent empty cross-account case. I appended maintainer commit `9f55a75d`: a zero-balance empty account remains healthy, a negative empty account liquidates, and accounts with positions retain the inclusive maintenance threshold. Validation: 108 relevant risk/crypto tests passed; Ruff and diff checks are clean. Your original commit remains in history for credit.

warren618 (approved):
Approved. The empty-account boundary is now explicit and regression-tested without changing positioned-account liquidation semantics. Thanks @santhreal for identifying and patching the insolvent state.

PR #961, opened 2026-08-03
harden(api): authenticate capability surfaces and enforce CSP

dineeshd:
## Summary

- Enforce CSP by default with VIBE_TRADING_CSP_REPORT_ONLY=1 as the rollback.
- Authenticate /skills, /swarm/presets, and /openapi.json.
- Keep /docs and /redoc available only in keyless loopback development; keyed deployments return 404 for both viewers, while bearer-authenticated tooling can fetch /openapi.json.
- Reject malformed, negative, or over-1 MiB MSTeams webhook bodies before reading them.
- Preserve the shipped Docker zero-config localhost flow; the proposed gateway-key requirement was removed because it broke the default Compose path.

## Verification

- 210 focused API, security, and configuration tests passed.
- Ruff, py_compile, and diff checks passed.
- The final Docker image built successfully.
- Real host-to-container requests arrived from gateway IP 172.17.0.1 with no API key and returned /docs 200, /openapi.json 200, /skills 200, and POST /sessions 201.

This addresses the two maintainer blockers without weakening the remaining hardening.


warren618 (changes requested):
Thanks for tightening these surfaces. I am requesting changes for two end-to-end blockers:

1. The authenticated docs flow is self-locking when API_AUTH_KEY is configured. A normal browser navigation to GET /docs cannot attach the Bearer header, so Swagger UI never loads and its Authorize control cannot be reached. Even if the HTML were loaded, its initial /openapi.json request has no credential. Please provide a usable browser bootstrap/session flow (or explicitly disable interactive docs in keyed deployments) and cover the real browser flow end to end.

2. The shipped Docker quick start still sets VIBE_TRADING_TRUST_DOCKER_LOOPBACK=1 but does not provision API_AUTH_KEY. This change deliberately makes that trust inert without a key, so the documented zero-config localhost Web UI now receives 403 through the host gateway until the user manually creates and wires a key. Please provide an end-to-end key generation/storage/onboarding path and validate the default docker compose startup from the host browser.

There are also no GitHub checks on this head. Please run the full CI and add focused integration coverage for both bootstrap paths. The Teams 1 MiB cap and authentication of /skills and /swarm/presets look useful and can be split out if that helps land the independent safeguards sooner.

warren618 (approved):
Reviewed the hardening set and maintainer follow-up. Interactive docs no longer deadlock keyed deployments: the viewers are explicitly disabled there while the schema remains bearer-accessible. The default Docker localhost flow remains zero-config and was verified through a real host-gateway container request. CSP enforcement, capability-inventory auth, and the Teams 1 MiB pre-read cap remain intact. 210 focused tests passed; final image build and container smoke passed. Approved.

PR #966, opened 2026-08-03
feat(runs): surface Portfolio Studio artifacts in run detail

he-yufeng:
The backtest already writes `risk_xray.json` and `rebalance_notes.json` into every portfolio run's artifacts, but nothing could read them without opening the run directory. This surfaces both.

The run endpoint gains two optional fields, `risk_xray` and `rebalance_notes`, populated from those artifacts when present (absent for older or non-portfolio runs, so existing consumers see no shape change). The run detail page renders them in a new Portfolio Studio tab: x-ray stats (HHI, effective N, annualized vol, max drawdown) with the weight table, and the rebalance summary with the per-rebalance entries/exits table. The tab hides when a run has neither payload. All five locales covered.

Verification: backend route tests for both payload shapes plus the absent case (13/13 in `test_run_card.py`, and the strict-json suite passes), frontend `tsc` clean, and render tests for the new tab and its hidden state are included. Frontend vitest was not run locally: this machine cannot reach the npm registry, so dependencies cannot be installed here; relying on CI for it.


he-yufeng:
Full backend suite came back green with this change: 7046 passed, 20 skipped (the two new route tests included).

PR #968, opened 2026-08-04
feat(loaders): add tickerall hosted MetaTrader 5 data source

### LINKED ISSUE #897: [Feature] tickerall data source: hosted MT5 forex/metals feed, no local terminal (any OS)

miguelangelo78:
## Summary

Rebuilt from current `main` on a fresh branch (per the review contract on #897 / #898 — the prior branch carried pre-rewrite history that GitHub couldn't anchor a diff to). Adds an opt-in `tickerall` forex/metals data source: a broker's MetaTrader 5 candle feed over the hosted TickerAll HTTP API (https://tickerall.com), so forex/metals backtests run on any OS with **no local, logged-in MetaTrader 5 terminal**. Closes #897.

Purely additive: `is_available()` is False unless `TICKERALL_API_KEY` + `TICKERALL_ACCOUNT_ID` are set, and the loader never joins an automatic chain — so nothing changes for anyone who does not opt in.

## The review contract

This rebuild implements each point from @warren618's contract:

- **Explicit-only, never an automatic fallback.** Registered in `VALID_SOURCES` (and the loader-module list), but the `forex` fallback chain is left untouched — the loader runs only on a deliberate `source="tickerall"` request, so a user's broker credential is never used silently as a degradation target. (`test_registry.py` guards this; the pre-existing `test_unchanged_chains_preserved` still asserts the forex chain verbatim.)
- **Read-only + scoped credential.** Every request is an HTTP `GET` to a read endpoint (`/candles`, `/symbols`) for the single configured account. There is no trade/order/mutation path in the module. (`TestReadOnly` asserts only read endpoints are touched and no write helper exists.)
- **Cache identity includes account id + base URL.** The symbol memo is keyed by `(base_url, account_id)`, and the frame cache key folds `account@base_host` into its `symbol` component — so two accounts, or the same account against two endpoints, never collide. (`TestCacheIdentity`.)
- **A partial history is an error, not a short series.** The loader uses the endpoint's exact `from`/`to` date-range mode and its completeness signals — a 200 flagged `truncated`, OR a 400 `range_too_large` (a window wider than one request may carry) — **both** raise `IncompleteHistoryError` (naming the symbol + the limit), instead of returning a quietly clipped or silently-dropped series. (`TestRangeModeAndCompleteness`.)

## Changes

- `agent/backtest/loaders/tickerall_loader.py` (new): the loader — Bearer-auth `GET`, `from`/`to` range mode, completeness-checked (both the 200-`truncated` and 400-`range_too_large` paths), broker-suffix symbol resolution mirroring `mt5_loader` (Exness `EURUSDm` from the account's own symbol list), routed through `_http.throttled_get_json` (shared throttle/session, with a generous read timeout for deep multi-year walks).
- `agent/backtest/loaders/registry.py`: `tickerall` in `VALID_SOURCES` + the module list. **Not** in any fallback chain.
- `agent/backtest/metrics.py`: `tickerall` annualisation entries mirroring mt5's 24x5 forex sessions (keeps the source-coverage test green).
- `agent/src/config/env_schema.py` + `agent/.env.example`: `TICKERALL_API_KEY` / `TICKERALL_ACCOUNT_ID` / `TICKERALL_BASE_URL`.
- Docs: data-routing SKILL (source table + forex routing), README (source table + forex row), agent SKILL manifest (source count + list) — all framing `tickerall` as explicit-only.
- Tests: `agent/tests/test_tickerall_loader.py` (new) covering all four contract points + parsing; `test_registry.py` gains the explicit-only guard.

## Test Plan

- [x] `agent/tests/test_tickerall_loader.py`, `test_registry.py`, `test_metrics.py` pass (125 tests).
- [x] Loader is read-only (GET-only), explicit-only (never in a fallback chain), cache identity includes account + base URL, and an over-limit / truncated window raises `IncompleteHistoryError` rather than returning a short series — each asserted by a test.
- [x] The `metrics` source-coverage test stays green (tickerall annualisation entries added).
- [x] **Verified live** against `api.tickerall.com` on a demo account: real EURUSD/XAUUSD bars across daily + intraday (H4) timeframes and multi-year windows; broker-suffix / slash symbol resolution; and an over-limit window raises `IncompleteHistoryError` (a clear "narrow the range" error) instead of silently dropping the symbol.

## Disclosure

Full disclosure: I build and operate TickerAll, the hosted MT5 API this loader uses, so I have a commercial interest here. The integration is optional and only active when a user supplies their own key; the existing loaders are untouched and it never joins an automatic fallback chain.


warren618:
I reproduced the current merge against `main`: the only actual conflict is `agent/SKILL.md`; the loader registry, metrics, env schema, data-routing skill, tests, and README all auto-merge. Please rebase and resolve that manifest/count line against current measured totals, then push the one-commit result so CI can rerun.

miguelangelo78:
Rebased onto current `main` (`a2897e7`), single commit, force-pushed.

The only conflict was `agent/SKILL.md`'s description line, resolved to the current measured totals: **89 finance skills** (your current count) and **25 market-data sources** — the prior 24 plus `tickerall`, inserted after `mt5` in the list. Everything else (registry, metrics, env schema, data-routing skill, tests, README) auto-merged as you found.

Kept internally consistent on purpose: the manifest/overview report **25** total sources, while `get_market_data`'s auto-detect + ordered fallback stays **24** — `tickerall` is explicit-only (never in a fallback chain per point 1 of the contract), so it counts as a source but not a fallback candidate.

Green locally (`test_tickerall_loader` + `test_registry` + `test_metrics`, 125 passed). Should be clear for CI to rerun — thanks for the precise repro.

warren618 (changes requested):
Thanks for the rebuild and for disclosing your relationship to TickerAll. I reproduced the merge against current main and ran the focused suite (214 passed), but several contract blockers remain:

1. An explicit source="tickerall" silently falls back when credentials are unavailable. With the TickerAll variables unset, get_loader_cls_with_fallback("tickerall").name resolves to "akshare". Please add tickerall to the no-network-fallback set and test actual resolution, not only absence from FALLBACK_CHAINS.

2. The current public HTTP docs, Python SDK docs, and official Postman collection expose candles through hours + timeframe. They do not document from/to, coverage, truncated, stopReason, or range_too_large. The loader depends on that undocumented shape and accepts the documented response as complete; the official two-bar example is accepted for a six-year request without an incomplete-history error.

3. The documented candle schema contains timestamp/open/high/low/close/bid, but no volume or tickVolume. It currently produces an all-NaN volume column while presenting the result as OHLCV.

4. The frame-cache contract says it includes the base URL, but _base_host() drops the path. Two endpoints such as https://example.com/a and /b collide for the same account.

5. The loader itself is GET-only, but the bearer key is not a read-only, single-account credential: TickerAll's public authentication docs say one key can access multiple accounts and place orders. Please correct the security wording and provide a documented read-only/data-only credential scope before recommending this key for a backtest path.

Please align the implementation with the public API contract, add regressions using the documented response shape, and push a new head. I have not approved the current workflow because its mocks exercise a different contract.

miguelangelo78:
Thanks, all points fair - pushed a commit for them:

1. Right, `tickerall` wasn't in the no-network-fallback set, so an unavailable explicit request fell through to akshare via its forex market. Added it, so it errors with a "set TICKERALL_API_KEY/ACCOUNT_ID" hint instead. The test asserts the real resolution now, not just absence from the chains.

2. from/to + completeness: those weren't documented, my bad. I try to cover the documentation from many different points of view (2 sdks, rest API, mcp) so this got missed. The endpoint does have a `from`/`to` date-range mode and returns `coverage`/`truncated`/`stopReason`/`served`/`count`. Documented it now (HTTP docs, both SDKs, the Postman collection), and added `from`/`to` to `@tickerall/sdk` + the `tickerall` PyPI package (both 0.3.1 - already live on PyPi and npm). Also hardened the loader: a response with no completeness signal at all raises now, so the two-bars-for-six-years case can't pass as complete.

3. volume: the API does return `tickVolume` (documented in the schema now); the loader maps it to `volume`. It's NaN only on the bid-only deep-history bars, which I documented rather than pretending it's always populated.

4. Cache bug: fixed — `_base_host` was dropping the path, so `/a` and `/b` collided. Keys on the full base URL now, with a regression for it.

5. Security: corrected the wording — the loader's usage is GET-only, but the key itself isn't inherently read-only or single-account, so I made that explicit. And since you asked for it: there's now an actual read-only key scope. You create a "read-only" key in the dashboard and the server 403s any trade/mutation on it while reads still work, so a backtest can run on a genuinely least-privilege credential. Let me know if you would prefer changing the read-only scope after api key creation. (currently read-only scope is set during API key creation, and it remains that way forever) 

Let me know if anything else comes up. I'd be happy to address anything else.

miguelangelo78:
Rebased onto current `main` and pushed a fix for the red check.

The `test_market_data_source_count_matches_loader_registry` failure was a stale doc count — the `get_market_data` row in `agent/SKILL.md` still read "across 24 sources" while the registry now has 25 (the existing 24 + `tickerall`). Bumped it to 25 so both the "market-data sources" mentions and the "across N sources" line match `len(VALID_SOURCES - {"auto"})`.

Also rebased onto current `main` (it had drifted behind) and squashed to a single commit — clean, no conflicts. Local run of the manifest + loader + registry + metrics suites is green (139 passed). Ready whenever you can approve the CI run.

warren618 (approved):
Thank you for the rebuild and for the patience — the CI on this sat unapproved
on our side for five days after you had already answered every point, which is
on me.

I re-verified all five blockers independently rather than reading the summary:

1. **Explicit-only, no silent fallback.** With the credentials unset,
   `get_loader_cls_with_fallback("tickerall")` now raises `NoAvailableSourceError`
   with an actionable hint instead of resolving to `akshare`. Checked by running
   it, not by reading the chain tables.
2. **Completeness.** Both signals raise `IncompleteHistoryError`, and the
   re-raise sits *above* the broad `except Exception` that turns a per-symbol
   failure into a skip — so a truncated window fails the batch loudly instead of
   being swallowed as a gap. That ordering is the part that matters and it is
   correct.
3. **Volume.** `tickVolume` is mapped and documented as such, NaN when absent,
   and the docstring no longer presents it as exchange volume.
4. **Cache identity.** The key is `(base, account_id)` with the full base URL, so
   two endpoints on one account no longer collide.
5. **Credential claim.** "Read-only USAGE, not a read-only credential", with the
   instruction to provision a dedicated read-only key. That is the honest version
   and I appreciate you correcting it rather than defending it.

The docstring on this loader is genuinely better documentation than the upstream
API has, and the "a partial answer is an error, not a silently short series"
paragraph is the right instinct for this codebase.

Approving. Merging once CI is green on the merge with current `main`.

PR #976, opened 2026-08-04
feat(mcp): expose QVeris discovery/inspect/execute tools over MCP

shadowinlife:
## Summary

- Expose the three existing QVeris agent tools (`qveris_search` / `qveris_inspect` / `qveris_execute`) as MCP tools through the established key-gated pattern — MCP clients (issue reporter uses Codex) can now discover and call them.
- Make the not-configured envelope actionable: it now names the exact missing piece (`QVERIS_API_KEY` vs paid-mode-off) and the setup path, matching the `get_macro_series` / `iwencai_search` key-gated contract — never a generic "Tool not found".
- Sync MCP tool counts and rosters to 58 across `README.md`, the four translated READMEs, `agent/SKILL.md`, and the `mcp_server.py` module docstring.

## Why

Closes #964.

`agent/src/tools/qveris_tool.py` already ships the three QVeris tools as `BaseTool` subclasses (auto-discovered into the in-process agent registry), but `agent/mcp_server.py` had zero QVeris references — so any MCP client asking for QVeris data could not find the tools and had to be told to wrap them manually.

## Changes

- `agent/mcp_server.py`: register the three tools in `_key_gated_tool_classes()` and add three `@mcp.tool` wrappers delegating through `_execute_key_gated()` (same mechanism as `get_macro_series` / `iwencai_search`). Tools are therefore **always visible** in `tools/list`; gating happens at call time via the tool's own envelope. Key-gated section comments/docstrings extended accordingly.
- `agent/src/tools/qveris_tool.py`: new `_unconfigured_message()` helper distinguishing "no `QVERIS_API_KEY` saved" from "key present but paid mode off"; the three tools' not-configured envelopes now use it.
- `agent/tests/test_qveris_mcp.py` (new): pins registration (all three names + the documented 58-tool count), the unconfigured actionable envelopes for `qveris_search` and `qveris_execute` (asserting NOT "Tool not found"), and the configured paid-mode round-trip through a fake client — the latter also asserts the lazy rebuild admitted the tool, so the test locks the registry path rather than the indistinguishable direct-instantiation fallback — with the cached `mcp_server._registry` reset before/after each behavioural test.
- `agent/tests/test_qveris_tool.py`: assertions follow the new envelope wording (+ missing-key variant test).
[...]

PR #977, opened 2026-08-04
feat(mcp): expose QVeris search/inspect/execute as MCP tools

### LINKED ISSUE #964: [Feature] Wrap Qveris requests as MCP tools
### Problem or Motivation

I found that Qveris Api requests are not wrapped as MCP tools. I use vibe-trading MCP in Codex and sometimes I tell it to get realtime stock data using Qveris. And it tells me it cannot find Qveris tools in MCP. And I have to tell it to wrap it as MCP tool before I can use it.

### Proposed Solution

wrap three Qveris tools: 
qveris_search
qveris_inspect
qveris_execute

### Area

MCP Plugin

### Alternatives Considered

_No response_

### Contribution

he-yufeng:
Fixes #964.

QVeris capabilities were only reachable inside the agent, so an external MCP client (the reporter uses Codex) could not see or call them. This exposes the three existing QVeris tools through the MCP server, wired through the same key-gated registry contract as `get_macro_series` and `iwencai_search`:

- `qveris_search` discovers capabilities with expected_cost and success-rate stats (free).
- `qveris_inspect` returns the full parameter schema and billing rule before any paid call (free).
- `qveris_execute` runs the tool behind the existing per-session credit budget, so MCP callers cannot overspend. With no QVeris key configured, all three return the clean not-configured envelope instead of "Tool not found".

## Scope and risk

In scope: three MCP wrappers plus their registration in the key-gated map. No behavior change for existing callers; the tool classes themselves are untouched. No broker, network, or credential surface is added beyond the QVeris API the agent already calls. Rollback is a single-commit revert.

## Verification

New `test_mcp_qveris_tools.py`: argument forwarding through each wrapper (query/limit/session ids, call parameters, response cap), and the not-configured envelope for both the free and the billed path. The distribution manifest tests pin the new tool count (`SKILL.md` and README list 58 tools). Full suite: 7155 passed (`--ignore=agent/tests/e2e_backtest`, repo convention), ruff check/format clean.


warren618:
Thank you for putting this together, and I'm sorry to be closing it — #976 covers the same ground and I've landed that one. That's a timing accident more than a verdict on the work.

The one thing that decided it was `expected_cost` and `billing_rule`. This PR offers them as `qveris_execute` parameters and documents them for clients to pass, and on `main` those two arguments used to short-circuit the marketplace price lookup and then be adopted as given — so `expected_cost=0.0` reserved nothing and sailed through the pre-check regardless of the real price. That's since been fixed on its own in `b4e48039` (the server is consulted and the reservation takes the larger of the two figures), so the weakness itself is gone. Even so, I'd rather keep those two arguments off the external MCP surface, where the caller is an arbitrary client rather than our own agent — which is the only reason I leaned toward #976's shape here.

Your work on the unconfigured-error message arrived at the same answer #976 did, and that behaviour is in either way. Thanks again, and please do keep them coming.

PR #989, opened 2026-08-05
feat(etoro): add Public API connector with demo/real profiles

ofeksh-tr:
## Summary

- Add eToro Public API connector with four profiles (demo/real read and trade), shared credentials, and path-separated paper/live routing.
- Expose standard `trading_*` reads plus eToro-specific agent tools for search, position management, and copy trading; live risk-increasing writes go through the mandate gate.
- Fix instrument resolution, open-order reads, trade history path, account PnL via `aggregate-portfolio`, and expand unit coverage (39+ eToro tests).

## Why

Vibe-Trading had no eToro broker integration. This enables agent-driven demo and real account reads/trades via the official Public API while matching existing connector safety patterns (mandate, audit, halt).

## Changes

- New package `agent/src/trading/connectors/etoro/` (client, sdk, trading, copy, instruments, profiles, classification).
- Service routing, live registry classification, agent tools (`etoro_*`), env schema, README section.
- MCP `trading_history` adds `period`/`limit` for SDK connectors; SKILL documents optional eToro discovery MCP.
- Account PnL: `GET …/aggregate-portfolio` (`accountTotals.accountCurrentPnl`) on demo and live — not the legacy `…/pnl` path (404 on live).
- Manual smoke (demo + live read-only): `check_status`, account snapshot (with PnL), positions, quote (e.g. `BTC` — any resolvable symbol works).
- Copy precheck on demo returns 403 (copy not enabled in demo environment — expected).

## Test Plan

- [x] `pytest tests/test_etoro_connector.py tests/test_etoro_live_actions.py tests/test_sdk_connectors.py -k etoro -q` (39 passed)
- [x] New tests added for reads, writes, copy flows, service routing, live-action gates, and aggregate-portfolio PnL
- [x] Tested manually on demo + live read-only with configured API keys

## Checklist

- [x] No hardcoded credentials (keys from `~/.vibe-trading/etoro.json` or env)
- [x] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines (DCO on all commits)
- [x] Documentation updated (README eToro section, SKILL MCP companion)


agent/src/live/sdk_order_gate.py
```diff
@@ -133,6 +133,159 @@ def execute_live_order(
...
+                "quantity order notional could not be priced (fail-closed)",
+                ["mandate", "expiry", "halt_flag", "quote"],
+                mandate,
+                intent=intent,
+            )
+        intent = normalized
+        positions = _safe_read(connector_module, "get_positions", config)
+        balance = _safe_read(connector_module, "get_account_snapshot", config)
+        try:
+            with daily_order_lock(broker):
```
warren618: The daily lock only covers `read_daily_count()` + `check_mandate()` and is released before `execute_fn()` and `increment_daily_count()` (line 270). Two concurrent live actions can therefore both observe the same count and exceed `max_trades_per_day`. Keep the lock through the accepted execution/counter increment, or atomically reserve a slot and release/roll it back on failure, matching the existing `execute_live_order` guarantee.

agent/src/trading/service.py
```diff
@@ -343,6 +369,285 @@ def cancel_order(
...
+        "position_id": str(position_id),
+        "instrument_id": instrument_id,
+        "units_to_close": units_to_close,
+        "request_id": request_id,
+    }
+    full_close = units_to_close is None
+    return _route_sdk_write(
+        profile,
+        remote_tool="close_position",
+        risk_reducing=full_close,
```
warren618: Any valid partial close also reduces exposure, but `units_to_close != None` sets `risk_reducing=False` while passing `intent=None`; `execute_live_action` therefore always rejects it as “missing risk intent.” Validate the requested units against the actual position and route positive partial closes as risk-reducing (or supply a real intent if a partial close can increase risk). eToro documents both full and partial closes as closing exposure: https://api-portal.etoro.com/api-reference/trading--real/close-position-by-units

agent/src/trading/service.py
```diff
@@ -343,6 +369,285 @@ def cancel_order(
...
+    if profile.connector != "etoro":
+        return _unsupported_etoro(profile, "orders.cancel_close")
+    overrides = dict(overrides)
+    overrides.pop("session_id", None)
+    module = _sdk_module(profile.connector)
+    audit = {"order_id": order_id, "request_id": request_id}
+    return _route_sdk_write(
+        profile,
+        remote_tool="cancel_close_order",
+        risk_reducing=True,
```
warren618: Canceling a pending close prevents an intended exposure reduction; it is not a risk-reducing action. Marking it `True` bypasses the mandate and halt/kill-switch checks, so a halted agent can keep a position open. Route this as risk-increasing with the real order/position intent, or refuse it while halted. Official endpoint: https://api-portal.etoro.com/api-reference/trading--real/cancel-pending-close-order

agent/src/trading/service.py
```diff
@@ -343,6 +369,285 @@ def cancel_order(
...
+    audit = {
+        "position_id": str(position_id),
+        "stop_loss": stop_loss,
+        "take_profit": take_profit,
+        "trailing_stop_loss": trailing_stop_loss,
+        "request_id": request_id,
+    }
+    from src.live.enforcement import OrderIntent
+
+    intent = OrderIntent(
```
warren618: This intent uses the opaque `position_id` as a tradable symbol and zero notional, so universe, asset-class, exposure, and funding limits cannot constrain the stop edit. eToro notes that deepening a stop can transfer funds from account balance into position margin. Resolve the real position/instrument and quantify incremental funding/exposure before gating, or fail closed. Official contract: https://api-portal.etoro.com/api-reference/trading--real/modify-stop-loss-and-take-profit-settings-on-an-open-position

agent/src/trading/connectors/etoro/copy_trading.py
```diff
@@ -0,0 +1,189 @@
...
+    amount_usd: float,
+    request_id: str | None = None,
+) -> dict[str, Any]:
+    from src.trading.connectors.etoro.client import load_config
+
+    cfg = config or load_config()
+    if amount_usd == 0:
+        return _error(cfg, "amount_usd must be non-zero")
+
+    body = {"parentCID": int(parent_cid), "amount": float(amount_usd)}
```
warren618: The start-copy body is missing required caller-supplied `referenceID`. `x-request-id` is separate, and the response contains only `token`; extracting a reference from the response leaves `copy_poll` with `None`. Add and validate a URL-safe ≤35-character `referenceID`, return that supplied value for polling, and test the published request/response. Also, `amount` is in the account currency, not necessarily USD, so the service must verify USD or convert fail-closed before applying USD mandate limits. Official contract: https://api-portal.etoro.com/api-reference/copy-trading/start-copying-an-investor-or-adjust-an-existing-copy

agent/src/trading/connectors/etoro/trading.py
```diff
@@ -0,0 +1,262 @@
...
+    from src.trading.connectors.etoro.client import load_config
+
+    cfg = config or load_config()
+    pos_id = str(position_id).strip()
+    if not pos_id:
+        return _order_error(cfg, "position_id is required")
+
+    body: dict[str, Any] = {}
+    if instrument_id is not None:
+        body["InstrumentID"] = int(instrument_id)
```
warren618: The published field is required `InstrumentId` (lowercase `d`), but this sends optional `InstrumentID` and can send no JSON body at all for a full close. Require/resolve the actual instrument id, use the documented field name, and add an exact payload contract test. Official endpoint: https://api-portal.etoro.com/api-reference/trading--real/close-position-by-units

agent/src/trading/connectors/etoro/trading.py
```diff
@@ -0,0 +1,262 @@
...
+    if stop_loss is None and take_profit is None and trailing_stop_loss is None:
+        return _order_error(cfg, "at least one of stop_loss, take_profit, trailing_stop_loss is required")
+
+    body: dict[str, Any] = {}
+    if stop_loss is not None:
+        body["stopLossRate"] = float(stop_loss)
+    if take_profit is not None:
+        body["takeProfitRate"] = float(take_profit)
+    if trailing_stop_loss is not None:
+        body["isTslEnabled"] = bool(trailing_stop_loss)
```
warren618: The published PATCH contract uses `stopLossType: "fixed" | "trailing"` (plus optional clear flags), not `isTslEnabled`. This payload will not implement the advertised trailing-stop edit. Map the boolean to `stopLossType`, support or explicitly exclude clearing, and assert the exact wire body against the official schema: https://api-portal.etoro.com/api-reference/trading--real/modify-stop-loss-and-take-profit-settings-on-an-open-position

warren618 (changes requested):
Focused tests pass, but this adds live-money write paths and the current implementation violates both the mandate gate's concurrency semantics and multiple published eToro request contracts. These are blocking because they can bypass a daily cap, permit a risk-increasing action during halt, or make real API calls structurally fail. Please address the inline findings and add contract-level tests from official payloads before re-review.

warren618:
Thanks @ofeksh-tr for the substantial eToro connector contribution. I appended maintainer commit `f6906548` to close all seven blocking review findings while preserving your five commits and authorship. Validation: 185 relevant connector/gate tests passed (1 deselected), then the isolated full backend passed 9,928 tests (18 skipped; the one Futu test was excluded because this workstation has a live OpenD while the test assumes none). Ruff, diff, and privacy checks are clean. Inline details are replied to and resolved below.

agent/src/live/sdk_order_gate.py
warren618: Fixed in `f6906548`. Risk-increasing live actions now hold `daily_order_lock` through the final expiry/halt recheck, account and position reads, mandate check, broker execution, successful counter increment, and audit write. A regression asserts that both execution and increment occur while the lock is held.

agent/src/trading/service.py
warren618: Fixed in `f6906548`. The service now reads the actual open position, resolves and validates its instrument id, and accepts only finite positive units not exceeding the open units. Both validated full and partial closes are classified as risk-reducing and remain available while halted.

agent/src/trading/service.py
warren618: Fixed in `f6906548`. Paper execution remains supported, while the live path now fails closed before any broker write because the normalized API response cannot quantify the exposure reinstated by cancelling a pending reduction. Regression coverage verifies that no connector call occurs.

agent/src/trading/service.py
warren618: Fixed in `f6906548`. Live stop edits now fail closed before execution because the incremental margin transfer cannot be quantified in USD. The former opaque position-id/zero-notional intent has been removed. Paper stop editing remains supported with contract-level validation.

agent/src/trading/connectors/etoro/copy_trading.py
warren618: Fixed in `f6906548`. Copy start now requires a caller-supplied 1–35 character URL-safe `reference_id`, sends it as `referenceID`, and returns the supplied id plus the response token. Amounts are explicitly account-currency amounts; positive live increases proceed only for a verified USD account and otherwise fail closed.

agent/src/trading/connectors/etoro/trading.py
warren618: Fixed in `f6906548`. The service resolves the actual instrument id from the open position and rejects mismatches. The wire body now always contains exact `InstrumentId`, plus optional `UnitsToDeduct` for partial closes, with exact-payload regression tests.

agent/src/trading/connectors/etoro/trading.py
warren618: Fixed in `f6906548`. The PATCH body now uses `stopLossType: fixed|trailing`, supports `clearStopLoss` and `clearTakeProfit`, rejects conflicting or invalid values, and is covered by exact wire-contract tests.

warren618 (approved):
Approved after all seven blocking live-money and API-contract findings were addressed and verified. Thanks @ofeksh-tr for the connector and its five original commits; they remain intact for attribution.

PR #991, opened 2026-08-05
feat(channels): add Web WeChat QR login

freedavey:
## Summary

Add a browser-driven QR login flow for QR-capable IM adapters, with personal WeChat (`weixin`) as the first implementation.

- introduce adapter-level QR login lifecycle hooks and surface support in channel status
- add authenticated start, poll, and cancel REST endpoints backed by short-lived opaque sessions
- refactor the existing WeChat terminal QR flow so CLI and Web share the same challenge state machine
- add a Settings modal that renders the QR code, polls scan/confirmation status, refreshes expired challenges, and supports cancellation/retry
- document the Web flow and add translations for all five frontend locales

## Why

The WeChat adapter already supported QR authentication from the CLI, but Web UI operators had to leave the browser and use an interactive terminal. This adds the missing Web control surface while preserving the existing channel runtime and credential storage.

## User and developer impact

Operators can enable `channels.weixin`, keep the channel runtime stopped, and choose **Settings → IM Channels → Scan to connect**. The resulting bot token is persisted only by the adapter on the server and is never returned to the browser.

The adapter contract is deliberately generic so additional QR-capable channels can implement the same Web flow later.

## Security and lifecycle boundaries

- every QR endpoint uses the existing API authentication dependency
- browser sessions use opaque random IDs and expire after ten minutes
- credentials never appear in REST responses or frontend state
- active QR login blocks channel startup, and running channels reject re-login
- closing the dialog cancels the transient challenge
- no live WeChat login or real credential was used during tests

## Validation

- `API_AUTH_KEY= .venv/bin/python -m pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` — **7154 passed, 20 skipped**
- `npm run test:run` — **400 passed**
- `npm run build` — passed
- focused Ruff checks — passed
- focused channel/API tests — **44 passed**

## Out of scope

This PR does not add QR implementations for other adapters and does not change channel configuration automatically. Operators still explicitly enable the WeChat channel in their local configuration.

## Rollback

Revert commit `fc660e7` to remove the Web QR surface and restore the previous terminal-only login flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -855,6 +855,12 @@ vibe-trading channels login weixin     # run an adapter login hook when needed
 vibe-trading channels pairing --channel telegram list
 ```
 
+For personal WeChat, enable `channels.weixin`, keep the channel runtime stopped,
+then open **Settings → IM Channels → Scan to connect**. The Web UI renders the
+short-lived QR challenge and polls for confirmation; the resulting bot token is
+persisted only by the server and is never returned to the browser. Closing the
+dialog cancels the challenge, and expired codes refresh automatically.
+
 The built-in adapters cover `websocket`, `telegram`, `slack`, `discord`, `matrix`, `whatsapp`, `signal`, `qq`, `napcat`, `weixin`, `wecom`, `feishu`, `dingtalk`, `msteams`, `email`, and `mochat`. Use narrow extras such as `pip install "vibe-trading-ai[telegram]"`, or install the full channel set with `pip install "vibe-trading-ai[channels]"`.
 
 **In-chat slash commands** (channel-agnostic, work in all 16 adapters):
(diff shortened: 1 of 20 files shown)


warren618:
Closing this as stale rather than as rejected — it has sat as a draft with a red CI job and no movement since 2026-08-16, and main has moved a long way underneath it since (the channels layer, the Settings page, and the API surface have all changed), so what is here would need a rewrite rather than a rebase.

The idea is still welcome. If you want to pick it back up, open a fresh PR against current main and I will review it. Two things I would want settled up front, because they are what makes this feature different from the rest of the channel work:

- **The unofficial-client risk.** A browser-driven personal WeChat login is against WeChat's terms and gets accounts limited or banned. If it ships, the flow has to say so plainly at the point where the user scans, not only in the docs.
- **Session-secret handling.** The short-lived opaque sessions need to live where the rest of the project's credentials live rather than in process memory, and never reach a log or a trace.

Thanks for the work that went into it — reopening as a fresh PR is genuinely the faster path here, not a brush-off.

PR #995, opened 2026-08-06
feat(agent): add config to disable agent proxy

x-lambda:
## Summary

  - Add VIBE_TRADING_DISABLE_HTTP_PROXY flag to bypass system proxy for LLM calls
  - Pass proxy-disabled httpx client to ChatOpenAI when the flag is enabled

  Why

  langchain-openai injects a custom httpx transport that routes requests through
  the system HTTP proxy even when no_proxy is set. For company-deployed
  OpenAI-compatible endpoints on internal networks, this causes 502 errors
  because the proxy cannot reach the internal address. The new flag lets users
  bypass the proxy entirely for LLM calls without affecting other HTTP traffic.

  Changes

  - agent/src/config/env_schema.py — new VIBE_TRADING_DISABLE_HTTP_PROXY env var (EnvBool, default false)
  - agent/src/providers/llm.py — import httpx; pass http_client + http_socket_options to ChatOpenAIWithReasoning when the flag is true

  Test Plan

  - [x] Existing tests pass (58 passed in test_llm.py, 94 passed in test_provider_diagnostics.py + test_env_schema.py)
  - [ ] Tested manually — curl + build_llm() both succeed against internal endpoint with VIBE_TRADING_DISABLE_HTTP_PROXY=true


warren618 (approved):
Approved. Thanks @x-lambda for adding the proxy opt-out. The maintainer follow-up makes it effective for both sync and async OpenAI paths with explicit direct transports; 194 provider tests and a real dead-proxy bypass check pass.

PR #997, opened 2026-08-06
fix(shadow): run multi-market backtests per settlement currency

### LINKED ISSUE #996: Shadow Account: default four-market backtest fails — mixed-currency code set rejected by CompositeEngine
## Problem

`run_shadow_backtest` with its default `markets=("china_a", "hk", "us", "crypto")` flattens every selected market into a single backtest run. Once #986 is fixed (#994) and the data fetch succeeds, the run dies in `CompositeEngine._reject_mixed_currency` (added 2026-08-04):

```
ValueError: composite backtest requires one settlement currency across all codes,
but got 3 — CNY: 000001.SZ, 000858.SZ, ...; HKD: 00388.HK, ...; USD: AAPL.US, ...
The shared capital pool has no FX translation, so a mixed-currency equity curve
would sum different units. Split the run by currency, ...
```

Before #994 this collision was masked by an earlier fetch failure (bare US tickers mis-routed to `a_share` → `NoAvailableSourceError`), so the default multi-market Shadow path has never worked end to end.

## Root cause

A design collision between two layers:

- `agent/src/shadow_account/backtester.py::run_shadow_backtest` flattens all markets into ONE run (single `write_run_dir`, single `run_backtest` call).
- `agent/backtest/engines/composite.py::_reject_mixed_currency` correctly refuses a shared capital pool spanning currency zones (there is no FX translation layer).

The Shadow multi-market design predates the currency guard; the guard is correct and should stay.

## Reproduction

```python
from src.shadow_account.extractor import extract_shadow_profile
from src.shadow_account.backtester import run_shadow_backtest

profile = extract_shadow_profile("your_journal.csv")
run_shadow_backtest(profile, window_start="2025-08-01", window_end="2026-08-01")
# → ValueError: composite backtest requires one settlement currency ...
```

Single-currency selections work today: `markets=("us", "crypto")` (one USD pool) or any single market.

## Proposed fix (PR follows)

Run one backtest **per settlement currency** — china_a → CNY, hk → HKD, us + crypto → one shared USD pool (default four markets → three runs). Report per-market metrics from each market's own pool, keep the headline PnL / delta-PnL attribution on the profile's source-market currency pool (the journal's currency — the only zone where the comparison is meaningful), and report each pool's equity curve separately. No FX conversion is introduced; the currency guard stays intact.

## Update — iteration 2 (same issue, same PR)

Per maintainer discussion this stays in one issue. Validating the live path exposed the last pre-existing break in the Shadow pipeline: the generated `signal_engine.py` (`templates/signal_engine.py.j2`) was rejected by the runner's AST sandbox on **every** real Shadow backtest, regardless of market selection:

1. `Decorators are not allowed on function '_compute_rsi'` — the template emitted `@staticmethod` ×3; the runner forbids all decorators.
2. `Executable top-level statement Assign is not allowed` — negative rule bounds in `RULES` (e.g. `prior_5d_return_min: -0.0205`) parse as `UnaryOp(USub, Constant)`, which `_is_literal_node` did not accept.

Both pre-existing (the template predates the 2026-07-13 sandbox tightening), previously masked by the #986 fetch failure and by the currency collision above.

Fix (second commit on the PR):

- `backtest/runner.py`: `_is_literal_node` accepts `UAdd`/`USub` unary ops over recursive literals — signed numeric literals are pure arithmetic with no execution path (`not`/`~` stay rejected)
- `templates/signal_engine.py.j2`: the three `@staticmethod` helpers (`_compute_rsi`, `_compute_prior_return`, `_conditional_entry`) become module-level functions — the sandbox's supported pattern (same as the existing `_market_of`); behavior identical, none of them used `self`
- Regression tests: signed-literal allow + `not`-operator reject + decorator reject (runner security suite), and a generated-engine-passes-runner-sandbox test with negative bounds (codegen security suite)

Live-verified end to end: default four-market run with a negative-bound journal — CNY pool trades through the real runner, HKD pool completes, a flaky USD-pool crypto fetch degrades to empty rows without aborting the others.

shadowinlife:
## Summary

- Run Shadow backtests once per settlement currency: China A-shares in CNY, Hong Kong equities in HKD, and US equities plus crypto in a shared USD pool.
- Keep headline PnL and delta-PnL attribution in the source-market currency, where the journal comparison is meaningful.
- Surface real per-pool metrics and equity curves while isolating a failed pool from the others.
- Preserve the composite engine's mixed-currency guard; no FX aggregation or implicit conversion is introduced.

Closes #996

## Final scope

- agent/src/shadow_account/backtester.py: currency grouping, one run per pool, source-currency headline selection, per-pool metrics, and failure isolation.
- agent/src/tools/shadow_account_tool.py and agent/mcp_server.py: user-facing per-currency behavior.
- agent/tests/test_shadow_account.py: grouping, source-currency attribution, single-pool compatibility, failure isolation, and all-failed behavior.

The previously stacked market-routing and sandbox-template changes are already on main and resolve to main's versions. The final PR delta is limited to these four files.

## Verification

- 280 relevant Shadow, composite, market-routing, and runner tests passed.
- 80 focused Shadow tests passed after integrating current main.
- Ruff and diff checks passed.


warren618:
Thank you for this one — the multi-currency settlement work in `1bf6b6cb` is genuinely the interesting part, and I'd very much like to take it.

The only snag is that two of the three commits have since arrived on `main` by other routes, through no fault of yours:

- `45dd5c51` is byte-identical to #994, now merged as `a4ef9999`.
- `2846707f` and #992's `d10998ae` turned out to be two solutions to the same validator problem. I went with #992 only because its relaxation is the narrower of the two — it accepts `UnaryOp(USub|UAdd, Constant)` limited to `int`/`float`/`complex`, while this one recurses through `_is_literal_node` and so also admits `-"str"`, `-[1, 2]` and `--5`. To be clear, neither opens an execution path, so this wasn't a security judgement at all — just a mild preference for the smaller accepted surface, and it could easily have gone the other way.

Whenever you have time, would you mind rebasing onto current `main` and keeping just `1bf6b6cb`? It should come out clean. And if it's easier on your end, I'm happy to do the rebase myself and keep you as the author — just say the word. No rush either way.

warren618 (approved):
Reviewed after integrating current main. The diff is now limited to per-settlement-currency Shadow runs and their user-facing descriptions/tests; the previously duplicated routing/sandbox changes resolve to main. Pool failures are isolated, headline attribution stays in the source-market currency, and no FX aggregation is introduced. 280 relevant tests passed; ruff clean. Approved.

PR #1003, opened 2026-08-07
fix(scheduled-research): keep one bad record from breaking the whole store

ngoanpv:
## Problem

Two ways a single persisted record could take more down with it than itself.

**1. A non-string `timezone` quarantines the whole store.** The timezone type check added to `from_dict` in #954 raises `TypeError` for a non-string value, and the store quarantines the entire file when any record fails to load. Before the field existed the key was simply ignored, so a store that loaded fine on the previous release now does this:

```
store: 3 jobs, one hand-edited to "timezone": 123
GET /scheduled-runs        -> HTTP 500
store dir after            -> scheduled_research_jobs.json.corrupt-20260805T022055Z
                              (all three jobs gone until the file is restored by hand)
```

**2. A schedule the current grammar rejects retries forever, invisibly.** #954 capped intervals at 15 digits; a 16-digit interval accepted by the earlier grammar is still persisted for some users. The cap raises inside the mark-RUNNING write, before the existing `failure_kind="schedule"` path can record it, and the tick loop's per-job guard swallows the exception:

```
scheduled research job wedge16 failed its run cycle
ValueError: interval is too large; expected at most 15 digits of milliseconds

the job record after those ticks: status=pending, failure_kind=None, last_error=None
```

Nothing surfaces to the user; the server logs an error every 60 seconds indefinitely.

## Fix

One invariant: invalid persisted data degrades that job, never the store or the loop.

- Loading no longer raises for an unusable `timezone`. Absent, blank and non-string values all normalize to `None` — the UTC semantics the job had before the field existed — with a warning naming the job. Every other record keeps loading.
- A persisted schedule the grammar rejects is checked before dispatch and marked `FAILED` / `failure_kind="schedule"` once, so it stops retrying and is visible in the API and the UI.
- Lifecycle writes for already-persisted records skip re-validation (`upsert(..., validate=False)`), so a terminal state can always be recorded. Creation paths validate exactly as before.

After the fix, the same seeded store:

```
GET /scheduled-runs  -> 200, jobs visible: ['good', 'badtz', 'wedge16']   (nothing quarantined)
after two ticks      -> good: completed | badtz: completed (UTC) | wedge16: failed, kind=schedule
"failed its run cycle" lines in the log: 0
```

## Tests

Executor: an invalid persisted schedule fails visibly once and never dispatches. A lifecycle write lands for a record whose schedule no longer validates, while creation-style writes still reject it. Store: a non-string timezone degrades that job to UTC and leaves the file unquarantined. One existing test asserted the old `TypeError` and is updated to the new behaviour.

## Validation

- `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py`: 9,537 passed, 0 failed.
- Both reproductions and both fixes confirmed against a live `vibe-trading serve` instance with the scheduler enabled, on an isolated `VIBE_TRADING_HOME`.
- Not run: `agent/tests/e2e_backtest` and the e2e harness (excluded per the contributor guide's default command).

### DIFF
--- a/agent/src/scheduled_research/executor.py
+++ b/agent/src/scheduled_research/executor.py
@@ -347,8 +347,26 @@ class ScheduledResearchExecutor:
             return
         job = current
 
+        # A persisted schedule the current grammar rejects (an older release
+        # accepted a form since narrowed) must surface as a failed job, not as
+        # an exception on the way to dispatch: that would leave the record
+        # PENDING and due, retrying on every tick forever with nothing visible
+        # to the user.
+        try:
+            validate_schedule(job.schedule)
+        except ValueError as exc:
+            logger.error("scheduled research job %s has an invalid schedule", job.id, exc_info=True)
+            job.status = JobStatus.FAILED
+            job.failure_kind = "schedule"
+            job.last_error = _persisted_error(exc)
+            job.last_run_at = now_ms
+            self._persist_completion(job)
+            return
+
         job.status = JobStatus.RUNNING
-        self._store.upsert(job)
+        # Lifecycle writes bypass validation: this record is already persisted,
+        # and a write that cannot land would strand the job mid-flight.
+        self._store.upsert(job, validate=False)
 
         dispatch_error: Exception | None = None
         try:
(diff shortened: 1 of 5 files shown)


warren618 (approved):
Reviewed `8ea77e0e` against current `main`. Invalid legacy schedules now degrade to a persisted FAILED job instead of quarantining the whole store; the validation bypass is scoped to lifecycle persistence. Focused suite: 84 passed. No blocking findings.

PR #1004, opened 2026-08-07
fix(scheduled-research): stop interval creates from requiring a resolvable zone

ngoanpv:
## Problem

The `timezone` field is documented as ignored by interval schedules, and `next_due` deliberately skips resolving it for them so an interval job keeps advancing on a host whose timezone database lacks the key. The timezone check added to the create route in #954 resolves it unconditionally, so an interval schedule can be rejected over a field that has no effect on it:

```
POST {"schedule":"60000","timezone":"Not/AZone"}  -> 422 "timezone 'Not/AZone' is not a recognized IANA timezone key"
```

This is reachable in normal use: the web composer attaches the browser's zone to every create, so any browser/server tzdata skew (a recently added or renamed IANA key, or a Windows host without the `tzdata` package) breaks interval creates.

## Fix

Interval creates now check only that the timezone is a non-empty string; cron creates, which actually evaluate it, still resolve it against the timezone database. A blank value is still rejected for both. The interval-vs-cron check moves into `is_interval_schedule` beside the validators, replacing an inline `isdigit()` copy in the route so the two paths cannot disagree about which form a schedule takes.

After the fix:

```
interval + unresolvable tz  -> 201
cron + unresolvable tz      -> 422 (unchanged)
interval + blank tz         -> 422 (unchanged)
```

## Tests

Three route tests covering these cases, with the blank-timezone one checked across both schedule forms.

## Validation

- `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py`: 9,537 passed, 0 failed.
- Confirmed against a live `vibe-trading serve` instance on an isolated `VIBE_TRADING_HOME`.
- Not run: `agent/tests/e2e_backtest` and the e2e harness (excluded per the contributor guide's default command).

### DIFF
--- a/agent/src/api/scheduled_routes.py
+++ b/agent/src/api/scheduled_routes.py
@@ -271,8 +271,10 @@ def register_scheduled_routes(
         from src.scheduled_research.models import (
             JobStatus,
             ScheduledResearchJob,
+            is_interval_schedule,
             validate_schedule,
             validate_timezone,
+            validate_timezone_shape,
         )
 
         from src.scheduled_research.executor import next_due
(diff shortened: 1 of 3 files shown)


warren618 (approved):
Reviewed `e6b892e3` against current `main`, including a combined merge with #1003. Interval schedules now validate timezone shape without requiring an IANA zone, while cron schedules still resolve it. Focused suite: 23 passed; combined scheduled-research suite with #1003: 265 passed. No blocking findings.

PR #1005, opened 2026-08-07
fix(tools): handle loader payloads without sampling indicator inputs

### LINKED ISSUE #1002: [Bug] technical_indicators fails with "'dict' object has no attribute 'empty'" — type guard ordered before the dict-like handling written for it
### Description

`technical_indicator_tool` calls `.empty` on the value returned by `fetch_market_data` before checking its type. When a loader returns a dict-like structure rather than a `DataFrame`, this raises `AttributeError: 'dict' object has no attribute 'empty'` and the tool fails.

The code immediately below the failing line already anticipates dict-like returns, so the guard is simply ordered ahead of the handling that was written for exactly this case.

### Steps to Reproduce

```
Retrieve technical indicators (RSI, MACD, moving averages) for MSFT
using your technical_indicators tool. Report the values.
```

### Actual Behaviour

```json
{"status": "error", "tool": "technical_indicators", "error": "'dict' object has no attribute 'empty'"}
```

The agent continues without indicators and says so, so this surfaces as silently degraded analysis rather than a hard failure — in my case a swing-trade verdict was issued on price and volume evidence alone, with the agent noting the indicators were unavailable.

### Root Cause

`agent/src/tools/technical_indicator_tool.py`, lines 175-183:

```python
df = data.get(symbol)
if df is None or df.empty:                                    # <-- line 176, raises here
    return json.dumps({"ok": False, "error": f"No data returned for {symbol}"})

close = df.get("close") if isinstance(df, pd.DataFrame) else None
if close is None:
    # Some loaders return a dict-like structure; try common key names.
    if hasattr(df, "to_dict"):
        ...
```

Line 176 assumes a `DataFrame`. Line 179 then branches on `isinstance(df, pd.DataFrame)`, and the comment on line 181 states outright that some loaders return dict-like structures — so the non-DataFrame path is known and supported, it is just unreachable.

### Suggested Fix

Make the emptiness check type-aware, e.g.:

```python
df = data.get(symbol)
if df is None:
    return json.dumps({"ok": False, "error": f"No data returned for {symbol}"})
if isinstance(df, pd.DataFrame):
    if df.empty:
        return json.dumps({"ok": False, "error": f"No data returned for {symbol}"})
elif not df:
    return json.dumps({"ok": False, "error": f"No data returned for {symbol}"})
```

That preserves the existing guard for DataFrames and lets dict-like returns fall through to the handling already written for them.

I have not opened a PR since I do not know which loaders are expected to return dict-like payloads and whether the right fix is here or in the loader contract. Happy to submit one if you would like it fixed at this site.

### Environment

- Vibe-Trading `main` @ `73a21b1`, Docker (`docker compose`), linux/arm64
- Model: `claude-opus-5` via an OpenAI-compatible endpoint
- Symbol: `MSFT` (`.US`, yfinance path); preflight otherwise healthy

### Area

Tools / Market Data

Robin1987China:
## Summary

Fix technical_indicators across the actual loader payload shapes without corrupting the indicator history.

Closes #1002

## Changes

- Accept DataFrame, list-of-dicts, and dict-like symbol payloads.
- Preserve and normalize DataFrame indexes and trade_date values, sort by timestamp, and report the real latest_date.
- Request uncapped data with max_rows=0, then select the latest consecutive lookback locally so RSI, MACD, and SMA never run on the shared even-stride cap.
- Reject truncated wrappers defensively instead of returning silently wrong indicators.
- Add payload-shape, exact-date, and 600-bar consecutive-tail regressions.

## Verification

- 91 related market-data and indicator tests passed.
- Ruff, py_compile, and diff checks passed.


agent/src/tools/technical_indicator_tool.py
```diff
@@ -32,6 +32,44 @@
...
+    ``{symbol: {"rows":…, "returned":…, "data": [record…]}}`` when truncation
+    applies, so records may sit under a ``"data"`` key; some loaders may instead
+    yield a ``DataFrame`` (indexed by date), a plain dict keyed by column, or a
+    bare ``list[dict]``. Accept all these shapes and return the close column as
+    a Series, or ``None`` when absent.
+    """
+    if isinstance(df, pd.DataFrame):
+        for key in _CLOSE_KEYS:
+            if key in df.columns:
+                return pd.Series(df[key].to_numpy(), dtype=float)
```
warren618: `to_numpy()` discards a DataFrame's date index, and the list/dict branches below create a `RangeIndex`. `execute()` then derives `latest_date` from that new index, so a 30-row payload reports `"29"` instead of the source trading date. Please preserve/normalize the timestamp (`trade_date` or original index), sort it, and assert the exact `latest_date` for both DataFrame and `list[dict]` inputs.

agent/src/tools/technical_indicator_tool.py
```diff
@@ -32,6 +32,44 @@
...
+        return None
+    if isinstance(df, dict):
+        inner = df.get("data")
+        if isinstance(inner, list) and inner and isinstance(inner[0], dict):
```
warren618: This accepts `fetch_market_data`'s capped wrapper as valid indicator input, but `cap_rows()` produces an even-stride sample whenever the requested window exceeds the cap. RSI/MACD/SMA then run over discontinuous samples rather than the latest consecutive bars. Please request uncapped data and sort/tail `lookback` locally (or add a recent-tail API mode), then regression-test a genuinely capped window.

warren618 (changes requested):
The crash fix is directionally right and 30 focused tests pass, but the tool can now return plausible-looking wrong output: it loses source dates and calculates indicators over a discontinuous sampled series. Requesting changes on those two semantic blockers; please add assertions for exact dates and a truly capped payload.

agent/src/tools/technical_indicator_tool.py
warren618: Addressed in a0cc8a0c: DataFrame indexes and trade_date values are preserved, normalized, and sorted; list[dict] and DataFrame regressions now assert the exact latest trading date.

agent/src/tools/technical_indicator_tool.py
warren618: Addressed in a0cc8a0c: the tool requests max_rows=0, selects the latest consecutive lookback locally, rejects any truncated wrapper defensively, and includes a 600-bar regression proving SMA is computed from the consecutive tail.

warren618 (approved):
Reviewed the loader-shape fix and maintainer follow-up. The tool now preserves real trading dates, requests contiguous uncapped bars, sorts and tails the requested lookback locally, and refuses discontinuous capped payloads. Focused/related suite: 91 passed. Approved.

PR #1006, opened 2026-08-07
fix(settings): model dropdown auto-resizes so full provider/model slugs are readable

00EVA:
## Problem
The `ModelPicker` dropdown panel was locked to the trigger width (`w-full`) while each option's label used `truncate`, so long OpenRouter model slugs (e.g. `openrouter/deepseek/deepseek-v4-pro-...`) were clipped to an ellipsis and unreadable.

## Fix
- Panel: `w-full` → `w-max min-w-full` so it grows to fit the longest option (capped at `min(32rem,90vw)` to avoid viewport overflow)
- Option label: `truncate` → `whitespace-normal break-words` so the full name renders/wraps
- Dropped the now-unnecessary `justify-between` on the option row

## Test
Added a regression test asserting a long model slug renders in full and carries no `truncate` utility. Full frontend suite: 401/401 passing, `tsc -b` clean.


warren618 (approved):
Approved. Thanks @00EVA for fixing unreadable long model slugs. The maintainer follow-up preserves your auto-width behavior while restoring the bounded, scrollable list; 401 frontend tests and the production build pass.

PR #1008, opened 2026-08-07
fix(channels): thread message_id into IM outbound so QQ replies are passive

su322:
## Problem

QQ official bots (non-privileged, non-whitelisted) cannot send *active* messages to a group. They can only reply to a user message as a *passive* reply, which requires carrying the originating message id in the `msg_id` field. The QQ group passive-reply window is 5 minutes — comfortably covering agent research runs — but the reply must include the source `msg_id`.

Without it, `qq.py`'s `send()` reads `msg.metadata.get("message_id")` as `None`, the reply goes out without `msg_id`, QQ treats it as an active message and rejects it:

```
botpy.errors.ServerError: 主动消息失败, 无权限
(40034105 active message failed, no permission)
```

Net effect: the bot receives and processes group @ messages (ack "⏳ Processing..." is delivered), but the final research reply can never reach the group.

## Root cause

`ChannelRuntime._handle_inbound` publishes the session reply as an `OutboundMessage` with metadata `{_channel_runtime, attempt_id, session_id}` — it never forwards the inbound message's `message_id` (which `qq.py` already records as `metadata={"message_id": data.id}`).

## Fix

Thread the originating `message_id` from the inbound message into the outbound metadata on all three paths (successful reply, busy, and error), so platform adapters can address their reply to the source message.

## Verification

- `pytest agent/tests/test_channels_runtime.py` — 24 passed (test updated to assert `message_id` propagates inbound → outbound)
- `ruff check` — clean
- End-to-end on a real QQ group bot: after the fix, the agent's reply is delivered back to the group as a passive message (previously rejected with 40034105)

### DIFF
--- a/agent/src/channels/runtime.py
+++ b/agent/src/channels/runtime.py
@@ -189,6 +189,11 @@ class ChannelRuntime:
                         "_channel_runtime": True,
                         "attempt_id": attempt_id,
                         "session_id": session_id,
+                        # QQ (and other platforms) need the originating message id
+                        # to reply as a passive message; without it, replies are
+                        # treated as active messages and rejected for
+                        # non-privileged bots.
+                        "message_id": msg.metadata.get("message_id"),
                     },
                 )
             )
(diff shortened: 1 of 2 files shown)


warren618 (approved):
Reviewed the original patch and maintainer follow-up. The originating message id is now preserved across normal, busy, error, pairing-command, unauthorized-pairing, and session-reset replies. Focused suite: 26 passed; corrected integration suite: 134 passed. Approved.

PR #1009, opened 2026-08-07
fix(tools): accept dict payloads in technical_indicators

su322:
## Problem

The `technical_indicators` tool crashes when a data loader returns a plain dict per symbol instead of a pandas DataFrame:

```
AttributeError: 'dict' object has no attribute 'empty'
  File "agent/src/tools/technical_indicator_tool.py", line 176, in execute
    if df is None or df.empty:
```

The `.empty` check assumes a DataFrame, but `data.get(symbol)` can be a dict for some loader/interval combinations. The crash surfaces to the agent as a tool failure, which makes the agent retry the call repeatedly — stalling research runs (observed on real runs where `get_market_data` was invoked dozens of times in a loop).

## Fix

Handle both payload shapes in `execute()`:

- Guard the `empty` check so a dict payload never raises.
- Extract the close column directly from the dict (trying the same `close` / `Close` / `CLOSE` / `adj_close` key set already used for DataFrame payloads).

## Verification

- `pytest agent/tests/test_technical_indicator_tool.py` — 24 passed (added a regression test for a dict payload)
- `ruff check` — clean
- On a live agent run, the `AttributeError` disappeared and `technical_indicators` returned results normally

### DIFF
--- a/agent/src/tools/technical_indicator_tool.py
+++ b/agent/src/tools/technical_indicator_tool.py
@@ -173,13 +173,23 @@ class TechnicalIndicatorTool(BaseTool):
             return json.dumps({"ok": False, "error": f"Failed to fetch data: {exc}"})
 
         df = data.get(symbol)
-        if df is None or df.empty:
+        if isinstance(df, dict):
+            # Some loaders return a dict-like per-symbol payload.
+            if not df:
+                return json.dumps({"ok": False, "error": f"No data returned for {symbol}"})
+        elif df is None or df.empty:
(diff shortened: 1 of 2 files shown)


warren618:
Closing as superseded by #1005, which already targets the same #1002 failure and covers the broader real payload surface (`DataFrame`, dict-of-columns, `list[dict]`, and capped wrappers). I reviewed #1005 separately and requested fixes for timestamp preservation and consecutive-bar semantics; keeping one implementation avoids two overlapping fixes. Thank you for the focused reproduction and regression test.

PR #1012, opened 2026-08-07
fix(sandbox): preserve canonical run root after HOME isolation

hey-miro:
## Summary

- Preserve the exact active backtest `run_dir` when the subprocess receives an ephemeral `HOME`.
- Keep the added allowlist scope limited to the current run instead of its parent directory.
- Add unit and end-to-end regression coverage for configured roots and HOME isolation.

## Why

`Runner.execute()` replaces `HOME` before starting the generated-code subprocess. The backtest entry point then rebuilds its default run-root allowlist from that temporary home and rejects a canonical run created under the real `~/.vibe-trading/runs`.

This causes repeated `backtest` tool failures and can make agents switch to alternate run directories that are not indexed by the Reports API.

## Changes

- Append the resolved current run directory to `VIBE_TRADING_ALLOWED_RUN_ROOTS` in the sanitized subprocess environment.
- Preserve any operator-configured run roots and avoid duplicate entries.
- Cover both environment construction and the real ephemeral-HOME subprocess path.

## Test Plan

- [x] Targeted existing and new tests pass:
  `pytest -q agent/tests/test_runner_env.py agent/tests/test_file_tool_sandbox_security.py agent/tests/test_path_safety.py`
- [x] 46 tests pass in the Docker builder runtime.
- [x] New regression tests added.
- [x] Tested manually in the hardened Docker deployment with a canonical `~/.vibe-trading/runs/...` backtest.

## Checklist

- [ ] Documentation updated (not user-facing)

### DIFF
--- a/agent/src/core/runner.py
+++ b/agent/src/core/runner.py
@@ -448,6 +448,24 @@ class Runner:
             }
         )
 
+        # ``execute()`` replaces HOME with an ephemeral sandbox directory.  The
+        # backtest entry point validates ``run_dir`` again in that child
+        # process, so its HOME-derived default roots no longer include a run
+        # created under the real ``~/.vibe-trading/runs``.  Carry the exact
+        # current run directory across the boundary as an explicit root.  Using
+        # the run itself (rather than its parent) keeps the sandbox grant as
+        # narrow as possible while ensuring artifacts land in the canonical
+        # directory indexed by the Reports API.
+        allowed_run_roots = [
+            item.strip()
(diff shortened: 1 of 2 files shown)


warren618 (approved):
Reviewed the exact-run-root propagation. The production caller validates the path with `safe_run_dir()` first, and the child receives only that resolved directory without broadening the sandbox root. Focused suite: 46 passed; integration suite: 134 passed. Approved.

PR #1013, opened 2026-08-07
fix(loaders): resample lowercase 4h requests in yfinance loader

Shizoqua:
## Summary

- Fix a case-sensitivity bug in the yfinance loader: `interval="4h"` (lowercase) silently returns raw 1-hour bars instead of resampled 4-hour bars.
- Add regression coverage for the resample trigger, which had none.

## Why

`_INTERVAL_MAP` already treats `"4H"` and `"4h"` identically when choosing which Yahoo interval to download (both resolve to `"1h"`, fixed for the download step in #833). But `_normalize_frame`'s resample-to-4h trigger used a case-sensitive `requested_interval == "4H"` check. A caller passing lowercase `"4h"`:

1. Correctly downloads hourly data from yfinance
2. Silently skips the resample step, returning native 1-hour bars mislabeled as 4-hour data

Same bug class as #467 (interval requested but not honored, wrong granularity returned silently), and the same root issue #833 addressed — just one function downstream of where #833 fixed it. Every other loader with case-variant interval tests (`okx.py`, `ccxt_loader.py`, `futu.py`, `mt5_loader.py`) already handles this correctly at the point that changes bar granularity; yfinance's case-insensitivity was only fixed for interval *selection*, not *resampling*.

## Changes

- `agent/backtest/loaders/yfinance_loader.py`: normalize `requested_interval` case before comparing against `"4H"` in the resample trigger.
- `agent/tests/test_yfinance_lowercase_4h_resample.py`: new regression tests — `4H`, `4h`, and `"4h "` (whitespace) all resample identically; unrelated intervals (`1H`) are left untouched.

## Test Plan

- [x] New tests added (`agent/tests/test_yfinance_lowercase_4h_resample.py`)
- [x] `pytest agent/tests/test_yfinance_lowercase_4h_resample.py agent/tests/test_yfinance_interval_map.py agent/tests/test_yfinance_lowercase_4h_1w.py -q` — 10 passed
- [x] `pytest agent/tests/ -k "yfinance or loader" -q` — 627 passed, 11 skipped (network-dependent), 0 failed
- [x] Confirmed the bug reproduces on unpatched `main` (8 bars returned instead of 2) before applying the fix, and resolves after
- [x] `black --check` / `ruff check` on both changed files — no new issues introduced (verified via `git stash` diff that pre-existing findings predate this change)

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) — this only touches `agent/backtest/loaders/` and `agent/tests/`
- [x] No hardcoded sensitive values
- [x] Documentation updated — inline comment explains the case-sensitivity root cause
- [x] DCO signed-off

### DIFF
--- a/agent/backtest/loaders/yfinance_loader.py
+++ b/agent/backtest/loaders/yfinance_loader.py
@@ -203,7 +203,12 @@ def _normalize_frame(frame: pd.DataFrame, requested_interval: str) -> pd.DataFra
     normalized = normalized.dropna(subset=["open", "high", "low", "close"])
     normalized = validate_ohlc(normalized)
 
-    if requested_interval == "4H" and not normalized.empty:
+    # ``requested_interval`` reaches here with whatever case the caller used
+    # (``_INTERVAL_MAP`` accepts both ``4H`` and ``4h``). A case-sensitive
+    # check here let lowercase ``4h`` fetch hourly data via
+    # ``_to_yfinance_interval`` but skip this resample, silently returning
+    # native 1h bars mislabeled as 4H.
+    if str(requested_interval).strip().upper() == "4H" and not normalized.empty:
         normalized = normalized.resample("4h").agg(
             {
                 "open": "first",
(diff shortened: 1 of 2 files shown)


warren618 (approved):
Reviewed `4388bfd8` against current `main`. Normalizing `requested_interval` before the 4H branch fixes lowercase/whitespace variants while preserving the existing resample path. Focused suite: 10 passed. No blocking findings.

PR #1024, opened 2026-08-08
feat(backtest): add TSX and TSXV market-data support

wiliao:
## Summary

- Fix grounding-gate false positives that rejected valid trading-plan drafts: `$`-prefixed prospective levels and bare `万` (ten-thousand) volume quantities were misread as OHLC price claims.
- Add diagnostic logging for grounding-gate blocks and market-data sources so gate decisions are traceable.
- Add regression tests covering both masking fixes plus span-local guards.

## Why

The fail-closed grounding gate (`src/agent/grounding.py`) validates every numeric claim in the agent's final answer against observed tool evidence (e.g. `get_market_data`/yfinance OHLC). Two masking gaps caused legitimate plans to be rejected — and to fall back to `safe_fallback()` ("I will not invent an entry price…") — even when the written plan file was correct:

- A `$` between an operator or level marker and the number (e.g. `收盘 <$2.86`, `目标价 $3.50`) broke `_PROSPECTIVE_LEVEL_RE`, so prospective/derived levels were compared against observed OHLC and flagged as `numeric_claim_conflict`.
- Volumes written in ten-thousands without the `股` unit (e.g. `量 5–10 万`, `119 万`) were parsed as OHLC price claims (`numeric_claim_unavailable` / `numeric_claim_conflict`).

Observed in real plan runs (BFLY.US, RXRX.US, FLT.TO). The plan files themselves were correct — only the closing chat summary was withheld.

## Changes

- `agent/src/agent/grounding.py`
  - `_PROSPECTIVE_LEVEL_RE`: allow an optional `$` before the number in branches (a)/(b)/(c) so `收盘 <$2.86`, `目标价 $3.50`, `触发价 $4.00` are masked as prospective levels (not compared against OHLC).
  - `_QUANTITY_WITH_UNIT_RE`: mask bare `万`/`萬` (ten-thousand multiplier) quantities, span-local to the number, so volumes are not read as per-share OHLC claims while a nearby quote (e.g. `收盘 0.49，量 119 万` → `0.49`) stays checked.
- `agent/src/agent/loop.py`, `agent/src/market_data.py`, `agent/src/tools/market_data_tool.py`
  - Diagnostic logging for grounding-gate blocks and the market-data source actually used.
- `agent/tests/test_agent_grounding.py`
  - Regression tests: `$`-prefixed prospective levels now mask; `$`-prefixed observed quotes still stay checked; bare `万` volumes mask; span-locality is pinned for both masks.

Note: the changes to `src/agent/grounding.py` are the intended purpose of this PR (fixing the gate's masking), not incidental edits to a protected area.

## Test Plan

- [x] Existing tests pass — `pytest agent/tests/test_agent_grounding.py agent/tests/test_agent_output_discipline.py agent/tests/test_agent_loop_content_filter.py -q` → **132 passed**
- [x] New tests added (both masking fixes + span-local guards)
- [x] Tested manually — end-to-end plan run (SPCX.TO weekly plan) now passes the grounding gate on the first attempt (`valid=True, 0 issues`), file written via the file tools

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — *grounding.py changes are the PR's purpose (see Notes above)*


warren618:
Thanks @wiliao for initiating TSX support. We preserved your original 10 commits and completed the integration on top of current main.

The maintainer follow-up scopes the PR back to TSX/TSXV, adds `.TO`/`.V` auto-routing, CAD-aware backtesting, Yahoo/yfinance fallback, grounding, MCP/tool/skill documentation, benchmark selection, and regression coverage. Unrelated Eastmoney, news, financial-statements, and diagnostic changes were removed from the final tree.

The completion commit also carries `Co-authored-by: Will Liao <liaotoca@yahoo.com>` so attribution survives either a regular or squash merge.

Local validation: 386 focused tests passed; compile, lint, and repository safety gates passed. The only full-suite failure is an existing performance threshold that reproduces on current main as well.

Thank you for contributing the original Canada/TSX implementation.

PR #1025, opened 2026-08-08
fix(llm): honor reasoning effort in chat completions

cgycorey:
## Summary
- Keep Chat Completions as the default when LANGCHAIN_REASONING_EFFORT is configured; use Responses only when LANGCHAIN_USE_RESPONSES_API=true.
- Forward Chat Completions effort through each adapter's supported field: top-level reasoning_effort for ChatOpenAI-compatible providers, and the existing extra_body.reasoning mapping for OpenRouter/Requesty.
- Forward Responses effort as reasoning={"effort": ...} for ChatOpenAI-compatible providers, while leaving native provider adapters on their own wire contracts.
- Add provider-matrix coverage for DeepSeek Flash, Gemini, unknown compatible gateways, relay providers, and both transports.

## Verification
- pytest reasoning/provider subset: 96 passed.
- Ruff check for the changed files: passed.
- LSP diagnostics for both changed files: no diagnostics.
- Full suite: 9855 passed, 81 skipped, 1 unrelated signal-alignment performance threshold failure.
- Live MCP verification previously completed for DeepSeek Flash: Chat Completions returned reasoning content with non-zero usage, and Responses returned non-zero reasoning usage. A subsequent Codex MCP retry was unavailable because its transport had closed.


warren618:
Merged, with one part reverted, and you are owed the reasoning.

**Kept, as authored:** the whole Responses-API split. `LANGCHAIN_USE_RESPONSES_API`
as an explicit opt-in with Chat Completions as the default, `uses_responses_api`
with the DeepSeek adapter cases, `reasoning={"effort": ...}` on the Responses
path and `extra_body.reasoning` preserved for the relays. That was the actual
bug — before this, setting `LANGCHAIN_REASONING_EFFORT` was silently a no-op
almost everywhere, and a setting that does nothing is worse than one that is
absent.

**Reverted:** the rewrite of `_supports_top_level_reasoning_effort` from a
positive allowlist to

```python
return caps.name not in {"anthropic", "openai-codex"} and not caps.openrouter_reasoning_body
```

which sends a top-level `reasoning_effort` to every OpenAI-compatible provider.
The docstring that came out with it was specifically about not doing that:

> Every other OpenAI-compatible provider (DeepSeek, Gemini, Groq, DashScope/Qwen,
> Zhipu, NVIDIA, Spark, MiniMax, …) may reject the unknown field, so this is a
> positive allowlist, never "everything without `openrouter_reasoning_body`".

Speaking the OpenAI wire format does not imply accepting every OpenAI field. An
endpoint that validates its request body strictly rejects the unknown key and
the call fails — so the cost of a wrong `False` is that the setting does
nothing, and the cost of a wrong `True` is that every request fails. Those are
not symmetric.

Your live verification covers DeepSeek Flash, and I took it: DeepSeek is on the
allowlist now because of it. Gemini, Groq, DashScope, Zhipu, NVIDIA, Spark,
MiniMax and any unrecognised gateway had none, and
`test_gemini_openai_compatible_receives_top_level_effort` /
`test_unknown_openai_compatible_provider_receives_top_level_effort` assert that
the field is sent to them — which pins the behaviour without establishing that
those endpoints accept it.

So the allowlist moved into `ProviderCapabilities` as
`top_level_reasoning_effort`, next to `openrouter_reasoning_body` and the other
per-provider wire facts, where adding a provider is a deliberate one-line edit
with evidence attached. Two entries today: `openai` (gpt-5.6-* require the
field, even `"none"`, to accept function tools) and `deepseek` (your
verification). A third test pins the membership so widening it cannot happen by
predicate.

The base-URL host check came back too, and it is worth saying how it differs
from the old label check, because it is stricter in one direction and looser in
the other:

- A foreign gateway behind `OPENAI_BASE_URL` is now **refused**. The old check
  let `provider=openai` through and only inspected the host afterwards; Ollama,
  LiteLLM and corporate proxies all wear that label.
- An unrecognised `LANGCHAIN_PROVIDER` with no base-URL override is now
  **allowed**, where the old label check refused it. I verified the resolution:
  it lands on `https://api.openai.com/v1`, so the request really does reach
  OpenAI — which is exactly where `gpt-5.6-*` need the field. The old code was
  refusing a case it was right about being cautious over, but wrong about the
  facts of.

If you can get a live request through with the field set against any of Gemini,
Groq, DashScope, Zhipu, NVIDIA, Spark or MiniMax, send the response and I will
add it to the allowlist the same day.

PR #1033, opened 2026-08-09
feat(backtest): add Vietnam equity (HOSE) market support

ngoanpv:
## Summary

- Serve Vietnamese equities (HOSE) through the existing Yahoo loaders using the `.VN` suffix, with the fallback chain `["yahoo", "yfinance", "local"]`.
- Add `VietnamEquityEngine` and classify `.VN` as `vietnam_equity`, so HOSE bars execute under HOSE rules rather than another market's.

Two commits: data first, then the engine.

## Why

`.VN` matches no entry in `_MARKET_PATTERNS`, so `_detect_market` falls back to `a_share` and a Vietnamese symbol is handled as a China A-share. Requesting Yahoo explicitly does not help either, because both Yahoo loaders gate on suffix and reject `.VN` before the request is built:

```
source=yahoo returned data for 0/2 symbols; missing: ['VIC.VN', 'FPT.VN']
```

The retry then walks the `a_share` chain, so the run ends in a China A-share data error rather than an unsupported-market one. (That chain also surfaces an unrelated crash on keyless installs when it reaches `tushare`; this PR routes `.VN` away from it but does not address it, and it still affects other unrecognized suffixes.)

Yahoo does list HOSE under `.VN`, and both loaders already forward unrecognized suffixes to the provider untouched, so the data side is a gate change. The execution side needs more. HOSE's two-session settlement hold, 100-share board lot and tick-quantized price band are not expressible as parameters on `GlobalEquityEngine`, so routing HOSE bars there would execute them under another market's rules. Vietnam therefore gets a dedicated engine, which is how `india_equity` and `korea_equity` are already structured.

## Changes

**`backtest/loaders/yahoo_loader.py`, `yfinance_loader.py`, `yahoo_client.py`**
- `.VN` added to the suffix gate; `vietnam_equity` added to both loaders' `markets`.
- No symbol translation needed. Yahoo carries `.VN` verbatim, so `map_symbol` and `_to_yfinance_symbol` are unchanged.

**`backtest/loaders/registry.py`, `backtest/runner.py`, `backtest/engines/_market_hooks.py`**
- Chain `vietnam_equity -> ["yahoo", "yfinance", "local"]`. Yahoo lists HOSE but not HNX or UPCoM, so the chain ends at `local` rather than a global source.
- `.VN` pattern and engine routing.

**`backtest/engines/vietnam_equity.py`** (new)
- Settlement: formally T+2. Since 2022 shares normally arrive before the afternoon session on T+2 and can be sold that afternoon, which is why it is often written T+1.5. Daily bars cannot represent that half-day boundary, so the engine holds two sessions, counted from the newest opening fill in the fill ledger. Counting from `Position.entry_bar_idx` releases later-bought shares early, because a same-direction increase folds into the open position without moving that index. Calendar-day arithmetic would release a Friday buy a session early. `vn_settlement_bars` makes the lag configurable.
- Price band: ±7% in normal sessions, from the reference price, quantized as the exchange publishes it (ceiling truncated down, floor rounded up). Checked against the prospective fill price before the fill, as `korea_equity` does, not against the decision bar's own close. Special sessions using ±20% need an explicit `price_limit`.
- Round lots of 100 shares. HOSE also accepts odd lots of 1-99 on a separate board that daily bars cannot represent, so odd-lot trading is not modelled and sub-lot orders round to zero.
- Costs: configurable brokerage per side, plus 0.1% personal income tax on gross sale proceeds for individual investors. Decree 253/2026/NĐ-CP keeps listed securities at `giá chuyển nhượng x 0,1%`; the 20% gain-based rate effective 2026-07-01 covers unlisted capital transfers, not exchange-traded shares.
- Long-only. `allow_short` is refused at construction. Circular 120/2020 defines a covered-short framework but routine cash-equity shorting is not operationally available.

**`backtest/engines/_market_hooks.py`, `composite.py`**
- `vietnam_equity -> VND` in the settlement-currency table. The marker it returned before was not only a label: the shadow-account backtester groups pools by it, names each pool's run directory after it and renders it into the headline currency note, so a HOSE run wrote a directory whose name contains a colon.
- `CompositeEngine` builds a Vietnam rule engine. Neither HOSE rule survives the stateless sub-engine contract — the hold needs positions and fills, the band needs the run's close panel — so both are evaluated against the composite's own state, mirroring the existing A-share T+1 interceptor. The rules are shared as module-level functions taking state and parameters separately, so the two callers cannot drift.

**Docs**: SKILL.md market entry and engine count, engine count in all five READMEs.

Known limit: Vietnamese tickers do not encode their exchange, so `.VN` cannot separate HOSE from HNX or UPCoM. Every `.VN` symbol gets HOSE's ±7% band. That is correct for everything Yahoo serves, but HNX bars supplied through `local` would need `price_limit=0.10`.

## Test Plan

- [x] Existing tests pass — full suite 10487 passed, 83 skipped
- [x] New tests added — 69 cases: suffix gate, symbol pass-through on both loaders, fallback chain, mocked daily fetch, `.VN` vs `.V` separation, tick grid and band arithmetic against a published worked example (122,301 ceiling rounds to 122,300; 106,299 floor to 106,300), settlement at each elapsed-session count, lot flooring, band-locked bars, cost stack, tick-quantized slippage. HTTP is mocked at the client boundary.

  Added in review: settlement after a same-direction increase (buy N, add N+1, partial and full sells refused on N+2, allowed on N+3), a reopened position starting its own clock, an end-to-end scale-in driven through `run_backtest` so `_execute_position_increase` runs for real, VND provenance through both the currency lookup and the shadow-account grouping that names run directories, and CompositeEngine dispatch plus both intercepted rules — settlement against the composite's positions and fill ledger, the band against its close panel with no `pre_close` column present.
- [x] Tested manually — `VIC.VN` + `FPT.VN`, 2026-02-03 to 2026-08-08, SMA(5/20), 1e9 VND.

Rebased onto `main` (`4a2f55d5`); the `agent/SKILL.md` counts were recomputed from code rather than resolved textually — `LOADER_REGISTRY` has 25 entries and the engines directory has 10 modules.

On the `main` this was first cut from (`281c8775`) that config does not reach the engine at all:

```
source=yahoo returned data for 0/2 symbols; missing: ['VIC.VN', 'FPT.VN']
akshare failed for VIC.VN: ('Connection aborted.', RemoteDisconnected(...))
  File "backtest/runner.py", line 1502, in fetch_data_map
    fallback_loader = LOADER_REGISTRY[fallback_source]()
  File "backtest/loaders/tushare.py", line 137, in __init__
    self.api = ts.pro_api(token)
Exception: api init error.
```

(With a Tushare token configured it returns an empty universe instead.) On this branch the same config completes:

```
final_value 1196340700.0   total_return 0.1963   sharpe 1.125   trade_count 6
```

Checked against the 12 fills that run produced: all whole 100-share lots, all prices on the 10/50/100-VND grid, and sell `holding_bars` of [5, 36, 3, 5, 18, 2]. The minimum of 2 is the settlement hold binding rather than being satisfied by chance.

Not covered: no intraday interval (the loaders map `4H` to Yahoo's `1h`; this engine models daily bars). HNX and UPCoM are untested, being unsupported on Yahoo. The ±20% band and non-default `vn_settlement_bars` have unit tests but no live run. `yfinance` is an unofficial Yahoo client, so that leg is best-effort under Yahoo's personal-use terms.

## Checklist


minh-vt:
LGTM ,but it would be great if you could also add support for the VNStock API.

warren618 (changes requested):
Thanks for grounding the HOSE rules and for keeping VNStock outside this patch. I reviewed head dd96d1d8. The single-market path is close, but three accounting/architecture gaps need to be fixed before we can claim Vietnam market support.

1. Vietnam has no settlement-currency mapping. `_detect_market("VIC.VN")` returns `vietnam_equity`, but `_MARKET_CURRENCY` in `agent/backtest/engines/_market_hooks.py:70-83` has no `vietnam_equity: VND`. `code_currency()` therefore returns `UNKNOWN:vietnam_equity`, which leaks into the composite currency guard and shadow-account reporting. Add VND and cover both direct currency lookup and downstream consumers.

2. CompositeEngine does not build a Vietnam rule engine. `_build_rule_engines()` in `agent/backtest/engines/composite.py:26-65` has no `vietnam_equity` branch, so `_rule_for()` can fall back to an unrelated first engine if this market reaches the composite path. Add the engine and make the price-band/settlement rules use CompositeEngine's shared bar and position state; the current stateless sub-engine contract cannot enforce either rule on its own. Do not silently disable HOSE rules in a cross-market configuration.

3. T+2 is wrong after same-direction scaling even in the single-market engine. `VietnamEquityEngine._is_settled()` checks only `Position.entry_bar_idx`, but BaseEngine's `_execute_position_increase()` preserves the original entry index while increasing size. A buy on N, increase on N+1, then full sell on N+2 is allowed even though the added shares have not reached T+2. Track settlement at fill/lot level, or conservatively block reductions until the newest remaining buy lot settles; partial reductions must never consume unsettled quantity.

Please add regressions for buy N / add N+1 / attempted partial and full sells on N+2 / allowed settlement on N+3, VND provenance, and CompositeEngine dispatch/state. Then rebase and recompute the `agent/SKILL.md` counts from current code rather than resolving that conflict textually. VNStock remains a separate follow-up and should not expand this PR.

ngoanpv:
Thanks — all three are fixed. Head is now `e440bae1`, rebased onto `4a2f55d5`.

**1. VND.** Added to `_MARKET_CURRENCY`. You were right that this reaches further than the composite guard, and further than I first credited: `src/shadow_account/backtester.py` groups pools by `code_currency`, names each pool's run directory `base_dir / currency` and renders the label into the headline `_currency_note`. A HOSE run therefore wrote a directory whose name contains a colon, which is not a legal path component on Windows. Covered at both ends — the direct lookup, and `_group_selection_by_currency` asserting a HOSE pool keys under `VND` with no `:` in any group key.

**2. CompositeEngine.** Added the branch, and both rules now read the composite's own state rather than being delegated. I kept the shape of the existing A-share T+1 interceptor rather than inventing a second pattern. The rules moved to module-level functions taking state and parameters separately, so `VietnamEquityEngine` and `CompositeEngine` share one implementation and cannot drift.

Worth being explicit about the band half: the reference price needs the run's close panel, which a stateless rule book does not have. Through the old delegation path a fill at the ceiling returned `True` and logged "band check is inactive"; through the interception it returns `False`. The regression uses `_close_arr` / `_code_to_col` with no `pre_close` column, so it fails if the band is ever silently handed back to the sub-engine.

**3. Settlement.** The clock now reads the newest opening fill from the fill ledger, which is immutable evidence of when each delta executed, rather than `Position.entry_bar_idx`. I took the conservative option: the whole position waits for its newest lot. The compressed position carries a weighted-average entry and no lot identity, so a partial reduction cannot be shown to consume settled shares only, and holding is the side that can never sell stock that has not arrived.

Regressions cover buy N / add N+1 / partial and full sells refused on N+2 / allowed on N+3, plus a reopened position starting its own clock. There is also an end-to-end test that drives a real scale-in through `run_backtest` so `_execute_position_increase` runs rather than the state being hand-built: with the old clock that run closes on bar 4, one session after the added lot, and on bar 5 with the fix.

**Rebase and counts.** `agent/SKILL.md` recomputed from code, not resolved textually — `LOADER_REGISTRY` has 25 entries and the engines directory has 10 modules, so the description keeps `main`'s 25-source list including `tickerall` and moves the engine count to 10. All five READMEs carry the same count.

Full suite: 10487 passed, 83 skipped, 69 new cases.

One thing you should weigh rather than take from me: the composite path is not reachable end to end today. `_reject_mixed_currency` refuses any book mixing `.VN` with another currency before dispatch, and no other market settles in VND, so the interception cannot currently be exercised by a real run. I implemented and tested it anyway on your reasoning — a market whose rules vanish in a composite is a trap for whoever relaxes that guard — and added a test pinning the refusal so the two invariants are visibly independent. If you would rather the guard be the documented invariant and the composite branch be dropped, say so and I will cut it.

Separately, and not in this PR: `china_a` and `india_equity` read `entry_time` / `entry_bar_idx` the same way for T+1, so a buy on N, an add on N+1 and a sell on N+1 disposes of the added shares same-day. Happy to open a follow-up if you want it fixed.

VNStock stays out, as agreed.

PR #1040, opened 2026-08-10
fix(cron): support comma-separated lists and hyphen ranges in schedule

santhreal:
## Bug

The cron schedule parser only handled `*`, `*/N`, and single integers. Comma-separated lists (e.g. `0,15,30,45`) and hyphen ranges (e.g. `9-17`) were rejected by `validate_schedule`, even though they are standard cron syntax.

## Fix

Extend `_CRON_FIELD_RE` and `_parse_cron_field` to support comma lists and hyphen ranges. `*` and `*/N` are accepted as whole-field forms only — mixing them into comma lists (e.g. `*/5,10`) is rejected because the parser cannot interpret step atoms inside a list, and accepting them would create a job that passes validation but permanently fails on execution.

## Tests

19 tests covering: validation of comma lists, ranges, mixed comma+range, inverted range rejection, out-of-range rejection, parser output for all forms, and `next_due` computation for comma/range schedules including weekend skipping.


santhreal:
Closing — upstream already implements comma-separated lists and hyphen ranges in the cron parser (see `_validate_cron_field` and `parse_cron_field` in `models.py`). This PR is a duplicate.

PR #1042, opened 2026-08-10
fix(background-tools): hold lock during all tasks dict access

santhreal:
## Bug

`BackgroundManager._execute()` mutated `self.tasks` (setting status, storing results) without holding `self._lock`, while `check()` read `self.tasks` concurrently. Under concurrent task execution, `check()` could see a partially-updated task dict or crash with a `RuntimeError` from dict mutation during iteration.

## Fix

Wrap all `self.tasks` access in `self._lock` in both `_execute()` and `check()`. The notification list append is also locked.

## Tests

7 tests covering: running status, completed status, unknown task, all-tasks summary, notification drain, concurrent check never sees partial update, and concurrent run+check no crash.


santhreal:
Closing — upstream already wraps all self.tasks access in self._lock in run(), _execute(), check(), and drain_notifications(). This PR is a duplicate.

PR #1044, opened 2026-08-10
fix(daily-count): guard read-modify-write with flock to prevent lost increments

santhreal:
## Bug

`increment_daily_count` performed a read-modify-write on the counter file without any lock. Two concurrent orders could both read the same count, increment to the same value, and write it back — losing one increment. This silently undercounts the daily order total, potentially exceeding the configured daily trade limit.

## Fix

Guard the read-modify-write with `fcntl.flock(LOCK_EX)` on a sibling lock file. The lock is held only across the counter file read + write, not across the broker call. The `fcntl` import is wrapped in `try/except ImportError` for Windows compatibility, following the existing pattern in `agent/src/channels/msteams.py`. `today = _utc_today()` is computed inside the lock so a thread that waits across the UTC midnight boundary stamps the count under the correct day.

## Tests

5 tests covering: increment returns increasing count, read returns current count, read returns 0 for new broker, 20 concurrent increments produce 20 (not fewer), and concurrent increments for different brokers don't interfere.


santhreal:
Closing — upstream already implements atomic read-modify-write with fcntl.flock (POSIX) and msvcrt (Windows), including a DailyOrderLockUnavailable exception and a context manager. This PR is a duplicate.

PR #1051, opened 2026-08-10
fix(etoro): runtime UI parity for SDK connector status

ofeksh-tr:
## Summary

- Add eToro `check_status` verify envelope (`configured`, `connection_state`, `error_code`, `credential_source`, `last_checked_at`) so `/live/status` and the Agent connector panel can surface eToro like other SDK brokers.
- Show broker-specific missing-credential guidance on `/runtime` for eToro (env vars + `~/.vibe-trading/etoro.json`) instead of Longbridge-only hints.
- Treat `broker_sdk` connectors as connected in the Agent Connector Runtime panel when `connection_state` is `connected`/`ready`, without OAuth on-ramp noise.
- Keep `configured: true` when credentials exist but the API returns auth errors (Longbridge parity).

## Why

eToro was merged as a connector (#989) but did not appear in the Agent UI and showed incorrect Longbridge setup text on `/runtime`. SDK key-based brokers were labeled “Not connected” with OAuth instructions even when credentials were valid.

Closes the runtime UI parity gap for eToro; read-only status remains the default on `/runtime` (`etoro-live-sdk-readonly`). Live trading still requires `etoro-live-trade` and a committed mandate.

## Changes

- **Backend:** eToro client/status envelope, SDK verification, live-route credential-source handling, and connector/runtime regressions.
- **Frontend:** `Runtime.tsx` setup hints, `RunnerStatus.tsx` SDK connected state, all five locales, and component/page regressions.

## Test Plan

- [x] Backend eToro connector/runtime selection: 43 passed.
- [x] Frontend RunnerStatus/Runtime selection: 38 passed.
- [x] i18n parity/interpolation selection: 20 passed.
- [x] GitHub CI: all five checks passed on exact head `e9f79ca5`.

## Risk / rollback

- **Risk:** Read-only diagnostics and UI only; no order-gate or mandate enforcement changes.
- **Rollback:** Revert the two commits on this branch.

## Checklist

- [x] No changes to protected agent-core directories.
- [x] No hardcoded secrets or credentials.
- [x] CONTRIBUTING.md style checks pass on changed files.


warren618 (changes requested):
Thanks for this — the core fix is real, and I verified it against `main`: eToro's `PAPER_GUARD` is `path_separated_key_bound` (`agent/src/trading/connectors/etoro/client.py:29`), but that value isn't in `_ENVIRONMENT_IDENTITIES` (`agent/src/api/live_routes.py:222`), so `_closed_vocabulary()` normalises it to `null` and the Runtime page renders the environment as **Unknown**. That one-line addition plus the new `configured` / `connection_state` / `error_code` fields close a genuine user-visible gap. I also checked that every new value lands inside the closed vocabularies in `live_routes.py`, that all five locales stay key-parity clean, and that the sentinel test really does prove no credential reaches the status payload.

Two things I'd like changed before merging:

**1. Please drop the unrelated reformatting.** 39 of the 61 Python hunks here are pure formatting (`+139/-41` — about 34% of the touched Python lines), reflowed to ~88 columns. This project's standard is `line-length = 120` (`pyproject.toml`) with `E501` ignored, and CI runs no format check, so nothing requires these hunks — they just make the substantive change hard to review and hard to bisect later. A few are also strictly worse, e.g. `"job id must be 1-128 characters of letters, digits, " "'_' or '-'"` in `scheduled_routes.py`.

**2. Please split out the `scheduled_routes.py` change.** It's unrelated to eToro. I checked what it actually does: the existing `-> None` handler already returns a correct empty 204 — the explicit `Response(status_code=204)` only drops a stray `Content-Type: application/json` header. That's a reasonable fix, but it deserves its own PR with a test asserting the header (`test_scheduled_routes.py` currently covers the status code and the empty body, but not the content type).

Once the diff is just the eToro runtime-parity change, this is good to merge.

ofeksh-tr:
> Thanks for this — the core fix is real, and I verified it against `main`: eToro's `PAPER_GUARD` is `path_separated_key_bound` (`agent/src/trading/connectors/etoro/client.py:29`), but that value isn't in `_ENVIRONMENT_IDENTITIES` (`agent/src/api/live_routes.py:222`), so `_closed_vocabulary()` normalises it to `null` and the Runtime page renders the environment as **Unknown**. That one-line addition plus the new `configured` / `connection_state` / `error_code` fields close a genuine user-visible gap. I also checked that every new value lands inside the closed vocabularies in `live_routes.py`, that all five locales stay key-parity clean, and that the sentinel test really does prove no credential reaches the status payload.
> 
> Two things I'd like changed before merging:
> 
> **1. Please drop the unrelated reformatting.** 39 of the 61 Python hunks here are pure formatting (`+139/-41` — about 34% of the touched Python lines), reflowed to ~88 columns. This project's standard is `line-length = 120` (`pyproject.toml`) with `E501` ignored, and CI runs no format check, so nothing requires these hunks — they just make the substantive change hard to review and hard to bisect later. A few are also strictly worse, e.g. `"job id must be 1-128 characters of letters, digits, " "'_' or '-'"` in `scheduled_routes.py`.
> 
> **2. Please split out the `scheduled_routes.py` change.** It's unrelated to eToro. I checked what it actually does: the existing `-> None` handler already returns a correct empty 204 — the explicit `Response(status_code=204)` only drops a stray `Content-Type: application/json` header. That's a reasonable fix, but it deserves its own PR with a test asserting the header (`test_scheduled_routes.py` currently covers the status code and the empty body, but not the content type).
> 
> Once the diff is just the eToro runtime-parity change, this is good to merge.

Done thanks for the feedback!

warren618 (approved):
Re-reviewed exact head `e9f79ca5`. Both prior blockers are resolved: unrelated formatting was removed and the scheduled-runs 204 fix was split into #1068. Local verification passed: 43 backend tests, 38 component/page tests, and 20 i18n parity/interpolation tests; repository CI is green. Approved.

PR #1059, opened 2026-08-10
Add Novita AI as an OpenAI-compatible LLM provider

jax-novita:
## Summary

- Register Novita AI (`https://api.novita.ai/openai`, Bearer API key) in the provider registry alongside the existing OpenAI-compatible providers
- Wire it through capabilities, legacy CLI onboarding, `.env.example`, and the README provider list

## Why

Novita AI exposes an OpenAI-compatible chat completions endpoint and follows the same `vendor/model` naming convention already used by several other providers in the registry. Adding it lets users select Novita from the built-in provider list instead of configuring a custom endpoint by hand.

## Changes

- `agent/src/providers/llm_providers.json`: add `novita` entry (`NOVITA_API_KEY` / `NOVITA_BASE_URL`, default model `moonshotai/kimi-k3`)
- `agent/src/providers/capabilities.py`: register `novita` capabilities; no reasoning-field flags since the v1 chat path doesn't expose any
- `agent/cli/_legacy.py`: add Novita to the interactive provider picker and env-var ordering
- `agent/.env.example`, `README.md`: document the new provider and its env vars
- `agent/tests/test_llm_provider_defaults.py`, `agent/tests/test_provider_diagnostics.py`: extend existing provider-catalog assertions to cover Novita
- `agent/tests/test_cli_init.py`: bump indices shifted by the new picker entry

Model IDs were verified against the live catalog at `https://api.novita.ai/openai/v1/models`.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q`: 6993 passed, 20 skipped)
- [x] New tests added (Novita coverage in `test_llm_provider_defaults.py`, `test_provider_diagnostics.py`, `test_cli_init.py`; 30 passed)
- [x] Tested manually: live call against `https://api.novita.ai/openai/v1/models` and a chat completion through `moonshotai/kimi-k3` both succeeded

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — this is a registry addition following the existing provider pattern
- [x] Documentation updated (README provider list, `.env.example`)


warren618:
Merged — this follows the add-a-provider pattern exactly, in all six places it
has to touch (`llm_providers.json`, `capabilities.py`, both env maps and
`_PROVIDER_CHOICES` in `_legacy.py`, the `_render_env_content` key list, and
`.env.example`). The note that Novita's v1 chat path exposes no reasoning
fields, so it rides the generic OpenAI-compatible path with no capability
flags, is exactly the right call and exactly the right place to record it.

Two conflicts against current `main`, both resolved toward main:

- The README provider line: main had added ModelScope while this branch added
  Novita AI. Kept both.
- `test_cli_init.py`: the branch bumped the hardcoded menu numbers 18/19
  because inserting a provider shifts the list. Main has since replaced those
  literals with `_provider_choice_number(<provider>)`, which looks the number up
  from `_PROVIDER_CHOICES` — so inserting a provider no longer requires touching
  them at all. Took the helper; your bumped literals are no longer needed.

I also added Novita AI to the other five READMEs. The same provider sentence
lives in all six and the count-consistency tests compare them to each other, so
five locales would otherwise have advertised a shorter list than the code ships.

PR #1060, opened 2026-08-11
Refine agent prompt wording and grounding with identity guidance plus support/resistance masking

wiliao:
## Summary

- Fix a grounding gate false-positive in `agent/src/agent/grounding.py` that treated technical support/resistance target labels as asserted observed price claims.
- Broaden `_PROSPECTIVE_LEVEL_RE` so markers like `支撑/阻力`, `压力位/压力线`, `support/resistance level/line/zone`, and `均值目标/平均目标` are masked as prospective levels.
- Add an identity-resolution guidance prompt and reference it from `agent/src/agent/context.py` so the agent reuses the locked symbol after `search_symbol`.
- Update the agent system prompt wording in `agent/src/agent/context.py` to satisfy output discipline tests and restore exact prompt discipline references.
- Preserve actual observed quote validation and prevent the agent from falling back to the safe-rejection message when a plan level is present.

## Why

- The agent rejected valid trading-plan prose because it parsed a support/resistance ladder level such as `9.38` as a purported observed price.
- That rejection triggered the fail-closed fallback text and blocked the update flow for file edits like `Trend/bfly-tech-trend-6-months.md`.
- The fix is in the agent grounding gate because this is the exact protected logic that decides whether a numeric claim is an observed price or a prospective level.

## Changes

- Updated `agent/src/agent/grounding.py`.
- Expanded the prospective-level regex `_PROSPECTIVE_LEVEL_RE` to recognize additional Chinese and English trading-plan markers.
- Added `agent/prompts/identity_resolution.md` with explicit symbol reuse guidance after `search_symbol`.
- Updated `agent/src/agent/context.py` to include the new prompt in the system guidance and restore precise output discipline wording.
- Updated `.gitignore` to allow `agent/prompts/` files to stay unignored.
- Added `agent/prompts/task_routing.md` as supporting task-routing guidance for agent workflows.
- Ensured the mask covers target/support/resistance language without weakening the validation on actual observed OHLC quotes.

## Note

- This change touches a protected agent-area file intentionally: the grounding gate is the exact logic that must distinguish prospective plan levels from observed OHLC price claims.
- The fix is contained to a single focus area and does not alter live order, MCP, or provider behavior.

## Test Plan

- Existing tests pass:
  - `python -m pytest agent/tests/test_agent_grounding.py -q`
  - Result: `79 passed`
- Regression coverage is provided by the grounding test module, which includes plan-level masking scenarios.

## Risk & Rollback

- [x] No live / broker / order-affecting behavior
- [x] No MCP / network / external-service behavior change
- [x] No secrets, tokens, or credentials introduced
- Rollback path: revert the commit that updates `agent/src/agent/grounding.py`

## Checklist

- [x] Protected area change (`agent/src/agent/grounding.py`) is explicitly justified in the PR body
- [x] No hardcoded secrets, API keys, or file paths
- [x] Code change is focused and limited to the grounding regex
- [x] Tests pass and no unrelated files were touched
- [x] No `Co-Authored-By:` or AI-assistant attribution trailers added
- [x] Documentation not needed beyond the PR description for this internal safety fix


warren618:
Merged, but with about half of it reverted, and you are owed the reasoning in full.

**Kept, as authored:** the grounding masks. Support/resistance ladders, historical extremes (`ATH`, 52-week, `以来最高`), document line citations, numbered headings, ratios and count nouns were all being read as observed price claims, and every new pattern is adjacency-anchored so a plain field reading like `8/12 高 149.60` stays checked. The 271 lines of tests are why I could verify that quickly. `_venue_mismatch_hints` is kept too — turning an `identity_mismatch` block into a hint that names the exact `search_symbol` call is a real improvement, and it stays additive rather than relaxing the gate.

**Kept:** the `skills/` namespace binding in `read_file_tool`. That one is an independent security fix and I would have taken it on its own: before it, `skills/<name>` was tried against `run_dir` first, so a decoy file the agent itself wrote there could shadow a bundled skill and be loaded as trusted guidance.

**Reverted:** `context.py`, the two `agent/prompts/*.md` files, and the `.gitignore` line. Three reasons that compound:

1. The rewritten Output Principles drop their operative clauses. "A number you cannot point at does not get written — not from memory, not from an earlier session" became "name the tool call"; the as-of requirement became "mention the information cutoff **when needed**", which hands the decision back to the model; the bans on prescribing position sizes and on widening scope disappeared; and principle 6 lost the requirement to state *which* principle is being refused and to then do the most useful compliant thing. Those clauses are the contract, not its wording.
2. Task Routing was replaced by a pointer to the prompt files. That converts always-present system guidance into an optional tool call, and `loop.py` truncates tool results at 10k, so the guidance becomes both skippable and truncatable.
3. The prompt files would not ship. `package-data` in `pyproject.toml` covers assets under `agent/src/`; `agent/prompts/` is neither a package nor listed, so a `pip install` would resolve the pointer to nothing. CI cannot catch that because CI runs from the checkout. (The `.gitignore` negation was also a no-op — `agent/prompts/` was never ignored.)

**One change of my own inside your masks:** the `_FX_RATE_RE` pair branch matched every `AAA/BBB <number>`, and `EUR/USD` is this project's canonical forex symbol — `backtest/engines/forex.py` names its pairs that way and `akshare_loader` accepts the slash form. So that branch stopped price validation for all of forex, and an invented rate would have passed the gate. It now requires an approximation marker, which a conversion basis carries and a quote does not: your real case, `远小于当前 USD/CAD≈1.36 的量级`, still masks, while bare `USD/CAD 1.36` is checked again.

If you want the identity-resolution and task-routing guidance as a product change, it is worth doing — as its own PR, with the prompts under `agent/src/` and registered in `package-data`, and without shortening the Output Principles. I would review that.

PR #1070, opened 2026-08-11
fix(etoro): crypto browse and flat market-data quotes

ofeksh-tr:
## Summary

- Fix crypto (and other asset-class) discovery: queries like `crypto` now browse `GET /api/v1/market-data/instruments?instrumentTypeId=10` instead of fuzzy search, which incorrectly returned HK equities.
- Centralize market-data paths (`/api/v1/market-data/instruments`, `/rates`, `/search`) so listing never hits the broken `/api/v1/instruments` route.
- Quotes and optional `include_rates` on type browse use the flat `/market-data/instruments/rates` endpoint on all profiles, including `etoro-live-sdk-readonly`.

## Test plan

- [x] `pytest agent/tests/test_etoro_connector.py` (37 passed)
- [x] Live smoke: `search_instruments("crypto")` returns `instrument_type_id=10` instruments (BTC, ETH, …) not HK tickers
- [x] Live smoke: `get_quote("BTC", profile=live-readonly)` succeeds via rates endpoint


warren618:
Merged. I checked the safety direction first, because the diff touches the
paper/live boundary, and it moves the right way: copy trading comes out of the
demo profile's capabilities and all five copy entry points gain a fail-closed
`_reject_copy_on_paper`. The order, execution, positions and info roots still
encode paper/live in the path.

I pushed one commit on top. Dropping the `/demo` prefix from `copy_root` leaves
it as the only root in `client.py` that takes a config and ignores whether it is
paper — it would hand demo credentials the real copy endpoint, and the only
thing preventing that is each of the four call sites remembering to check first.
That is a convention where the rest of the file has a structure. `copy_root`
now raises `EtoroConfigError` on a paper config; your per-function error
envelope still fires first and stays the user-facing message, and this sits
underneath it as the backstop.

The instrument-type browse, the flat market-data rates path, and the copy
capability split all went in as authored. 59 connector tests pass.

PR #1072, opened 2026-08-11
Add local multi-broker portfolio dashboard and read-only connection framework

goatyyc:
## Summary

Add a reusable, local-first portfolio dashboard that aggregates read-only holdings from multiple broker or exchange connections.

This builds on the existing Vibe-Trading connector abstraction instead of introducing a separate credential system: the repository ships the framework, schemas, UI, and adapter templates, while each developer configures credentials and optional plugins only on their own machine.

## Why

The existing connector layer can read from an active trading profile, but it does not provide a consolidated account overview across multiple providers. This change adds that missing portfolio-management layer without expanding into trade execution.

## Changes

- Add an editable Portfolio page with total assets, daily change, allocation charts, per-account status, holdings, CSV export, and manual refresh.
- Add connection management and generic reconnect actions.
- Add a generic local connection registry with normalized IDs and profile validation.
- Store secrets through the operating-system credential store; repository configuration contains only non-secret connection metadata.
- Add a fail-closed local plugin interface and scaffold for new read-only providers.
- Add portfolio aggregation, normalization, snapshots, FX conversion, and sanitized AI-analysis context.
- Reuse and extend existing IBKR MCP, Longbridge SDK, and Binance SDK read paths.
- Add authenticated REST routes and a read-only MCP portfolio tool.
- Document setup, extension points, local file locations, and security boundaries in English and Chinese.

## Security and scope

- No API keys, OAuth tokens, account identifiers, or personal filesystem paths are committed.
- Local plugins may declare only known read capabilities; order or trade capabilities are rejected.
- Credentials remain in the local OS credential store and are never returned by the settings API.
- AI context excludes credentials, account IDs, connection names, local paths, and exact quantities.
- Order placement, cancellation, transfers, withdrawals, and server-side/shared credential storage are deliberately out of scope.

## Test plan

- `10437 passed, 83 skipped` — complete backend test suite
- `61 passed` — redaction, connection registry, and authenticated connection-route tests
- `4 passed` — Portfolio frontend tests
- Ruff, Black check for added Python files, staged diff checks, and repository grep gates passed
- Staged secret and personal-path scan returned zero findings

## Notes

The implementation starts with IBKR, Longbridge, and Binance, but the connection schema and plugin scaffold are provider-neutral. A new provider can be added locally without committing its credentials or account configuration.


warren618:
Landed on main as the base of #1171 — with follow-up commits on top rather than a rebase request. Thank you @goatyyc: the connection-instance model, the immutable snapshot store, per-source provenance and the read-only plugin manifest were the right shape, and they are in as you wrote them. What changed on the way in:

- A source that fails to refresh is now an **error excluded from the totals** (`status: error`, `last_success_at`, `complete: false`), never a carried-forward cache — that is the rule #1171 settled on.
- Every UI string goes through `t()` in all seven locales (the page was `zh ? … : …` ternaries, and `de` was missing the key); backend messages are English.
- `keyring` is the optional `[keyring]` extra; plugin discovery is cached on manifest mtime; a discovery-only profile is no longer a selectable source.
- `portfolio_summary` returns `risk_xray_args` so the existing `portfolio_risk_xray` is fed rather than reimplemented, and `vibe-trading portfolio show|refresh|sources` gives the same snapshot in the terminal.
- **Not landed, deliberately:** the IBKR-specific OAuth changes in `src/tools/mcp.py` (browser User-Agent, pinned metadata, authorization-URL rewrite), the `/v1/api/mcp-public` seed, and the guessed IBKR read tool names. #1126 found the documented endpoint correct, and none of us can verify the tool names without an IBKR account. Only the provider-neutral `interactive_oauth=False` guard stayed. If you can verify against a real account, please open that part as its own PR with the request/response evidence — it is the one piece that still needs exactly that to be reviewed.

PR #1074, opened 2026-08-11
fix(providers): normalize Ollama runtime URL at source

### LINKED ISSUE #1069: Ollama provider fails with NotFoundError: 404 page not found on every agent run, despite passing preflight and working via direct curl
**Environment:**
- Vibe Trading commit: 8452a84 (main, includes #970)
- Ollama: v0.32.6
- Model: qwen2.5:3b (also reproduced with llama3.3:latest)
- OS: Ubuntu on WSL2, Windows 11
- .env: LANGCHAIN_PROVIDER=ollama, OLLAMA_BASE_URL=http://localhost:11434

**Symptom:**
Every agent run fails instantly (~180ms, 0 steps) with:
provider_stream_error provider=ollama model=qwen2.5:3b: NotFoundError: 404 page not found

Traceback originates at agent/src/agent/loop.py:817 -> agent/src/providers/chat.py:397 (stream_chat).

**What works (ruled out as causes):**
- Preflight check: OK | LLM (ollama) | qwen2.5:3b via http://localhost:11434
- Direct curl http://localhost:11434/v1/chat/completions (plain): succeeds
- Same endpoint with tools param: succeeds
- Same endpoint with tools + stream: true: succeeds
- Confirmed no duplicate/stray .env values, no ~/.vibe-trading/.env overriding config

**Repro:**
1. Configure agent/.env for Ollama as above with a locally pulled model
2. vibe-trading serve --port 8899, open web UI
3. Send any prompt
4. Immediate 404 failure

Since a hand-built request matching the same shape (tools + streaming) succeeds against Ollama directly, the discrepancy appears to be in how chat.py/llm.py constructs the actual request (headers, extra_body params, or endpoint path) versus a standard OpenAI-compatible call.

warren618:
## Summary
- Normalize Ollama base URLs to `/v1` at the centralized credential boundary.
- Remove the duplicated environment-only normalization so preflight, diagnostics, synchronized environment variables, and runtime construction share one resolved endpoint.
- Add URL-variant coverage and an actual `build_llm()` constructor regression.

## Root cause
`_sync_provider_env()` normalized the configured root, but `build_llm()` called `get_llm_credentials()` again and passed the original unnormalized URL to `ChatOpenAIWithReasoning`, producing requests against the wrong path.

## Test Plan
- Provider-focused pytest selection: 90 passed.
- Ruff and `py_compile`: passed.

Closes #1069

### DIFF
--- a/agent/src/providers/capabilities.py
+++ b/agent/src/providers/capabilities.py
@@ -312,7 +312,10 @@ def get_llm_credentials(
         Reads dynamic env vars via ``os.getenv`` — not part of ``EnvConfig``.
         When no base URL is set in the environment, falls back to the provider
         catalog's ``default_base_url`` so a provider set without an explicit
-        ``*_BASE_URL`` still hits its canonical endpoint.
+        ``*_BASE_URL`` still hits its canonical endpoint. Ollama URLs are
+        normalized here to its OpenAI-compatible ``/v1`` root so diagnostics,
+        preflight, environment synchronization, and runtime construction all
+        consume the same endpoint.
     """
     caps = get_provider_capabilities(provider, model)
     key_env, base_env = caps.api_key_env, caps.base_url_env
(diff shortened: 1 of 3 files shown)


Paulaham:
Confirmed fixed on my end after pulling main post-merge (PR #1074). Reinstalled fresh, ran with LANGCHAIN_PROVIDER=ollama / qwen2.5:3b, and the agent now runs correctly instead of hitting the 404 on every request. Thanks for the quick turnaround, @warren618 — appreciate the fix.

PR #1075, opened 2026-08-11
fix(ibkr): select market-data tier; report starved quotes as no_data

jay79-boop:
## Summary

- `reqMarketDataType` is never called anywhere in the codebase, so TWS serves every request on tier 1 (live), which requires a paid per-instrument market-data subscription.
- Without that subscription IBKR answers error 354, no tick ever arrives, and `get_quote` returns `status: "ok"` with every price field `null` — a silent failure passed through the tool layer to the agent as if it had succeeded.
- Adds a validated `market_data_type` connector config field (default `3`, the free 15-minute delayed feed), applies it ahead of `reqMktData` and `reqHistoricalData`, and returns `status: "no_data"` with a diagnostic when the ticker never fills.

## Why

A read-only research connector that reports success while handing back null prices is worse than one that errors: nothing downstream can distinguish "this instrument is quiet" from "you are not entitled to this data." Anyone running the IBKR connector without a market-data subscription hits this on their first quote, and the response gives no indication why.

TWS defaults to live data. Tier 3 (delayed) is free and needs no subscription, which makes it the appropriate default for a research connector; users who hold subscriptions can set `market_data_type: 1`.

No existing issue covers this — happy to open one first if you prefer that order.

## Changes

**`agent/src/trading/connectors/ibkr/local.py`**

- `MARKET_DATA_TYPES` / `DEFAULT_MARKET_DATA_TYPE` constants documenting the four TWS tiers.
- `IBKRLocalConfig.market_data_type` — accepts either the numeric code or the tier name, reads both `market_data_type` and `marketDataType` keys, and rejects codes TWS does not define.
- `_apply_market_data_type()` — best-effort by design: no-ops when the SDK lacks `reqMarketDataType`, and swallows errors so a tier hint can never fail a read that would otherwise succeed.
- `_wait_for_tick()` now returns whether data actually arrived. Private helper, so no API change.
- `get_quote()` returns `status: "no_data"` plus an `error` naming the likely causes when the ticker never fills; the response also reports which tier served it.
- `get_historical_bars()` applies the same tier hint.

**`agent/tests/test_ibkr_local.py`** — 7 new tests (17 cases with parametrization) covering tier ordering before both request calls, the starved-ticker path, best-effort degradation against old SDK builds, and config validation.

## Decision worth your review

**This changes effective default behavior.** Previously TWS's own default (live) applied; now delayed does. Users with paid subscriptions get delayed data until they set `market_data_type: 1`.

I chose that because the silent-null failure is the worse outcome and it affects everyone without a subscription, whereas the new default degrades data freshness for users who can trivially opt back in. If you would rather keep live as the default and rely on the new `no_data` signal to surface the problem, that is a one-line change to `DEFAULT_MARKET_DATA_TYPE` and I will push it.

## Deliberately out of scope

- `get_historical_bars` gets the tier hint but still returns `status: "ok"` with an empty `bars` list when IBKR sends nothing. Same class of problem, but empty-vs-starved is murkier for historical data and I did not want to widen a bug fix.
- No changes to order placement, mandate enforcement, kill-switch, or audit-ledger paths.
- No README or wiki updates. Say the word if the new config field should be documented there and I will add it.

## Test Plan

- [x] New tests added
- [ ] Tested manually against a live TWS session — **not done, see risk notes**

Verification run on this branch:

- `agent/tests/test_ibkr_local.py` goes from 14 to 31 tests; all 17 new cases pass.
- One pre-existing test, `test_cli_connector_routes_to_handler`, fails in my sandbox for lack of optional deps (`scipy`/`pandas` via the factor stack). It fails **identically on unmodified `main` in the same environment**, so it is environmental and not a regression from this change.
- `ruff check` on both changed files reports the same 4 findings before and after this change — no new lint errors introduced.
- `bash tools/ci_grep_gates.sh` passes.
- `python -m compileall` on the changed module is clean.

## Risk notes

Per `AGENT_CONTRIBUTOR_GUIDE.md`:

- **Live/broker risk:** none to order flow. This touches only the read path (`get_quote`, `get_historical_bars`). No order placement, mandate, kill-switch, or audit-ledger code is modified. `reqMarketDataType` is a session-scoped hint to TWS about which data tier to serve and cannot place, modify, or cancel anything.
- **Network:** unchanged — the existing loopback socket to TWS/IB Gateway. No new endpoints, no outbound calls.
- **Secrets/credentials:** none touched. The connector still holds no IBKR credentials.
- **MCP/deployment:** no changes.
- **Not validated against real Interactive Brokers infrastructure.** Every test runs against the repo's fake IB. No TWS or IB Gateway session was available, so the delayed-data path is correct-by-construction and reviewed, not empirically proven. That is the main thing I would want a maintainer with a live session to sanity-check.

## Rollback

Single commit touching two files. `git revert` restores prior behavior exactly; there are no migrations, no persisted state, and no config file rewrites. A narrower rollback is available by setting `DEFAULT_MARKET_DATA_TYPE = 1`, which restores the old effective default while keeping the `no_data` diagnostic.

## Checklist

- [x] No hardcoded values (API keys, file paths, magic numbers) — tier codes are named constants
- [x] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines; commit carries the DCO `Signed-off-by:` trailer
- [x] Documentation updated (if user-facing change) — config field is documented in the `IBKRLocalConfig` docstring; happy to extend to README on request

## Disclosure

This change was AI-assisted, prepared and reviewed per `AGENT_CONTRIBUTOR_GUIDE.md`. The bug was found by reading the connector against the IBKR API docs rather than by hitting it at runtime; the reasoning and the tests are in the diff for review on their merits.

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/agent/src/trading/connectors/ibkr/local.py
+++ b/agent/src/trading/connectors/ibkr/local.py
@@ -32,6 +32,19 @@ PROFILE_DEFAULT_PORTS = {
     "live-readonly": 7496,
 }
 
+# TWS market-data tiers. Type 1 (live) requires a paid market-data subscription
+# for the instrument; without one TWS answers error 354 and never sends a tick.
+# Types 3/4 fall back to the free 15-minute delayed feed, so they are the safe
+# default for a research/backtest connector.
+MARKET_DATA_TYPES = {
+    1: "live",
+    2: "frozen",
+    3: "delayed",
+    4: "delayed-frozen",
+}
+
+DEFAULT_MARKET_DATA_TYPE = 3
+
 
 class IBKRDependencyError(RuntimeError):
     """Raised when the optional ``ib_async`` package is not installed."""
@@ -57,6 +70,9 @@ class IBKRLocalConfig:
         account: Optional account code to filter requests.
         timeout: Connection timeout in seconds.
         readonly: Always passed as true when the SDK supports it.
+        market_data_type: TWS market-data tier (1 live, 2 frozen, 3 delayed,
+            4 delayed-frozen). Defaults to delayed so quotes work without a
+            paid market-data subscription.
     """
 
     host: str = "127.0.0.1"
@@ -66,6 +82,7 @@ class IBKRLocalConfig:
     account: str | None = None
     timeout: float = 8.0
     readonly: bool = True
+    market_data_type: int = DEFAULT_MARKET_DATA_TYPE
 
     @classmethod
     def from_mapping(cls, data: Mapping[str, Any] | None = None) -> "IBKRLocalConfig":
@@ -90,6 +107,9 @@ class IBKRLocalConfig:
             account=_clean_optional_str(payload.get("account")),
             timeout=float(payload.get("timeout") or 8.0),
             readonly=bool(payload.get("readonly", True)),
+            market_data_type=_clean_market_data_type(
+                payload.get("market_data_type", payload.get("marketDataType"))
+            ),
         )
 
     def with_overrides(
@@ -100,9 +120,12 @@ class IBKRLocalConfig:
         client_id: int | None = None,
         profile: str | None = None,
         account: str | None = None,
+        market_data_type: int | None = None,
     ) -> "IBKRLocalConfig":
         """Return a copy with CLI/tool overrides applied."""
         payload = asdict(self)
+        if market_data_type is not None:
+            payload["market_data_type"] = market_data_type
         if profile is not None:
             payload["profile"] = profile
             if port is None:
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks, and the diagnosis is right: TWS defaults to the live tier, so without a subscription the ticker never fills and the caller just gets nulls after the full poll timeout.

One follow-up landed on top (`b23aabf0`). `_apply_market_data_type()` swallowed both the missing-attribute path and any exception, while the response still echoed the requested tier back — and a quote served on TWS's own default tier is indistinguishable from one served on the tier we asked for, so the response could say `delayed` while holding live-tier data, and the starved-ticker message would then blame a subscription that may not be the cause. `market_data_type` is now `market_data_type_requested` plus `market_data_type_applied` (null when it did not take), with a warning naming the reason, and the no-data diagnosis differs depending on whether the tier was actually in force. The history path carries the same provenance, which it previously omitted. Your test that an old SDK build must not break a read is deliberately preserved — applying the tier stays best-effort, it just is not silent any more.

PR #1077, opened 2026-08-12
fix(agent): 修复东方财富研报接口要求必传时间参数的 HTTP 400 错误

zzz607:
为研究研报工具添加 `beginTime` 和 `endTime` 可选参数，允许用户按研报发布日期过滤结果，默认保持向后兼容的最近两年窗口。同时修复东方财富接口要求必传时间参数的 HTTP 400 错误。

## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/src/tools/research_reports_tool.py
+++ b/agent/src/tools/research_reports_tool.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 import json
 import logging
+from datetime import datetime, timedelta
 from typing import Any
 
 from backtest.loaders._http import (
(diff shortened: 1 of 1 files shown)


warren618:
Merged — thanks, this was a live breakage: the endpoint returns HTTP 400 for the parameter set `main` was sending, so `get_research_reports` was failing for every A-share symbol until your fix.

One follow-up landed on top (`9d2c1002`). Eastmoney answers a reversed window (`beginTime > endTime`) with HTTP 200 and `hits: 0`, and this tool turns an empty `data` array into `no research coverage found for '<symbol>'` — so a reversed range would have been reported as a fact about the company instead of an error about the request. That case is now rejected before the HTTP call, and the default-window tests freeze `now()` so the trailing two-year default is asserted exactly rather than approximately.

zzz607:
是的，你是对的，我给漏掉了这种情况

PR #1082, opened 2026-08-12
fix(backtest): report actual post-fill positions

AndyLongest:
## Summary

Separate optimizer intent from execution truth in backtest artifacts and downstream analytics.

- write post-fill, mark-to-market weights to `artifacts/positions.csv`
- preserve requested weights in `artifacts/target_positions.csv`
- use actual positions for invested-weight metrics and risk x-ray inputs
- expose both artifacts through the run API
- recommend `position_adjustment: "rebalance"` for generated target-weight strategies

## Problem

A target can move from 20% to 80% while market constraints, lot rounding, fees, or insufficient cash produce a different executed book. Previously `positions.csv` stored the target frame, so reports could claim 80% even when the portfolio still held roughly 20%. The same target values also flowed into risk diagnostics.

## Validation

- `127 passed`: base-engine, risk-xray, and run-card/API tests
- Ruff checks passed

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -409,6 +409,11 @@ class BaseEngine(ABC):
         self.fill_records: List[FillRecord] = []
         self.trades: List[TradeRecord] = []
         self.equity_snapshots: List[EquitySnapshot] = []
+        # Per-bar, post-fill portfolio weights.  These are deliberately kept
+        # separate from ``target_pos``: market rules, lot rounding, fees, and
+        # insufficient cash can all make the executed book differ from the
+        # optimiser's request.
+        self.actual_position_snapshots: List[tuple[pd.Timestamp, Dict[str, float]]] = []
         self._bar_idx: int = 0
         self._active_symbol: str = ""  # set by _rebalance/_close_position for subclass use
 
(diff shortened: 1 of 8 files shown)


warren618:
Merged. I pushed one commit onto the branch rather than sending it back, since the change is right and the gap was small.

`target_positions.csv` was written to disk and wired into the REST response, but not added to `_ARTIFACTS_SPEC` in `src/core/runner.py` — and `Runner` builds the artifact map it returns by walking that spec, so "expose both artifacts" held for the API path and not for the runner path, which is the one downstream consumers use to compare optimiser intent against executed fills. The test I added asserts the direction that matters — every artifact the engine writes is declared — so the next artifact added without registration fails there rather than going unnoticed.

On the rest: I traced the weight units before trusting them, because `_actual_weights` divides by leverage via `_calc_margin` while targets looked like notional fractions. They agree — `_plan_open_order` multiplies by leverage when sizing, so `margin / equity` really is the target weight's unit. I also checked that every bar reaches the snapshot append (the four `continue` statements are all in inner per-symbol loops) and that both `_actual_positions_frame` call sites receive the same code list. Full engine matrix passes: 537 tests across all twelve engines plus risk x-ray and metrics.

PR #1084, opened 2026-08-12
feat(reports): add interactive backtest research dashboard

AndyLongest:
## Summary

Add an optional research-report view for completed strategy backtests while preserving the existing candlestick and trade-detail views.

The dashboard includes:

- report-style strategy identity derived from symbol, prompt, run card, and date range
- headline return/risk/trading KPIs
- normalized equity versus benchmark
- drawdown with rolling Sharpe
- realized trade P&L
- trade ledger and complete metric table
- direct `Full Report` routing to the dashboard
- quiet local report-server bootstrap for CLI backtests

The run ID is retained as secondary traceability metadata instead of being promoted as the report title.

## Runtime integration

Backtests produced in a detached allowed run directory are archived into the active run after successful tool completion, so the CLI URL and web API refer to the same reportable run. A report server is reused when available or started quietly on a free loopback port.

## Validation

- backend: `48 passed`
- frontend: `10 passed`
- TypeScript/Vite production build passed
- Ruff checks passed

This PR does not replace the existing candlestick chart. It adds a separate `Research dashboard` tab and makes completed-run report links open that tab.


warren618:
Merged. The dashboard itself is a real addition and the charts, KPI strip and trade ledger all land as authored. Four commits went on top of the branch, and two questions are left open in #1094 rather than decided here.

**Removed: the automatic report server.** `cli/report_server.py` ran `cli._legacy serve` — the whole API, not a report view — with `start_new_session` / `CREATE_NEW_PROCESS_GROUP`, stdout and stderr discarded, and nothing that ever terminated it: `_SPAWNED_PROCESSES` was appended to and never read, with no `atexit`. Every run that produced a backtest could leave a loopback API listening after the command exited, unauthenticated whenever `API_AUTH_KEY` is unset — which is a supported local setup. It was called from a result-printing path and blocked there up to 30s with its child's output discarded, so the CLI just appeared to hang. Its readiness probe asserted only HTTP 200, so any unrelated local service answering that path would have been printed to the user as their dashboard, and it defaulted to port 8000 while the rest of the project uses 8899.

The metric summary is worth keeping, so it stayed: the CLI prints the metrics, the run id, and then *names* the dashboard URL with a note to run `vibe-trading serve`. A test asserts nothing spawns a process from the print path, since that is the part that must not come back.

**CLI labels are English.** They were hard-coded Chinese (`✓ 回测完成`, `总收益率`, `报告:` …) and the CLI has no translation layer to route through, so an English user saw Chinese. `load_backtest_metrics` was also a second `metrics.csv` reader, so it is now `_read_metric_values` beside the existing `_read_metrics`.

**Frontend strings go through the locale catalogue.** The component had its own switch — `tr(en, zh)` keyed off `i18n.language` — which imported the real i18n instance but bypassed the catalogue, so ja/ko/ar/es users got English and the parity test could not see the strings at all. Separately, `inferStrategyName` returned Chinese *unconditionally* (`MACD 趋势策略`, `RSI 择时策略`, …), so an English user asking to backtest a MACD strategy got a Chinese report title. 37 keys now live under `runDashboard` plus `runDetail.dashboard` in all six locales, and the `RunDetail` test asserts both directions: an English UI must not leak a Chinese title, and a zh UI must actually produce one.

**`_archive_backtest_result` re-validates its paths** with `safe_run_dir`. `backtest` already refuses a run_dir outside the allowed roots, so the source could not be arbitrary — but this function reads that path back out of a *tool result* rather than from the validated arguments, and a copy loop should not depend on a guarantee made in another module.

**One fix that CI caught before I did:** dropping the unused `TOOL_RESULT_LIMIT` import from `loop.py` was correct, but two test modules imported the constant *through* `loop.py` and collection failed. They now import from `src/config/limits.py`, matching the two test modules that already did.

I also re-merged current `main` after #1082 landed — the branch's earlier merge used a stale ref, so CI had been validating a combination that no longer existed, and #1082 changes metrics shared by every engine.

Thanks for this one — the dashboard is the part users will actually see, and it is good.

PR #1085, opened 2026-08-12
fix(alpha): surface strict alpha t-stats in bench JSON and HTML report

jay79-boop:
categorise_strict() gates on alpha_t_full, alpha_t_train, alpha_t_test and the random-control baseline, but the CLI row projection kept only the IC fields, so a strict run reported a category with no way to see the numbers behind it. The HTML report never carried them either.

Forward the four fields when present and add a Strict gate section to the report. Non-strict runs are byte-identical: the fields are only copied when the row has them, and the report section is gated on the strict flag.

## Summary

-

## Changes

-

## Test Plan

## Checklist

### DIFF
--- a/agent/src/factors/cli_handlers.py
+++ b/agent/src/factors/cli_handlers.py
@@ -80,6 +80,17 @@ _REPO_ROOT = Path(__file__).resolve().parents[3]
 # fail at universe load, not produce a Korean benchmark.
 _UNIVERSE_CHOICES = ["csi300", "sp500", "btc-usdt"]
 
+# Per-row fields that only ``bench_runner_strict`` produces. They are the
+# statistics ``categorise_strict`` actually gates on, so a strict run that
+# omits them reports a verdict the reader cannot check. Forwarded to both the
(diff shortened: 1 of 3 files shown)


jay79-boop:
Small correction to the "filed as an issue" line above. I went looking for prior
art on the categorisation before opening anything and found #649 (closed via
#657), which established the current behaviour deliberately — a confirmed factor
must clear the threshold in both train and test. So there is no issue to file,
and this PR should not be read as arguing against that gate.

The gate is right; only its evidence was invisible. With these fields exposed,
`academic_carhart_mom` on sp500 2020-2025 with `--oos-split 2023-01-01` reads
`alpha_t_full 2.7697 / alpha_t_train 0.7931 / alpha_t_test 3.0179` — correctly
not `confirmed_alive` under #649's rule, but now distinguishable from a factor
whose `alpha_t_full` is 0.011, which the JSON previously rendered identically.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1088, opened 2026-08-13
Docker build add feishu telegram channel

birdxs:
## Summary

1.编辑Dockerfile文件，使得Docker镜像编译过程增加飞书、telegram频道依赖
2.创建docker-build.yml，利用github actions编译docker镜像同时推送至ghcr和dockerhub，该工作流为手动触发，每次工作前手动输入标签会自动打上latest标签。工作流同时会把项目介绍同步推送至dockerhub项目主页。

## Why

原Dockerfile默认无飞书、telegram频道支持

## Changes

1.修改了Dockerfile文件，增加飞书、telegram频道支持
2.创建了docker-build.yml文件

## Test Plan

1.通过github actions编译docker镜像测试通过
2.本地拉取镜像部署项目，成功开启飞书、telegram频道

## Checklist

- [ √] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion
- [ √] No hardcoded values (API keys, file paths, magic numbers)
- [√ ] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x ] Documentation updated (if user-facing change)

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,106 @@
+name: Build and Push Docker Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      image_tag:
+        description: 'Image tag (e.g., v1.0.0, latest)'
+        required: true
+        default: 'latest'
+
+env:
+  GHCR_REGISTRY: ghcr.io
+  DOCKERHUB_REGISTRY: docker.io
+  GHCR_IMAGE_NAME: ${{ github.repository }}
+  DOCKERHUB_IMAGE_NAME: ${{ secrets.DOCKERHUB_USERNAME }}/${{ github.event.repository.name }}
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thanks for wiring up the multi-arch build and the Docker Hub description step.

One follow-up landed on top (`adaeeb56`), mostly to bring it in line with conventions the other workflows already follow:

- All seven actions are pinned to commit SHAs with a version comment, matching the four existing workflows in this repo; `moby/buildkit:master` and the custom `driver-opts` block are gone (that also drops an unneeded `network=host`).
- `id-token: write` was removed — there is no OIDC consumer in the job (no cosign, attestation or provenance step).
- The channel SDKs no longer come from `pip install -e ".[extras]"`. `lark-oapi`, `qrcode` and `python-telegram-bot` are in neither `requirements-lock.txt` nor its source, so that line resolved them and their transitive dependencies unhashed at build time, six lines below the comment stating dependencies come from the hash-pinned lock. They now come from a new `requirements-channels-lock.txt` (16 pins, all hashed, generated under the main lock as a constraint so no shared package diverges in version), the project install is `--no-deps`, and CI dry-runs both locks so the new one cannot rot unnoticed.
- The `workflow_dispatch` input was interpolated directly into the Summary step's `run:` body; it now goes through `env`, and the tag is validated against the Docker tag grammar before any registry work.

One thing worth knowing before the first dispatch: a newly created GHCR package is private by default, so the `docker pull` line the summary prints only works for authenticated users until the package visibility is set to public once. The summary now says so.

PR #1091, opened 2026-08-13
feat(frontend): add tearsheet tab with monthly returns heatmap and top-N drawdowns

### LINKED ISSUE #1090: [Feature] Backtest tearsheet tab: monthly returns heatmap, annual returns, and top-N drawdown breakdown
### Problem or Motivation

After a backtest finishes, the Run Detail page shows the equity/drawdown curve and scalar metrics, but the standard tearsheet views that every comparable platform ships are missing:

- **Monthly returns heatmap** — the canonical tearsheet element in QuantConnect, pyfolio, and QuantStats, and the first thing a fund manager checks for return seasonality and consistency. Today there is no monthly breakdown at all.
- **Annual returns bar chart** — standard in every tearsheet; currently only derivable by reading the equity curve.
- **Top-N drawdown episodes** — only the single max-drawdown scalar is surfaced. Peak/trough/recovery dates, per-episode depth, and recovery duration are not shown anywhere, even though they are the core of drawdown risk review.

The data is already there: every backtest writes `artifacts/equity.csv`, and the `GET /runs/{run_id}` response already carries the full rows (`artifacts_equity_csv`), so this is purely a frontend-rendering gap.

### Proposed Solution

Add a **Tearsheet** tab to the Run Detail page (hidden when a run has no equity data) that derives everything client-side from the existing equity payload:

1. **Equity curve annotated with the top-N drawdown zones** (peak→recovery shaded regions with rank labels).
2. **Monthly returns heatmap** — years × months grid with signed per-cell percentages plus a full-year column; locale-aware up/down color convention consistent with the existing candlestick chart.
3. **Annual returns bar chart** — per-year bars colored by sign with value labels.
4. **Top drawdowns table** — rank, depth, peak/trough/recovery dates, peak→trough and trough→recovery durations, with an explicit "not recovered" state.

No backend changes, no new dependencies — all derivations are pure frontend functions over the already-delivered equity rows, so the feature is unit-testable in isolation.

### Area

Web UI

### Alternatives Considered

- Compute the breakdown server-side and extend the run response with a dedicated tearsheet payload — rejected for now: `artifacts_equity_csv` already delivers the full series, so a frontend derivation avoids any API surface change and keeps the math unit-testable. A server-side payload could follow later if other surfaces (IM channels, PDF export) need the same numbers.
- Render the monthly heatmap inside the existing Chart tab — a dedicated tab keeps the tearsheet self-contained and matches how the page already organizes Studio/Validation/Run Card views.

### Contribution

- [x] I'm willing to submit a PR for this feature

shadowinlife:
## Summary

- Adds a **Tearsheet** tab to the Run Detail page: monthly returns heatmap (years × months + full-year column), annual returns bar chart, top-5 drawdown table, and an equity curve annotated with the ranked drawdown zones.
- All numbers are derived client-side from `artifacts_equity_csv` (the full equity.csv rows the run response already delivers) — no backend change, no new dependency.
- New keys in all six locales; month axis labels come from `Intl.DateTimeFormat`, so no per-locale month keys were needed.

## Why

Closes #1090.

The Run Detail page shows the equity curve and scalar metrics, but none of the tearsheet views every comparable platform (QuantConnect, pyfolio, QuantStats) ships: a monthly returns heatmap, annual returns bars, and a top-N drawdown breakdown with peak/trough/recovery dates. These are the standard post-backtest review elements — monthly seasonality, year-by-year consistency, and which drawdowns actually hurt and for how long.

## Changes

- `frontend/src/lib/tearsheet.ts` — pure derivation functions: `normalizeEquitySeries` (accepts the string `artifacts_equity_csv` rows or `EquityPoint[]`), `computeMonthlyReturns` (first-month base = first observation, gap months → null, no padding outside the data range, non-positive base → null), `computeAnnualReturns`, `computeTopDrawdowns` (running-peak episodes: depth, peak/trough/recovery times, day spans), `toDrawdownZones`.
- `frontend/src/components/charts/MonthlyReturnsHeatmap.tsx` — ECharts heatmap with signed in-cell percentages, symmetric diverging color scale built from the theme's locale-aware up/down colors, and a full-year column.
- `frontend/src/components/charts/AnnualReturnsChart.tsx` — per-year bars colored by sign with value labels and a zero line.
- `frontend/src/components/charts/TopDrawdownsPanel.tsx` — drawdown table (rank, depth, peak, trough, recovery, peak→trough and trough→recovery durations, "not recovered" state) matching the existing Studio tab table styling.
- `frontend/src/components/charts/EquityChart.tsx` — one additive optional prop `drawdownZones` rendering ranked markArea regions on the equity grid; existing call sites are byte-identical when the prop is absent.
- `frontend/src/pages/RunDetail.tsx` — `TearsheetTab` wired in right after the Chart tab, hidden when the run has no equity data.
- `frontend/src/lib/api.ts` — declares `artifacts_equity_csv` on `RunData` (the field was already in the backend response, just untyped on the frontend; it is the uncapped counterpart of the 1000-row `equity_curve`).
- Six locale files (`en`, `zh-CN`, `ja`, `ko`, `ar`, `es`) — 15 `runDetail.*` keys each.

## Test Plan

- [x] Existing tests pass (`cd frontend && npm run test:run` — 427/427 across 50 files; backend untouched)
- [x] New tests added — 21 vitest cases in `frontend/src/lib/__tests__/tearsheet.test.ts`: hand-computed monthly/annual returns, first-period base convention, gap-month nulls without padding, non-positive base, two-episode drawdown detection with exact depths/times/day spans, unrecovered tail, n-limit, unparseable timestamps, zone fallback.
- [x] Tested manually — `npm run build` (tsc + vite) clean; dev server against a real 751-row multi-year run (2023–2026): tab renders in light and dark mode, zh-CN locale shows red-up/green-down chart colors consistent with the candlestick chart, annual returns compound to the reported total return, table max drawdown matches the metrics card, and all duration day counts were verified against the dates.

## Checklist

- [x] Documentation updated (if user-facing change) — UI strings added to all six locales; no separate docs page exists for Run Detail tabs


shadowinlife:
## Visual preview — Tearsheet tab

Screenshots from a real backtest run (run `159659_dip_buy_20260604`, ~750 equity rows, 2023-04 → 2026-06 window), showing the new **Tearsheet** tab in Run Detail.

| Light | Dark |
|---|---|
| Tearsheet light | Tearsheet dark |

What's shown:

- **Monthly returns heatmap** — year × month grid with per-cell returns and a YTD column; cells outside the backtest window stay empty, colors follow the locale convention (zh-CN: red = gain, green = loss).
- **Annual returns bar chart** — one bar per calendar year with signed value labels.
- **Top-N drawdown episodes** — ranked table with depth, peak/trough/recovery dates, peak→trough and trough→recovery durations; an ongoing episode is marked as not yet recovered.

The tab only appears when a run carries equity data (`artifacts_equity_csv` or `equity_curve`), and renders in all six UI locales.

warren618:
Merged, and thank you — the tearsheet is a real gap closed, and the separation
between `lib/tearsheet.ts` (pure, tested) and the three chart components is the
right shape.

I pushed two commits onto the branch rather than sending it back.

**The month-boundary bug.** Daily backtests write date-only timestamps —
pandas drops `00:00:00` when every bar is midnight, so `equity.csv` reads
`2024-01-30,100.0,0.0`. A bare `"YYYY-MM-DD"` is specified to parse as **UTC**
midnight, while `getFullYear()` and `getMonth()` read **local** time. West of
UTC the whole series shifts back a day, so each month-boundary bar was
attributed to the previous month:

| TZ | `2024-02-01` resolves to |
|---|---|
| Asia/Shanghai, UTC | 2024-02-01 |
| America/New_York, America/Sao_Paulo | **2024-01-31** |

Run against the real artifact shape in `America/New_York`, February returned
`0.090090` (121/111) where the correct value is `0.090909` (120/110), and
January `0.110000` instead of `0.100000`. `computeAnnualReturns` had the same
problem at the year boundary.

What makes this one worth spelling out: it is correct in UTC and in UTC+8, so
it is right on CI and right on my machine, and wrong for every user in the
Americas. Your 20 tests passed in both zones. The fix pins date-only input to
local midnight, matching how the `"YYYY-MM-DD HH:MM:SS"` form already parsed,
and the four new tests run under a pinned negative-offset zone — asserting the
pin took effect first, so they cannot quietly stop testing anything.

**The heatmap's 13th column.** It carries the compounded full-year return, and
the y-axis is already a list of years, so a header reading just "Year" sits
next to a row labelled "2024" without saying how they differ. zh-CN, ja and ko
already rendered it as full-year/annual; en, es and ar now match.

Everything else went in as authored.

PR #1092, opened 2026-08-13
fix(grounding): auto-recover identity and price evidence with bounded budget

### LINKED ISSUE #1081: Grounding safe fallback stops instead of automatically recovering identity and price evidence
## Description

For market-sensitive analysis, the agent sometimes ends with the deterministic fallback:

> I could not safely lock the instrument identity or price evidence, so I did not produce a trading conclusion. Please confirm the candidate symbol and venue.

The Chinese deployment emits the equivalent message:

> 当前无法安全确认标的身份或价格证据，因此没有生成交易结论。请确认候选证券代码和交易所后再继续。

If the user replies with a simple “continue”, the agent usually resumes, calls the missing resolver or market-data tool, and completes the analysis. This suggests that the missing evidence is often recoverable automatically and does not require a genuine user decision.

## Reproduction

1. Ask the built-in agent to analyze an instrument named by company/fund name and provide a market conclusion.
2. Let the model attempt a final answer before identity and price evidence have both been grounded.
3. The grounding gate rejects the draft and provides correction feedback.
4. After repeated rejected drafts, the loop emits `safe_fallback()` and terminates.
5. Reply “continue”. The next run often calls `search_symbol`, then `get_market_data`, and successfully completes the original request.

## Current implementation

Observed on commit `652917e`:

- `agent/src/agent/grounding.py:619` — `correction_prompt()`
- `agent/src/agent/grounding.py:636` — `safe_fallback()`
- `agent/src/agent/grounding.py:669` — deterministic Chinese fallback
- `agent/src/agent/loop.py:901` — final-answer grounding validation
- `agent/src/agent/loop.py:932` — retries only while `validation_count < 3`
- `agent/src/agent/loop.py:935` — otherwise emits `safe_fallback()` as final output
- `agent/src/agent/context.py:121` — identity-before-market-data policy

The safety policy itself is valuable and should remain. The issue is that a recoverable missing-evidence state is converted into a terminal user-facing response.

`correction_prompt()` currently also says:

> If evidence is unavailable or conflicting, say so and ask for clarification; do not guess.

This treats `unresolved`, `ambiguous`, `conflicting`, and exhausted data sources too similarly. For `unresolved`, the next safe action is often deterministic and read-only.

## Expected behavior

When the missing evidence can be obtained through read-only tools without ambiguity, the agent should automatically continue the original task:

1. `identity_not_locked` + `unresolved` → call `search_symbol` in a separate assistant tool turn.
2. Exactly one authoritative candidate → lock and reuse the exact canonical symbol and venue.
3. Missing price evidence → call `get_market_data` in the next turn and use its existing bounded provider fallback.
4. Evidence obtained → regenerate and validate the final answer.
5. Ask the user only when multiple equally credible candidates remain, identity evidence conflicts, all permitted data sources are exhausted, or the next action requires trading/approval/publication authority.

The agent should not ask “whether to continue” for an unambiguous read-only recovery step.

## Suggested design

Introduce a bounded grounding-recovery state separate from rejected-draft count, for example:

```text
draft rejected
  -> identity unresolved: search_symbol
  -> unique identity locked: get_market_data
  -> price evidence obtained: resume original task
  -> final answer passes grounding: return
```

Possible independent budgets:

```python
MAX_GROUNDING_RECOVERY_ROUNDS = 6
MAX_SYMBOL_RESOLUTION_ATTEMPTS = 2
MAX_PRICE_EVIDENCE_ATTEMPTS = 3
```

The exact limits are less important than distinguishing:

- rejected prose drafts;
- resolver attempts;
- market-data provider attempts;
- genuine ambiguity requiring user input.

The correction message should give executable next actions and explicitly say not to finalize or ask the user to continue while safe recovery remains available.

## Safety constraints to preserve

- Keep `search_symbol` and dependent market-data calls in separate assistant tool-call turns.
- Reuse the exact resolver-locked symbol and venue; do not silently rewrite suffixes such as `.SS` and `.SH`.
- Do not invent prices or entry levels without observed evidence or an explicit derivation.
- Keep bounded retries and fail closed after all allowed sources are exhausted.
- Do not auto-authorize order placement, approval, publication, or other state-changing actions.
- Ask for clarification when two or more equally credible identities remain.

## Suggested regression tests

1. A premature market conclusion is rejected and automatically followed by `search_symbol`.
2. A unique resolver result is followed by `get_market_data` in a separate turn.
3. After price evidence arrives, the original analysis completes without the “confirm and continue” fallback.
4. Multiple equally credible candidates still request user clarification.
5. Exhausted market-data sources produce a bounded safe fallback without an infinite loop.
6. Symbol/venue mismatch and same-batch resolver-consumer calls remain blocked.

## Environment

- Self-hosted Docker deployment
- Vibe-Trading commit: `652917e`
- LLM: DeepSeek through an OpenAI-compatible endpoint
- The behavior is reproducible in the built-in agent flow and is not caused by an external trading execution request.

Shizoqua:
## Summary

- When a draft is rejected for a recoverable state (unresolved instrument identity, or missing price evidence on a locked identity), the loop now drives the original task forward with read-only tool turns instead of stopping at the three-draft cap and handing the user the 'confirm and continue' safe fallback.
- Recoverable-only issue sets get an executable next action from the correction (call \search_symbol\ alone this turn, then \get_market_data\ in a separate turn once identity locks) instead of the prose-only closing line that steered toward asking for clarification.
- Recovery uses its own bounded budget (recovery rounds, symbol resolution attempts, price evidence attempts), so recoverable and non-recoverable rejections no longer share the rejected-draft counter.

## Why

Closes #1081. The grounding safe fallback stopped runs that could still recover deterministically, wasting the user's time and ending in a needless fail-closed message while read-only evidence tools remained available. Because \search_symbol\ and \get_market_data\ are the only tools involved, safety is preserved: separate resolver/consumer turns stay enforced by \uthorize_tool_call\, not by prompt text, and the terminal fallback remains once the recovery budgets are spent.

## Changes

- \gent/src/agent/grounding.py\: add \MAX_GROUNDING_RECOVERY_ROUNDS\, \MAX_SYMBOL_RESOLUTION_ATTEMPTS\, \MAX_PRICE_EVIDENCE_ATTEMPTS\; track recovery budgets on the ledger; \
ecovery_action()\, \
ecord_recovery()\, \
ecovery_prompt()\, \
ecovery_summary()\; correction prompt names the next read-only tool instead of asking for clarification when a recovery step remains.
- \gent/src/agent/loop.py\: on a rejected validation, attempt a bounded read-only recovery turn before appending the correction prompt; skip the fallback path until recovery is exhausted.
- \gent/tests/test_grounding_recovery.py\: 13 tests covering recovery instruction selection across unresolved/locked/ambiguous/conflicting states, budget bounds (including that widening a budget to unlimited breaks a test), prompt content, summary state, and the loop recovery path.

## Test Plan

- [x] Existing tests pass: \gent/tests/test_grounding_recovery.py\ (13), \	est_agent_grounding.py\ (90), \	est_grounding_nan_volume.py\, \	est_agent_loop_trace.py\, \	est_agent_loop_terminal_state.py\, \	est_loop_helpers.py\ — 190 passed.
- [x] New tests added

## Checklist

- [ ] No hardcoded values (API keys, file paths, magic numbers) — budgets are named constants


warren618:
Merged. This lands on the one part of #1081 that survived retesting: the
correction prompt was prose-only and `validation_count` was a single counter
shared by recoverable and non-recoverable rejections. Splitting them and naming
an executable next tool is the right fix, and I checked the safety property you
claim in the description — `authorize_tool_call` blocks any symbol-bearing call
while `frozen_status != "locked"`, so the recovery prompt cannot talk the model
past the identity gate, and `recovery_action` only returns `get_market_data`
once identity is locked. `conflicting`, `ambiguous` and `invalidated` correctly
get no recovery at all.

One commit pushed on top, on the tests rather than the behaviour.

`test_total_recovery_rounds_are_bounded` spent the round budget by calling
`record_recovery("search_symbol")` `MAX_GROUNDING_RECOVERY_ROUNDS` times — which
also spends the symbol budget. The `None` it asserted came from the symbol cap
(2), not the round cap (6), so setting `MAX_GROUNDING_RECOVERY_ROUNDS` to 9999
left all 13 tests green. The description says widening a budget to unlimited
breaks a test; that holds for the two per-action budgets — I mutated each and
both are caught — but not for this one.

It is also unreachable as shipped: `recovery_action` hands out at most 2 + 3 = 5
recoveries, so `_recovery_rounds >= 6` never fires. That is fine as an outer
backstop for when the per-action caps are raised, but it has to be pinned as
one. The round cap is now exercised with both per-action caps patched out of the
way, and a second test records what actually binds today and asserts the
backstop stays at or above it. Both loops are driven by literal ceilings rather
than by the constant under test — a loop written as
`range(MAX_GROUNDING_RECOVERY_ROUNDS)` passes for any value of that constant,
including infinity, which is how the original slipped through.

All three budgets now fail a test when widened.

PR #1097, opened 2026-08-14
feat: positions structure visualization — weight pie/treemap + sector distribution in Run Detail

### LINKED ISSUE #1098: [Feature] Positions tab in Run Detail: weight pie/treemap, sector distribution and weight evolution
### Problem or Motivation

Position weights today only reach users as the raw `positions.csv` run artifact. For anyone holding a diversified book — especially a long/short book across asset classes — the first post-backtest questions cannot be answered inside the product:

- **No concentration view** — how much weight sits in each name and how concentrated the book is. Wind (万得) opens its portfolio module on 持仓结构/行业分布 charts as the default buy-side monitoring view; Bloomberg PORT's first screen is a position breakdown by sector/asset class with pie and treemap renderings; QuantConnect backtest reports and the pyfolio/QuantStats tearsheets (the de-facto open-source standard) render portfolio-weight plus sector-exposure charts; JoinQuant (聚宽) and RiceQuant (米筐) show 持仓分布 + 行业分布 on every backtest result page.
- **No sector/asset-class tilt view** — industry distribution is a standard buy-side monitoring chart and mandatory disclosure in fund quarterly reports; today it requires exporting the CSV to another tool.
- **No weight evolution over time** — how the book's composition shifted across the backtest horizon is invisible; only per-date rows exist in the CSV.

The data already exists — every run produces `positions.csv` with filled per-date weights — but nothing renders it, and pie/treemap series are not even registered in the frontend's tree-shaken ECharts bundle.

### Proposed Solution

A new **Positions** tab in Run Detail (visible when the run's `positions.csv` carries symbol columns), rendering the existing artifact with no new data pipeline:

**Frontend**
- **Weight distribution** — per-symbol pie ↔ treemap toggle (treemap auto-defaults above 8 symbols) with a date slider over the whole backtest horizon; short legs render as labelled slices and the uninvested remainder as a cash slice, so the pie always shows the gross book.
- **Sector distribution** — horizontal net-exposure bars; starts from an offline asset-class grouping (A-share/HK/US/crypto/…) and upgrades to Eastmoney Shenwan industries once resolved.
- **Weight evolution** — stacked area of per-symbol weights over time (downsampled, top-10 legend + merged rest).
- Register `PieChart`/`TreemapChart` in the shared ECharts bundle; i18n keys in all six locales.

**Backend**
- `GET /runs/{run_id}/positions/sectors` — read-only classification over the run's `positions.csv` symbols: asset class via the existing market detection, A-share industry via an Eastmoney Shenwan lookup. Resolved **on demand** (explicit button, never an automatic external call), cached in `artifacts/sector_map.json`, with per-symbol failure isolation so one failed lookup degrades to the asset-class view instead of aborting; non-A-share symbols never reach Eastmoney.

Semantics: pie = **gross** composition (|weight| slices incl. shorts + cash); sector bars = **net** exposure per industry/asset class — the two charts answer different questions on purpose. Research only; does not place orders.

### Area

Web UI / frontend (+ backend HTTP route)

### Alternatives Considered

- Keep weights CSV/table-only: rejected — concentration and sector-tilt review is the point of a buy-side surface, and every comparable platform renders it.
- Client-side industry classification only: partially adopted — the offline asset-class grouping ships with no network calls, but A-share Shenwan industry resolution requires a server-side lookup, hence the on-demand endpoint.
- Resolve industries automatically when a run completes: rejected — external calls must stay explicit and user-triggered; resolution is a button with a cache, never a background side effect.

---

### Scope note (2026-08-15)

The PR resolving this issue (#1097) has evolved into a **comprehensive optimization across all five P0 visualization panels** defined by the frontend capability-gap analysis, rather than the Positions panel alone:

1. **Tearsheet** (#1091, merged) — tearsheet equity-chart data memoized (no more chart re-init on parent re-renders).
2. **Factor Research** (#1099, merged) — factor artifact scan offloaded to the FastAPI threadpool; IC-correlation tooltip XSS fixed; `FactorResearchPanel ↔ RunDetail` circular import resolved.
3. **Positions** (this issue) — full feature as proposed above, plus the maintainer-review contract (gross-basis presence/cash semantics, signed auto/symmetric axes, abs-ranked non-cancelling evolution remainder, bounded threadpool sector resolution) and tooltip XSS hardening.
4. **Options Lab** (#1096, merged) — payoff route gained the same unexpected-exception envelope as the chain route.
5. **Attribution** (gap-analysis P0 #4) — new panel delivered in the same PR: `GET /runs/{run_id}/attribution` with factor alpha/beta decomposition and Brinson-Fachler attribution (table, waterfall, rolling beta/alpha, cumulative chart), computed entirely from persisted artifacts.

Cross-cutting: a shared `escapeHtml` now guards every artifact-derived string interpolated into ECharts tooltip HTML across all five panels (ECharts renders custom formatter output via `innerHTML` unescaped; artifact content is attacker-influenceable via crafted run directories, and the API key in `localStorage` made that an escalation path).

shadowinlife:
## Summary

Adds a **Positions** tab to the Run Detail page — portfolio-book structure visualization rendered from the existing `positions.csv` artifact (no new data pipeline): per-symbol weight pie/treemap with a date slider, sector/asset-class net-exposure bars, and a weight-evolution stacked area. Closes #1098.

Pie = **gross** composition (|weight| slices incl. shorts + cash); sector bars = **net** exposure per industry/asset class. A long/short pair in one industry nets to 0 on the bars while both legs stay visible on the pie — the two charts answer different questions on purpose. Research only; does not place orders.

## Round 2: a comprehensive optimization across all five visualization panels

After maintainer review and a five-agent post-implementation audit (goal verification, hands-on QA, code quality, security, context mining), this PR grew from the Positions panel into a **comprehensive optimization pass over all five P0 visualization panels** the gap analysis defined (Tearsheet, Factor Research, Positions, Options Lab, Attribution). One patch, five panels:

| # | Panel | Where it shipped | What this patch contributes |
|---|-------|------------------|------------------------------|
| 1 | **Tearsheet** | #1091 (merged to main) | Tearsheet equity-chart data memoized — the chart no longer disposes and re-initializes on every parent re-render |
| 2 | **Factor Research** | #1099 (merged to main) | `GET /runs/{id}/factor` artifact scan offloaded to the FastAPI threadpool (was blocking the event loop with up to 8 MB of synchronous reads); CorrelationMatrix tooltip XSS fixed; `FactorResearchPanel ↔ RunDetail` circular import resolved (`RunCardPanel`/`RunCardStat` extracted to `components/run/RunCard.tsx`) |
| 3 | **Positions** | this PR | Full maintainer-review contract (gross-basis presence/cash semantics, signed auto/symmetric axes, abs-ranked non-cancelling evolution remainder, bounded threadpool sector resolution) + tooltip XSS hardening + symlink-safe sector cache + shared bounded resolver executor |
| 4 | **Options Lab** | #1096 (merged to main) | `POST /options/payoff` gains the unexpected-exception envelope (502 `{"ok": false, ...}`) already documented for the chain route — no more raw 500s |
| 5 | **Attribution** | **new in this PR** | The fifth P0 panel, treated as a bug per the capability-gap analysis: Brinson-Fachler + factor attribution rendered from persisted artifacts (details below) |

### Cross-panel security fix (the audit's blocking finding)

ECharts 6 renders **custom tooltip formatter output via `innerHTML` without escaping** (verified against the installed `TooltipHTMLContent`), while artifact-derived strings — `positions.csv` column headers and date cells, factor bundle directory names, equity timestamps — are attacker-influenceable through a crafted run directory. Because the API key lives in `localStorage`, one hovered tooltip could escalate to full API control, including live-trading endpoints.

Fix: a shared `escapeHtml()` now wraps every artifact-derived interpolation in every panel's chart tooltips (Positions pie/sector/evolution, factor IC-correlation matrix, tearsheet equity, attribution charts), with a regression test. Pure-numeric and i18n-static interpolations are untouched.

### New: Attribution panel (panel 5)

`GET /runs/{run_id}/attribution` (new route module `attribution_routes.py`, registered under the api_server thin-assembler line budget) computes performance attribution **entirely from persisted run artifacts** — no data re-fetched — and degrades per section with notes instead of raising:

- **Factor section** — OLS of portfolio daily returns on the benchmark series: beta, per-period and annualized alpha with t-statistic, R², observation count, 60-bar rolling beta/alpha series, and cumulative portfolio-vs-benchmark compounding. Benchmark is the run's configured ticker when explicit, otherwise an auto equal-weight basket of the run's codes (`mode` reports which).
- **Brinson section** — single-period Brinson-Fachler via `quantlib.attribution`: allocation / selection / interaction with a per-sector table that ties out to the active return. Mode is `symbol` (auto benchmark), `asset_class` (explicit benchmark across asset classes, with a Cash row and buy-and-hold approximation note), or `invested_cash` degradation.
- **Frontend** — Attribution tab in Run Detail: alpha/beta/R²/IR/tracking-error stat cards, Brinson table with signed coloring, allocation→selection→interaction waterfall, rolling beta/alpha chart, cumulative attribution chart; loading/error/no-data states; full six-locale i18n.
- Bounds follow the factor-endpoint precedent: ≤500-point series with the last point pinned, 6-decimal float rounding, per-field degradation, `run_in_threadpool` offload.

## Screenshots

| Light — resolved industries | Dark — resolved industries | Dark — offline asset-class fallback |
|---|---|---|
| light | dark-resolved | dark-fallback |

## Changes

**Backend**
- `GET /runs/{run_id}/positions/sectors`: read-only classification over the run's `positions.csv` symbols — asset class via the existing `_detect_market`, A-share industry via `resolve_industry_board` (push2 `slist?spt=1`). Resolved on demand, cached in `artifacts/sector_map.json` (`?refresh=1` bypasses), per-symbol degradation, non-A-share symbols never reach Eastmoney. Blocking work runs in the FastAPI threadpool under a bounded contract: ≤200 network lookups, a shared ≤4-worker executor, the shared client's 15 s per-request timeout, symlinked `artifacts/` rejected, `total_symbols`/`symbol_limit` reported.
- `GET /runs/{run_id}/factor`: artifact scan now offloaded via `run_in_threadpool` (identical response shape).
- `GET /runs/{run_id}/attribution`: new, as described above.
- `POST /options/payoff`: unexpected tool exceptions now return the 502 error envelope instead of a raw 500.

**Frontend**
- **Weight distribution** — per-symbol pie ↔ treemap toggle (auto-defaults treemap above 8 symbols) with a date slider; short legs render as labelled slices, the uninvested remainder as a gross-basis cash slice (omitted when fully committed or levered); card subtitles state each chart's denominator (weights are margin ÷ NAV).
- **Sector distribution** — horizontal net-exposure bars with signed auto/symmetric axes; offline asset-class grouping upgrading to Eastmoney Shenwan industries on demand; >12 groups merge into an Other bucket.
- **Weight evolution** — stacked signed area, ranked by average absolute exposure, remainder split into non-cancelling long/short series.
- **Attribution** — new tab (stat cards, Brinson table, waterfall, rolling beta/alpha, cumulative chart).
- Security: shared `escapeHtml` across all chart tooltip formatters that interpolate artifact strings.
- Quality: tearsheet chart-data memoization; `hasPositions` date-column predicate shared with the parser; resolve-button gating aligned with the backend's `a_share` default for unknown symbols; `FactorResearchPanel ↔ RunDetail` circular import removed; `useChartLifecycle` promoted to a shared hook and adopted by the panel charts.
- `PieChart`/`TreemapChart` registered in the shared ECharts bundle; i18n keys in all six locales.

## Verification

- Backend: `pytest agent/tests/api/test_positions_sectors.py agent/tests/test_attribution_api.py agent/tests/test_factor_api.py agent/tests/test_options_routes.py agent/tests/test_api_infrastructure.py` green (attribution suite covers auto/explicit benchmark modes, symbol/asset-class Brinson, degradation notes, 404/400, downsampling with last point pinned).
- Frontend: `vitest run` green incl. escapeHtml regression, XSS-sink coverage, attribution tab states, short/neutral/levered book semantics; `npm run build` (incl. `tsc -b`) clean.
- Five-agent audit: goal verification PASS, hands-on QA 38/38 scenarios PASS (real Eastmoney/Yahoo flows included), code quality and security findings resolved by this patch.
- Live smoke against a 5-name A-share run: industries resolved (银行Ⅱ/白色家电/白酒Ⅱ/保险Ⅱ), cache hit 47 ms vs 5.2 s cold.


shadowinlife:
**Why position-structure charts are table stakes — prior art across comparable products:**

- **Wind (万得)** — portfolio module opens on 持仓结构/行业分布 charts (pie + bar) as the default buy-side monitoring view; sector allocation is one of its most-used screens
- **Bloomberg PORT** — position breakdown by sector/asset class with pie and treemap renderings is the first screen portfolio managers see; the treemap specifically exists to make weight concentration readable at a glance
- **QuantConnect** — backtest reports render a portfolio-weight pie plus sector exposure; pyfolio/QuantStats tearsheets (the de-facto open-source standard this repo's tearsheet work follows) include sector exposure charts
- **JoinQuant (聚宽) / RiceQuant (米筐)** — every strategy backtest result page shows 持仓分布 (pie) and 行业分布 bars; it is the expected output of any backtest in the CN quant ecosystem
- **Morningstar / fund quarterly reports** — holdings pie + industry allocation are mandatory disclosure visuals; buy-side users read them daily

Vibe-Trading currently exposes filled weights only as a raw `positions.csv` artifact, so a user holding a diversified book cannot see concentration or sector tilt without exporting to another tool. This PR closes that gap with the same data the backend already produces.

warren618 (changes requested):
The Positions tab is a valuable addition, but the current implementation contradicts the PR's own gross-pie/net-sector contract for short books. I reviewed head 4a93a1d6 and am requesting changes on the following correctness and runtime issues.

1. `latestHoldingDate()` at `frontend/src/lib/positions.ts:87-95` decides that a date has holdings from the signed sum. A +50% / -50% market-neutral book sums to zero, so the entire tab renders as no positions. Presence must be based on finite gross exposure, `sum(abs(weight))`.

2. `withCashSlice()` at lines 160-168 computes cash as `1 - sum(signed weights)`, while the pie renders position slices with `abs(weight)`. A +50% / -50% book therefore shows 100% gross positions plus 100% invented cash. Derive residual cash from the same gross/margin basis when the artifact contract makes that valid; otherwise omit the inferred cash slice rather than fabricating it. The displayed denominator and labels must state whether the chart is capital, gross exposure, or normalized composition.

3. Both signed charts hide shorts. The sector x-axis and evolution y-axis are hard-coded to `[0, 1]` at `PositionsTab.tsx:438-441` and `514-517`. Evolution ranking also uses signed average weight at lines 139-147, pushing persistent shorts out of the top set, and the single `rest` series cancels omitted longs against omitted shorts. Use signed auto/symmetric axes, rank by average absolute exposure, and preserve long and short remainder separately (or another non-cancelling representation).

4. The async route calls synchronous Eastmoney work directly. `get_run_positions_sectors()` at `agent/src/api/runs_routes.py:445-474` invokes `_build_positions_sector_map()`, which loops over every symbol and performs a blocking lookup at lines 104-110 with no symbol cap. One large portfolio can block the event loop for many network timeouts. Run the blocking path in FastAPI's threadpool or make the route synchronous, and add a bounded symbol/concurrency/timeout contract.

Please add tests for short-only, market-neutral, levered/mixed long-short, negative sector/evolution axes, non-cancelling remainder, and a large symbol list. Rebase after #1099 is merged because both PRs modify Run Detail and the same six locale files. #1098 should close only with the corrected merge.

shadowinlife:
Thanks for the precise review — all four points verified against the code and fixed in 2fd7ec68 (rebased onto current main, which also resolves the conflicts with #1099).

**1. `latestHoldingDate()` signed presence test** — fixed. Presence is now decided by finite gross exposure `sum(|weight|) > 0.001`; a +50%/−50% market-neutral book renders normally. Regression test added.

**2. `withCashSlice()` fabricated cash** — fixed. The engine contract is `weight = direction × margin_value / equity` with `sum(|weight|) ≤ 1` (`backtest/engines/base.py`), so the residual is now derived on the same gross basis: `cash = 1 − sum(|weight|)`, omitted entirely when the book is fully committed or levered (`sum(|weight|) ≥ 1`) instead of inferred from the signed sum. Denominator semantics are now stated on every card via subtitles (`basisGross`: slices = |weight|, margin ÷ NAV; `basisNet`: signed weight ÷ NAV, negative = net short), keys added in all six locales. Regression tests: market-neutral (no cash slice), short-only (cash = 1 − gross), mixed long-short, levered (no cash).

**3. Signed charts hiding shorts** — fixed on all four sub-points:
- Sector x-axis and evolution y-axis now use a signed auto/symmetric range computed from the data (`[0, nice(max)]` when everything is non-negative, symmetric `[−nice, +nice]` once any exposure is negative) instead of hard-coded `[0, 1]`.
- Evolution ranks by average **absolute** exposure, so persistent shorts stay in the top-10 legend set.
- The single cancelling `rest` series is split into separate `Other (long)` / `Other (short)` remainder series (each emitted only when non-zero).
- Tests cover the symmetric net-short axis, the zero-based long-only axis, and the non-cancelling remainder.

**4. Async route blocking the event loop** — fixed. `get_run_positions_sectors` now runs `_build_positions_sector_map` via `run_in_threadpool`, and the resolution is bounded by an explicit contract (documented at the constants and in both docstrings): at most **200** network industry lookups per resolve (excess symbols degrade to asset-class grouping and land in `unresolved`), at most **4** concurrent workers, and the shared Eastmoney client's **15 s** per-request socket timeout. The envelope now reports `total_symbols` / `symbol_limit`. Test added: a 250-symbol A-share book makes exactly 200 mocked network calls and degrades the capped tail.

**Verification after the fix**
- `pytest agent/tests/api/test_positions_sectors.py` → **13 passed**; `tests/test_factor_api.py + test_api_infrastructure.py + test_sector_tool.py + test_security_auth_api.py` → 126 passed (rebase regression check).
- `vitest run` → **539 passed** (55 files), incl. the new market-neutral / short-only / levered / negative-axis / non-cancelling-remainder cases.
- `npm run build` (incl. `tsc -b`) clean; `ruff check` clean.

shadowinlife:
## Round 2 pushed (9da94eef) — comprehensive optimization across all five panels

Following a five-agent post-implementation audit (goal verification, hands-on QA 38/38, code quality, security, context mining), this PR now extends beyond the Positions panel into a comprehensive pass over all five visualization panels. **Several fixes deliberately touch chart code that is already on `main`** (from the merged panel PRs), so they are called out explicitly here:

**Fixes to code already on main:**

| Already-merged code | Problem | Fix in this PR |
|---|---|---|
| Factor endpoint `GET /runs/{id}/factor` (#1099) | Artifact scan (≤8 MB reads + CSV parsing) ran synchronously on the event loop | Offloaded via `run_in_threadpool` (+ regression test) |
| `CorrelationMatrix` tooltip (#1099 + the Correlation page) | Factor directory names interpolated unescaped into ECharts tooltip HTML | `escapeHtml` applied |
| `EquityChart` tooltip (used by the Tearsheet, #1091) | Equity timestamps interpolated unescaped into tooltip HTML | `escapeHtml` applied |
| Options payoff route (#1096) | Unexpected tool exception leaked a raw 500 | 502 error envelope, matching the chain route (+ test) |
| Tearsheet equity chart data (#1091) | Rebuilt every render → full chart dispose/re-init | Memoized |
| `FactorResearchPanel ↔ RunDetail` circular import (#1099) | Fragile cycle via `RunCardPanel`/`RunCardStat` | Extracted to `components/run/RunCard.tsx` |
| Chart lifecycle boilerplate (7 components across #1091/#1096/#1099) | Duplicated init/ResizeObserver/dispose blocks | Promoted to shared `useChartLifecycle` hook |

**Cross-panel security fix (the audit's blocking finding):** ECharts 6 injects custom tooltip formatter output via `innerHTML` without escaping, while artifact-derived strings (positions.csv headers/date cells, factor directory names, equity timestamps) are attacker-influenceable through a crafted run directory — and the API key in `localStorage` made that an escalation path to full API control including live-trading endpoints. A shared `escapeHtml` now guards every artifact-derived tooltip interpolation across all panels (unit-tested).

**Positions hardening (this PR's own surface):** sector cache rejects symlinked `artifacts/`; industry resolution now reuses one shared bounded executor instead of a per-request pool.

**New: Attribution panel** (gap-analysis P0 #4, treated as a bug): `GET /runs/{run_id}/attribution` (factor alpha/beta decomposition + Brinson-Fachler in symbol/asset_class/invested_cash modes, computed entirely from persisted artifacts, per-section degradation, ≤500-point downsampling, threadpool offload) with a full Run Detail tab — stat cards, Brinson table, waterfall, rolling beta/alpha, cumulative chart — and six-locale i18n. Registered under the `api_server.py` thin-assembler budget (399/400 lines).

**Verification:** frontend 554 vitest tests passed + `tsc` + build clean; backend contract suites 89 passed (attribution 7/7, positions 15, factor 14, options 11, infrastructure 42); ruff clean.

he-yufeng:
Read the sectors route closely. One logic slip in the cap: in _build_positions_sector_map the budget check is `index < _POSITIONS_SECTOR_MAX_SYMBOLS`, which counts positions across the whole symbol list rather than the A-shares actually heading to the network. A mixed book with more than 200 non-A-share names in front (say a US-heavy cross-asset run) pushes every A-share past index 200 and gets zero industry resolution, while the 200-lookup budget sits completely unused. The cap test only exercises an all-A-share list, where list position and lookup count happen to coincide, so it stays green either way. Gating on `len(a_share_to_resolve) < _POSITIONS_SECTOR_MAX_SYMBOLS` matches the documented contract (at most 200 network lookups, excess degrade), and a mixed-book regression case would lock it in.

One smaller note in the same function: the symlink guard covers the artifacts directory but not the cache file itself, so a crafted `sector_map.json` symlink would still be followed by write_text. A `cache_path.is_symlink()` check next to the directory check closes it.

Commenting only, not approving: I have not read the full diff end to end.

shadowinlife:
Thanks — both points are valid and both fixed in 6742ea99.

**1. Cap counts list position instead of A-share lookups** — confirmed, real bug. `index < _POSITIONS_SECTOR_MAX_SYMBOLS` gates on the symbol's position in the full list, so a US-heavy cross-asset book with 200+ non-A-share names in front pushes every A-share past index 200 and resolves zero industries while the 200-lookup budget sits unused — contradicting the documented contract. Fixed exactly as suggested: the gate is now `len(a_share_to_resolve) < _POSITIONS_SECTOR_MAX_SYMBOLS`, so the budget counts A-share network lookups only.

Two regression cases lock it in:
- `test_mixed_book_resolves_a_shares_after_non_a_share_prefix` — 205 US symbols followed by 5 A-shares: all 5 resolve (5 mocked lookups), `unresolved == []`. On the old code this makes **0** network calls and leaves all 5 unresolved (verified by reverting the fix and re-running).
- `test_mixed_book_cap_still_bounds_a_share_lookups` — 205 US + 250 A-shares: exactly 200 lookups, the A-share tail degrades to `unresolved`, so the cap still binds on mixed books.

**2. Symlink guard misses the cache file itself** — confirmed. The guard covered `artifacts/` but not `sector_map.json` inside it, so a planted cache symlink was still followed by both the cache read (`_load_json_file`) and the rewrite (`write_text`) — a write primitive outside the run directory. The check now covers both, next to the directory check as suggested: a symlinked cache file is rejected the same way as a symlinked artifacts dir (the codebase's reject-don't-follow policy, same as the factor scan), returning the `no positions artifact` note.

Regression: `test_symlinked_cache_file_rejected_without_write_through` — real artifacts dir + real `positions.csv`, only `sector_map.json` is a symlink to a sentinel target; asserts the note envelope, zero network calls, and the sentinel target left byte-identical (old code overwrote it).

**Verification**
- `pytest agent/tests/api/test_positions_sectors.py` → **18 passed** (15 prior + 3 new); all 3 new tests fail against the pre-fix code, pass with it.
- `pytest agent/tests/api/test_positions_sectors.py agent/tests/test_attribution_api.py agent/tests/test_factor_api.py agent/tests/test_options_routes.py agent/tests/test_api_infrastructure.py` → **92 passed**.
- `ruff check` clean on both touched files.

warren618 (changes requested):
Thanks for the thorough rework — all four review points are fixed at 6742ea99, each with regression coverage. latestHoldingDate gates on gross exposure (positions.ts:102), withCashSlice derives cash from 1 − Σ|w| and omits it when levered (positions.ts:190), the sector and evolution axes are signed/symmetric with ranking by average absolute exposure and separate Other(long)/Other(short) series (PositionsTab.tsx:56,174,205), and the sector route now runs in the threadpool with a 200-symbol A-share cap, 4 shared workers, disk cache, and symlink rejection (runs_routes.py:473,1012). The requested test matrix — market-neutral, short-only, levered/mixed, negative axes, non-cancelling remainder, large symbol list — is all present, locale key parity holds across all six languages, and the escapeHtml/useChartLifecycle consolidations are a genuine improvement.

What still blocks: the attribution module is a whole second feature (783-line core, new route, four charts) unrelated to #1098 — please split it into its own PR. It also repeats the signed-sum cash bug we just blocked: symbol mode computes cash as 1 − Σ signed weights (attribution_core.py:583), inventing 100% cash for a market-neutral book, and asset-class mode divides weight-weighted returns by the signed class weight (attribution_core.py:633/643), which explodes when long and short legs nearly cancel. Move it out, fix those two semantics with tests, and ship it separately.

Minor nits: the new attribution charts should use the useChartLifecycle hook this PR introduces (they hand-roll init/ResizeObserver/dispose instead); treemap label color and palette hexes are hardcoded; a symlinked positions.csv is read while the cache file is symlink-rejected.

The positions-visualization part itself is otherwise ready once the attribution split lands.

he-yufeng:
Confirmed both fixes in 6742ea99, and both read correct:

1. The budget now counts A-share lookups (`len(a_share_to_resolve) < _POSITIONS_SECTOR_MAX_SYMBOLS`), so a mixed book with a long non-A-share prefix still resolves its A-shares. `test_mixed_book_cap_still_bounds_a_share_lookups` pins the bound, which is the case that regressed before.
2. The symlink check now rejects a symlinked `sector_map.json` inside a real artifacts dir, not just a symlinked dir, and `test_symlinked_cache_file_rejected_without_write_through` covers the write-through path.

Not approving yet since I have only read the sector-map portion of the diff closely; the chart components and the other routes deserve a pass too. From this corner it looks good.

shadowinlife:
Ran a full 5-lane post-implementation review (goal verification, hands-on QA, code quality, security, context mining) across the final state of this PR. It surfaced one blocking crash plus several hardening tails — all fixed in 564c783d.

**Blocking (fixed)**
- `attribution_core.py` — asset-class Brinson divided by a net-zero class weight: a market-neutral pair inside one asset class (+0.5/−0.5, exactly what pair-trading runs produce) raised `ZeroDivisionError` and 500-ed the whole `/runs/{id}/attribution` endpoint, losing the factor section with it. Reproduced end-to-end, now degrades to the invested/cash fallback. Same family: `benchmark_equity[0] <= 0` is now guarded with a skip note (the factor path already guarded it). Regression tests: `test_attribution_market_neutral_pair_degrades_instead_of_crashing`, `test_attribution_benchmark_equity_starting_at_zero_skips_brinson`.

**Security hardening (fixed)**
- Artifact-derived symbols (`positions.csv` headers / `config.json` codes) are interpolated into `ohlcv_<symbol>.csv` paths; path separators, `..` sequences and NUL bytes are now rejected so crafted artifact names cannot escape the artifacts directory. Regression: `test_attribution_artifact_symbol_path_traversal_is_contained` (plants a leakable file outside `artifacts/`, fails on pre-fix code).
- `_read_positions_symbols` now rejects a symlinked `positions.csv`, aligning with the artifacts-dir and cache-file guards. Regression: `test_symlinked_positions_csv_rejected_without_read`.
- `sector_map.json` cache is written atomically (temp file + `os.replace`) so concurrent `refresh=1` requests cannot tear it.

**Consistency tails (fixed)**
- The three attribution charts (waterfall / cumulative / rolling beta-alpha) now use the shared `useChartLifecycle` hook this PR introduces, instead of hand-rolling the same init/ResizeObserver/dispose boilerplate (~60 lines deduped).
- `escapeHtml` applied to the CandlestickChart trade-marker tooltip fields (`trades.csv` side/price/qty/reason) — the last artifact-derived tooltip sink on the Run Detail page, same XSS class this series eliminates cross-panel.
- Misleading Brinson skip note corrected (missing ohlcv artifacts no longer reported as "positions.csv could not be parsed"); `_attribution_equal_weight_returns` docstring aligned with actual partial-coverage behavior.
- DCO: the previous tip commit was missing `Signed-off-by`; the branch was rebased to add it (old tip `6742ea99` → `3b950995`, content unchanged — verified by empty tree diff).

**Verification**
- Backend: `pytest test_attribution_api.py test_positions_sectors.py test_factor_api.py test_options_routes.py` → **54 passed**; all 4 new regression tests fail against the pre-fix code (verified by temporary revert).
- Frontend: `vitest run` → **57 files / 554 passed** (covers the hook migration via AttributionTab tests); `tsc -b` clean; `ruff check` clean on all touched backend files.

shadowinlife:
Metadata cleanup: per CONTRIBUTING.md ("Do NOT add Co-Authored-By: trailers or AI-assistant attribution lines"), the AI attribution trailers have been stripped from the two commits that carried them; the branch was rebased accordingly (previous tip `f8c5fe65` → `564c783d`, tree unchanged — empty diff). All five commits keep their `Signed-off-by:` and the DCO requirement is met. Head is now `564c783d`.

shadowinlife:
Visual QA pass (head `1b92f892`): rendered all five P0 panels against a fixture run with real backtest artifacts + synthetic factor bundles, captured every tab with Playwright, and cross-checked the visuals against API ground truth plus independently recomputed values.

**Verdict: all panels render correctly.** Numbers verified exact in the KPI strip, tearsheet (monthly heatmap incl. the −6.2% Oct cell), positions donut/treemap/industry bars (43/22/15/10/10 after eastmoney industry resolution), Brinson table arithmetic (allocation + selection + interaction = active return), waterfall, rolling beta/alpha, cumulative returns, factor IC stats/correlation matrix, and options payoff (long call + iron condor breakevens/max P&L/Greeks).

The QA surfaced three cosmetic defects, fixed in `1b92f892`:
1. **Legend swatch mismatch** — attribution cumulative & rolling charts set only `lineStyle.color`, so legend swatches fell back to the default palette; `itemStyle.color` now matches the line color on every series.
2. **Factor long-short spread format** — rendered as a bare decimal (`0.4168`); now a signed percent (`+41.68%`), and signed IC stats (IC mean / IR / spread) take positive/negative tones via `RunCardStat`.
3. **Options payoff dead zone** — the default nice x-domain extended to zero, leaving blank space left of the curve; axis is now pinned to the curve's data extents.

Plus regression tests: new `ChartOptionConsistency.test.tsx` asserts legend/line color parity and the payoff axis extents; factor panel test updated to the signed-percent format. Frontend suite 557/557, `tsc -b` clean. Each fix was re-verified visually after the change.

Note: the console `TypeError: Cannot read properties of null (reading 'layerStack')` on tab switches is an upstream zrender-6 dispose race (echarts 6.0.0–6.1.0, unfixed on master): `dispose()` nulls the painter internals without invalidating an in-flight progressive paint frame. Console-only — the DOM is already torn down, no visual impact, no leak; the `useChartLifecycle` dispose sequence here is correct.

Issue #1016, opened 2026-08-08
desktop: validate a signed 0.3.0 to 0.3.1 updater flow


QCYTSN:
## Context

This is the updater follow-up explicitly kept out of Windows packaging PR #1015. It must not enable an updater or release feed until signing and release ownership are settled.

## Dependencies

- #1015 reviewed and merged;
- an agreed Authenticode signing identity;
- an agreed release repository/feed owner;
- signed 0.3.0 and 0.3.1 test artifacts.

## Required end-to-end validation

Run the complete flow on a clean Windows machine or reverted VM snapshot:

- install and launch signed 0.3.0;
- configure representative local settings and an encrypted test credential;
- discover a signed 0.3.1 update from the real candidate feed;
- download and verify publisher signature and expected release metadata;
- stop the backend/watchdog without leaving a Python PID or loopback listener;
- apply the update and relaunch 0.3.1;
- verify application version, frontend/backend startup, preserved user data, and readable encrypted credentials;
- simulate an interrupted or failed update and document a tested recovery path;
- verify that a tampered, unsigned, wrong-publisher, or downgraded artifact is rejected.

## Gate

The updater remains disabled by default until the signed 0.3.0 -> 0.3.1 run passes and recovery instructions are documented. This issue is validation/release work, not part of PR #1015.

warren618:
This is correctly scoped and correctly gated, and it is blocked on decisions
that are mine rather than on work that is yours. Rather than leave it sitting, here
is where each dependency actually stands:

- **#1015 merged** — done.
- **Authenticode signing identity** — not settled. This is the real blocker.
  The project has no code-signing certificate and no organisational identity to
  issue one under, and an unsigned or self-signed updater is worse than no
  updater: it trains users to accept an unverified binary, which is the exact
  failure mode this issue's last checklist item is meant to prevent.
- **Release repository/feed owner** — follows from the above. Not settled.
- **Signed 0.3.0 / 0.3.1 test artifacts** — cannot exist until the identity does.

So the honest status is: the validation plan is accepted as written and I do not
want it changed, but it cannot run yet, and no amount of work on the updater code
moves it. I would rather say that plainly than leave you waiting on a review.

Two things that are unblocked and useful in the meantime, if you want them:

- The **interrupted/failed-update recovery path** and the **backend/watchdog
  shutdown leaving no Python PID or loopback listener** can both be built and
  tested unsigned. The shutdown one especially — we have already had a real
[...]

PR #1101, opened 2026-08-14
feat(desktop): add dormant update safety boundary

QCYTSN:
## Summary

- add a strict, PID-scoped backend/watchdog shutdown result for future update handoff;
- add dormant Windows candidate verification and interrupted-attempt recovery primitives;
- document and test the tampered / unsigned / invalid-signature / wrong-publisher / downgraded rejection matrix.

## Why

Refs #1016.

The signed `0.3.0 -> 0.3.1` run is still blocked on maintainer-owned signing
identity and release-feed decisions. The issue identified two useful pieces that
can be completed without those decisions: proving that desktop cleanup never
kills unrelated Python processes, and making the future verification/recovery
contract executable before a real certificate exists.

## Changes

- `BackendManager.stopForUpdate()` succeeds only after the exact owned backend
  PID, watchdog PID, and loopback listener are all gone.
- Graceful-shutdown and parent-death smoke tests each keep an unrelated Python
  sentinel alive while the owned process tree is removed.
- Candidate verification checks semantic-version monotonicity, expected
  SHA-256, Windows Authenticode status, and an explicit allowlist of SHA-256
  signer-certificate fingerprints.
- Rejections use stable codes; an empty or malformed publisher allowlist fails
  closed.
- A same-directory atomic journal records only `verified -> backend-stopped ->
  installer-launched` and reconciles completed/interrupted attempts without
  following arbitrary artifact paths.
- Windows CI runs the deterministic rejection/recovery suite and a real
  Authenticode adapter check against a signed Windows system binary.
- `UPDATE_SAFETY.md` records the rejection matrix, recovery matrix, remaining
  enablement gates, and rollback limitation.

## Deliberately out of scope

- no updater call site, update UI, background check, release feed, or download;
- no installer launch and no change to the signed/unsigned packaging commands;
- no publisher identity, certificate, or release owner is selected;
- no claim that a partially modified NSIS installation can already be rolled back;
- no changes under `agent/src/agent/`, `agent/src/session/`, or
  `agent/src/providers/`.

The updater remains disabled by default. The real signed clean-Windows upgrade
and interruption run tracked in #1016 remains the release gate, so that issue
remains open.

## Test Plan

- [ ] Existing full backend suite (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`) — not run locally because this PR does not change Python/backend code; repository CI remains authoritative.
- [x] New tests added.
- [x] Tested manually on Windows.

Local checks completed:

```text
npm ci
npm run build
npm run test:update-safety
npm run test:locales
npm run test:backend-resolution
npm audit --audit-level=high
node scripts/smoke-lifecycle.mjs
node scripts/smoke-parent-death.mjs
git diff --check upstream/main...HEAD
```

Observed results:

- TypeScript production build passed;
- npm audit reported zero vulnerabilities;
- real Windows Authenticode inspection returned `Valid` with a SHA-256 certificate fingerprint;
- all synthetic acceptance/rejection and journal-recovery cases passed;
- graceful update handoff removed the owned backend/watchdog/listener while an unrelated Python PID survived;
- killing only the Electron main PID removed the owned backend/listener while a second unrelated Python PID survived.

## Risk and rollback

The verifier and journal have no runtime caller, so this PR cannot check,
download, or apply an update. The lifecycle behavior adds bounded post-shutdown
verification; normal application shutdown continues to use the existing path.
Rollback is a single revert of this commit. No user data migration or external
service state is introduced.

## Checklist

- [x] No hardcoded values (API keys, user file paths, publisher identities)
- [x] Documentation updated
- [x] Commit contains the required DCO `Signed-off-by:` trailer

### DIFF
--- a/.github/workflows/desktop-windows.yml
+++ b/.github/workflows/desktop-windows.yml
@@ -48,6 +48,10 @@ jobs:
         working-directory: desktop/electron
         run: npm ci
 
+      - name: Verify dormant update rejection and recovery policy
+        working-directory: desktop/electron
+        run: npm run test:update-safety
+
       - name: Verify encrypted credential storage
         working-directory: desktop/electron
         run: npm run smoke:credentials
--- a/desktop/electron/README.md
+++ b/desktop/electron/README.md
@@ -8,6 +8,12 @@ credential storage. This layer does not contain an updater, optional IM
 adapters, personal WeChat pairing, or changes to the agent loop and provider
 behavior.
 
+It also contains dormant, fail-closed building blocks for a future signed
+updater: Authenticode/publisher/version verification, an interruption-recovery
+journal, and a strict backend shutdown result. They do not check a feed,
+download an artifact, launch an installer, or enable updates. See
+[UPDATE_SAFETY.md](UPDATE_SAFETY.md).
+
 ## What the shell does
 
 - Enforces a single desktop application instance.
@@ -22,6 +28,8 @@ behavior.
 - Reports startup failures and exposes retry and log-folder actions.
 - Requests authenticated graceful shutdown before asking the watchdog to
   terminate the owned Windows process tree as a fallback.
+- Exposes a strict update-handoff shutdown call that succeeds only after the
+  owned backend PID, watchdog PID, and loopback listener are all gone.
 - Terminates the Python process tree if the Electron main process is killed
   without running JavaScript shutdown handlers.
 - Localizes desktop-owned loading, status, error, dialog, and menu text in
@@ -55,11 +63,15 @@ The Windows lifecycle suite used by CI can be run from the same directory:
 
 ```powershell
 npm run smoke:lifecycle
+npm run test:update-safety
 ```
 
 It verifies localized message parity, graceful shutdown, authentication, and
 the parent-death path by force-terminating only the Electron main PID before
-checking that the Python process and loopback listener are gone.
+checking that the Python process and loopback listener are gone. Both graceful
+and parent-death tests keep a separate Python sentinel alive throughout the
+owned-process cleanup. The update-safety test exercises the rejection matrix
+and every journal recovery phase without enabling an updater.
 
 Backend resolution is intentionally narrow. It checks an explicit
 `VIBE_TRADING_EXECUTABLE` override first, then exact application/resource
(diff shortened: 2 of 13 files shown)


warren618 (dismissed):
Thanks for the careful dormant boundary and the strong Windows test matrix. I reviewed head d8723223 and the current green checks. I am requesting changes because the existing tests do not cover four failure modes that can make this boundary attest more than it has proved.

1. Failed shutdown discards the ownership state needed to recover. In `desktop/electron/src/backend-manager.ts:194-216`, `stop()` records incomplete evidence but still clears the watchdog handle, backend PID, base URL, error state, and log stream before `stopForUpdate()` throws. If taskkill fails or the listener survives, the installer is correctly blocked, but the manager can no longer retry against the exact owned process tree. Please retain the ownership state until all three predicates are true, or return a typed failure/recovery handle that preserves those identifiers. Add a regression test where shutdown evidence remains incomplete and a second cleanup attempt can target the same owned PIDs without touching the sentinel process.

2. HTTP failure is not evidence that the listener is gone. `listenerAcceptsConnections()` at lines 477-485 treats any 500 ms `/health` timeout or protocol error as `listenerClosed=true`. A live listener that accepts TCP but hangs, resets HTTP, or responds slowly therefore passes the update gate. Please probe TCP acceptance/port ownership rather than application health, and add a test with a listening-but-hung endpoint.

3. `UpdateRecoveryJournal.begin()` has a check-then-rename race. Lines 55-60 check that `pending-update.json` is absent, while lines 138-150 later rename a unique temp file over the final path. Two concurrent `begin()` calls can both pass the read and the later rename can replace the first attempt. Please reserve the journal atomically (exclusive create, lock, or equivalent CAS) and test that exactly one concurrent begin succeeds and its record is not overwritten.

4. Candidate acceptance is not tied to one immutable byte sequence. `verifyWindowsUpdateCandidate()` hashes the artifact at line 81 and inspects Authenticode later at line 91 through the same mutable path. A replacement between those reads can make the returned hash and signature describe different files; a replacement after both reads can invalidate a later launch. Please make the staged artifact immutable for the verification/launch contract, or add a final identity/hash verification after signature inspection and require the exact same recheck immediately before launch. Cover a deterministic mutation case.

Please also bound the synchronous PowerShell probes with a timeout so a stuck host cannot hang the Electron main process indefinitely.

Separately, the current body sentence `does not close #1016` is parsed by GitHub as a closing keyword. I will correct that wording independently; #1016 must remain open for the real signed 0.3.0 -> 0.3.1 upgrade and interruption run.

warren618 (approved):
Verified on head 3012fe98. The requested safety gaps are covered by regression tests: retained shutdown ownership and retry, TCP listener probing, atomic journal reservation, locked candidate identity with pre-launch hash recheck, and bounded PowerShell probes. All required CI checks are green, including Windows source lifecycle and NSIS packaging.

PR #1105, opened 2026-08-14
Fix: agent run reliability (grounding false-rejections, final-answer gate, LLM timeouts)

wiliao:
## Summary

- **Grounding gate (`agent/src/agent/grounding.py`)**: stop rejecting correct drafts.
  - `_timestamp_matches_claim_date` now matches report-style weekday/intraday-suffixed
    date cells (`08-10(一)`, `08-10(周一)盘中`, `08-10盘中`) via a new `_claim_date_tuple`
    helper; previously such cells matched no evidence row and every price in the row
    was reported `numeric_claim_unavailable`.
  - `_CSV_FILENAME_SUFFIX_MAP` maps `_US` → `.US`, so bash-written US OHLC CSVs
    (`INTC_US.csv`) are ingested instead of silently contributing no evidence.
  - `_SHORT_DATE_RE` also masks dash-form dates (`08-10`) in prose, not only slash form,
    so the date digits are not mistaken for price claims.
  - New `_ORDER_LEVEL_RE` masks order/limit syntax — `100 @ $3.50`, `100 股 @ 4.00`,
    GTC and labelled orders (`挂单/限价单/限价/委托/订单/买入价/卖出价`) — so order
    levels and share counts are treated as prospective, not observed, quotes.
  - `correction_prompt` now tells the model to *remove* rejected claims instead of
    repeating them in a different format.
- **Final-answer gate (`agent/src/agent/loop.py`)**: never release tool-call markup as an answer.
  - New `_looks_like_tool_call_syntax()` catches DSML/tool-call markup (plain, ASCII-bar,
    and fullwidth-bar forms) emitted as prose on the forced-text final iteration; the answer
    is retried with a plain-text instruction, or replaced with a neutral message when no
    budget remains. New `tool_call_syntax_in_answer` trace event.
  - A single empty LLM response (no text, no tool calls) is retried once with a nudge;
    only a second consecutive empty response fails the run, so a transient provider
    empty completion cannot kill a run that has already written its deliverable.
  - Revision cap raised 3 → 4 when iteration budget remains, so the run stops surrendering
    early after rejected drafts.
  - Degraded outcomes (safe-fallback released) now set `degraded: true` plus a human
    reason in the trace `end` event and the run result, so the UI can distinguish
    "delivered" from "answered with a fallback".
  - Wrap-up nudge now tells the agent to finish outstanding file writes first instead of
    "stop calling tools", which is fatal for file-update tasks.
- **LLM call timeouts (`agent/src/config/env_schema.py`, `agent/src/agent/loop.py`, `agent/src/providers/chat.py`)**:
  a silent provider stall used to freeze a run indefinitely.
  - New `VIBE_TRADING_LLM_TIMEOUT_SECONDS` (default 300s).
  - Both `stream_chat` calls bound the provider call and pass an `idle_timeout_s`
    per-delta watchdog; a stall now surfaces as a retryable `ProviderStreamError` instead
    of hanging.
  - The auto-compact summary call is bounded; on timeout/failure compaction degrades to
    hard truncation with a transcript note instead of freezing the loop.

## Why

The recurring "update trading plan / trend report" task failed repeatedly in production on 2026-08-14:

1. The INTC 6-month trend report update failed **3 times in a row** with
   `numeric_claim_unavailable` rejections of correct prices — the report's own date
   format (`08-10(一)`) was never matched to the evidence, and the US OHLC CSV written by
   the agent (`INTC_US.csv`) was never ingested.
2. A run ended on the forced-text path by releasing raw tool-call markup (fullwidth-bar
   DSML mojibake) as its "final answer".
3. A run **silently stalled for 31 minutes** with no LLM activity after a tool failure —
   the auto-compact summary call had no timeout; only a manual interrupt ended it.
4. The RXRX weekly update was rejected because a **GTC order summary**
   (`100 @ $3.50 / 100 @ $4.00`) was treated as observed quotes (`numeric_claim_conflict`),
   and the run then **failed on a transient empty LLM response** 16 seconds after it had
   already written and verified the target file.

These are agent-reliability fixes in the protected grounding/answer logic; no trading
order, broker, or external-service behavior is touched.

### `agent/src/agent/grounding.py`

- `_CSV_FILENAME_SUFFIX_MAP`: added `("_US", ".US")`.
- `_SHORT_DATE_RE`: added dash-form branch (`08-10`) with the same bounds/fencing as the
  existing slash form.
- New `_ISO_CLAIM_DATE_PREFIX_RE` / `_YEARLESS_CLAIM_DATE_PREFIX_RE` / `_claim_date_tuple()`;
  `_timestamp_matches_claim_date` compares against the extracted (month, day) instead of
  requiring a full-cell match.
- New `_ORDER_LEVEL_RE` (quantity @ price, and order-keyword + price) applied in
  `_numbers_without_dates_or_percent` alongside the existing prospective-level mask.
- `correction_prompt`: added "remove or reword every rejected claim" guidance.

### `agent/src/agent/loop.py`

- New `_llm_timeout_seconds()` (config + module-level override, mirroring
  `_tool_timeout_seconds`).
- `_looks_like_tool_call_syntax()` + rejection/retry/fallback path in the no-tool-call
  branch; `tool_call_syntax_in_answer` trace event.
- Empty-response retry: `MAX_CONSECUTIVE_EMPTY_RESPONSE_SKIPS = 1`; the first empty
  completion appends a nudge and retries; the second consecutive empty breaks with the
  existing `empty_model_response` reason.
- Revision cap: `revision_cap = 4 if max_iterations - iteration >= 3 else 3`.
- `_released_fallback` flag; `degraded: true` + reason in `end` event and result dict.
- Wrap-up nudge wording: finish outstanding file writes first.
- Both `stream_chat` calls pass `timeout=` and `idle_timeout_s=`; auto-compact summary
  call passes `timeout=` and degrades to hard truncation on failure.

### `agent/src/providers/chat.py`

- `stream_chat` gained `idle_timeout_s`; raises a retryable `ProviderStreamError` when no
  delta arrives within the budget.

### `agent/src/config/env_schema.py`

- Added `vibe_trading_llm_timeout_seconds` (alias `VIBE_TRADING_LLM_TIMEOUT_SECONDS`,
  default 300.0).

## Note

- `agent/src/agent/grounding.py` and `agent/src/agent/loop.py` are protected agent-area
  files; changes are deliberately scoped to the grounding gate, the final-answer gate,
  and LLM-call timing — none weaken identity gating or the numeric-claim safety checks
  (fabricated prices are still rejected; missing currency/source is still rejected).

## Test Plan

- New regression tests:
  - `test_agent_grounding.py::test_weekday_suffixed_claim_dates_match_evidence`
  - `test_agent_grounding.py::test_run_dir_ohlc_csv_us_filename_maps_symbol`
  - `test_agent_grounding.py::test_order_level_prices_are_not_observed_quotes`
  - `test_agent_loop_dsml_tool_calls.py::test_agent_loop_never_releases_tool_call_syntax_as_a_final_answer`
  - `test_agent_loop_content_filter.py::test_empty_content_no_filter_retries_once_then_breaks`
  - `test_agent_loop_content_filter.py::test_single_empty_response_recovers_with_nudge`
  - `test_agent_loop_terminal_state.py::test_empty_model_response_returns_specific_reason` (updated)
  - `test_chat_llm_streaming.py::test_stream_idle_timeout_aborts_a_stalled_stream`
  - `test_chat_llm_streaming.py::test_stream_idle_timeout_allows_a_flowing_stream`
  - `test_loop_helpers.py::test_llm_timeout_seconds_default_and_override`
- Verified in the sandboxed environment (tmp-path-dependent tests could not execute there;
  no regressions in the runnable subset):
  - `tests/test_agent_grounding.py tests/test_agent_output_discipline.py` → **84 passed**
  - `tests/test_chat_llm_streaming.py tests/test_loop_helpers.py` → **56 passed**
- End-to-end: the exact drafts rejected in production now pass the gate (08-10(一) date
  cells; GTC order summary `100 @ $3.50 / 100 @ $4.00`); fabricated prices and
  missing-currency drafts are still rejected.

## Risk & Rollback

- [x] No live / broker / order-affecting behavior
- [x] No MCP / network / external-service behavior change
- [x] No secrets, tokens, or credentials introduced
- Rollback path: revert the commits touching `agent/src/agent/grounding.py`,
  `agent/src/agent/loop.py`, `agent/src/providers/chat.py`, `agent/src/config/env_schema.py`.

## Checklist

- [x] Protected area changes (`grounding.py`, `loop.py`) explicitly justified in the PR body
- [x] No hardcoded secrets, API keys, or file paths
- [x] Code change is focused: four related reliability areas, no unrelated files touched
- [x] Tests pass; new regression tests added for each fix
- [x] No `Co-Authored-By:` or AI-assistant attribution trailers added
- [x] Documentation: see `Docs/Fix-Proposal-Agent-Run-Reliability-Remaining.md` for the
  remaining (P1–P3) issues tracked for future implementation


warren618:
This is four independent pieces of work in one branch, and the review would go much faster split up. Concretely:

**1. The grounding fixes are already on main — please drop those hunks.** They are also the only reason this PR now conflicts. All four landed today, by a slightly different route in two cases:

- report-style date cells (`08-10(一)`, `08-10盘中`) and the `_US` CSV stem → `335eb0ef`
- the dash-form trading day → `b22c974e`. Note the shape: masking every `MM-DD` would swallow a genuine price range, since `目标价 10-20 元` is written exactly like an October date. A zero-padded month reads as a date on its own; 10/11/12 have to carry a weekday or session marker.
- order/limit syntax (`100 @ $3.50`) → `f00e7a5a`, narrowed in `03e1ff65`. A bare `@ <price>` branch turned out to be too wide: dates are masked first, so an observed close written `收盘 2026-08-10 @ 8.20` arrived as `@ 8.20` and stopped being checked. An order level now needs its quantity or an order label.

**2. The loop reliability work is the part only you have**, and it is the reason this PR matters. I verified all three against the source before writing this: neither `stream_chat` call passes a timeout, so a silent provider stall does hang a run; `loop.py` breaks out and fails the run on a single empty response; and there is no guard anywhere against tool-call markup being released as the final answer. Please keep this as the PR, on its own — it touches the protected agent core and deserves a review that is not sharing a diff with three other things.

**3. The Canadian `.TO`/`.V` venue-sibling fallback** (`market_data.py`) is self-contained and unrelated to any of the above. As its own PR it can land quickly.

**4. The cmd.exe heredoc guard** (`bash_tool.py`) is likewise self-contained and small.

One cleanup either way: `agent/tests/_iso1.txt`, `_iso2.txt`, `_iso4.txt`, `_iso5.txt`, `_iso6.txt` and `_bash_written.md` look like run artifacts that got committed by accident.

The production traces in your description are what made these findable — thank you for writing them up in that detail.

PR #1107, opened 2026-08-15
feat(runs): add a positions tab to run detail

he-yufeng:
Closes the chartable slice of #1098.

Run Detail gains a Positions tab, fed entirely by artifacts the run response already carries (`positions.csv` / `target_positions.csv`), so there is no backend change:

- Latest Weights: gross/net exposure cards, a donut for long-only books (it flips to signed bars as soon as any short leg shows up), plus the full weight table
- Weight Evolution: stacked area across rebalances, top 8 symbols by latest weight plus an Other bucket
- Target vs Actual: latest optimizer targets against post-fill weights, sorted by absolute drift, so blocked or rounded fills stay auditable

Out of scope, deliberately: the sector/asset-class tilt from #1098 (the product has no industry mapping source today, and deriving one from symbol prefixes would be wrong), and a treemap (the donut and bars answer the same concentration question without another render path).

Test plan: new unit tests for the parsing/drift helpers plus two RunDetail tests (tab render with drift math, and hidden-when-absent gating): 19 passed. `tsc -b` clean, `npm run build` clean. The full frontend suite shows 32 failures in apiAuth/useDarkMode/useSSE/SettingsQVeris/Layout that reproduce identically on clean main in this environment (verified with a stash), untouched by this change.

Frontend-only, no broker/MCP/network surface. Rollback is reverting the single commit.

### DIFF
--- a/frontend/src/components/charts/PositionWeightsChart.tsx
+++ b/frontend/src/components/charts/PositionWeightsChart.tsx
@@ -1,0 +1,88 @@
+import { useEffect, useRef } from "react";
+import i18n from "@/i18n";
+import { getChartTheme } from "@/lib/chart-theme";
+import { echarts } from "@/lib/echarts";
+import { useThemeDark } from "@/lib/theme-store";
+import type { WeightPoint } from "@/lib/positions";
+
+interface Props {
+  points: WeightPoint[];
+  symbols: string[];
+  height?: number;
+}
+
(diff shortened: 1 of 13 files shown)


he-yufeng:
Closing as duplicate: #1097 already implements this (positions tab from positions.csv, closes #1098) and is further along in review. I missed it in my pre-work dedup pass, my fault. Will put review energy there instead.

PR #1117, opened 2026-08-16
feat(i18n): add German (Deutsch) language support

1psconstructor:
Add de.json with 1216 translated keys for the full UI. Register German in SUPPORTED_LANGUAGES and localeLoaders.

## Summary

-

## Changes

-

## Test Plan

## Checklist


warren618:
Merged — German is the seventh UI language. Your key parity was exact: 1216 keys against `en.json`, none missing, none extra, which is rarer than it should be.

Two commits went onto your branch before the merge, both about coverage rather than content:

1. **The parity suite could not see German.** `i18n.test.ts` iterated a hardcoded map of locale imports, so `de.json` was covered by no test at all — perfect parity and zero verification. The suite now derives its locale set from the `locales/` directory, and a new case pins that set against `SUPPORTED_LANGUAGES` in both directions, so an eighth language cannot ship untested either.
2. **The PR would have failed CI once it ran.** A separate case asserted the exact ordered list of six codes; it now includes `de`.

I also changed one string: `scheduled.noActiveMandate` read "dann committed werden", a passive clause with no subject. It now says who commits what.

PR #1118, opened 2026-08-16
fix(tests): sandbox the suite's config root before collection on Windows

### LINKED ISSUE #1116: [Bug] Test suite escapes its sandbox on Windows and writes to the real ~/.vibe-trading, including the live audit ledger
## Description

**What happened:** Running the project's own test suite on Windows wrote into the **real** `~/.vibe-trading/` config root instead of a temporary one. Among other things it appended three synthetic `order_rejected` records to the live, hash-chained audit ledger at `~/.vibe-trading/live/audit.jsonl` — one per full suite run — and created a `live/audit_chain.jsonl` containing a single NUL byte.

**What I expected:** A test run to leave the user's config root untouched. The suite already has guards for exactly this, and two of them are among the failures — they are not incidental, they *are* the mechanism:

- `agent/tests/test_runner_env.py::test_prepare_sandbox_home_reexposes_only_loader_paths`
- `agent/tests/test_loader_retry_helpers.py::test_stubbed_config_cannot_enable_cache_or_redirect_root_into_cwd`

Because the redirection does not hold, a large block of the 40 failures I see on Windows are downstream of it rather than independent — e.g. every `test_governance.py` failure reports `LedgerCorruptionError: ledger chain already broken at index=0 ... malformed_json`, which is the suite reading back the real NUL-byte `audit_chain.jsonl` it created on an earlier run.

**Why this is worth more than the usual "tests are messy on Windows":** the audit ledger's entire value is that it is append-only and tamper-evident, and it is the artefact a live deployment is audited against. After three test runs mine held three entries describing order rejections that no human and no agent ever initiated. Nothing was traded and nothing could have been — every record is the mandate gate correctly denying an `mt5` `place_order` with `"no valid mandate on file"`, there is no broker connector configured, and there are no broker credentials on the machine. The integrity problem is with the *record*, not with any position. But a user who later runs `verify` against that ledger, or who ships it as evidence, has a corrupted chain and three fabricated denials with no way to tell them from real ones.

I found this while preparing #1115 and running the suite to check that PR for regressions.

## Steps to Reproduce

On Windows, with the repo checked out and the dev extra installed:

1. Confirm the config root is absent or note its contents: `ls ~/.vibe-trading/live` (mine did not exist).
2. `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py -q`
3. `ls -R ~/.vibe-trading/live`

After the run, the following exist that did not before:

```
~/.vibe-trading/live/audit.jsonl                        one order_rejected record per suite run
~/.vibe-trading/live/audit_chain.jsonl                  1 byte, NUL
~/.vibe-trading/live/alpaca/.order_submit.lock
~/.vibe-trading/live/robinhood/oauth/                   (empty)
~/.vibe-trading/live/runtime/heartbeats/robinhood.heartbeat
~/.vibe-trading/runs/2026MMDD_HHMMSS_*/                 one run directory per suite run
~/.vibe-trading/reports/alpha_bench_*.html              4 files
~/.vibe-trading/sessions/session-1/{trace.jsonl,run_manifest.json}
~/.vibe-trading/fonts/NotoSansCJKsc-Regular.otf         downloaded during the run
~/.vibe-trading/shadow_accounts/                        (empty)
~/.vibe-trading/workspace/                              (empty)
```

The count scales with runs: I ran the full suite three times and got exactly three ledger records and three run directories.

Note the font download — the run also makes a network call, which may be worth separating out.

## Error Logs

```shell
# One of the three records written to the REAL ledger (formatting mine):
{"audit_id": "la_379b20a8...", "ts": "2026-08-16T07:57:56.668+00:00", "session_id": "",
 "kind": "order_rejected", "outcome": "blocked",
 "gate_decision": {"allowed": false, "decision": "deny", "checked_limits": ["mandate"]},
 "server": "mt5", "remote_tool": "place_order", "error": "no valid mandate on file"}

# The knock-on effect, once a NUL-byte chain file exists in the real root:
E   src.governance.ledger.LedgerCorruptionError: ledger chain already broken at
    index=0 seq=None reason=malformed_json: Expecting value: line 1 column 1 (char 0)

# The two guards that are themselves failing:
FAILED agent/tests/test_runner_env.py::test_prepare_sandbox_home_reexposes_only_loader_paths
FAILED agent/tests/test_loader_retry_helpers.py::test_stubbed_config_cannot_enable_cache_or_redirect_root_into_cwd
```

For reference, the same command on `main` (`9806936`) gives me `40 failed, 10368 passed, 91 skipped, 9 errors`. I have not tried to work out how many of those 40 disappear once the isolation holds, but the governance block (16) and several of the symlink/permission ones look like candidates.

## Interface

Python API — specifically the project's own `pytest` suite, not a user-facing entry point.

## LLM Provider

anthropic (not exercised by this — no model call is involved in the failure)

## Version

Repository `main` at `9806936`. Installed distribution alongside it is `vibe-trading-ai 0.1.13`.

## Environment

Windows 11 (10.0.26200) / Python 3.13.15 (python.org, in a venv)

---

**Suggested direction, offered tentatively since I do not know the history here.** The cleanest fix is probably a session-scoped `autouse` fixture in `agent/tests/conftest.py` that points the config root at `tmp_path_factory` and fails loudly if anything resolves outside it, rather than each test redirecting for itself — that turns "isolation quietly did not apply" into an error at collection time. A narrower fix would be to make the two guard tests above hard-fail the run on Windows instead of joining a block of forty, since as it stands their failure is easy to read as platform noise. Happy to put a PR together for whichever you'd prefer, or to test a patch on Windows if that is more useful — I have the environment in front of me and this reproduces every time.

lorenzozanee:
## Summary

- Install a sandboxed config root at conftest import time: `VIBE_TRADING_HOME`, `HOME` and `USERPROFILE` now point at a per-session temp directory before pytest collects any test module.
- On Windows the suite previously resolved the real `~/.vibe-trading` config root (live mandate + audit ledger) because `Path.home()` reads `%USERPROFILE%` and ignores `$HOME`, so the historical `HOME`-based isolation idiom was inert there.
- Module-level installation is required because many modules bake path constants at import time (`loop.RUNS_DIR`/`SESSIONS_DIR`, `goal`/`session` `_DB_PATH`, `helpers.ENV_PATH`, memory dirs, ...), which resolve during collection — before any fixture runs.

## Why

Closes #1116.

Running the project's own test suite on Windows wrote into the real `~/.vibe-trading`, appending synthetic `order_rejected` records to the live, hash-chained audit ledger at `~/.vibe-trading/live/audit.jsonl` and creating a NUL-byte `audit_chain.jsonl`. The audit ledger is append-only and tamper-evident; a test run corrupted it in a way that a later `verify` cannot distinguish from real entries.

## Changes

- `agent/tests/conftest.py`: install the sandbox at module import time (before collection), with a session teardown and a guard that fails loudly if the sandbox is not active.
- `agent/tests/test_config_root_sandbox.py` (new): pins the sandbox contract, including the Windows `HOME`/`USERPROFILE` case and the audit-ledger path.
- `agent/src/core/runner.py` + `agent/tests/test_runner_env.py`: fall back to copying the re-exposed loader paths when symlink creation fails on Windows (privilege 1314), so the ephemeral-HOME guard holds there.
- Per-test `monkeypatch.delenv("VIBE_TRADING_HOME")` where a test asserts the default-home fallback; `USERPROFILE` added to the loader-cache-root guard; `src.live.paths.get_runtime_root` patch added to the mt5 fixture; explicit `VIBE_TRADING_HOME` for the daily-count subprocess env.

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q`)

Local validation: the two named guard tests and the new regression file fail on the base commit (no sandbox active: `VIBE_TRADING_HOME` unset) and pass after. Full-suite A/B against base `9806936f`: 9367 passed / 180 failed / 42 collection errors (submitter) vs 9353 / 191 / 42 (base) — 0 new failures, 11 previously-failing tests now pass; the remaining failures are pre-existing optional-dependency gaps. The real `~/.vibe-trading` tree is byte-identical before and after a full run.

## Checklist

- [ ] Code follows CONTRIBUTING.md guidelines

### DIFF
--- a/agent/src/core/runner.py
+++ b/agent/src/core/runner.py
@@ -129,12 +129,27 @@ def _prepare_sandbox_home(real_home: Path | None) -> Path:
                 src = src_root / rel
                 if not src.exists():
                     continue
+                dst = dst_root / rel
                 try:
-                    (dst_root / rel).symlink_to(src, target_is_directory=src.is_dir())
+                    dst.symlink_to(src, target_is_directory=src.is_dir())
                 except OSError:
-                    # Best-effort: a loader that can't find its config just falls
-                    # back to a live fetch / disabled cache — never a hard break.
-                    pass
+                    # Symlinks need privileges on Windows, so they can fail even
+                    # when nothing is fundamentally wrong. Fall back to a copy:
+                    # the loader cache/config paths stay visible to the subprocess
+                    # (the whole point of the re-exposure) without privileges.
+                    # Copies are per-run snapshots that the existing sandbox
+                    # cleanup removes, so the opt-in cache stops persisting across
+                    # runs only on hosts that cannot create symlinks.
+                    try:
+                        if src.is_dir():
+                            shutil.copytree(src, dst, dirs_exist_ok=True)
+                        else:
+                            shutil.copy2(src, dst)
+                    except OSError:
+                        # Best-effort: a loader that can't find its config just
+                        # falls back to a live fetch / disabled cache — never a
+                        # hard break.
+                        pass
             try:
                 os.chmod(dst_root, 0o755)
             except OSError:
--- a/agent/tests/conftest.py
+++ b/agent/tests/conftest.py
@@ -2,7 +2,11 @@
 
 from __future__ import annotations
 
+import atexit
+import os
+import shutil
 import sys
+import tempfile
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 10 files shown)


warren618 (changes requested):
Thanks for the thorough write-up and for fixing the root cause rather than the symptom. Installing VIBE_TRADING_HOME/HOME/USERPROFILE at conftest import time is the right mechanism: it runs before any src import (no rootdir conftest, no pytest11 plugins import src), and both get_runtime_root() and direct Path.home() sites — including the unpatched src.live.paths binding that wrote the real ledger in #1116 — resolve into the sandbox. The runner.py copy fallback correctly restores the test_prepare_sandbox_home guard that failed on Windows, and its exception handling is complete (shutil.Error is an OSError).

One blocking consistency gap before merge: the _client fixtures in agent/tests/test_etoro_runtime.py:16 and agent/tests/test_longbridge_runtime.py:34 still patch only Path.home, but the runtime root now resolves env-first through get_runtime_root() (see src/trading/connectors/longbridge/credentials.py:55), so their tmp_path redirection is silently inert and those files fall into the session-shared sandbox — exactly what the new TEST DISCIPLINE comment forbids. Please add monkeypatch.delenv("VIBE_TRADING_HOME", raising=False) to both fixtures (the other twelve sites already got it), or patch the get_runtime_root binding the way test_mt5_connector.py:206 does.

Two nits: (1) on Windows the copy fallback copies the entire opt-in loader cache per run, outside the subprocess timeout, and the subprocess writes are discarded — consider a size guard or a directory junction, and document the cost; (2) the PR's CI runs are still action_required, so please attach a full-suite Windows run (or approve the pending workflow runs) before merge.

With the fixture fix this is ready to merge.

warren618:
Thanks — the diagnosis is right, and the mechanism you named (`ntpath.expanduser` reads `%USERPROFILE%` and never `$HOME`, so the suite's `HOME`-based idiom is inert on Windows) checks out against the source. Two things I found verifying it, so I pushed a follow-up to this branch rather than sending it back.

**It is not Windows-only.** On macOS with a default environment, `pytest agent/tests/test_mt5_connector.py` appends one `order_rejected` record to the real `~/.vibe-trading/live/audit.jsonl`, and `test_alpha_bench_meta_survivorship.py` writes an `alpha_bench_*.html` into the real `~/.vibe-trading/reports/`. `conftest.py` had no config-root sandbox at all — its only autouse fixture snapshots `os.environ` — so any module that baked `Path.home() / ".vibe-trading"` at import time resolved against the real home on every platform. Windows is worse, not different in kind.

**`VIBE_TRADING_HOME` was the wrong lever.** `get_runtime_root()` consults it *before* falling back to `Path.home()`, so pointing it at the sandbox also outranked every per-test `monkeypatch.setenv("HOME")` / `monkeypatch.setattr(Path, "home", ...)` and collapsed per-test roots into one shared directory. That is what the 13 `monkeypatch.delenv` additions were compensating for, and one case slipped through: `test_shadow_account.py::test_find_by_journal_hash_returns_latest` failed 2 of 3 runs here, reading a profile a previous test in the same file had saved. It also preserves an exported `VIBE_TRADING_HOME` — the leak the sandbox exists to close. The follow-up **deletes** the override instead of setting it, leaving `Path.home()` as the single point of control; per-test redirection then works on every platform with no sandbox awareness, so `test_agent_config.py`, `test_api_live_runtime.py` and `test_channel_media_roots.py` are untouched again.

Also fixed: **resolve the sandbox root** (`tempfile.mkdtemp()` returns `/var/folders/...` on macOS, a symlink to `/private/var/folders/...`; the unresolved spelling makes guards that compare a resolved path against a `Path.home()`-derived prefix read it as an escape — `test_mcp_alpha_zoo_contract.py::test_alpha_bench_forwards_selection` failed 3/3 on macOS). **Keep the per-user site directory importable for children** — the suite hands its `HOME` to every subprocess and `pip install --user` puts numpy/pandas under a directory derived from `HOME`; tool auto-discovery logs the import failure and continues, so the child came up with 57 of 94 tools and surfaced as `Tool 'options_pricing' not found` in `test_mcp_server_smoke.py`. **An outcome guard** — the existing assertions check the redirect is *installed*, which is a proxy; `pytest_sessionfinish` now snapshots the real live ledgers before the redirect and fails the session if they moved.

Verified on macOS: full suite `7 failed, 10525 passed` against a clean `main` baseline of `7 failed, 10534 passed` — the same seven pre-existing `test_swarm_run_metadata` failures, no new ones. `audit.jsonl` and `audit_chain.jsonl` are byte-identical (sha256) before and after a full run, `reports/` and `runs/` unchanged, both hash chains still verify; the same command on `main` appends a record.

PR #1119, opened 2026-08-16
fix(quantlib): survive long-horizon discount underflow in xirr and money-weighted return

pengpengyi92:
# Draft PR Body — Vibe-Trading `xirr` discount-underflow fix

Branch: `fix/quantlib-xirr-underflow`
Title: `fix(quantlib): survive long-horizon discount underflow in xirr and money-weighted return`

---

## Summary

- `xirr()` and `money_weighted_return()` crash with `ZeroDivisionError: float division by zero` on cash-flow streams spanning more than ~51 years.
- The NPV kernel now treats a discount factor that *underflows* to zero as a diverging term and keeps the bisection bracketing sign-exact.
- No behavior change for any existing input; three focused regressions added.

## Why

The IRR bisection always evaluates the NPV at its lower bracket (`rate = -0.999999`, `base = 1e-6`). For horizons beyond `308 / log10(1e6) ≈ 51.34` years, `base**exponent` leaves the normal float range: it either underflows to `0.0` and the term `amount / 0.0` raises, or it stays subnormal so the term overflows to `inf` and two opposite-signed flows make `math.fsum` raise `ValueError: -inf + inf in fsum`. XIRR exists precisely for long, irregular streams (long-dated instruments, historical series, pension-style cash flows), so the crash is a real input class, not a theoretical edge:

```text
xirr([(date(1965,1,1), -1000.0), (date(2025,1,1), 3000.0)])
# before: ZeroDivisionError: float division by zero
# after:  0.0184661...  == 3**(365/21915) - 1
```

## Changes

- `agent/src/quantlib/performance.py`: in `_npv`, two classes of diverging term are now grouped by exponent (largest first) and represented by a sign-preserving finite stand-in (`_NPV_DIVERGED_MAGNITUDE = 1e300`): a discount factor that underflows to zero, and a subnormal factor whose term leaves the float range (the latter previously handed `fsum` infinities and raised `ValueError: -inf + inf in fsum` for opposite signs). The solver compares signs only, so bracketing stays exact, `math.fsum` cannot overflow, and the behavior is platform-independent. The existing `OverflowError -> 0.0` policy and every finite-term computation are untouched.
- `agent/tests/quantlib/test_performance.py`: five red-before regressions — 60-year gain, 60-year loss, `money_weighted_return` path, 52-year subnormal window, and a three-flow stream with opposite-signed long-horizon terms — asserted against closed-form rates.

## Test Plan

- [x] Existing tests pass: `py -3 -m pytest agent/tests/quantlib -q` → `1007 passed, 61 skipped` (baseline before change: `1002 passed, 61 skipped`).
- [x] New tests added (5), including one that fails on the pre-fix kernel with `ValueError: -inf + inf in fsum`.
- [x] Tested manually: 60-year and 52-year `xirr` probes match `3**(365/days)-1` to 1e-9 relative; opposite-sign stream zeroes its NPV residual to < 1e-6.
- Not run (unchanged code): `agent/tests/e2e_backtest`, frontend build.

## Checklist

- [x] No hardcoded values (the stand-in magnitude is a documented module constant).
- [x] Code follows CONTRIBUTING.md guidelines (ruff clean, diff whitespace clean, no unrelated reformatting).
- [x] Documentation updated: module docstrings for `_npv` and the new constant; no user-facing surface change.

## Risk and rollback

- Risk: none. Pure quantlib math, no broker/MCP/network/live-order surface.
- Out of scope (deliberately): `fundmath.py` keeps its own log-space NPV kernel and its documented non-finite contract; this PR does not touch it.
- Rollback: `git revert <sha>` restores the previous behavior; no data migration.

### DIFF
--- a/agent/src/quantlib/performance.py
+++ b/agent/src/quantlib/performance.py
@@ -147,6 +147,13 @@ _IRR_TOLERANCE = 1e-12
 #: Hard iteration cap for the bisection, so a pathological input cannot spin.
 _IRR_MAX_ITERATIONS = 400
 
+#: Magnitude returned in place of a diverging NPV term. At ``base`` just above
+#: zero and a horizon beyond ~51 years the discount factor underflows to 0.0 and
+#: the true term diverges to +/-inf. The bisection compares signs only, so a
+#: sign-preserving finite stand-in keeps the bracketing exact without letting
+#: :func:`math.fsum` overflow.
+_NPV_DIVERGED_MAGNITUDE = 1e300
+
 
 class UnsolvableRateError(ValueError):
     """Raised when a cash-flow stream admits no single internal rate of return."""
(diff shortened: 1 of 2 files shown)


warren618:
Verified and merging. Reproduced on `main` through the agent tool surface: `quantlib_call` → `performance.xirr` over 1965-01-01 → 2025-01-01 raises `ZeroDivisionError`, and a three-flow stream with opposite-signed long-horizon terms does the same. Both return correct rates after — the 60-year case gives `0.018466100…`, matching the closed form `3**(365/21915) - 1` to the digit.

Worth stating explicitly: `_solve_irr` only ever uses `_npv`'s sign, never its magnitude, so the finite stand-in is safe and the product overflowing to ±inf still compares correctly. Grouping by exponent before deciding the sign is right for `base < 1`, where the largest exponent dominates. Two control inputs (25-year, 52-year) return bit-identical values before and after, which backs the no-behaviour-change claim.

PR #1121, opened 2026-08-17
fix(quantlib): refuse non-finite DCF inputs instead of a silent negative share price

Robin1987China:
## Summary

The DCF valuation engine now refuses non-finite inputs at every public builder instead of silently returning a valuation built on them.

## Why

`run_dcf(beta=inf)` returned `value_per_share = -1.7` with no error — a silently wrong per-share number an agent would trade on. The package's own rule is that an unusable input makes the model NOT RUNNABLE, but the WACC build, FCFF bridge, and terminal-value computation read their inputs with `float(...)` and never check finiteness, and the `__post_init__` reconciliation net (`math.isclose`) returns `True` for `inf == inf`.

## Changes

- `agent/src/quantlib/valuation/dcf.py`: add `_require_finite` (mirroring `_require_nonnegative`) and apply it at the entry of `wacc`, `fcff_bridge`, `terminal_value`, `discount_fcff`, and `equity_bridge` (`enterprise_value`).
- `agent/tests/quantlib/valuation/test_dcf.py`: seven regression tests covering `inf`/`nan` refusal per builder, the end-to-end `run_dcf` path, and negative-but-finite acceptance.

## Test Plan

- [x] ruff clean on both files
- [x] `pytest tests/quantlib -q` → 1009 passed, 61 skipped (7 new tests)
- [x] finite-input worked example unchanged (`value_per_share = 10.42805...`)

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (docstring Raises updated)

### DIFF
--- a/agent/src/quantlib/valuation/dcf.py
+++ b/agent/src/quantlib/valuation/dcf.py
@@ -251,6 +251,34 @@ def _require_nonnegative(value: float, name: str, model: str) -> float:
     return numeric
 
 
+def _require_finite(value: float, name: str, model: str) -> float:
+    """Check a value is a finite number, refusing NaN and infinity.
+
+    Args:
+        value: The candidate value.
+        name: Field name for the error message.
+        model: Model name for the error message.
+
+    Returns:
+        ``value`` as a float.
+
+    Raises:
+        ValuationError: If the value is not a finite number. A non-finite
+            input here would otherwise flow through the valuation arithmetic
+            and surface as a silently wrong per-share number, which is exactly
+            the outcome the package's no-silent-defaults rule exists to prevent.
+    """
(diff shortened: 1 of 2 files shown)


warren618:
Merged, and preferred over a competing patch for the same issue. Two reasons worth naming: you guarded `discount_fcff`, `equity_bridge` and `sensitivity_grid` as well as the three obvious builders — a non-finite `wacc_rate` reaching `discount_fcff` directly would still have produced a real-looking share price — and you updated the `Raises:` sections. In this package the docstring is part of the fix: the whole contract is that an unusable input makes a model NOT RUNNABLE, so a guard nobody can find from the signature is only half of it.

PR #1125, opened 2026-08-17
[Bug] In the final verification response, the method "numbers_without_dates_or_percent" had a problem in extracting the prices, which resulted in the correct report also being rejected.

### LINKED ISSUE #1122: [Bug] In the final verification response, the method "numbers_without_dates_or_percent" had a problem in extracting the prices, which resulted in the correct report also being rejected.
### Description

Method "_numbers_without_dates_or_percent" is used to extract the problematic prices from the report. For example, in the sentence "4. 价格跌破SMA20", the extracted result is [4]. For another example, "4. **技术面风险**: 若跌破观测支撑位0.980 CNY (2026-07-14最低)", the extracted result is [4.0, 2026.0, 7.0, 14.0]. The extracted prices will be verified against the previous evidence. If the verification fails, the issue will be recorded and ultimately lead to the large model refusing to answer.

### Steps to Reproduce

Analyze the stock price trend directly in the dialog box of the web UI.

Image

here is the trace
[trace.json](https://github.com/user-attachments/files/31132533/trace.json)

### Error Logs

```shell

```

### Interface

Web UI (vibe-trading serve)

### LLM Provider

_No response_

### Version

v0.1.13

### Environment

python3.13

nslatour:
### Description

Method "_numbers_without_dates_or_percent" is used to extract the problematic prices from the report. For example, in the sentence "4. 价格跌破SMA20", the extracted result is [4]. For another example, "4. **技术面风险**: 若跌破观测支撑位0.980 CNY (2026-07-14最低)", the extracted result is [4.0, 2026.0, 7.0, 14.0]. The extracted prices will be verified against the previous evidence. If the verification fails, the issue will be recorded and ultimately lead to the large model refusing to answer.

###

## Changes

Modified 2 file(s): agent/tests/test_agent_grounding.py, agent/src/agent/grounding.py

Closes #1122

## Checklist

- [x] Code follows the project's style guidelines
- [x] Changes have been tested locally
- [x] Documentation updated if needed

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -201,7 +201,17 @@ _RATE_FORMULA_IDENTITY_RE = re.compile(
     r"\b[01](?=\s*[-+]\s*(?:[A-Za-z_][A-Za-z0-9_]*_?rate\b|[^\d\s()+*/=-]{0,12}(?:成本率|费率|税率|滑点率)))",
     re.IGNORECASE,
 )
-_DATE_RE = re.compile(r"\b(?:19|20)\d{2}[-/]\d{1,2}[-/]\d{1,2}\b")
+# "\b" is a word boundary between "\w" and non-"\w", but Python's regex module
+# treats CJK ideographs as "\w" too, so a trailing "\b" never matches where a
+# date directly abuts Chinese prose with no separating space or punctuation:
+# "(2026-07-14最低)" left the date unmasked, and its 2026/7/14 were extracted
+# as candidate prices and rejected against the observed OHLC range. Anchoring
+# on the same ASCII-only lookaround as ``_NUMBER_RE``/``_SHORT_DATE_RE`` keeps
+# the date from merging with an adjacent digit or letter run while still
+# matching when CJK text touches either side.
+_DATE_RE = re.compile(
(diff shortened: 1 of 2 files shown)


warren618:
Thanks for chasing this one down — the diagnosis is right, and it matches what I reproduced locally: `\b` is Unicode-aware, so `最` counts as a word character, `2026-07-14最低` has no boundary after `14`, and the whole date survives the mask.

I merged #1132 for it. It reaches the same behaviour by a different anchor — `re.ASCII` on the existing pattern rather than explicit lookarounds — and I checked the two against each other before deciding: identical results on every case I tried, including `x2026-07-14`, `2026-07-140` and `版本2026-07-14版`. What decided it was the test coverage. #1132 pins both directions: the glued date extracts nothing, **and** a real quote standing next to a date still extracts (`'on 2026-07-14 close was 8.20 CNY' -> [8.2]`). A one-sided mask test passes just as happily when the mask swallows too much, which is the failure mode this file has had before.

Two small things for next time, offered as useful rather than as the reason for closing: the issue reference in the new test's docstring reads `#1786964822`, which is a timestamp rather than an issue number (it should be #1122); and CI never ran on this branch, so neither of us could see the suite pass.

Closing as superseded by #1132. The bug you reported is fixed on main. Thank you.

PR #1128, opened 2026-08-17
fix(alpha-bench): wire sector tags into SP500 panel

cgycorey:
## Problem

Nineteen alpha101 alphas are gated behind `requires_sector=True` (industry-neutralized formulas). The SP500 bench panel never attached a `sector` DataFrame, so the registry pre-rejected all 19 — the user sees `n_skipped=19` with no explanation.

The underlying compute functions already have a degraded global-demean fallback (the `_ind_neutralize` helper), but the registry gate blocks them before compute is reached.

## Fix

`_load_sp500_panel` now attaches a same-shape `sector` DataFrame of per-code sector labels after the vwap synthesis. The labels come from `yfinance.Ticker.info['sector']` — free, no API key, and the panel loader already has per-code metadata access.

- Strips `.US` suffix before calling `yfinance.Ticker` (yfinance expects bare symbols)
- Best-effort: per-code fetch failures are silently skipped; a total failure degrades to the pre-existing panel with a warning
- Same shape as `close` (date index × code columns) — exactly what `_ind_neutralize` expects

## Verification

- 4 regression tests pass (same-shape DataFrame, graceful failure when yfinance is down, `_meta` blob survives, no sector key when yfinance returns None)
- E2E on 8-name universe: 3 of 4 sector-gated alphas now compute correctly (alpha101_048 hits >95% NaN — the 4th NaN alpha alongside 019/039/052 on this universe)
- Ruff clean

## Notes

- Combined with the earlier fundamental-zoo PR (#1127), the SP500 bench panel now provides: OHLCV + vwap + fund:* (5 columns) + sector. The only remaining unwired columns are `amount` (gtja191, CN-only) and `advN` (computed internally from volume).
- Stale panel pickles in `~/.vibe-trading/cache/sp500_*.pkl` must be deleted to force re-fetch with the sector DataFrame.
- The alpha101_019/039/048/052 NaN alphas are a separate issue (formula/data-coverage, not panel wiring).

### DIFF
--- a/agent/src/tools/alpha_bench_tool.py
+++ b/agent/src/tools/alpha_bench_tool.py
@@ -501,6 +501,30 @@ def _load_sp500_panel(start: str, end: str) -> dict[str, pd.DataFrame]:
     if all(k in panel for k in ("open", "high", "low", "close")):
         panel["vwap"] = (panel["open"] + panel["high"] + panel["low"] + panel["close"]) / 4.0
 
+    # Per-code sector tags for alpha101 industry-neutralized alphas.
+    # The registry gates 19 alpha101 alphas behind requires_sector=True;
+    # their compute() calls _ind_neutralize() which expects a same-shape
+    # DataFrame of sector labels.  yfinance Ticker.info is free.
+    try:
+        import yfinance as yf
+
+        close_cols = panel["close"].columns
+        sector_map: dict[str, str] = {}
+        for code in close_cols:
+            try:
+                info = yf.Ticker(code.removesuffix(".US")).info
+                sector = info.get("sector")
+                if sector:
+                    sector_map[code] = sector
+            except Exception:  # noqa: BLE001 — per-code best-effort
+                pass
+        if sector_map:
+            panel["sector"] = pd.DataFrame(
+                {c: sector_map.get(c, "") for c in close_cols},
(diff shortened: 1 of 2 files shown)


warren618:
Thank you — the diagnosis is exactly right: nineteen alpha101 alphas declare `requires_sector`, `registry.py:338` refuses them when the panel has no `sector` frame, and the SP500 panel never attached one, so the bench reported `n_skipped=19` on every run.

I landed a fix on main (`99861eb2`) that takes a different route to the same place. The constituent list already comes from a Wikipedia table, and that table carries a `GICS Sector` column beside `Symbol` — 503 rows, 11 sectors, no nulls when I checked it. Reading the column we were already throwing away costs no extra request and no per-name lookup, where `yfinance.Ticker(...).info` in a loop is up to 500 sequential calls on a bench path, with the rate-limit exposure that comes with it.

Two details from your PR that I kept: the `.US` suffix has to come off before the symbol is used for lookup, and a partial result must degrade rather than pretend. On the second point I ended up stricter — below 90% sector coverage the tag stays off entirely, because one large "unknown" bucket demeans as a single group, which is the global-demean fallback the alphas already have, reported as industry neutralization.

Closing as landed by another route. The bug you found was real and nobody else had spotted it — thank you.

PR #1129, opened 2026-08-17
fix(tools): report partial registry construction

### LINKED ISSUE #1124: build_registry() silently returns a partial tool registry when a tool module fails to import
## Description

`build_registry()` swallows a tool module's import failure, logs one line, and returns a **partial** registry. There is no signal at the call site that the tool surface is incomplete, so the failure surfaces much later as an apparently unrelated "tool not found".

Found while verifying #1118. With the dependency that 37 tool modules share made unimportable, the registry silently dropped from **94 tools to 57** and the only evidence was a single truncated stderr line:

```
margin_trading_tool: Unable to import required dependencies:
```

The visible symptom was several layers away — the MCP server completed its handshake, `tools/list` returned a catalogue that passed the smoke test's `>= EXPECTED_MIN_TOOL_COUNT` check, and then:

```json
{"status": "error", "error": "Tool 'options_pricing' not found"}
```

Nothing in that message suggests a dependency problem, and `tools/list` succeeding makes it look like a routing bug rather than a registry that came up half-built. It took a subprocess probe printing `len(tool_names)` to find the real cause.

This is a real user-facing failure mode, not only a test concern: a user with one broken optional dependency gets a quietly reduced tool surface and an error that points at the wrong thing.

## Suggested direction

Nothing here should hard-fail — a broken optional dependency should not stop the agent from starting. But the degradation should be *legible*:

- Record which modules failed and why on the registry object, and surface the count in the startup log at WARNING (`registered 57 tools; 37 modules failed to import`), not one line per module buried in a subprocess's stderr.
- Make "tool not found" say so when the name belongs to a module that failed to import, instead of the generic message.
- Do not truncate the underlying `ImportError` — the missing package names are the actionable part and they were cut off here.

Worth a guard test: a registry built with one module forced to fail must report the failure, and the count assertion should compare against the code rather than a fixed number (see #1116-adjacent lesson about counting tests measuring proxies).

er-s-an:
## Summary

- retain module import and tool registration failures on `ToolRegistry`
- emit one aggregate warning when local registry construction is partial
- make missing-tool responses explain matching startup failures while preserving partial startup

## Root cause

Tool discovery logged and discarded import failures. Callers received an apparently healthy partial registry, and later requests only saw a generic `Tool not found` error.

## Validation

- `PYTHONPATH=agent python3 -m pytest agent/tests/test_tool_registry_diagnostics.py agent/tests/test_tool_registry_security.py -q` — 4 passed
- `ruff check --ignore E402 agent/src/agent/tools.py agent/src/tools/__init__.py agent/tests/test_tool_registry_diagnostics.py`
- `python3 -m compileall -q agent/src/agent/tools.py agent/src/tools/__init__.py agent/tests/test_tool_registry_diagnostics.py`

The broader MCP regression module was not fully runnable in the local system Python because `fastmcp` and the optional web-search dependency are not installed.

Fixes #1124

### DIFF
--- a/agent/src/agent/tools.py
+++ b/agent/src/agent/tools.py
@@ -56,6 +56,8 @@ class ToolRegistry:
 
     def __init__(self) -> None:
         self._tools: Dict[str, BaseTool] = {}
+        self._import_failures: Dict[str, str] = {}
+        self._registration_failures: Dict[str, str] = {}
 
     def register(self, tool: BaseTool) -> None:
         """Register a tool."""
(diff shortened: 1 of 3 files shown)


warren618:
Merged — thank you. The shape was right where it counts: keeping the `_discover_subclasses()` signature meant three call sites across `test_investor_lenses_skill.py` and `test_institutional_commands.py` stayed untouched, and the exact-match lookup (`name in {module_name, module_name.removesuffix("_tool")}`) will not blame a failed module for an unrelated typo.

I pushed one commit to your branch before merging, so you are not surprised by it: the per-module discovery failure is back at WARNING rather than DEBUG. The aggregate line reports *how many* sources failed; only the per-module record says *which*, and #1124 is precisely about an operator not being able to see that. A guard test now pins the level, and I checked it bites — restoring DEBUG fails it.

Issue #1123, opened 2026-08-17
test_factor_api patches os.scandir process-wide, so a teardown-time directory scan cascades


warren618:
## Description

`agent/tests/test_factor_api.py::test_factor_scan_caps_directory_entries` replaces `os.scandir` **process-wide** rather than on the module under test:

```python
monkeypatch.setattr(runs_routes.os, "scandir", GuardedScandir)
```

`runs_routes.os` *is* the global `os` module (`runs_routes` does a plain `import os`), so this is `setattr(os, "scandir", ...)`. The replacement counts entries across a shared closure and asserts a cap of 7:

```python
def __next__(self):
    nonlocal entries_seen
    assert self._iterator is not None
    entry = next(self._iterator)
    entries_seen += 1
    assert entries_seen <= entry_limit
    return entry
```

`monkeypatch` does undo it at teardown, so `main` is unaffected in a normal run — this is latent fragility, not a live defect. But the blast radius is the whole interpreter for the duration of the test, and it is only contained by teardown ordering. Any `pytest_runtest_teardown` hook that scans a directory runs *before* the fixture finalizers, so it receives `GuardedScandir`, raises on the 8th entry, aborts the teardown chain, and the undo never happens. `os.scandir` then stays replaced for the rest of the session.

I hit exactly this while instrumenting the suite for #1116: a teardown hook calling `Path.iterdir()` produced ~3000 cascading errors from ~40% of the run onward, and pytest's own `cleanup_dead_symlinks` and the `cleanup_numbered_dir` atexit handler both died inside the test's closure with `AssertionError: assert 6342 <= 7` — the counter had kept incrementing across the entire session.

## Suggested direction

Patch the name the code under test resolves rather than the shared module — bind `scandir` in `runs_routes` (`from os import scandir`, patch `runs_routes.scandir`), or patch the traversal helper directly. Either keeps the guard's reach to the module it is testing.

Low priority: no user-facing impact, and the current test passes.

PR #1130, opened 2026-08-17
fix: refuse non-finite DCF inputs + track registry import failures

### LINKED ISSUE #1120: run_dcf silently returns a negative share price for non-finite inputs
`run_dcf` in `src/quantlib/valuation/dcf.py` accepts non-finite inputs and returns a valuation that looks real instead of refusing.

Repro (the worked example from `test_dcf.py`, with `beta=inf`):

```python
run_dcf(inputs, capital_structure_basis="target",
        discounting_convention="mid_year", terminal_value_method="perpetuity_growth")
# -> value_per_share = -1.7, enterprise_value = 0.0   (no error raised)
```

Same gap in the lower-level builders:

- `wacc(beta=inf)` returns `wacc = inf`
- `terminal_value(terminal_year_ebitda=nan)` returns `nan` fields
- `fcff_bridge(ebit=[inf, ...])` emits a non-finite `fcff`

The engine's own rule (`contracts.py`) is that an unusable input makes the model NOT RUNNABLE. Non-finite values slip through because `wacc()`, `fcff_bridge()` and `terminal_value()` read their numeric inputs with `float(...)` and never check `math.isfinite`, and the `__post_init__` reconciliation net uses `math.isclose`, which returns `True` for `inf == inf` — so `inf` sails straight through the arithmetic re-check.

Note the balance-sheet side is already guarded: `equity_bridge()` rejects non-finite magnitudes via `_require_nonnegative`, and `diluted_shares` goes through `require_positive`. The WACC / FCFF / terminal-value inputs are the uncovered gap.

A negative per-share number is worse than a crash here — it looks exactly like a real valuation and an agent would trade on it.

### LINKED ISSUE #1124: build_registry() silently returns a partial tool registry when a tool module fails to import

cgycorey:
Closes #1120, closes #1124

## 1. DCF non-finite input guards (#1120)

NaN/Inf inputs slipped through `float()` casts in `wacc()`, `fcff_bridge()`, and `terminal_value()`. The `__post_init__` reconciliation used `math.isclose` where `inf == inf` is True, so `run_dcf` returned a plausible-looking negative share price instead of refusing.

**Fix:**
- Add `_require_finite()` helper (for rates/ratios — allows negative values, only rejects NaN/Inf)
- Add `_guard_sequence_finite()` helper (for forecast sequences)
- Wire guards into all three functions + market values

**Verification:** 98 DCF tests pass (66 existing + 32 new). The exact repro from #1120 (`beta=inf`) now raises `ValuationError` instead of returning `value_per_share=-1.7`.

## 2. Registry import failure tracking (#1124)

`build_registry()` silently swallowed import failures — a broken optional dependency produced a quietly reduced tool surface and a generic 'tool not found' error that pointed at the wrong thing.

**Fix:**
- `ToolRegistry` gains `failed_imports` dict (module_name → error)
- `_discover_subclasses` returns `(classes, failures)` tuple
- `build_registry` logs a single WARNING with count of failed modules
- 'tool not found' errors now hint when the module failed to import
- Instantiation failures also recorded

**Verification:** 4 guard tests + 1576 existing tool tests pass. Ruff clean.

### DIFF
--- a/agent/src/agent/tools.py
+++ b/agent/src/agent/tools.py
@@ -5,9 +5,9 @@ from __future__ import annotations
 import json
 import logging
 from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional
 
 logger = logging.getLogger(__name__)
-from typing import Any, Dict, List, Optional
 
 
 class BaseTool(ABC):
@@ -56,6 +56,8 @@ class ToolRegistry:
 
     def __init__(self) -> None:
         self._tools: Dict[str, BaseTool] = {}
+        # module_name -> error message for modules that failed to import
+        self.failed_imports: Dict[str, str] = {}
 
     def register(self, tool: BaseTool) -> None:
         """Register a tool."""
@@ -73,7 +75,14 @@ class ToolRegistry:
         """Execute a tool and guarantee a valid JSON return value."""
         tool = self._tools.get(name)
         if not tool:
-            return json.dumps({"status": "error", "error": f"Tool '{name}' not found"}, ensure_ascii=False)
+            hint = ""
+            if self.failed_imports:
+                # Check if the tool name matches a known module pattern
+                for module_name, error in self.failed_imports.items():
+                    if name in module_name or module_name.replace("_tool", "") in name:
+                        hint = f" (module '{module_name}' failed to import: {error.split(chr(10))[0]})"
+                        break
+            return json.dumps({"status": "error", "error": f"Tool '{name}' not found{hint}"}, ensure_ascii=False)
         try:
             return tool.execute(**params)
         except Exception as exc:
--- a/agent/src/quantlib/valuation/dcf.py
+++ b/agent/src/quantlib/valuation/dcf.py
@@ -250,6 +250,56 @@ def _require_nonnegative(value: float, name: str, model: str) -> float:
         )
     return numeric
 
+def _require_finite(value: float, name: str, model: str) -> float:
+    """Check that a rate or ratio input is a finite number.
+
+    Unlike ``_require_nonnegative``, this allows negative values — rates
+    and ratios (beta, growth rates, etc.) can legitimately be negative.
+    It only rejects NaN, Inf, and non-numeric inputs.
+
+    Args:
+        value: The candidate number.
+        name: Field name for the error message.
+        model: Model name for the error message.
+
+    Returns:
+        ``value`` as a float.
+
+    Raises:
+        ValuationError: If the value is not a finite number.
+    """
+    try:
+        numeric = float(value)
+    except (TypeError, ValueError) as exc:
+        raise ValuationError(f"{model}: {name} must be a number, got {value!r}") from exc
+    if not math.isfinite(numeric):
+        raise ValuationError(
+            f"{model}: {name} must be a finite number, got {numeric!r}"
+        )
+    return numeric
+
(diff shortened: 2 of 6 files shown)


warren618:
Thank you for taking both of these on — the diagnosis in each half is right, and both halves are now fixed on main. I went with the two focused PRs instead, and the reasons are about shape rather than correctness, so they are worth writing down.

**#1120, non-finite DCF inputs — merged #1121.** It guards the three entry points you did and also `discount_fcff`, `equity_bridge` and `sensitivity_grid`. A non-finite `wacc_rate` handed straight to `discount_fcff`, or a non-finite `enterprise_value` reaching the net-debt bridge, would still have produced a real-looking per-share number. It also updates the `Raises:` sections, which this package leans on: its whole contract is that an unusable input makes a model NOT RUNNABLE, so the docstring is part of the fix rather than a note about it.

**#1124, partial registry — merged #1129.** Two differences decided it. It keeps the `_discover_subclasses()` signature, which three call sites across `test_investor_lenses_skill.py` and `test_institutional_commands.py` depend on — changing it to a tuple is what forced this PR to edit two unrelated test files. And its lookup is exact (`name in {module_name, module_name.removesuffix("_tool")}`) rather than substring: `if name in module_name or module_name.replace("_tool", "") in name` will blame a failed `sec_filings_tool` for any requested name that happens to sit inside it, and a confidently wrong explanation is worse than the generic one it replaces.

The other reason is plain: one PR per issue. Bundling #1120 and #1124 means neither can be merged or reverted without the other, and the DCF half was ready well before the registry half.

Both issues are now closed by the merges above. If you would like another one, #1123 is open and unclaimed.

PR #1131, opened 2026-08-18
fix(mcp): attach market-data provenance to get_market_data responses

Shizoqua:
## Summary

- `get_market_data()` now passes `include_provenance=True` to `fetch_market_data_json()`, so `_provenance`, including `volume_unit`, is present in the MCP response.
- This aligns the MCP path with `MarketDataTool.execute()` and the existing MCP docstring, which already instructs clients to inspect `_provenance.volume_unit`.

## Why

`get_market_data()` documents `_provenance.volume_unit` as the source of volume-unit information, but its call to `fetch_market_data_json()` omitted `include_provenance=True`.

As a result, the `_provenance` block was absent from MCP responses even though `MarketDataTool.execute()` already enables it for the same market-data path.

Without `_provenance.volume_unit`, MCP consumers cannot reliably distinguish volume units across sources and markets.

Out of scope: market-data normalization, loaders, source selection, and provenance generation itself.

## Changes

- Pass `include_provenance=True` from `get_market_data()` to `fetch_market_data_json()`.
- Add regression coverage confirming `_provenance.volume_unit` is included.
- Add a control confirming the existing per-symbol row payload remains unchanged.

## Test Plan

- [x] Regression test fails on unpatched `main` and passes after the fix
- [x] Row-shape control passes before and after the fix
- [x] Adjacent MCP/market-data suites: 86 passed
- [x] Broader MCP/market-data selection: 353 passed, 5 skipped
- [x] Full suite: 10,455 passed, 90 skipped, 2 unrelated failures reproduced on unpatched `main`
- [x] `ruff check` passes on both changed files
- [x] No errors introduced by this change in `mypy`

## Checklist

- [x] No hardcoded values
- [x] Documentation: N/A, the existing docstring already describes the provenance contract
- [x] DCO signed-off

## Risk

Low. The response change is additive only. Existing market-data rows remain unchanged; the MCP response now also includes the provenance block already supported by the underlying market-data path. No broker, credential, order, payment, wallet, or live-trading behaviour is changed.

## Rollback

Single call-site option plus its regression test; revertible with a normal `git revert`.

### DIFF
--- a/agent/mcp_server.py
+++ b/agent/mcp_server.py
@@ -1591,6 +1591,7 @@ def get_market_data(
         interval=interval,
         max_rows=max_rows,
         loader_resolver=_get_loader,
+        include_provenance=True,
     )
 
 
--- a/agent/tests/test_get_market_data_provenance.py
+++ b/agent/tests/test_get_market_data_provenance.py
@@ -1,0 +1,73 @@
+"""Regression test — get_market_data (MCP) must attach the _provenance
+block its own docstring promises.
+
+Pre-fix: fetch_market_data_json() was called without include_provenance=True,
+so _provenance (including volume_unit) was silently omitted from every MCP
+response, while the internal MarketDataTool.execute() call site (used by the
+agent/swarm loop for the same request) always passed include_provenance=True
+and got the block. An MCP client reading the tool's own docstring, which
+tells it to check `_provenance.volume_unit` before comparing volume values,
+got a response with no such field. Post-fix: the MCP path passes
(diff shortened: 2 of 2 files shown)


warren618:
Merged. The asymmetry was exactly as you described it: `market_data_tool.py:105` passes `include_provenance=True`, `mcp_server.py:1586` did not, and the docstring twelve lines above told clients to read `_provenance.volume_unit` — a field the response never contained. A documented contract that the code does not honour is worse than an undocumented one, because it is the one people build against.

PR #1132, opened 2026-08-18
fix(grounding): mask ISO dates that run into CJK text

Robin1987China:
## Summary

The grounding gate's price extractor now masks ISO dates that run straight into CJK text, instead of reading their digits as three candidate prices.

## Why

`_DATE_RE` used `\b` as a word boundary, but `\b` is Unicode-aware so CJK letters count as word characters. A date glued to Chinese text — e.g. `(2026-07-14最低)` — has no boundary after `14`, so the date survived the mask and contributed `2026`, `7`, `14` as candidate prices that rejected a correct report (reported in #1122 and again in #1125).

## Changes

- `agent/src/agent/grounding.py`: compile `_DATE_RE` with `re.ASCII` so `\b` is a byte boundary (CJK is no longer a word character). ASCII-token boundaries such as `x2026-07-14` keep their old behavior.
- `agent/tests/test_agent_grounding.py`: two regressions — a unit test on `_numbers_without_dates_or_percent` and an end-to-end `validate_final_answer` test.

## Test Plan

- [x] ruff clean on both files
- [x] `pytest tests/test_agent_grounding.py -q` → 131 passed (2 new)
- [x] related grounding suites (`test_agent_output_discipline`, `test_grounding_recovery`, `test_swarm_grounding`) → 101 passed

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (comment explains the re.ASCII rationale)

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -201,7 +201,12 @@ _RATE_FORMULA_IDENTITY_RE = re.compile(
     r"\b[01](?=\s*[-+]\s*(?:[A-Za-z_][A-Za-z0-9_]*_?rate\b|[^\d\s()+*/=-]{0,12}(?:成本率|费率|税率|滑点率)))",
     re.IGNORECASE,
 )
-_DATE_RE = re.compile(r"\b(?:19|20)\d{2}[-/]\d{1,2}[-/]\d{1,2}\b")
+# re.ASCII keeps ``\b`` a *byte* word boundary. Without it, ``\w`` is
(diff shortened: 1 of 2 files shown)


warren618:
Merged. The `re.ASCII` anchor is the right call — I compared it against the explicit-lookaround alternative on every case I could think of, including `x2026-07-14`, `2026-07-140` and `版本2026-07-14版`, and they agree everywhere.

What decided it was your test carrying both directions: the glued date extracts nothing **and** a real quote next to a date still extracts. A one-sided mask test passes just as happily when the mask swallows too much, which is the failure mode this file has had before.

PR #1139, opened 2026-08-18
fix(autopilot): preserve backtest validation evidence when linking hypotheses

Shizoqua:
## Summary

- Preserve a run card's `validation` evidence when `link_autopilot_backtest` links a backtest to a hypothesis.
- Thread the existing validation block through `HypothesisRegistry.link_backtest()` alongside the performance metrics.
- Keep walk-forward, Monte Carlo, and bootstrap evidence available to the downstream research workflow instead of silently dropping it.

## Why

`write_run_card()` stores robustness results such as walk-forward, Monte Carlo, and bootstrap validation under the run card's top-level `validation` key.

`LinkAutopilotBacktestTool` previously read only `metrics`, and `HypothesisRegistry.link_backtest()` had no validation parameter.

As a result, a linked hypothesis could retain the backtest's headline performance metrics while losing the validation evidence needed to judge whether that performance is robust or potentially overfit.

This change carries the existing validation evidence through the link operation without changing how the validation itself is calculated.

## Changes

- Read `validation` alongside `metrics` in `LinkAutopilotBacktestTool.execute()`.
- Pass the validation block through `HypothesisRegistry.link_backtest()`.
- Persist the validation evidence on the linked run card.
- Add regression coverage confirming validation evidence survives the linking workflow.

## Test Plan

- [x] Regression test fails on unpatched `main` and passes after the fix
- [x] Adjacent tests: 25 passed
- [x] Broader hypothesis/autopilot selection: 46 passed, 5 skipped
- [x] No new `ruff` issues introduced by this change
- [x] No new `black` formatting issues introduced by this change
- [x] No new `mypy` errors introduced by this change

## Checklist

- [x] No hardcoded values
- [x] Documentation: N/A, no doc enumerates this tool's response schema
- [x] DCO signed-off

## Risk

Low. This preserves existing validation evidence when linking a backtest to a hypothesis. It does not change backtest calculations, validation algorithms, broker behaviour, credentials, orders, payments, wallets, or live-trading behaviour.

## Rollback

The change is limited to validation evidence propagation and its regression coverage and can be reverted with a normal `git revert`.


he-yufeng:
Nice catch. One sibling surface still drops the same key: the generic `link_backtest` tool in hypothesis_tool.py has no `validation` parameter, so an agent linking a card through that surface cannot carry validation through even when it read the card itself. The registry now persists the key when it is given, so the fix there is just the optional passthrough plus a test.

Happy to follow up with that small patch, or leave it to you if you want it.

PR #1142, opened 2026-08-18
fix(swarm): close the per-task ChatLLM to stop pooled connections leaking

### LINKED ISSUE #1141: Swarm worker leaks a pooled HTTP connection per task: ChatLLM is never closed
## Summary

`agent/src/swarm/worker.py`'s `run_worker` builds a fresh `ChatLLM` (`agent/src/providers/chat.py`) for every worker task:

```python
# 2. Create LLM
llm = ChatLLM(model_name=agent_spec.model_name)
```

`ChatLLM.__init__` calls `build_llm()` (`agent/src/providers/llm.py`), which constructs a brand-new LangChain chat model adapter (`ChatOpenAI` in the common OpenAI-compatible path, e.g. DeepSeek via `base_url=https://api.deepseek.com`) — each with its own pooled `httpx.Client` underneath. `ChatLLM` has no `close()`/`aclose()`/`__del__`/context-manager method, and `run_worker` never releases the client it built; `llm` just falls out of scope at each of the function's ~9 `return WorkerResult(...)` points.

Since a `ChatOpenAI`/`openai.OpenAI` instance isn't guaranteed to be promptly refcount-collected (there's enough internal cross-referencing that collection often waits on a cyclic-GC pass rather than happening immediately), the underlying pooled connection can outlive the task and accumulate as a `CLOSE-WAIT` socket. In a long-running swarm deployment (5 agents/run, runs firing every ~30 min) this adds up over hours until the process is restarted.

## Where I saw it

Running a self-hosted swarm preset continuously, `ss -tnp` showed growing `CLOSE-WAIT` entries owned by the `vibe-trading` process pointing at the configured LLM provider, that only cleared on a service restart — the same failure shape as an earlier, since-fixed leak in a downstream watcher process that held one long-lived unbounded `httpx.AsyncClient` for its whole run loop. Here it's the opposite shape: many short-lived clients (one per task) that are each individually never closed.

## Suggested fix

Give `ChatLLM` a `close()` that best-effort closes whatever the active provider adapter exposes (`root_client`/`client` cover `ChatOpenAI` and its OpenAI-compatible subclasses, which is the path most deployments without the optional native `langchain-deepseek`/`langchain-anthropic` extras installed will hit), and have `run_worker` call it in a `finally` block regardless of which return point fires.

I already applied this as a local patch to my own install (`vibe-trading-ai==0.1.12`) and confirmed both files still import/compile correctly:

- `ChatLLM.close()` in `providers/chat.py` — probes `root_client`/`client` for a callable `close`, swallows failures (never let a close error mask the task's real result).
- `run_worker` in `swarm/worker.py` — split into a thin wrapper (`run_worker`) that owns a `try/finally` around the renamed original body (`_run_worker_impl`), threading an `llm_holder: list` through so the impl's many early returns don't each need their own cleanup call; the wrapper closes everything in `llm_holder` in `finally`.

Happy to open a PR with this if useful — wanted to flag it first in case there's a preferred approach (e.g. a context-manager `ChatLLM.__enter__`/`__exit__` instead) or a reason it's been left as-is.

Echoandelementwebsites:
Closes #1141

## Goal

Stop `run_worker` from leaking a pooled HTTP connection every time it builds a `ChatLLM` for a swarm task. Observed as growing `CLOSE-WAIT` sockets on a long-running `vibe-trading` process (5 agents/run, runs firing every ~30 min) that only cleared on a full service restart.

## Affected areas

- `agent/src/providers/chat.py` — adds `ChatLLM.close()`.
- `agent/src/swarm/worker.py` — `run_worker` split into a thin wrapper (unchanged public name/signature) around the original body, renamed `_run_worker_impl`, threading an `llm_holder: list` through so the wrapper can close the built `ChatLLM` in a `finally` block regardless of which of the function's ~9 `return WorkerResult(...)` points fired.
- No changes to broker connectors, mandate/order-gate logic, MCP entry points, auth, or any network-facing server surface. No config, `.env`, or dependency changes.

## Out of scope

- Not touching provider adapters other than the `root_client`/`client` attribute probe (i.e. `ChatOpenAI` and its OpenAI-compatible subclasses — the path taken without the optional native `langchain-deepseek`/`langchain-anthropic` extras). If those adapters store their client under different attribute names, `close()` silently no-ops for them rather than raising; extending coverage is left for a follow-up if needed.
- Not fixing the pre-existing `ruff --select ALL` findings already present in the surrounding, unmodified code in these two files (line-length, complexity, blind-except patterns) — left untouched per `CONTRIBUTING.md`'s guidance against mixing unrelated formatting cleanup into a focused PR.
- Not adding a regression test for the leak itself (asserting a closed connection pool isn't easily observable at the unit level without a live provider or a fake with pool introspection); verification is the manual `ss -tnp` check below.

## Test plan

- [x] `python -m py_compile` on both changed files.
- [x] Both modules import cleanly (`from src.providers.chat import ChatLLM`, `from src.swarm.worker import run_worker, _run_worker_impl`).
- [x] `ruff check` on both files — no new findings introduced by the changed lines (pre-existing findings in untouched code left as-is; see Out of scope).
- [x] This exact patch has been running against a `vibe-trading-ai==0.1.12` production install (backported to that release's slightly different `ChatLLM.__init__`) since before this PR was opened.
- [ ] Not run: `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` — I don't have a full local checkout of this repo in this environment (this PR was built via the GitHub API against fetched file content, not a local clone), so I couldn't execute the suite here. Flagging for a maintainer/CI run before merge rather than claiming a result I didn't produce.

## Risk

- **Live/broker/order risk**: none. No order-gate, mandate, kill-switch, or broker-write code is touched.
- **MCP/network risk**: none. No MCP server, route, or network-facing behavior changes — this only affects when an already-open provider HTTP client is closed, not what it sends or where.
- **Secret risk**: none. No credentials, `.env`, or token-handling code touched.
- **Deployment risk**: low. Purely additive (`close()` is new; nothing previously called it) plus a call-site refactor that preserves `run_worker`'s existing name and signature, so no caller changes are needed. The only new runtime behavior is that a swarm task's LLM client is now closed after each task instead of leaking — no change to task inputs, outputs, or control flow.

## Rollback / close path

Single revertible change: `git revert` this merge commit restores the prior (leaking) behavior with no data migration, no config change, and no other code depending on the new `close()` method or the `_run_worker_impl` split. If a regression surfaces post-merge (e.g. a provider adapter whose `close()` call has a side effect worse than the leak it fixes), reverting is the preferred fix per `AGENT_CONTRIBUTOR_GUIDE.md` rather than a broader refactor.

### DIFF
--- a/agent/src/providers/chat.py
+++ b/agent/src/providers/chat.py
@@ -296,6 +296,31 @@ class ChatLLM:
             reasoning_effort=runtime_cfg.langchain_reasoning_effort.strip().lower(),
         )
 
+    def close(self) -> None:
+        """Best-effort close of the underlying provider HTTP client.
+
+        build_llm() constructs a fresh provider adapter (and its own pooled
+        httpx client) on every call, and nothing previously released it --
+        callers (e.g. src/swarm/worker.py's run_worker, which builds one
+        ChatLLM per task) just let it fall out of scope, leaking one pooled
+        connection each. LangChain's chat model wrappers expose no uniform
+        close API across providers, so this probes the attribute names
+        actually used by the adapters build_llm() can return (root_client/
+        client, covering ChatOpenAI and its OpenAI-compatible subclasses,
+        which is the path most deployments without the optional native
+        langchain-deepseek/langchain-anthropic extras installed will hit)
+        and swallows any failure -- called from cleanup, where a close
+        error must never mask the task's real result.
+        """
+        for attr in ("root_client", "client"):
+            client = getattr(self._llm, attr, None)
+            close = getattr(client, "close", None)
+            if callable(close):
+                try:
+                    close()
+                except Exception:
+                    logger.debug("ChatLLM.close: failed to close %s", attr, exc_info=True)
+
     def chat(self, messages: List[Dict[str, Any]], tools: Optional[List[Dict[str, Any]]] = None, timeout: Optional[int] = None) -> LLMResponse:
         """Call the LLM synchronously.
 
--- a/agent/src/swarm/worker.py
+++ b/agent/src/swarm/worker.py
@@ -383,6 +383,47 @@ def run_worker(
     grounding_block: str = "",
     agent_config: AgentConfig | None = None,
 ) -> WorkerResult:
+    """Run a single worker task and always close the ChatLLM it builds.
+
+    Thin wrapper around :func:`_run_worker_impl`: the impl builds a fresh
+    ``ChatLLM`` per call (see its step 2) and nothing previously closed the
+    provider HTTP client backing it, so every swarm task leaked one pooled
+    connection -- CLOSE-WAIT sockets accumulating on the vibe-trading process
+    over the life of a run. ``llm_holder`` is how the impl hands the instance
+    back out for cleanup without threading a close call through its many
+    early ``return WorkerResult(...)`` points.
+    """
+    llm_holder: list[ChatLLM] = []
+    try:
+        return _run_worker_impl(
+            agent_spec,
+            task,
+            upstream_summaries,
+            user_vars,
+            run_dir,
+            event_callback,
+            include_shell_tools,
+            grounding_block,
+            agent_config,
+            llm_holder,
+        )
+    finally:
+        for llm in llm_holder:
+            llm.close()
+
+
+def _run_worker_impl(
+    agent_spec: SwarmAgentSpec,
+    task: SwarmTask,
+    upstream_summaries: dict[str, str],
+    user_vars: dict[str, str],
+    run_dir: Path,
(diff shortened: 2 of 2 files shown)


warren618:
Closing in favour of #1145, which fixes the same leak (#1141).

Both PRs close the per-task `ChatLLM`. This one calls `.close()` unconditionally, which fails against every test double in the swarm suite — CI recorded 24 failures, all `AttributeError: '_StubChatLLM' object has no attribute 'close'`. #1145 guards the call and the full Python suite passes (10724 passed).

Thanks for the report in #1141 — that's what got this found, and the fix is landing.

PR #1148, opened 2026-08-19
fix(grounding): resolve unit-confusion false positives in price-claim validation

549236606-oss:
## Summary

Fixes four categories of unit-confusion false positives in GroundingLedger price-claim validation.

## Changes

- Fix A: _is_explicit_derivation accepts = without keyword
- Fix B: _QUANTITY_WITH_UNIT_RE extended to mask qty/premium/multiplier etc
- Fix C: _compare_price_claim skips comparison when scale differs >100x
- Fix D: safe_fallback routes unit-conflict cases to no-tool-evidence message

## Test plan

- [x] 25 new unit tests
- [x] Full existing grounding test suite (66/66) passes
- [x] Docker end-to-end verification
- [ ] Real conversation test


warren618:
Thanks — the grounding false positives you're describing are real, but this branch can't land as it stands. Its workflows were never approved so none of this ran in CI; I merged it into main locally and ran the suite.

**Blocker: Fix C reopens the gate it is meant to tighten.** `test_a_comparison_with_an_invented_quote_is_still_rejected` now fails. In `_compare_price_claim`:

```python
if ratio > 100 or ratio < 0.01:
    return None
```

`return None` means "no issue", so the claim is accepted. With observed evidence at 1.171 and a claimed 999.99 the ratio is 854, and the invented quote passes. The effect is that the further a hallucinated price is from anything observed, the likelier it survives — the opposite of what the gate is for. A unit conflict needs its own outcome, not the accept path.

**Please also split the branch.** Beyond the stated grounding work it carries:

- all of #1135 (now merged, so it should drop out on rebase);
- `docker-compose.yml` + `scripts/install-broker-sdks.sh` — bind mounts of your own working copy, with a comment naming a local path. Those belong in a local override file, not in the repo;
- `agent/src/agent/context.py` — a system-prompt policy change that relaxes when the agent may answer without price evidence. That deserves its own PR and its own discussion;
- `agent/src/market_data.py` — a 30s deadline plus `max_fallback_attempts` 5→3, neither mentioned in the description. Dropping two fallback attempts makes resolution less robust, and it collides with #1105.

**One piece I'd take right away:** the `check_status` envelope fix in `futu/sdk.py`. Adding `connection_state`/`last_checked_at` to match the longbridge shape fixes a real Web UI bug — `/live/status` renders a working Futu connection as unavailable without them — and `test_futu_check_status_envelope.py` fails on main today. Send that as its own PR and it goes in immediately.

Rebased onto current main with only the grounding work and a different handling of the unit-conflict case, I'll review again promptly.

he-yufeng:
The grounding half of this is strong. Dropping bare `trade` from the actionable regex, masking count contexts before canonical symbols, and gating the "price conflict" fallback on an actual `numeric_claim_conflict` are all the right calls, and the 100x unit-scale gate in `_compare_price_claim` kills the premium-vs-OHLC false positives cleanly. Happy to see this part land.

Two things need to change before this can merge, though:

1. Please split the PR. The grounding fixes, the seven new futu read endpoints, and the docker-compose / runtime-install changes are three separate reviews with very different risk profiles, and bundling them makes each one harder to verify.

2. The compose changes weaken the deployment's security model and read like local customizations rather than upstream defaults: `read_only: true` is removed from the service, the entrypoint is forced to root (`user: "0:0"`) so a script can `pip install` packages at every container start, and the new bind mounts reference a personal host path. Runtime `pip install` also bypasses the hash-pinned requirements lock, which is the supply-chain control this repo currently has. If the goal is making the futu SDK available without a long rebuild, installing it in the Dockerfile inside the pinned requirements flow (or declaring an optional extra) gets there without dropping rootfs immutability and without a root entrypoint.

The futu endpoints themselves look read-only and envelope-consistent from a first pass. Once they are split out I'll give them a proper read.

warren618:
Final ping before this gets closed for inactivity. The two reviews above (2026-08-19) stand: rebased to the grounding-only work with the unit-conflict case getting its own outcome rather than the accept path, this is wanted; and the futu `check_status` envelope fix merges immediately as its own PR. There has been no response for over a week. If nothing moves by 2026-09-03 I will close this to keep the queue honest — closing is not a rejection, and it reopens the moment a split branch shows up.

warren618:
Closing per the 2026-08-27 notice. There has been no response since 2026-08-19
and the deadline was 2026-09-03.

This is not a rejection, and nothing here is wasted. To restate what stands:

- The grounding work is wanted. Dropping bare `trade` from the actionable regex,
  masking count contexts before canonical symbols, and gating the price-conflict
  fallback on an actual `numeric_claim_conflict` are all correct. The one change
  needed is that a unit-scale conflict gets its own outcome rather than the
  accept path — `return None` there means "no issue", so the further a
  hallucinated price sits from anything observed, the likelier it survives.
- The futu `check_status` envelope fix has since landed separately, so that part
  is already in main and you don't need to carry it.

A rebased branch with only the grounding work reopens this immediately. Thanks
for the report and the analysis behind it.

Issue #1157, opened 2026-08-20
[Feature] add failed task replay/resume for swarm DAG runs


SiMinus:
### Problem or Motivation

When a swarm run fails after some upstream tasks have completed, users currently need to rerun the whole preset. This wastes LLM time and can discard useful completed artifacts.

### Proposed Solution

I propose adding a replay/resume path that:

- keeps completed upstream tasks and artifacts
- resets the failed task and downstream dependent tasks
- re-executes only the affected DAG subgraph
- leaves completed independent branches untouched

Initial scope:

- support failed/cancelled swarm runs only
- CLI/API entry point can be discussed
- no changes to the ReAct agent loop
- no live trading or broker behavior involved

If this direction sounds reasonable, I’d be happy to work on a focused first PR for the minimal replay path.

### Area

Swarm (new preset or workflow)

### Contribution

- [x] I'm willing to submit a PR for this feature

warren618:
Shipped in #1158 (main `76c8c8b4`), and thank you for the shape of this request — the scope you drew is the scope that landed, including the two constraints that made it reviewable: failed/cancelled runs only, and no changes to the ReAct loop.

One thing I owe you directly: you offered to take the first PR and @cgycorey opened one nine hours later. That is not how I would want a contribution offer to be answered, and if you have appetite for the follow-up work below it is yours without competition.

What landed:

- `retry_run(run_id, resume=True)` on MCP and `POST /swarm/runs/{run_id}/retry?resume=true` on REST. `resume` defaults `false`, so existing retry behaviour is byte-identical.
- Completed tasks carry forward with their summary, `completed_at`, worker iterations and artifacts — artifacts are copied into the new run directory so the record is self-contained and the old run stays deletable.
- Kept tasks emit `task_resumed` with `source_run_id`, and their summaries feed re-executed dependents through `input_from`.

Two guards worth naming, because they go past the issue's scope and I think they are what makes the feature trustworthy rather than merely working:

- A completed task is **not** kept when its definition changed since the original run — its agent, prompt template, DAG wiring, or the agent spec that executes it (role, system prompt, tools, skills, model, retries, timeout). A run-level model or provider swap invalidates kept results too, for agents that do not pin their own model.
[...]

PR #1158, opened 2026-08-20
feat(swarm): replay/resume failed runs keeping completed task artifacts

cgycorey:
Implements the minimal replay path for #1157.

## What
- `retry_run` (MCP) and `POST /swarm/runs/{run_id}/retry` accept `resume=true`: the new run carries every completed task from the failed/cancelled run forward — status, summary, `completed_at`, and artifacts (re-homed into the new run directory so the record is self-contained) — and re-executes only the failed/cancelled subgraph.
- Kept tasks are skipped by the layer executor, emit a `task_resumed` event with `source_run_id`, and their summaries feed re-executed dependents via `input_from` (no empty-upstream regression: the 5/27 gating stays intact).
- Guards: artifact paths escaping the source run dir are dropped; a missing/unreadable prior task store degrades to a full fresh re-run (logged).
- `resume` defaults `false` — existing retry behavior is byte-identical.

## Design note
DAG semantics make closure math unnecessary: a completed task can never depend on a failed one, so "keep everything completed, reset everything else" is exactly the failed-subgraph re-execution.

## Test plan
- New `agent/tests/test_swarm_replay.py`: keeps-completed-and-reruns-subgraph (failed + cancelled parametrized), artifact re-homing + escape-path drop, `task_resumed` event, degraded-fresh fallback.
- Extended `agent/tests/test_swarm_retry.py`: resume plumbing through the MCP tool.
- Full swarm suite: `313 passed`; `ruff check` clean on all changed files.

### DIFF
--- a/agent/mcp_server.py
+++ b/agent/mcp_server.py
@@ -2684,7 +2684,7 @@ def reap_stale_runs() -> str:
 
 
 @mcp.tool
-def retry_run(run_id: str) -> str:
+def retry_run(run_id: str, resume: bool = False) -> str:
     """Retry a failed, stale, or cancelled swarm run.
 
     Re-launches a brand-new run with the same preset and variables as the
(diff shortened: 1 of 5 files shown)


cgycorey:
Review fixes applied (HEAD `0cec9e4b`), all with regression tests:

1. **`artifacts/../run.json` clobber** — artifact re-homing now rejects `..` components lexically and requires the *resolved* source and destination to land inside the `artifacts/` subtree on both sides (previously only constrained to the run root). `test_replay_rejects_dotdot_artifact_clobber`.
2. **Internal symlink (e.g. `artifacts/analyst -> tasks`)** — the same subtree rule prevents re-homing from redirecting onto task/control files. `test_replay_source_symlink_artifact_dropped`.
3. **Changed agent definitions kept stale results** — the keep guard now also compares the agent spec (system prompt, tools, skills, model, retries, timeout) alongside the task fields. `test_replay_does_not_keep_changed_agent_definition`.
4. **Changed upstream did not invalidate dependents** — the overlay now drops any kept task whose transitive upstream re-executes, so a dependent's summary can never be stale relative to fresh upstream output. `test_replay_reexecutes_dependents_of_reexecuted_upstream`.
5. **Partial-task-store contract** — documented (and tested) that a wholly missing/unreadable store degrades to a fresh run while an individually missing task simply re-executes: `test_replay_partially_missing_store_keeps_present_completed`.

Swarm test suite: 327 passed; lint clean.

Issue #1126, opened 2026-08-17
[Bug] Documented IBKR MCP endpoint returns HTTP 400


sykuang:
### Description

**What happened:**

The README and seeded `IBKR_MCP_SERVER_SEED` configure the official IBKR connector with:

```text
https://api.ibkr.com/v1/api/mcp
```

After completing OAuth authorization, connector discovery fails because that documented endpoint returns HTTP 400:

```shell
httpx.HTTPStatusError: Client error 400 Bad Request for url https://api.ibkr.com/v1/api/mcp
```

This prevents the documented `ibkr-live-official-mcp-readonly` profile from discovering or calling IBKR tools.

**Expected:**

Following the README configuration and running the documented authorization command should produce a usable read-only IBKR MCP connection.

### Steps to Reproduce

1. Use the README/seeded IBKR MCP configuration with `https://api.ibkr.com/v1/api/mcp`.
2. Run:
   ```bash
   vibe-trading connector authorize ibkr-live-official-mcp-readonly
   ```
3. Complete IBKR OAuth authorization.
4. Attempt connector/tool discovery.
5. Observe HTTP 400 from `https://api.ibkr.com/v1/api/mcp`.

### Error Logs

```shell
httpx.HTTPStatusError: Client error 400 Bad Request for url https://api.ibkr.com/v1/api/mcp
```

### Interface

CLI (`vibe-trading`)

### LLM Provider

Not applicable; failure occurs during MCP OAuth/transport setup.

### Version

local main checkout

### Environment

macOS / Python 3.11

### Possible Solution

IBKR currently publishes a public MCP endpoint at:

```text
https://api.ibkr.com/v1/api/mcp-public
```

A draft implementation is available for review: https://github.com/sykuang/Vibe-Trading/pull/2

The draft changes the seeded/documented endpoint to `/mcp-public` and adds the narrow IBKR OAuth compatibility needed for authorization, reconnect, token refresh, and MCP initialization. With that implementation, live verification discovered 34 tools and completed a read-only `get_account_summary` call.

warren618:
Investigated. The endpoint and your configuration are both fine — the 400 is IBKR rejecting the token, not a malformed MCP request.

Probing `https://api.ibkr.com/v1/api/mcp` directly gives a clean contrast:

| request | response |
|---|---|
| no `Authorization` header | **401** |
| `Authorization: Bearer <anything unusable>` | **400** |

[...]

PR #1178, opened 2026-08-22
fix(ibkr): point official read-only MCP seed at /mcp-public endpoint

c020627:
# fix(ibkr): point official read-only MCP seed at /mcp-public endpoint

## Problem

The official IBKR MCP read-only profile (`ibkr-live-official-mcp-readonly`) is seeded and
documented with `https://api.ibkr.com/v1/api/mcp`. After completing OAuth, connector/tool
discovery fails because that endpoint returns HTTP 400:

```
httpx.HTTPStatusError: Client error 400 Bad Request for url https://api.ibkr.com/v1/api/mcp
```

This blocks the documented read-only probe from discovering or calling IBKR tools.
(Reported in #1126.)

## Solution

IBKR publishes the public MCP endpoint at `https://api.ibkr.com/v1/api/mcp-public`. This PR:

- Updates the seeded default `IBKR_MCP_SERVER_SEED["url"]` in `agent/src/config/schema.py`
  — the actual config value used at runtime.
- Updates the two matching docs: `README.md` ("Official IBKR MCP read-only probe") and
  `agent/SKILL.md`.
- Aligns the two seed assertions in `agent/tests/test_ibkr_mcp_seed.py` and
  `agent/tests/test_mcp_oauth_schema.py` (these assert the seed URL).

## Affected areas

- `agent/src/config/schema.py` — `IBKR_MCP_SERVER_SEED`
- `README.md`, `agent/SKILL.md` — documentation
- `agent/tests/test_ibkr_mcp_seed.py`, `agent/tests/test_mcp_oauth_schema.py` — tests

## Out of scope

- No OAuth compatibility code, no new tools, no order-path changes. The read-only `mcp.read`
  wildcard discovery probe, its scopes, and the live order gate are unchanged.
- Translation files (`README_*.md`) are intentionally left for a follow-up sync.
- This is a config/doc correction only; it does not add the broader IBKR OAuth client-approval
  flow described in the issue (that is a larger enhancement and out of scope for this minimal fix).

## Test plan

- `python -m py_compile agent/src/config/schema.py` and the two test modules — pass.
- Targeted validation (local, pydantic only):
  `AgentConfig.model_validate({"mcpServers": {"ibkr": IBKR_MCP_SERVER_SEED}})` yields
  `server.url == "https://api.ibkr.com/v1/api/mcp-public"` with `scopes == ["mcp.read"]` and
  `enabled_tools == ["*"]` — verified.
- The two updated assertions now match the corrected seed URL.
- The full suite (`pytest agent/tests/test_ibkr_mcp_seed.py agent/tests/test_mcp_oauth_schema.py`)
  was not executed here because it requires the heavy `fastmcp`/`mcp` dev environment; the changed
  lines are pure string literals matching the seed, so they pass once that environment is installed.

## Risk

- No live/broker/MCP write, no network call, no secret, no deployment, no CI-secret change.
- The change only repoints a read-only profile's URL to IBKR's public MCP endpoint. The profile
  stays read-only (`mcp.read`, wildcard discovery probe); the live order gate is untouched.
- Existing users who already wrote the old URL into `~/.vibe-trading/agent.json` must re-run
  `vibe-trading connector configure ibkr-live-official-mcp-readonly --yes` (or re-seed) to pick up
  the corrected endpoint.

## Rollback

- Documentation/config-only change; revert with `git revert <sha>`.

## DCO

- Every commit is signed off (`Signed-off-by:`).

---

Refs #1126 (the OAuth client-registration failure reported there stays open)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1495,7 +1495,7 @@ endpoint in read-only mode. Add this to `~/.vibe-trading/agent.json`:
   "mcpServers": {
     "ibkr": {
       "type": "streamableHttp",
-      "url": "https://api.ibkr.com/v1/api/mcp",
+      "url": "https://api.ibkr.com/v1/api/mcp-public",
       "auth": {
         "type": "oauth",
         "scopes": ["mcp.read"],
--- a/agent/SKILL.md
+++ b/agent/SKILL.md
@@ -256,7 +256,7 @@ Add Interactive Brokers' official MCP endpoint as a read-only external server:
   "mcpServers": {
     "ibkr": {
       "type": "streamableHttp",
-      "url": "https://api.ibkr.com/v1/api/mcp",
+      "url": "https://api.ibkr.com/v1/api/mcp-public",
       "auth": {
         "type": "oauth",
         "scopes": ["mcp.read"],
--- a/agent/src/config/schema.py
+++ b/agent/src/config/schema.py
@@ -226,7 +226,7 @@ ROBINHOOD_MCP_SERVER_SEED: dict[str, object] = {
 # config must pin an explicit tool allowlist plus pass the live order gate.
 IBKR_MCP_SERVER_SEED: dict[str, object] = {
     "type": "streamableHttp",
-    "url": "https://api.ibkr.com/v1/api/mcp",
+    "url": "https://api.ibkr.com/v1/api/mcp-public",
     "auth": {
         "type": "oauth",
         "scopes": ["mcp.read"],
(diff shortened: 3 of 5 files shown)


warren618:
Merged — thank you. IBKR's own AI-integration page gives `https://api.ibkr.com/v1/api/mcp-public` as the URL to enter in an MCP client, which settles it. I synced the five translated READMEs in the same landing so the seed and all seven documentation sites agree.

One change to the linkage: this does not close #1126. The repoint is necessary, but @sykuang's 22 Aug run shows the probe on main is still stopped earlier, at IBKR rejecting FastMCP's dynamic client registration with 403 — that is tracked there and needs its own PR. So the reference is "refs", not "closes".

PR #1179, opened 2026-08-22
feat(web): support multi-file drag-and-drop and paste uploads

AirHua-byte:
## Summary

- Let Web chat attach up to five files in one turn.
- Add multi-select, drag-and-drop, and clipboard image/file paste.
- Replace the PDF-only menu label with a generic upload action while preserving legacy attachment history.

## Why

The upload endpoint and agent tools already support documents, data files, and images, but the composer presents a PDF-only, single-file workflow. Users cannot drag files into the chat or paste screenshots from the clipboard.

This is a focused Web attachment proposal based on the existing upload API; there is no linked issue.

## Changes

- Add a five-file attachment queue with removable, accessible filename chips and attachment-only send support.
- Accept files from the picker, drag-and-drop, and clipboard paste while reusing the existing 50 MB per-file upload API.
- Align client-side blocked extensions with the backend and stop advertising YAML types that the API rejects.
- Encode multiple uploaded paths in a structured prompt envelope and keep legacy single-file history readable.
- Render all attachment filenames in message history without exposing server paths.
- Update all seven locales, README capability/API text, and interaction/parser tests.

## Test Plan

- [x] Existing tests pass (`npm test -- --run`): 59 files, 579 tests.
- [x] New tests cover multi-file drop, clipboard image paste, attachment-only submit, five-file limit, history chips, prompt round-trip, malformed envelopes, and legacy history.
- [x] Tested manually in headed Chromium: picker multi-upload, clipboard PNG upload, drag target, English/Chinese labels, HTTP 200 upload responses, and zero console errors/warnings.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion.
- [x] Documentation updated (if user-facing change).


warren618:
Merged — thank you. Verified locally: vitest 60 files / 586 tests, `tsc -b` clean, the locale parity suite green across all seven files. One follow-through in the landing: `agent.uploadPdf` had no remaining reader once the menu label moved to `agent.uploadFile`, so it is removed from all seven locales. I also mirrored the two README sentences (capability row, `/upload` API row) into the five translated READMEs.

PR #1180, opened 2026-08-22
fix(session): recover interrupted replies after service restart

AirHua-byte:
## Summary

- Checkpoint streamed assistant text while an attempt is running.
- Reconcile orphaned pending/running attempts after restart as an explicit interrupted transcript entry.
- Make session JSON updates atomic and preserve the reply/attempt commit window.

## Why

A service restart currently drops the in-memory assistant stream. The user message and attempt stay on disk, but no assistant reply is appended and the attempt can remain running indefinitely.

This is a focused reliability proposal based on a locally reproduced restart gap; there is no linked issue.

## Changes

- Add throttled partial-response checkpoints with atomic JSON replacement.
- Add an `interrupted` terminal attempt state and idempotent startup reconciliation.
- Commit the user-visible reply before `attempt.json` so startup can finish either side of that two-file write.
- Keep explicit user cancellation distinct from event-loop shutdown cancellation.
- Add restart, commit-window, cancellation, and atomic-write regression coverage.
- Document partial-response recovery behavior.

## Test Plan

- [x] Existing tests pass (`python -m pytest -q agent/tests`): 11,042 passed, 93 skipped.
- [x] New tests added for checkpoints, restart reconciliation, idempotency, graceful shutdown, reply-first recovery, and atomic JSON writes.
- [ ] Tested manually (not run against a live in-progress agent to avoid interrupting operator work; process lifecycle paths are covered by automated service recreation/cancellation tests).

## Checklist

- [ ] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion.
  - `agent/src/session/` is protected and there was no prior maintainer discussion; this is intentionally opened as a Draft for direction.
- [x] Documentation updated (if user-facing change).

### DIFF
--- a/README.md
+++ b/README.md
@@ -799,7 +799,7 @@ Docker publishes the backend on `127.0.0.1:8899` by default and runs the app as 
 > [!NOTE]
 > **Using Ollama with Docker:** the container reaches a host-side Ollama via `host.docker.internal`, not `localhost` (inside the container `localhost` is the container itself). `docker-compose.yml` defaults `OLLAMA_BASE_URL` to `http://host.docker.internal:11434`; export `OLLAMA_BASE_URL` (or set it in a top-level `.env`) to point elsewhere. This relies on the `host-gateway` mapping in `extra_hosts`, which requires **Docker Engine ≥ 20.10 / Compose v2** (provided automatically on Docker Desktop).
 
-Your data survives updates: persistent memory, the cross-session search index, user-created skills, shadow accounts, broker connector config, web sessions, backtest runs, swarm history, and uploads all live in named Docker volumes, so `git pull && docker compose up --build` keeps them. They are deleted only by `docker compose down -v`.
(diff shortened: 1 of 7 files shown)


warren618:
Merged — thank you, and noting that I reviewed this as the protected-area change it is (`src/session/`), since the Draft status you described in the checklist was not set on the PR.

What convinced me: `SessionService` is constructed exactly once per process (`api/state.py`), so "any pending/running attempt at construction time belongs to a previous process" holds; the checkpoint only reacts to `text_delta` / `stream_reset`, both of which `AgentLoop` really emits; the append-message-then-attempt.json ordering makes the startup reconciliation correct on either side of the crash window, and the tests cover exactly that window. Full suite 11232 passed.

Two limits worth writing down:
1. The single-process assumption is per storage root, not per machine. Two servers sharing one `~/.vibe-trading/sessions` (two `serve` ports, or Docker plus local) would have the second start mark the first's in-flight attempt interrupted and append a recovery message the first server then follows with the real reply. No data is lost, but it is a spurious transcript entry. A cheap guard — an owner pid or heartbeat on the attempt that recovery checks — would be a welcome follow-up if you want it.
2. `os.replace` over an open `session.json` on Windows raises `PermissionError`; CI's pytest job runs on Ubuntu only, so that path is unverified here. If a Windows user reports attempt updates failing, this is the first place to look.

PR #1182, opened 2026-08-22
fix(providers): preserve shared HTTP clients during cleanup

AirHua-byte:
## Summary

- Keep LangChain's cached HTTP clients alive when a short-lived `ChatLLM` is closed.
- Close only HTTP clients explicitly created and owned by Vibe-Trading.
- Handle synchronous and asynchronous cleanup without un-awaited coroutine warnings.

## Why

This is a follow-up to #1141 and #1153. With current `langchain-openai`, separately constructed `ChatOpenAI` adapters can borrow the same cached sync and async HTTPX clients. The existing `ChatLLM.close()` closes `root_client` and `root_async_client` as though every adapter owns them.

After any one-shot path such as automatic title generation or image vision closes its wrapper, later requests can reuse the already-closed cached client and fail persistently with `OpenAIConnectionError: Connection error` (caused by `RuntimeError: Cannot send a request, as the client has been closed`) until the service restarts. The async root client is also closed without awaiting its coroutine.

## Changes

- Add an explicit private ownership marker to the OpenAI-compatible adapter.
- Leave the marker empty for LangChain-managed default transports, treating them as borrowed process resources.
- Mark the explicit proxy-free sync/async HTTPX clients created by `build_llm()` as owned.
- Deduplicate close targets and consume async cleanup correctly from both sync and async call sites.
- Preserve the prior best-effort cleanup behavior for native adapters that do not expose the ownership marker.
- Add lifecycle regression tests for borrowed clients, owned clients, async cleanup, deduplication, native-adapter compatibility, and transport construction.

## Test Plan

- [x] Full backend suite: `11,126 passed, 93 skipped`.
- [x] Focused provider/session/swarm suite: `150 passed`.
- [x] Ruff passed on all changed Python files.
- [x] New lifecycle test file passes Black. The existing changed files also fail Black unchanged on `origin/main`, so this PR avoids unrelated whole-file formatting.
- [x] Manual OpenRouter A/B: three default-client streams succeeded across close boundaries; two explicitly owned direct-client streams succeeded and closed both sync/async clients; two post-restart streams succeeded and the API remained healthy.

## Risk and rollback

This changes only provider HTTP-client lifecycle. It does not change proxy settings, provider selection, retries, status-code handling, credentials, sessions, trading, MCP, dependencies, or deployment behavior. Genuine transient network/provider failures remain visible. Rollback is a single revert of this commit.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion. Provider lifecycle cleanup was discussed in #1141 and implemented in #1153; this is a reproduced follow-up correction.
- [x] Documentation updated if user-facing (not applicable; lifecycle behavior only).

### DIFF
--- a/agent/src/providers/chat.py
+++ b/agent/src/providers/chat.py
@@ -5,7 +5,9 @@ ChatLLM is designed specifically for the AgentLoop ReAct cycle.
 
 from __future__ import annotations
 
+import asyncio
 import html
+import inspect
 import logging
 import os
 import re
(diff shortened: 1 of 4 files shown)


warren618:
Merged — thank you, and this one is real: langchain-openai 1.4.1 builds the default transports through `_cached_sync_httpx_client` / `_cached_async_httpx_client` (an `lru_cache` keyed on base URL and timeout), so `close()` on one wrapper was closing the client every later `ChatOpenAI` on that base URL still holds. Your five tests exercise stub adapters; I added one that constructs two real `ChatOpenAIWithReasoning` instances on the same base URL, proves they share the cached clients, closes one wrapper and asserts the sibling's transport is still open — it fails on main (`is_closed is True`) and passes with your change. Full suite 11232 passed.

One thing I did not verify, flagged rather than assumed: `langchain-deepseek`'s native `ChatDeepSeek` adapter has no ownership marker and keeps the legacy cleanup. If it also borrows the cached default transports, the same symptom would remain on that path. Not a regression from this PR — noting it in case you or anyone with that adapter installed wants to check.

AirHua-byte:
Thanks for the careful review and for adding the real shared-transport regression test.

I noticed your note about the native `langchain-deepseek` path and checked it locally against both the minimum supported release (1.0.0) and the current release (1.1.0). `ChatDeepSeek` creates separate `openai.OpenAI` and `AsyncOpenAI` clients per instance, so neither the sync nor async transports are shared. Closing one `ChatLLM` wrapper closed only its own transports, while a sibling remained usable.

So this ownership concern does not reproduce on the currently released 1.x versions. Thanks for flagging it.

PR #1184, opened 2026-08-23
fix(valuation): refuse non-finite comps and three-statement inputs

### LINKED ISSUE #1183: comps and three-statement engines silently accept non-finite inputs
`_compute_multiple` (comps) and `check_balance_sheet` (three-statement) still let NaN/inf through — the same gap that #1120/#1121 closed in DCF.

Repro:

```python
from src.quantlib.valuation.comps import _compute_multiple
_compute_multiple(float("nan"), 100.0, peer_name="P", multiple_name="ev_ebitda")
# -> (nan, None)  — a nan multiple that is *included* in the distribution and drags the median to nan
```

```python
from src.quantlib.valuation.threestatement import check_balance_sheet, BalanceSheet
check_balance_sheet(BalanceSheet(period=0, cash=float("nan"), net_working_capital=1.0,
    ppe=2.0, revolver_balance=1.0, paid_in_capital=2.0, retained_earnings=0.0))
# -> passes silently  — abs(nan) > tolerance is always False, so a nan balance sheet sails through the "hard" balance check
```

Both comparisons are NaN-blind: `denominator <= 0.0` / `numerator <= 0.0` in `_compute_multiple`, and `abs(residual) > tolerance * scale` in `check_balance_sheet`. The DCF engine and the contracts helpers (`require_positive`, `_require_nonnegative`) already refuse non-finite inputs; these two sibling engines are the gap.

Robin1987China:
Closes #1183

## Summary

The comparables and three-statement valuation engines now refuse non-finite inputs everywhere they enter the arithmetic, instead of silently propagating them into a multiple, an EV bridge, a calendarised metric, a balance-sheet check, or a projection.

## Why

Five entry points let a NaN/inf through, all the same class of bug #1120/#1121 closed in DCF:

- `_compute_multiple`'s `<= 0.0` guards are NaN-blind, so `(nan, None)` was returned and the `nan` multiple poisoned the cross-sectional median.
- `enterprise_value` / `equity_value_from_enterprise_value` returned a non-finite EV/equity value.
- `calendarise_metric` returned a non-finite LTM/calendar-year metric.
- `check_balance_sheet`'s `abs(residual) > tolerance` is false for `nan`, so a non-finite balance sheet passed the "hard" check.
- `project_three_statement` surfaced a NaN driver as a misleading `ConvergenceError` ("did not converge") rather than a non-finite refusal.

## Changes

- `agent/src/quantlib/valuation/comps.py`: `_compute_multiple` excludes a non-finite side with `non_finite_numerator` / `non_finite_denominator`; `_bridge_delta` + both bridge functions and `calendarise_metric` raise `ValuationError` on non-finite inputs.
- `agent/src/quantlib/valuation/threestatement.py`: `check_balance_sheet` raises `BalanceSheetError` on a non-finite residual; `project_three_statement` validates every opening figure and driver entry up front via a new `_require_finite` helper.
- Tests in `test_comps.py` and `test_threestatement.py`.

## Test Plan

- [x] ruff clean on all touched files
- [x] `pytest tests/quantlib/valuation -q` → 191 passed (10 new)
- [x] `pytest tests/quantlib -q` → 1077 passed, 64 skipped

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (docstrings updated)

### DIFF
--- a/agent/src/quantlib/valuation/comps.py
+++ b/agent/src/quantlib/valuation/comps.py
@@ -91,6 +91,7 @@ independently importable.
 
 from __future__ import annotations
 
+import math
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from types import MappingProxyType
@@ -251,7 +252,8 @@ def calendarise_metric(
         The aligned :class:`CalendarisedMetric`.
 
     Raises:
-        ValuationError: If `policy` is not one of `CALENDARISATION_POLICIES`.
+        ValuationError: If `policy` is not one of `CALENDARISATION_POLICIES`,
+            or if any supplied period value is not a finite number.
         MissingInputError: If `periods` lacks a field this policy needs.
     """
     if policy not in CALENDARISATION_POLICIES:
@@ -260,6 +262,17 @@ def calendarise_metric(
             f"one of {CALENDARISATION_POLICIES}"
         )
     model = f"comps.calendarise_metric[{company_name}.{metric_name}:{policy}]"
+
+    for name, val in (
+        ("last_full_fiscal_year", periods.last_full_fiscal_year),
+        ("current_year_to_date", periods.current_year_to_date),
+        ("prior_year_to_date", periods.prior_year_to_date),
+        ("next_full_fiscal_year", periods.next_full_fiscal_year),
+    ):
+        if val is not None and not math.isfinite(val):
+            raise ValuationError(
+                f"{model}: {name} must be a finite number, got {val!r}"
+            )
 
     if policy == "ltm":
         require_inputs(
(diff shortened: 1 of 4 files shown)


warren618:
Merged to main. Verified the mechanism before merging: `require_inputs` still runs first in `project_three_statement` (threestatement.py:675), so a *missing* field keeps raising `MissingInputError` and the new finite checks only see present fields — the contract ordering survives. `import math` was already in scope in both modules. The exclusion-record approach in `_compute_multiple` (excluded + reasoned, never averaged) is exactly how the non-positive cases were handled, so the two failure classes now read the same way downstream. Full suite green locally. Thanks for pairing the issue with the fix and for the NaN-blind `abs(residual) > tolerance` repro — that one-line predicate hid a whole class of silent passes.

PR #1185, opened 2026-08-23
fix(market-data): validate codes, date range, and source before fetching

cgycorey:
## Problem

`get_market_data` passed kwargs straight to the loader layer with no input validation. Four failure modes:

1. **Empty `codes` list** returned a silent `{}` — indistinguishable from "no data for these symbols", so an agent keeps going on empty results instead of fixing the call.
2. **`start_date > end_date`** or malformed dates burned the *entire fallback source chain* (all 5+ loaders, several seconds of network calls) before surfacing a misleading `_unresolved` that looks identical to a bad-symbol error.
3. **Missing required parameter** (e.g. no `start_date`) raised a raw `KeyError` envelope: `{"error": "\u0027start_date\u0027"}`.
4. **Unknown `source`** (e.g. `source="bogus"`) silently resolved to `_unresolved` after the same chain burn.

## Fix

Validate in `execute()` before any network call, mirroring the existing `get_fundamentals` tool pattern:

- `codes` must be a non-empty list of non-empty strings (stripped)
- `start_date`/`end_date` must be strict `YYYY-MM-DD` calendar dates (regex + `fromisoformat`; the latter alone accepts compact `YYYYMMDD` on Python 3.11+, which loaders reject)
- `start_date <= end_date`
- `source` must be in the declared enum

All invalid calls now fail fast with a clean `{"ok": false, "error": ...}` envelope and never touch the network.

## Tests

9 new tests in `tests/test_market_data_tool.py` covering every validation branch. Full suite passes (8946 passed; the one failure is a pre-existing machine-load-sensitive perf benchmark, reproduced on a clean tree).

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -166,9 +166,17 @@ def _align(
     Returns:
         (dates, close_df, positions_df, returns_df)
     """
-    # Build unified sorted date index from all symbols' trading calendars
+    # Build unified sorted date index from all symbols' trading calendars.
+    # sort+dedup instead of np.unique: unique's hash-based path costs ~15ms
+    # for a 250k-element merge of 50 calendars; sort+dedup is ~5ms and the
+    # sorted result is bit-identical (same values, same order).
     indexes = [data_map[c].index for c in codes]
-    merged = np.unique(np.concatenate([index.asi8 for index in indexes]))
+    merged = np.concatenate([index.asi8 for index in indexes])
+    merged.sort()
+    keep = np.empty(merged.shape[0], dtype=bool)
+    keep[0] = True
+    np.not_equal(merged[1:], merged[:-1], out=keep[1:])
+    merged = merged[keep]
     common_tz = indexes[0].tz
     if all(index.tz == common_tz for index in indexes) and common_tz is not None:
         dates = pd.DatetimeIndex(pd.to_datetime(merged, utc=True).tz_convert(common_tz))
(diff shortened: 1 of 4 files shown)


cgycorey:
Second commit: the MCP wrapper in mcp_server.py called fetch_market_data_json directly, bypassing the new validation — so the MCP surface (the primary external entry point) still had all three failure modes. Routed it through registry.execute("get_market_data", ...) like factor_analysis and the other wrappers, so both surfaces share one validated path. Output is unchanged: same loader resolver (get_loader is the shared default) and same include_provenance=True. 209 MCP tests pass.

warren618:
Merged to main with one carve-out: the `backtest/engines/base.py` `_align` rewrite (np.unique → sort+adjacent-dedup) did not go in. Three reasons: it isn't mentioned anywhere in the PR description; `keep[0] = True` raises IndexError when the merged calendar is empty, where `np.unique` returns an empty array; and the comment's rationale ("unique's hash-based path") doesn't match numpy — `np.unique` sorts, it doesn't hash, so if the speedup is real it comes from skipping unique's extra work, not from avoiding hashing. If the ~10ms matters in your workloads, please send it as its own PR with a benchmark and an empty-input guard and we'll take it on its merits.

Everything else is in as-is: the validation ordering (codes → dates → range → source → interval → max_rows), the strict `YYYY-MM-DD` check (good catch on `fromisoformat` accepting compact dates on 3.11+), the `_INTERVAL_CANON` case-normalization (`_VALID_INTERVALS` has no case collisions, verified), and deriving the source enum from `VALID_SOURCES` so the tool can never silently reject a registered loader again — the regression test that locks enum == registry is the part that keeps this fixed. One follow-up we did on top: the reroute left `fetch_market_data_json` imported-but-unused in `mcp_server.py`; dropped.

PR #1187, opened 2026-08-23
feat(scheduled): add agent-confirmed research tasks

AirHua-byte:
## Summary

- Add one `scheduled_research` Agent Tool for scheduler status, jobs, playbooks, and create/cancel proposals.
- Keep Agent mutations behind deterministic confirmation: Web confirmation card, CLI `y/N`, and exact IM confirmation replies.
- Align scheduled jobs with title, source, schedule, timezone, end boundary, delivery target reference, and delivery receipt fields while retaining compatibility with existing jobs.
- Add channel-agnostic configured delivery targets without exposing raw chat/user ids to the Agent or confirmation UI.
- Record generic adapter acknowledgements as `accepted`; record `sent` only when a provider receipt is available, currently implemented end to end for Feishu.
- Expose scheduler status, proposal commit/discard, and job detail APIs, with localized Web UI states.

## Safety boundaries

- Tool proposals never directly create or cancel jobs.
- Proposals expire after 15 minutes.
- Commit requires the scheduler executor to be enabled and running.
- Provider destination ids remain in operator configuration and persisted job state, not Agent-facing payloads.
- Existing jobs load with backward-compatible defaults.

## Validation

- Relevant backend suite: 354 passed.
- Frontend page and locale suite: 45 passed.
- TypeScript `tsc --noEmit` passed.
- Vite production build passed.
- Ruff and Black checks passed for the new and relevant changed Python files.
- Registry smoke test confirmed exactly one scheduled tool: `scheduled_research`.
- Local end-to-end validation covered proposal, confirmation, execution, Feishu delivery receipt, expiration, and cancellation.
- Additional manual acceptance was completed through the running application.

## Out of scope

- Additional scheduling tools.
- Provider-specific delivery receipts beyond Feishu.
- Redesigning the existing direct scheduled-task administration form.

## Risk and rollback

The change adds persisted proposal and job fields, new confirmation endpoints, and a delivery receipt state. Feishu now raises when its client is unavailable instead of silently treating a send as successful.

Rollback by reverting this commit. Existing scheduled jobs remain readable because the added fields have defaults; uncommitted proposal files are inert without the new endpoints.


warren618:
Merged to main. The proposal/commit separation is the right shape: the tool can only write an inert, TTL'd proposal file; every commit goes through an authenticated surface action; and `commit_proposal` refusing to run when the executor isn't enabled+running closes the "confirmed into a void" case. The delivery-target indirection (operator config holds the raw id, agent and confirmation UI see ref/label/channel) and the accepted-vs-sent receipt distinction both survive review; so does refusing to mark Feishu sends successful when the client is missing.

One fix landed on top (cfa9c36f): the IM confirmation only matched `确认`/`取消`, and the appended confirmation card was Chinese-only — the 16 adapters serve both audiences, and the channel runtime's deterministic strings default to English elsewhere (e.g. the busy message). It now accepts `confirm`/`cancel` (case-insensitive) alongside the Chinese tokens, renders the card in English naming both spellings, and there's a runtime test covering all four tokens plus free-text passthrough. Docs mirrored across the six READMEs. Also noted while reviewing: the IM confirmation path had no test at all — worth covering the surfaces you add, not only the store underneath them.

PR #1188, opened 2026-08-23
fix(feishu): persist QR login credentials

AirHua-byte:
## Summary

- Persist Feishu/Lark QR-authorized app credentials before reporting login success.
- Preserve existing agent configuration and atomically replace the JSON file with owner-only permissions.
- Reduce Feishu SDK log noise and document the persisted login behavior.

## Scope

This PR only fixes Feishu/Lark channel login persistence. It does not include the Scheduled Research Tool or scheduled delivery changes.

## Validation

- `pytest -q agent/tests/test_feishu_login_persistence.py agent/tests/test_feishu_parse_md_table_edge_columns.py`
  - 3 passed
- Ruff passed for the changed Python files.
- `git diff --check` passed.

## Risk and rollback

The QR flow now writes credentials to `~/.vibe-trading/agent.json` after authorization. The write uses a private same-directory temporary file, fsync, atomic replacement, and `0600` permissions.

Rollback by reverting this commit. Existing saved credentials can remain in place or be removed manually by the operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1035,6 +1035,10 @@ vibe-trading channels login weixin     # run an adapter login hook when needed
 vibe-trading channels pairing --channel telegram list
 ```
 
+`vibe-trading channels login feishu` saves the QR-authorized app credentials to
+`~/.vibe-trading/agent.json` with owner-only file permissions before reporting
+login success.
+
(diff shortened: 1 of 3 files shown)


warren618:
Merged to main. Verified the two hazards this class of change usually hides before merging: `_read_config_file` is a plain `json.loads` (no env expansion, no comments), so the read-modify-write can't bake expanded values into the file — and the explicit `.json`-only guard means a YAML config is refused rather than rewritten with its comments stripped. The atomic write (same-dir tmp, fsync, `os.replace`, 0600 before close) is the right sequence. README sentence mirrored to the five translations.

PR #1189, opened 2026-08-23
docs(skills): fix historical_var quantile formula in risk-analysis

youngjincho02-arch:
SKILL.md described the order-statistic index as `floor((1-confidence)*n)`, but `src/quantlib/risk.py`'s `_tail_index` actually implements `ceil((1-confidence)*n)-1`. The two agree at realistic VaR confidence levels (0.90/0.95/0.99) but not in general - updating the doc to match the code.


warren618:
Merged to main. Verified against the source before merging: `_tail_index`'s docstring (risk.py:133) states `ceil((1 - confidence) * n) - 1`, and its own comment even explains why the floor form agrees at every confidence level anyone uses — which is exactly why this doc bug could sit unnoticed. Docs that state a formula the code doesn't implement are worse than no formula; thanks for the precision.

PR #1195, opened 2026-08-24
feat(quantlib): add Heston (1993) stochastic volatility option pricing

santhreal:
### Summary
Implements semi-analytical European option pricing under the Heston (1993) stochastic volatility model via numerical quadrature of the continuous characteristic function (Lewis 2001 / Albrecher et al. 2007 formulation).

### Changes
- Added `agent/src/quantlib/volatility.py` implementing `heston_price`, `heston_characteristic_function`, and `heston_feller_condition`.
- Exposed functions via `QuantlibCallTool` in `agent/src/tools/quantlib_tool.py`.
- Added unit and property tests in `agent/tests/quantlib/test_volatility.py` testing against published literature benchmarks (Moodley 2005), put-call parity, and convergence to Black-Scholes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,7 +614,7 @@ Intraday bars: 1m / 5m / 15m / 30m / 1H / 4H / 1D. 15 metrics + benchmark compar
 </details>
 
 <details>
-<summary><b>Quant Library</b> <sub>286 tested functions across 19 modules, callable from every transport</sub></summary>
+<summary><b>Quant Library</b> <sub>289 tested functions across 19 modules, callable from every transport</sub></summary>
 
 `src/quantlib` holds one tested implementation of each piece of finance math the
 agent needs. Skills **import** these rather than carrying formulas inside
--- a/README_ar.md
+++ b/README_ar.md
@@ -610,7 +610,7 @@ LONGBRIDGE_ACCESS_TOKEN=...
 </details>
 
 <details>
-<summary><b>Quant Library</b> <sub>286 دالة مختبَرة عبر 19 وحدة، قابلة للاستدعاء من كل المسارات</sub></summary>
+<summary><b>Quant Library</b> <sub>289 دالة مختبَرة عبر 19 وحدة، قابلة للاستدعاء من كل المسارات</sub></summary>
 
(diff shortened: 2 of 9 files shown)


warren618:
Merged, along with #1196, #1197, #1198 and #1203 — with one correction on top, which I'm flagging because it changed a number rather than a style.

`heston_price` used `e^{-iuk}` in the Lewis integrand where the formula needs `e^{+iuk}`. That is not a small numerical difference: it inverts the volatility skew. With ρ = −0.6 (the leverage effect), the merged version priced a 120-strike call at 4.668 and a 90-strike call at 13.719; Monte Carlo on the same parameters gives 4.246 and 14.204 — rich where it should be cheap, in both directions. With the sign corrected, all four cases I checked land inside two standard errors of MC, and the implied-vol smile slopes the way a negative correlation requires.

Worth saying why the suite couldn't see it, because the tests you wrote were reasonable ones:

- as `sigma_v → 0` the centred characteristic function is **real**, so `Re[e^{+iuk}·φ] == Re[e^{-iuk}·φ]` and the Black-Scholes limit is bit-identical under either sign (I measured: 21.646231 both ways);
- the Moodley benchmark and the put-call parity check are both at the money, where `k = 0` and the factor is exactly 1.

So every property in the file was structurally blind to it. Two tests now pin the skew direction — ρ < 0 must make low strikes the expensive ones, and must cheapen OTM calls — and the sign carries a comment marking it load-bearing. Reverting the character fails exactly those two and nothing else, which is the check that was missing.

The rest all checked out. HRP reproduces López de Prado's own recursive bisection exactly on 4, 5, 7 and 10 assets (max difference 0.0); the copula CDFs, tail-dependence coefficients and Kendall-τ calibrations are each correct; Roll, Amihud, Kyle and the VPIN bucketing are right; and the barrier finite-difference Greeks match `bs_greeks` to 1e-9 in the far-barrier limit.

One housekeeping note for the batch: all five PRs bumped the Quant Library badge from the same 286 base (289 / 290 / 294 / 290 / 287), and none touched the module half — so four new modules landed while it still said nineteen. Measured after the merges: **306 functions across 23 modules**, now set in all six READMEs with the four new modules added to the table. The badge check only ever asserted the function number, so it now asserts both.

PR #1200, opened 2026-08-24
fix(quantlib): guard near-zero variance in OU fit and pair hedge ratio

santhreal:
### Summary
Guards against singular regressions and unhandled zero-variance inputs in Ornstein-Uhlenbeck parameter estimation, half-life computation, and pair hedge ratio fitting.

### Changes
- Added checks rejecting flat / zero-variance series ($\sigma \le 10^{-12}$) and enforcing $\ge 3$ aligned observations in `agent/src/quantlib/timeseries.py` (`compute_half_life`, `fit_ornstein_uhlenbeck`, and `find_hedge_ratio`).
- Added regression test `test_timeseries_constant_series_rejection` in `agent/tests/quantlib/test_timeseries.py`.

### DIFF
--- a/agent/src/quantlib/timeseries.py
+++ b/agent/src/quantlib/timeseries.py
@@ -221,7 +221,7 @@ def compute_half_life(spread: pd.Series) -> float:
     frame = pd.concat({"delta": delta, "lag": lagged}, axis=1).dropna()
     if len(frame) < 3:
         raise ValueError(f"compute_half_life needs at least 3 lag/delta pairs, got {len(frame)}")
-    if frame["lag"].std(ddof=0) == 0:
+    if float(frame["lag"].std(ddof=0)) <= 1e-12:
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Read this against current main. The constant-series case is already guarded: all three functions raise on `std == 0` with a clear error. What this changes is only the near-zero band in (0, 1e-12], and I don't see the motivating input: a spread series with std 1e-13 is not something the pipeline produces from real prices. If you hit this in practice, what was the input?

Two smaller things. The new test wraps its whole body in try/except ImportError and skips, so it can silently never run; move the import to the top and let it fail loudly. And the error-message rewording in find_hedge_ratio is unrelated churn, please revert that line.

santhreal:
Yes, this is not a realistic scenario. This is theoretical. I will close it out.

PR #1203, opened 2026-08-24
feat(quantlib): add finite-difference sensitivity Greeks for barrier options

santhreal:
### Summary
Implements central finite-difference calculation of sensitivity Greeks (Delta, Gamma, Theta, Vega, Rho) for single-barrier options.

### Changes
- Added `barrier_greeks` in `agent/src/quantlib/options.py` supporting all 8 standard barrier option configurations with Greek scaling consistent with `bs_greeks`.
- Added comprehensive unit tests in `agent/tests/quantlib/test_options.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,7 +614,7 @@ Intraday bars: 1m / 5m / 15m / 30m / 1H / 4H / 1D. 15 metrics + benchmark compar
 </details>
 
 <details>
-<summary><b>Quant Library</b> <sub>286 tested functions across 19 modules, callable from every transport</sub></summary>
+<summary><b>Quant Library</b> <sub>287 tested functions across 19 modules, callable from every transport</sub></summary>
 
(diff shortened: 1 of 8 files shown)


he-yufeng (commented):
This one is a real implementation and the finite differences are correct where the price is smooth. The hard case for barrier Greeks is near the barrier, and that is where this breaks: with dS = 1e-4 * S, any spot within that distance of H puts one stencil leg across the barrier, where the knock-out value collapses to the rebate, so delta and gamma come out wrong exactly where barrier Greeks matter most. The test only covers the far-from-barrier case, which is the easy one.

Could you add either a one-sided stencil when the stencil would cross H, or at minimum a documented limitation plus a near-barrier test pinning the current behavior? Happy to approve once the hard case is handled or honestly bounded.

PR #1204, opened 2026-08-24
fix(quantlib): support multi-segment test sets and guard empty training partitions in CV

santhreal:
### Summary
Fixes combinatorial purged cross-validation splits where holding out multiple non-contiguous test blocks caused purge & embargo to treat the entire span as test, purging all training observations. Also guards against empty training folds under heavy embargo.

### Changes
- Updated `_apply_purge_and_embargo` and `detect_boundary_leakage` in `agent/src/quantlib/crossvalidation.py` to identify and process individual contiguous test segments.
- Added explicit guard raising `ValueError` when purge and embargo eliminates all training observations in a fold.
- Added regression tests in `agent/tests/quantlib/test_crossvalidation.py`.

### DIFF
--- a/agent/src/quantlib/crossvalidation.py
+++ b/agent/src/quantlib/crossvalidation.py
@@ -178,27 +178,32 @@ def _apply_purge_and_embargo(
     n = label_ends.size
     starts = np.arange(n)
     test_positions = np.flatnonzero(test_mask)
-    first_test, last_test = int(test_positions[0]), int(test_positions[-1])
-
-    # The test set's own footprint runs from its first observation to the last
-    # index any of its labels reaches -- not merely to its last row.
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Verified this one end to end, and it is a real fix. Reproduction on current main with two non-contiguous test blocks ({2} and {9}) and one-period labels: the span logic purges 8 of 12 observations, wiping the gap training points 4..7; this branch purges only the 4 adjacent ones, which is the correct semantics. The contiguous-segment split in _apply_purge_and_embargo is sound, the empty-test-mask early return fixes a latent index-out-of-range on the old first_test lookup, and the empty-training-fold guard is a good addition. test_crossvalidation.py passes (36) on the branch.

Issue #1205, opened 2026-08-24
[Feature] UK equities (LSE .L) market support in market data + backtest routing


cgycorey:
## Feature request: UK market support

The loader layer and MCP `get_market_data` support US, HK, India, Korea, Canada, Vietnam, and crypto — but not UK equities. LSE (`.L`) symbols currently resolve to `_unresolved`.

### Current failure (the motivation)

```jsonc
// get_market_data({codes: ["VOD.L"], ...}) with source=auto
{ "_unresolved": ["VOD.L"] }
```

Root cause: `_SOURCE_PATTERNS` has no `.L` rule, so UK symbols fall through to the `tushare` default. `_detect_market` also defaults them to `a_share`. The symbol is then routed down the China fallback chain (tencent → mootdx → eastmoney → baostock → akshare → tushare → local), which cannot serve LSE — several seconds of network attempts later it surfaces `_unresolved`, indistinguishable from a genuinely dead ticker.

Forcing `source=yahoo` works today:

```text
VOD.L  → OK 15 bars (close 116.60, 2026-08-04)
SHEL.L → OK 15 bars
HSBA.L → OK 15 bars
```

### Proposal (working fix, locally verified)

Treat UK as a first-class market with the same parity as Canada/US:

1. `src/market_data.py` — `_SOURCE_PATTERNS`: add `^[A-Z0-9&.\-]+\.L$` → `yahoo` (mirrors Canada `.TO/.V`); yahoo_loader `_is_supported` gate: add `.L`/`.IL`
2. `backtest/engines/_market_hooks.py` — `_MARKET_PATTERNS`: add `.L` → `uk_equity`; `_MARKET_CURRENCY`: `uk_equity: GBP` (currently returns `UNKNOWN:uk_equity`)
3. `backtest/loaders/registry.py` — `FALLBACK_CHAINS`: add `"uk_equity": ["yahoo", "yfinance", "local"]` (mirrors `ca_equity`)
4. `backtest/runner.py` — `_MARKET_TO_SOURCE`: add `uk_equity: yahoo` (currently backtests silently fall to the `tushare` engine); composite dispatch → `GlobalEquityEngine(market="uk")`
[...]

Issue #1062, opened 2026-08-11
[Bug] A-share volume unit inconsistency across fallback chain (lots vs shares, 100x) and missing unit metadata


shadowinlife:
## Description

**What happened:**

A-share loaders return `volume` in each source's **native unit with no normalization**, and the units differ across the fallback chain by a factor of exactly 100:

| Loader (A-share fallback chain position) | Native volume unit |
|---|---|
| `tencent` (chain head) | lots (手, 1 lot = 100 shares) |
| `mootdx` | lots (手) |
| `eastmoney` | lots (手) |
| `baostock` | **shares (股)** |
| `tushare` | lots (手) |

Because `fetch_market_data` walks the chain by availability, the **same query can return volume values differing by 100× between two runs** — e.g. when `tencent` is transiently throttled and the request degrades to `baostock`. Neither the data envelope, `_provenance`, nor the `get_market_data` tool description carries any unit metadata, so:

1. The built-in agent and external MCP clients (Claude Desktop, opencode, Cursor, ...) have no way to know which unit a returned `volume` is in;
2. A batch query where some symbols are served by `tencent` and others degrade to `baostock` produces a **mixed-unit dataset**, silently corrupting any cross-symbol volume analysis (distribution stats, liquidity screens, volume-based factors);
3. Downstream consumers (backtest signal engines using volume, swarm workers, scheduled research) inherit the same ambiguity.

**What I expected:**

`volume` normalized to a single canonical unit at the loader boundary (consistent with the project's existing centralized loader-boundary guards, e.g. the OHLC sanity check), and/or at minimum explicit unit metadata in `_provenance` plus the `get_market_data` tool description so consumers can interpret the values correctly.

## Steps to Reproduce

Full end-to-end reproduction on `main` @ `1bf1d8b4` (v0.1.13+):

**Environment:** macOS 15 (arm64), Python 3.12, no API keys required (both sources are keyless).

**Step 1 — install the optional baostock dependency:**

```bash
pip install baostock
```

**Step 2 — fetch the same symbol/date through both loaders:**

```python
import sys; sys.path.insert(0, "agent")
from backtest.loaders.tencent_loader import DataLoader as Tencent
from backtest.loaders.baostock_loader import DataLoader as Baostock

code, start, end = "600519.SH", "2026-07-31", "2026-07-31"

t = Tencent().fetch([code], start, end)
b = Baostock().fetch([code], start, end)

[...]

PR #1206, opened 2026-08-24
feat(market-data): UK (LSE .L) equity market support

cgycorey:
## Review fix: yfinance currency (91aa4015)

**Finding**: the yfinance fallback inferred currency from the `.L` suffix and ÷100'd every LSE line — but `.L` is not uniformly GBp. Verified against Yahoo `get_history_metadata()['currency']` (2026-08-29): VOD.L=GBp, HSBA.L=GBp, IUSA.L=GBp, SGLN.L=GBp, **VUSA.L=GBP, VUSD.L=USD**. A backtest via the yfinance path would corrupt every absolute quantity (initial-capital sizing, lot rounding, min-notional, cross-market composite weights) while returns survive.

**Fix**: the yfinance path now reads Yahoo's declared currency (`Ticker.history_metadata['currency']`, same field the direct-Yahoo path already passes to `scale_pence_to_currency`) and scales **only** on GBp/p. A missing currency fails closed — "do not scale", never "assume pence" from the suffix. GBp names (VOD.L) still ÷100.

**Tests** (in `test_yfinance_uk_scale.py`): GBP-quoted `.L` (VUSA.L) unscaled; USD-quoted `.L` (VUSD.L) unscaled; absent currency unscaled (fail-closed); GBp still scaled.

Branch rebased onto `origin/main` (`e90b6c6c`); 16+1 commits.

## References

- **Issue**: #1205 — [Feature] UK equities (LSE .L) market support
- **Tax rule sources**: gov.uk "Tax on buying shares" + HMRC Stamp Taxes manual STSM014030 (FA86/S99(13) rounding), STSM041270 (AIM exemption), STSM101060 (ETF exemption)

## Scope

UK LSE (`.L`) equity market support per #1205. Full parity: routing, classification, settlement currency, price units, engine dispatch, statutory SDRT, plus the journal/shadow and financial-statements downstream consumers. (Irish ISE `.IL` explicitly out of scope.)

## What this adds

UK equities (LSE `.L`) currently fall through `_SOURCE_PATTERNS` to the `tushare` default, get classified `a_share`, and die on the China fallback chain with a misleading `_unresolved` (several seconds of network attempts, indistinguishable from a dead ticker). This makes UK a first-class market with full parity to Canada/US:

1. **Routing** — `_SOURCE_PATTERNS`: `.L` → `yahoo`; `FALLBACK_CHAINS`: `uk_equity: [yahoo, yfinance, local]`
2. **Classification** — `_MARKET_PATTERNS` → `uk_equity`; `_detect_submarket` → `uk`
3. **Settlement** — `code_currency` → `GBP` (was `UNKNOWN:uk_equity`)
4. **Engine** — composite/runner dispatch → `GlobalEquityEngine(market="uk")`; never silent CryptoEngine fallthrough (2x set literals in runner, + `_MARKET_TO_SOURCE`)
5. **SDRT cost model** — 0.5% Stamp Duty Reserve Tax on purchases, penny rounding with exact-1/2p rounds-up (FA86/S99(13)), `Decimal ROUND_HALF_UP`; config key `uk_stamp_tax`; trade-side detection is is_open-aware (opens: direction>0=buy; closes: -direction — closing a long is a sale, covering a short is a purchase, matching base.py's position-side close path); validated against HMRC manual worked examples (13.4547->13.45, 13.455->13.46)
6. **Volume units** — yahoo/yfinance loaders declare `uk_equity: shares` (was null/undeclared, breaking the #1065/#1062 provenance contract)
7. **Price units** — Yahoo serves LSE names in **pence (GBp)**: VOD.L was ~117p, HSBA.L ~1500p; the engine priced a GBP 1.17 share at GBP 117 (100x notional error on positions and SDRT). Normalize OHLC /100 at the loader boundary (chart-meta `currency == "GBp"` on the Yahoo path, suffix rule on the yfinance fallback) so every consumer sees GBP values matching `code_currency`; provenance declares `currency_conversion: "GBp->GBP (/100)"`; loader cache version bumped v4->v5 (pre-normalization parquet never served again, #1062 precedent)
8. **Journal/shadow consumers** — `_infer_market_from_symbol`: `.L` → `"uk"` (was `"other"`, silently degrading shadow-analysis price features to NaN); shadow `_MARKET_KEY_MAP`/labels map `uk` → `uk_equity`, `"UK equity"`
9. **Financial statements** — \`.L\` routed to Yahoo quoteSummary in \`get_financial_statements\` (income/balance/cashflow annual histories + financialData/defaultKeyStatistics indicators, \`{raw,...}\` flattened to numbers); was rejected with a suffix error. \`get_fundamentals\` stays US-only by contract (SEC XBRL PIT panels — Yahoo has no filed-date PIT history), failing loud for UK

## Verified

- Live: `VOD.L` close now **GBP 1.1715** (was 117.15), `HSBA.L` **GBP 15.184** (was 1514.6) — yahoo direct
- Economics: 1,000 sh VOD.L -> GBP 1,171.5 notional, SDRT GBP 5.86 (was 585.75) — the 100x unit bug is gone end to end
- 386 affected tests pass locally (UK suites: yahoo client/loader, yfinance scale, market-data provenance, support + backtest smoke, registry, journal, shadow, financial statements, stock profile; plus upstream settings/source-order suites); **CI green on all 9 checks** (Python 3.11 + 3.14 tests, package, hash locks, Windows lifecycles)

## Known scope notes

- Fee model: SDRT is the only exchange-mandated UK cost modeled; broker commissions are caller-config (same as US zero-commission default)
- SDRT exemptions (UCITS ETFs, eligible AIM shares, gilts, new issues) are the caller's concern, not per-symbol detection
- **Scope note**: Yahoo's Irish ISE surface is `.IR`/EUR (BIRG.IR serves EUR data); routing it requires a separate market class (EUR currency, unverified Irish tax rules) — intentionally out of scope. This PR covers LSE `.L` only

## Follow-ups (not included)

- Live market-open specs (`MARKET_SPECS` has no London session; missing market fails loud, never silent)
- Symbol-search UK suffix hints


warren618:
Thanks — the UK work itself reads well, and the parts I checked (SDRT on the purchase side, GBp→GBP normalisation at the loader boundary, `.L` routing through the fallback chain, statements / profile / journal) are the right shape.

I can't merge it as it stands, because it isn't only UK support. The branch is up to date with `main` at `80ffdda4`, which already carries #1232 and #1233 — and it still changes `src/live/runtime/` by +502/−54, a second implementation of the halt-sweep and flatten-latch work that landed on 2026-08-27:

- `_halt_episode` goes from "the per-broker sentinel wins" to "the newest sentinel wins", and the latch record accumulates an `episodes` list instead of binding one.
- `mark_sweep_fired` moves from *before* the sweep (`runner.py:585` on `main`) to *after* it, gated on a new `side_effects_attempted` flag.

The second one changes the kill switch's no-retry rule. It may well be the right change, but it cannot be reviewed inside a market-data PR, and the diff that would carry it is titled "UK equity market support".

**The underlying point is legitimate and I want the fix.** I reproduced both gaps against current `main`:

1. A broker read that returns an error envelope (`{"status": "error", ...}`) is iterated as a mapping, so the sweep dies with `AttributeError: 'str' object has no attribute 'get'`. #1232 taught the *submit* responses about envelopes; the reads never learned.
2. Because `mark_sweep_fired` runs before `flatten_and_cancel`, a sweep whose reads both fail still latches — so the kill switch's action is silently skipped for that episode, and neither a later tick nor a restart retries it.

So: please rebase onto `main` and drop `10a438e6`, `bd20789c`, `0ea29b36` and `bd93142` from this branch, leaving the UK commits. Send those four as their own PR against `main`, stating those two behaviours as the thing being fixed with a test for each, and I'll review it promptly. Splitting also means the UK support isn't held behind a live-safety review.

warren618:
Thank you for doing the split — the four live commits are gone, the scope is LSE-only, and CI is 9/9. I also want to note the `_LOADER_CACHE_VERSION` bump to v5: pre-normalization pence entries becoming unreachable is exactly right and easy to forget.

I cannot merge it yet, because the yfinance path infers the currency from the ticker suffix:

```python
if is_gbp_pence_symbol(symbol):          # ^[A-Z0-9&.\-]+\.L$
    normalized, _ = scale_pence_to_currency(normalized, "GBp")
```

`.L` is not uniformly GBp. Live Yahoo `get_history_metadata()['currency']`, checked 2026-08-29:

| ticker | currency |
|---|---|
| VOD.L | GBp |
| HSBA.L | GBp |
| IUSA.L | GBp |
| SGLN.L | GBp |
| **VUSA.L** | **GBP** |
| **VUSD.L** | **USD** |

So a backtest on `VUSA.L` or `VUSD.L` through the yfinance fallback gets every OHLC price divided by 100. Returns survive, since the whole book is scaled together, but every absolute quantity does not — initial-capital sizing, lot rounding, min-notional checks, and any cross-market composite run that weights this sleeve against a correctly-scaled one.

The direct-Yahoo path in this same PR already does the right thing: it passes the chart meta's real `currency` into `scale_pence_to_currency`, so a GBP or USD line falls through untouched. yfinance exposes the same field. Please read the declared currency there too, and fail closed when it is absent rather than falling back to the suffix — a missing currency should mean "do not scale", not "assume pence".

Worth a test with a `.L` symbol whose source currency is `GBP` and one whose currency is `USD`, both asserting the prices come through unscaled. Everything else in the PR I am happy with.

warren618:
Follow-up review found a second-order currency issue in my earlier guidance: allowing a USD-quoted `.L` line through unscaled is still unsafe because the downstream `uk_equity` accounting contract is statically GBP. The follow-up now treats `.L` only as a venue marker: declared GBp/p is normalized to GBP, declared GBP passes unchanged, and USD/other/missing currency is omitted before accounting. Conversion provenance is per symbol and survives loader-cache round trips; cache version is bumped so older unsafe `.L` entries cannot be served. The follow-up also wires UK whole-share flooring and `slippage_uk`, and strengthens the smoke tests so they use loader-normalized GBP values and actually assert integral quantities. Verification: 627 related tests passed; Ruff and compilation checks passed.

Issue #1207, opened 2026-08-24
[Roadmap] Financial correctness hardening: live gates, backtest engines, shadow-account chain


he-yufeng:
I spent this week going through the money paths end to end: live mandate gates, backtest engines, data loaders, and the shadow-account chain. Three read-only audit passes, then I re-verified every finding below line by line against current main. Filing this as the tracking roadmap. I will send the fixes as a series of small test-backed PRs, one per finding, each referencing this issue.

## Verified correct, please do not "fix" these

- Daily engine timing: signals are `shift(1)` per symbol calendar and fill at the next bar open; sizing uses open equity; optimizers read strictly past data. No lookahead in the daily path.
- `BaseEngine` accounting: capital and equity stay consistent across open / add / reduce / close in both directions, with proportional margin and commission release.
- A-share single engine: T+1, no shorting, limit-band checks use the historical base price only, stamp duty sell-side only, commission floor correct.
- strict perpetual path: funding sign, settlement dedup, maintenance-margin brackets, adverse-extrema liquidation. Good shape.
- `quantlib` (XIRR, TWR, VaR conventions, waterfalls) and `backtest/metrics.py`: zero/negative prices and empty series are deliberately handled.
- Slippage direction, tick-grid rounding toward the adverse side (KR/VN), tax single vs double-sided rules (CN/HK/IN/KR/VN), PIT fundamentals at daily granularity.
- Live idempotency: deterministic client order ids, no-retry writes, cancel audit. `expires_at` handling is consistently UTC and fail-closed.
- `scheduled_research`: cron evaluation (DST gaps, fold), interval advance, backoff cap, verdict `previous` embedding. No numeric hazards.

## Phase 0: live gate fail-opens (real money at stake, small diffs)

[...]

PR #1209, opened 2026-08-24
fix(alpaca): sign position quantity by side for the mandate gate

he-yufeng:
## Summary

- Alpaca short positions now carry a signed quantity (`qty` negated when `side` is short), matching what the mandate gate reads.

## Why

Part of #1207 Phase 0. Alpaca reports `qty` as an absolute magnitude with direction in `side`, but `_position_signed_market_value` reads quantity as signed (its own fixtures use negative quantities for shorts). An Alpaca short therefore booked as positive exposure: each additional sell moved the computed exposure toward zero, so `max_total_exposure_usd` and `max_leverage` stopped binding on the short side, and a buy-to-cover read as adding long.

## Changes

- `connectors/alpaca/sdk.py`: negate `qty` when the position side is short; longs pass through untouched.
- New connector tests pinning the sign, plus one proving the gate reads an Alpaca-shaped short row as negative exposure.

## Test Plan

- [x] Existing tests pass (`pytest tests/test_alpaca_position_sign.py tests/test_alpaca_tap_routing.py tests/test_mandate_enforcement.py -q`, 46 passed)
- [x] New tests added (5, mock-only)
- [ ] Tested manually: not run against a live Alpaca account; the mapping is covered by mock tests.

Risk and rollback: the positions payload now shows negative quantity for shorts instead of a bare magnitude; that is the convention the gate and the tiger connector already share. Rollback is reverting the one mapping branch.

## Checklist

- [x] Documentation updated (if user-facing change) — internal mapping fix, nothing user-facing to document

### DIFF
--- a/agent/src/trading/connectors/alpaca/sdk.py
+++ b/agent/src/trading/connectors/alpaca/sdk.py
@@ -847,7 +847,11 @@ def _obj_get(obj: Any, name: str, default: Any = None) -> Any:
 
 
 def _position_to_dict(item: Any) -> dict[str, Any]:
-    side = str(_obj_get(item, "side", ""))
+    # alpaca-py's Position.side is a (str, Enum) member whose str() is
+    # "PositionSide.SHORT", not "short" — read .value so the direct-SDK path
+    # emits the same lowercase token the TAP JSON path does.
+    side_raw = _obj_get(item, "side", "")
+    side = str(getattr(side_raw, "value", side_raw)).strip().lower()
     quantity = _obj_get(item, "qty")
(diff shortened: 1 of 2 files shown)


warren618:
Merged with one addition (5e394a31): on the direct-SDK path alpaca-py's `Position.side` is a `(str, Enum)` member, and `str()` of it is `"PositionSide.SHORT"`, so the `== "short"` check only fired on the TAP JSON path. The commit reads `.value` before comparing and adds an enum-typed regression test. The exposure-math analysis in the PR body was exactly right — thanks.

PR #1213, opened 2026-08-25
fix(live): own unresolved Alpaca submissions before broker writes

Elfsa-Miranda:
## Summary

- Durably record ownership of one unresolved Alpaca submission before the broker write begins.
- Block every later risk-increasing Alpaca write while that marker is unresolved or invalid.
- Clear ownership only after an exact broker acknowledgement, durable admission accounting, and durable audit.

## Motivation

An Alpaca submission can reach the broker and then lose its response because of a timeout, connector exception, empty acknowledgement, or process crash. Without durable pre-write ownership, a restart cannot safely distinguish “not submitted” from “submitted but response lost.”

This change establishes a crash-safe, fail-closed boundary around the existing production order gate. It does not claim exactly-once execution and does not retry uncertain writes.

## Changes

- Add a strict, redaction-safe, per-broker pending-action marker.
- Persist and fsync the marker before calling Alpaca.
- Reuse the existing per-broker submission lock to serialize marker creation and the broker call.
- Generate one agent-owned `client_order_id` and pass it through both direct SDK and TAP submission paths.
- Require exact echoed `client_order_id` plus a nonempty broker order ID before clearing the marker.
- Retain the marker after exceptions, errors, empty/mismatched acknowledgements, accounting failures, audit failures, or clear failures.
- Extend the existing classic audit writer with an opt-in durable-fsync requirement.
- Exercise the behavior through the real trading service/live gate path.

## Tests

- 117 focused mandate, halt, readonly, runtime, reconciliation, and order-gate tests passed.
- Verified marker-before-write ordering and audit-before-clear ordering.
- Verified marker persistence failure produces zero broker calls.
- Verified timeout/exception/restart blocks a second write.
- Verified corrupt markers fail closed.
- Verified empty broker ID, missing client ID, and mismatched client ID retain the marker.
- Verified direct SDK and TAP receive the same externally owned client ID.
- `compileall`, Ruff, and `git diff --check` passed.

## Safety

- No uncertain broker write is retried.
- Marker persistence failure guarantees zero broker submissions.
- A marker is cleared only after exact acknowledgement and durable local evidence.
- Missing, malformed, or contradictory evidence remains unresolved.
- Existing mandate, halt, risk, readonly, redaction, and broker-lock behavior remains in force.
- Tests use fakes and fixtures only; no real Broker write or credential is used.

## Out of Scope

- Exact Broker lookup after a response-lost submission.
- Fill and position recovery.
- Cancel recovery or cancel finality.
- Automatic revalidation or automatic continuation.
- Support for brokers other than Alpaca.
- Exactly-once execution claims.

## Rollback

Revert this PR to restore the previous Alpaca submission path. Before rollback, halt live trading and preserve/reconcile any existing pending-action marker against exact Broker truth. Do not delete or bypass unresolved evidence merely to complete the rollback.

### DIFF
--- a/agent/src/live/audit.py
+++ b/agent/src/live/audit.py
@@ -251,6 +251,7 @@ def write_live_action(
     event_callback: EventCallback | None = None,
     trace_writer: _TraceWriterLike | None = None,
     chain: bool = True,
+    require_durable: bool = False,
 ) -> dict[str, Any]:
     """Fan a redacted live-action record out to up to three sinks (SPEC §5).
 
@@ -290,6 +291,7 @@ def write_live_action(
             exact historical record shape and ledger file. Raises
             :class:`src.governance.ledger.LedgerCorruptionError` if that
             chain ledger's existing history is already broken.
+        require_durable: Raise when the classic ledger fsync fails.
 
     Returns:
         The redacted record dict — identical to what was written to every sink —
@@ -312,6 +314,8 @@ def write_live_action(
             os.fsync(handle.fileno())
         except OSError as exc:
             _warn_fsync_failure(exc, path)
+            if require_durable:
+                raise
 
     if chain:
         from src.governance.ledger import append_record
--- a/agent/src/live/pending_action.py
+++ b/agent/src/live/pending_action.py
@@ -1,0 +1,175 @@
+"""Crash-safe ownership marker for one unresolved broker side effect."""
+
+from __future__ import annotations
+
+import json
+import math
+import os
+import uuid
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Literal, Mapping, Self
+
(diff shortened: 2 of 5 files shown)


Elfsa-Miranda:
This is PR 1 of a three-PR safe-submit-recovery series.
Review and merge order:
1. This PR — durably own an unresolved Alpaca submission before the Broker write.
2. PR 1221 — recover the exact submission by the agent-owned client order ID and account for it idempotently.
3. PR 1222 — recover exact partial/full fills using Broker order truth and signed position evidence.
This PR is independently safe and useful if the follow-up PRs never merge:
an uncertain outcome remains durably owned and blocks every later
risk-increasing Alpaca write. It does not depend on a later PR to prevent a
duplicate submission.
The follow-up PR links will be added here after they are opened.

PR #1226, opened 2026-08-25
fix(backtest/forex): treat metals as metals (pip size, lot size, spreads)

P1Piyush:
## Problem

`XAUUSD` normalises to `XAU/USD` and then goes down the generic FX path, which assumes a `0.0001` pip and a 100,000-unit standard lot. Metals are neither. Three things follow on any metals backtest:

**1. Cost is ~1,460x too small.** `_pip_value()` checks only whether the *quote* is JPY, so `XAU/USD` returns `0.0001`. `XAU/USD` is also absent from `_SPREAD_PIPS`, so it takes `_DEFAULT_SPREAD_PIPS = 2.0`. The resulting charge:

| | pip | spread | per side | round trip |
|---|---|---|---|---|
| before | $0.0001 | 2.0 | $0.00013 | **$0.00026** |
| after | $0.10 | 3.2 | $0.19 | **$0.38** |

A gold backtest is effectively frictionless today. That matters more than it sounds: on M1 gold strategies the difference between frictionless and costed is routinely the difference between break-even and clearly negative, so the engine will report an edge wherever none exists.

**2. Every realistic gold position rounds to zero.** `round_size()` rounds to the nearest 1,000 units. Position size is in ounces for metals, so the smallest expressible gold position becomes 1,000 oz (~$4M notional at current prices) and anything under that silently becomes `0`.

**3. Swap is understated 1,000x.** `calc_forex_swap` computes `lots = pos.size / lot_size` with `lot_size = 100_000`; a gold lot is 100 oz.

## Change

Adds `_METAL_SPECS` mapping each metal to `(pip size, units per standard lot)` for XAU/XAG/XPT/XPD, and routes `_pip_value()`, `round_size()` and the swap call through it. Metal spreads are added to `_SPREAD_PIPS`, expressed in that metal's own pips.

The `XAU/USD` figure of 3.2 pips ($0.32) is the **median measured from Dukascopy tick data, 2017–2024**. The other three are typical retail quotes and are commented as such — anyone with better data should override via the existing `spread_pips_override`.

**FX behaviour is unchanged**: same pip sizes, same 1,000-unit rounding, same swap lots. `round_size()` reads `self._active_symbol`, following the existing precedent in `apply_slippage()`, and `_active_symbol` is set before every `round_size()` call site in `base.py`.

## Tests

13 new cases in `TestMetals`, including explicit regressions for both bugs (`test_gold_pip_is_not_fx_pip`, `test_gold_position_under_1000_oz_survives_rounding`) and a guard that FX rounding is untouched.

```
$ PYTHONPATH=agent pytest agent/tests/test_forex_engine.py agent/tests/test_execution_causality.py
56 passed
```

Those are the two test files that reference the forex engine. I could not run the full suite locally (unrelated collection errors from missing optional deps in my environment), so CI should be the judge on the rest.

### DIFF
--- a/agent/backtest/engines/forex.py
+++ b/agent/backtest/engines/forex.py
@@ -4,7 +4,7 @@ Market rules:
   - 24x5 (Mon Sydney open to Fri NYC close)
   - Spread replaces explicit commission (bid-ask)
   - Leverage: 50:1 to 500:1 (configurable)
-  - Standard lot = 100,000 units of base currency
+  - Standard lot = 100,000 units of base currency (metals differ: see _METAL_SPECS)
   - Swap (overnight rollover interest) at daily close
   - No price limits, no restrictions on direction
   - PnL in quote currency (converted via exit price for cross pairs)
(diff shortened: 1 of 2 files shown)


he-yufeng (approved):
Verified this locally before approving. On current main, `XAU/USD` gets pip size 0.0001 (three orders under the metal's own pip) and a 5 oz gold position rounds to zero outright, so gold backtests today are both frictionless and lossy on sizing. On this branch the same probes give 0.1 and 5.0, and the full forex suite passes (44 tests).

The segment math in `_metal_base`/`_lot_units` reads right, FX paths are unchanged and pinned, and the spread figure lands in a realistic range for retail spot gold. One non-blocking note: for metals, `_lot_units` now ignores a user-configured `lot_size` and always uses the metal spec. That is the sensible default (a "lot" of gold is 100 oz by definition), but if someone tuned lot_size to model a non-standard metals contract, this silently discards their setting. Fine as is; just calling it out.

warren618 (approved):

PR #1227, opened 2026-08-26
docs(skills): add Chinese reference (SKILL.zh-CN.md) alongside English

sambazhu:
Local-only bilingual support: each ai-berkshire skill now has a SKILL.zh-CN.md (Chinese methodology reference) next to the English SKILL.md (which the agent loads). thesis-tracker.zh-CN.md carries the macro red line (demographic / consumption / generational) in Chinese.

Not in the upstream PR — PR stays English per review; this is for the local Chinese user's reference.

## Summary

-

## Changes

-

## Test Plan

## Checklist


warren618:
Before reviewing: the PR body reads like a note for a local-only branch — "Not in the upstream PR — PR stays English per review; this is for the local Chinese user's reference" — yet it is opened against upstream, so I want to check what you intended. If the upstream submission is deliberate, two things to know: the agent only ever loads `SKILL.md`, so `SKILL.zh-CN.md` files would live in the tree as reference documents no code path reads; and the repo's language convention so far is English skills/docs with i18n confined to UI strings. That is a policy question we would need to settle before merging the first per-language skill doc, not an objection to the translations themselves. If it was meant to stay local, just close it — no harm done. If it was meant for upstream, please update the body to describe the actual proposal and we will take the policy question from there.

warren618:
Closing this, based on your own note that it is not intended for upstream. If I read that wrong, say so and I will reopen.

For what it is worth: a Chinese `SKILL.zh-CN.md` beside each English one is not something I want to carry, because the two will drift and there is no test that can catch a translation going stale the way the locale-parity suite catches a missing UI string. The skills are read by the model, which handles English fine and answers in the user's language already.

Issue #1030, opened 2026-08-09
[Feature] Read-only Binance USD-M Shadow Account reconciliation (planning)


honginp:
## Problem or Motivation

#1019 completed strict historical position resizing and closed the remaining backtest work tracked by #952. The historical engine can now model isolated and cross-margin state, funding, liquidation, and same-direction scaling, but it still has no exchange-grounded observation layer.

The next gap is a safe way to capture a Binance USD-M account observation and compare it with the same local accounting/risk contracts used by backtest, shadow simulation, and what-if analysis. The result should be drift evidence only: exchange data must never silently overwrite local model state or trigger an account action.

This issue proposes two explicit roles:

- `CrossMarginRiskModel`: deterministic local model used by historical backtest, shadow simulation, and what-if analysis.
- `BinanceAccountSnapshot`: immutable, timestamped representation of read-only values reported by Binance USD-M, normalized for reconciliation with the local model.

### Locked scope

- Binance USD-M linear `*/USDT` perpetuals.
- Isolated and cross margin.
- One-way position mode.
- Single-asset USDT collateral model.
- Read-only account observation and deterministic drift reporting.
- Default-off live adapter; fixture-only CI.

### Core contracts

`BinanceAccountSnapshot` should be frozen and contain:

- observation/server timestamps and source/provenance metadata;
- wallet/account-level balances reported by the exchange;
- normalized per-symbol signed quantity, entry price, leverage, margin mode, isolated margin where applicable, and exchange-reported unrealized/initial/maintenance amounts;
- canonical USD-M symbols;
- completeness, staleness, and fidelity flags;
- schema/config version or content hash sufficient to reproduce reconciliation assumptions.

A `ReconciliationReport` should contain:

- account-level field deltas;
- per-symbol field deltas;
- missing and unexpected positions;
- tolerance configuration/version;
- stale, incomplete, and unsupported-state findings;
- deterministic severity/status and machine-readable evidence.

The exchange snapshot is evidence, not a second accounting ledger. It must not mutate `AccountState`, automatically correct local state, or be treated as proof that the historical model exactly reproduces Binance's liquidation engine.

### Data flow

1. Read only the allowlisted USD-M account/position observations required by the snapshot.
2. Validate observation timestamps, coherence, completeness, and staleness.
3. Normalize supported isolated/cross one-way USD-M state into `BinanceAccountSnapshot`.
4. Evaluate the local model at the same observation boundary using versioned maintenance/risk assumptions.
[...]

PR #1230, opened 2026-08-26
feat(backtest): add Binance USD-M drift evidence artifacts

honginp:
## Summary

- add deterministic Binance USD-M drift evidence for complete cross and isolated account snapshots
- fail closed for incomplete, unsupported, stale, or unsupported-provenance observations
- append strict-JSON audit records and atomically replace a concise latest summary
- preserve tolerance, source/schema/config provenance, comparison counts, and fidelity flags

## Why

Shadow 1 supplied immutable reconciliation contracts and Shadow 2 supplied the strict read-only observation boundary. This slice records their output as reviewable drift evidence without creating a second accounting ledger or an account action path.

Progresses #1030.

## Changes

- add a pure evidence builder over the existing reconcile_binance_account contract
- validate supported snapshot schema/profile/configuration hash before a comparison can be complete
- derive drift severity from numeric, symbol, and structural findings instead of trusting caller fields
- reject malformed or contradictory records before writing artifacts
- write artifacts only under an explicit run_dir/artifacts path: binance_drift.jsonl and binance_drift_summary.json
- add cross, isolated, rejection, tampering, strict-JSON, determinism, and import-boundary tests

## Test Plan

- 218 targeted tests passed across drift evidence, reconciliation, USD-M observation, crypto-engine regression, and write guardrails
- ruff check and ruff format --check passed for changed files
- py_compile and git diff --check passed

Full repository pytest was not run locally; CI remains authoritative for the full suite.

## Safety / Network Risk

- no network access in the evidence module or tests
- no credentials or account identifiers in artifacts
- no connector, order, alert, remediation, resize, or liquidation action path
- exchange evidence never mutates AccountState or RiskSnapshot
- CI uses synthetic fixtures only

## Fidelity Notes

- liquidation_engine_assessment is always not_assessed
- this compares reported account and position fields only
- rejected observations preserve why comparison was invalid; they do not become drift verdicts
- tolerance defaults remain explicit, versioned evidence and should be calibrated against real read-only observations separately

## Out of Scope

Live execution, alerting, dashboards, automatic correction, open orders, hedge mode, multi-assets/non-USDT collateral, portfolio margin, COIN-M, and exact Binance liquidation-engine reproduction.

## Rollback

Revert this commit. The slice has no migration and is not wired into an engine or live runtime.

### DIFF
--- a/agent/backtest/binance_shadow_evidence.py
+++ b/agent/backtest/binance_shadow_evidence.py
@@ -1,0 +1,400 @@
+"""Deterministic Binance USD-M drift artifacts with no account action path."""
+
+from __future__ import annotations
+
+from dataclasses import asdict
+import json
+import math
+from pathlib import Path
+import re
+import tempfile
+from typing import Any, Mapping
+
+import pandas as pd
+
+from backtest.binance_account_reconciliation import (
+    BinanceAccountSnapshot,
+    ReconciliationTolerance,
+    reconcile_binance_account,
+)
+from backtest.perpetual_risk import AccountState, RiskSnapshot
+
+
+EVIDENCE_SCHEMA_VERSION = "binance-usdm-drift-evidence-v1"
+SUMMARY_SCHEMA_VERSION = "binance-usdm-drift-summary-v1"
+SUPPORTED_SNAPSHOT_SCHEMA = "binance-usdm-account-observation-v1"
+SUPPORTED_SOURCE_PROFILE = "binance-live-sdk-readonly"
+REQUIRED_SNAPSHOT_FIDELITY_FLAGS = frozenset({"client_observation_time", "sequential_signed_reads"})
+_SHA256 = re.compile(r"[0-9a-f]{64}")
+_EVIDENCE_FIELDS = frozenset(
+    {
+        "schema_version",
+        "status",
+        "severity",
+        "has_drift",
+        "expected_at",
+        "observed_at",
+        "source",
+        "source_profile",
+        "snapshot_schema_version",
+        "snapshot_configuration_hash",
+        "tolerance",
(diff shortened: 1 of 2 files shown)


he-yufeng (approved):
Verified locally before approving. The new module is read-only by construction: it consumes an already-fetched `BinanceAccountSnapshot`, builds the comparison record, and writes artifacts under the run dir, so nothing in the order or account-action path is touched. The fail-closed rejections (incomplete snapshot, unsupported provenance, timestamp skew) all land as structured records instead of exceptions, which matches the trust-layer contract we want here.

Ran the new suite against this branch: 19/19 pass, and the surrounding binance/shadow/reconciliation area stays green (196 passed, 5 skipped). The validation has real teeth — drift is re-derived from the comparison rows and cross-checked against `has_drift`/`severity`, and the per-row `within_tolerance` is re-verified from the deltas, so the artifact cannot silently contradict itself. `ruff check` clean on both files.

warren618 (approved):
Verified: pure-additive slice — no non-test importer of `binance_shadow_evidence` anywhere in `agent/`, so the no-engine/no-live-runtime claim holds by construction. Imports are stdlib + pandas + the two existing backtest contracts from #1229 (no network surface, no connector, no `datetime.now` — timestamps stay caller-supplied). The validator re-derives drift from the findings and rejects records whose severity/has_drift disagree, which makes the artifacts tamper-evident rather than trust-the-caller. Writes land only under the explicit `run_dir/artifacts` (append-only JSONL + atomic summary replace, `allow_nan=False`). Ran the module's 19 tests plus the reconciliation/USD-M-readonly/shadow/runtime-reconcile set locally: 165 passed.

PR #1231, opened 2026-08-27
Feat/market source priority

sambazhu:
## Summary

- Adds per-market data-source priority: every market's fallback-chain order is now visible and reorderable from the Settings page (`Settings → Data Source Priority` card), with hot-apply — no restart needed.
- Backed by one env var per market, e.g. `MARKET_DATA_ORDER_A_SHARE=tushare,tencent,mootdx,...`, persisted to the user env file.
- **All default chains are unchanged** — this is a pure opt-in capability; without an override the behavior is byte-for-byte identical to `main`.

## Why

The hardcoded `FALLBACK_CHAINS` order is a sensible default for free-tier users (e.g. tushare sits last in `a_share` to spare free quota), but users with paid/preferred subscriptions may want a different priority per market — e.g. a tushare subscriber wants `tushare` first for A-shares. Today that requires editing code; this PR makes it a settings toggle.

## Changes

**Backend**
- `backtest/loaders/registry.py`: `MARKET_DATA_ORDER_<MARKET>` parsing/validation (value must be a permutation of the market's default chain — reordering allowed, adding/dropping sources is not; invalid values warn and keep the default), snapshot-gated `refresh_source_order_overrides()` that reassigns chains in place (dict identity preserved), applied at import time, `_ensure_registered()`, `fetch_market_data()` start, and after the Settings PUT.
- `src/market_data.py`: `source="auto"` fetches now start from the override order when one is active; explicit `source` and `fallback_chain_provider` keep priority.
- `src/api/settings_routes.py`: GET returns `source_orders` (default vs effective vs override per market); PUT validates, persists, syncs `os.environ`, and hot-applies; `null`/`[]` clears an override.
- `src/config/env_schema.py`: 12 documenting `market_data_order_*` fields.

**Frontend**
- New `SourcePrioritySettings` card: market selector, numbered reorderable list (↑/↓), Default/Custom badges, per-market reset, save-all semantics (default-equal markets send `null` to clear residue), invalid-override warning.
- `lib/api.ts` types; i18n for all 7 locales.

**Tests**
- New `test_source_order_overrides.py` (12 tests: no-op idempotence, in-place reorder visibility, invalid-value handling, snapshot gating, accessor coverage).
- `test_market_data.py` +4 (auto honors override head; explicit src still first; `local:` prefix untouched; chain-provider hook wins).
- `test_settings_api.py` +5, `test_env_schema.py` +12-field assertion, new `SourcePrioritySettings.test.tsx` (6 tests).
- `tests/conftest.py` scrubs `MARKET_DATA_ORDER_*` from the shell env so developer environments can't leak into default-chain assertions.

## Test Plan

- [x] Existing tests pass — full backend suite `11229 passed / 0 failed / 92 skipped` (clean env)
- [x] New tests added — see above; frontend `592/592` incl. 6 new component tests
- [x] All 6 CI grep gates pass (env reads go through `src.config.accessor.get_env_value`)
- [x] Manual e2e on a live server: reorder + save via Settings page → immediate effect on next `auto` fetch (provenance switched to the promoted source); quota-exhausted source falls through the chain seamlessly; reset restores default order and clears the env entry
## Summary

-

## Changes

-

## Test Plan

## Checklist


he-yufeng (commented):
Useful knob, and the permutation-only validation is the right constraint. One interaction worth thinking about before this lands: #1207 items 6 and 7 are about the chain itself, not the order. Today yahoo delivers raw OHLC while tencent/eastmoney/etc. return adjusted data, and a partial batch stops the walk down the chain. Once users can reorder sources per market, the caliber mix becomes user-facing in many more combinations: a user who puts yahoo first for us_equity gets unadjusted prices where the default chain hides that. It might be worth stamping the adjustment provenance on the result (or at least documenting the caliber caveat next to the setting) so a reordered chain does not silently change backtest numbers. Not blocking on my side, I am fixing items 6 and 7 upstream of this either way.

sambazhu:
Thanks @he-yufeng — agreed on all counts. The permutation-only constraint was chosen exactly to keep this a pure reordering knob, but you're right that reordering makes the caliber mix user-facing in many more combinations (yahoo-first us_equity being the sharpest example).

Since the caliber unification (#1207 items 6/7) is landing upstream of this, I've taken the documentation half in e132aec7: the setting card now carries an explicit caveat (all locales) that sources differ in adjustment basis — yahoo/yfinance raw OHLC vs tencent/eastmoney/tushare adjusted — so a reordered chain can change backtest numbers until provenance stamping lands. The stamp itself we'll leave to your fix to avoid duplicating the mechanism.

he-yufeng:
Good call splitting it that way: the caveat lands the user-facing part now, and the stamp belongs with the caliber fix itself. I will pick up items 6 and 7 next on my side, and this setting gets provenance for free once that lands.

warren618 (approved):
Reviewed against main and verified locally beyond the suite. Design checks: the `_DEFAULT_CHAINS` snapshot is taken before any mutation and the refresh is snapshot-gated, so an override-free process only ever reassigns content-equal copies once at import — default behavior is preserved; `is_valid_source_order` is multiset equality, so add/drop/duplicate all fail; `get_env_value` is an uncached passthrough, which is what makes the PUT's os.environ sync + refresh actually hot-apply; the three guards in `fetch_market_data` (explicit source, chain-provider hook, no-network entry points) are all in place. Ran the four backend suites (194 passed) and the new component test (6 passed); all seven locales have exact key parity. Also did a real-server open-box pass: GET shows default==effective, PUT reorder hot-applies (tushare promoted to head), a dropped-source order is refused with 400 before anything persists, and null restores the default.

Two non-blocking notes: clearing an override leaves a `MARKET_DATA_ORDER_X=` empty-value line in the dotenv rather than dropping the key (inert, just residue), and the zh-CN caliberCaveat drops BaoStock from the adjusted-sources list that en/es/de carry. Neither blocks; feel free to follow up or leave as is.

Nice touch surfacing the adjustment-basis caveat in the UI — that matches an open correctness thread on our side (#1207 finding 6).

he-yufeng:
The caliber fix you were holding the provenance stamp for is up in #1287: both Yahoo paths now serve adjusted prices, so the chain is uniform and there is no raw source left for the caveat to flag. Your settings card text stays accurate as is; the stamp itself is no longer needed on my side.

PR #1232, opened 2026-08-27
fix(live): fail closed on broker error envelopes during the halt sweep

he-yufeng:
## Summary

- The kill-switch sweep treated any non-exception broker response as success, but `MCPServerAdapter.call_tool` converts a failed remote call into a `{"status": "error"}` envelope instead of raising.
- Error envelopes now count as failures in both the cancel pass and the flatten pass: they land in `report["errors"]` and get an `error` audit outcome instead of a fake `accepted`.

## Why

Refs #1207 (Phase 0, item 3). During a halt, a flaky broker connection produced a compliant-looking audit trail while the resting order stayed live, which is exactly the failure the sweeper exists to prevent. Verified against current main: `call_tool` returns the envelope on exception (`agent/src/tools/mcp.py`), the production `_submit` wiring passes it straight through (`agent/src/api/live_routes.py`), and the sweep only branched on exceptions (`agent/src/live/runtime/flatten.py`).

## Changes

- `flatten._error_envelope_message`: recognizes the adapter's error envelope and extracts its message.
- `_cancel_resting_orders` and `_flatten_open_positions`: an error envelope is recorded in `report["errors"]` and audited with outcome `error` (the envelope itself is kept as `broker_response`), never appended to `cancelled_order_ids` / `flatten_orders_submitted`. Success payloads without the error status are untouched.

## Test Plan

- [x] New regression tests: an error envelope on cancel keeps the order out of `cancelled_order_ids`, records the failure, and writes an `order_rejected` audit carrying the envelope; same for the flatten pass.
- [x] `pytest agent/tests/test_runtime_flatten.py -q`: 10 passed.
- [x] Live safety set (`test_sdk_order_gate`, `test_mandate_enforcement`, `test_killswitch_blocks_orders`, `test_readonly_default`): 112 passed. `test_runtime_runner` + `test_api_live_runtime`: 47 passed.
- [ ] Full suite and e2e not run; change is confined to the sweep's response handling. `ruff check` clean on both files.

Risk: no new broker surface, the sweep still never retries and never touches live accounts in tests. Behavior change is fail-closed only: responses previously misread as success now surface as errors. Rollback is a plain revert.

### DIFF
--- a/agent/src/live/runtime/flatten.py
+++ b/agent/src/live/runtime/flatten.py
@@ -191,6 +191,21 @@ def _cancel_resting_orders(
                 str(exc),
             )
             continue
+        envelope_error = _error_envelope_message(response)
+        if envelope_error is not None:
+            report["errors"].append(
+                {"phase": "cancel", "order_id": order_id, "error": envelope_error}
+            )
+            _audit(
+                broker,
+                _sweep_tool(broker, "cancel_order", _CANCEL_TOOL),
+                f"cancel order {order_id}",
+                request,
+                response,
+                "error",
+                envelope_error,
+            )
+            continue
         report["cancelled_order_ids"].append(order_id)
         _audit(
             broker,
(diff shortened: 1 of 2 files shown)


warren618 (approved):
Independently verified against main: `call_tool` returns the `{"status": "error"}` envelope on both exceptions (mcp.py) and `is_error` results (`_normalize_call_tool_result`), the sweep's `_submit` wiring passes it straight through, and both sweep passes branched only on exceptions. Success payloads are pinned to `status: "ok"`, so the check cannot misread a success. Ran the flatten suite (10) plus the live safety set (169) locally; a mutation removing the envelope check fails exactly the two new regression tests. Fail-closed only, no relaxation anywhere.

PR #1233, opened 2026-08-27
fix(live): persist the flatten latch across runner restarts

he-yufeng:
## Summary

- The HALT sentinel is a file and survives restarts, but the runner's `_flatten_fired` latch did not. A restart with flatten orders still working replayed the whole sweep on the next tick: a cancel pass plus a fresh market order per position, which can flip the account from long to net short.
- The latch is now persisted next to the per-broker HALT sentinel, bound to the halt episode that caused it.

## Why

Refs #1207 (Phase 0, item 4). Verified against current main: `live/halt.py` keeps HALT as a filesystem sentinel, while `LiveRunner.__init__` resets `_flatten_fired = False` on every boot (`runtime/runner.py`), so nothing stops the replay after a restart.

## Changes

- `runtime/sweep_latch.py`: records the sweep's firing in `<runtime_root>/live/<broker>/FLATTEN_FIRED`, keyed to the episode identity (the sentinel's `tripped_at`, falling back to the file mtime for a hand-touched sentinel). Atomic write, corrupt or stale records read as not-fired.
- `runner._run_preemptive_sweep`: checks the persisted latch and marks it before firing. A fresh trip writes a new sentinel, so clearing HALT and re-tripping re-arms the sweep with no manual cleanup.

## Test Plan

- [x] New tests (`test_sweep_latch.py`, 8): same-episode suppression, clear-and-retrip re-arming, hand-touched sentinel via mtime, global-sentinel visibility, corrupt latch fallback, and two runner-level restart scenarios (no replay after restart, fires again on a new episode).
- [x] `pytest agent/tests/test_sweep_latch.py -q`: 8 passed. Runner + flatten + live safety set (`test_runtime_runner`, `test_runtime_flatten`, `test_api_live_runtime`, order gate, mandate enforcement, killswitch, readonly): 167 passed.
- [ ] Full suite and e2e not run; change is confined to the sweep latch path. `ruff check` clean on the touched files (runner.py carries two pre-existing F401s from upstream, untouched).

Risk: fail-closed direction only. The latch suppresses at most a sweep that already fired in the same halt episode; it can never block a new episode's sweep, and it is written before firing so a crashed sweep is never retried, matching the existing no-retry rule. No broker surface added. Rollback is a plain revert.

### DIFF
--- a/agent/src/live/runtime/runner.py
+++ b/agent/src/live/runtime/runner.py
@@ -46,6 +46,7 @@ from src.live.mandate.model import Mandate
 from src.live.mandate.store import load_mandate
 from src.live.runtime.flatten import flatten_and_cancel
 from src.live.runtime.jobstore import JobStore
+from src.live.runtime.sweep_latch import mark_sweep_fired, sweep_already_fired
 from src.live.runtime.liveness import write_heartbeat
 from src.live.runtime.scheduler import Job, Scheduler
 from src.live.runtime.triggers import Trigger, due_now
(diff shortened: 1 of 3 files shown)


warren618 (approved):
Independently verified against main: the HALT sentinel persists (`halt.py`) while `_flatten_fired` resets in `LiveRunner.__init__`, so a restart mid-flatten replayed the sweep — real bug. Checked the episode pairing: `read_halt(broker)` reads only that broker's sentinel (no fallback), so path/payload pairs in `_halt_episode` can't mismatch, and since the runner gates on `halt_flag_set`, a firing sweep always has a live sentinel to bind to. Latch is written before firing (crashed sweep never retried) and keyed to `tripped_at`/mtime, so a new trip always re-arms. Ran the latch suite + runner/flatten/api + live safety set locally (183 total); a mutation ignoring the episode binding fails exactly the two re-arm tests. Suppression is provably at-most-same-episode.

PR #1236, opened 2026-08-27
fix(skills): correct relative links in SKILL.md reference tables

ethanstoner:
## The bug

Seven skills write their bundled-file references with a redundant skill-name prefix. From `agent/src/skills/sec-edgar/SKILL.md`:

```markdown
[sec_edgar_client transport](sec-edgar/references/sec_edgar_client.md)
```

GitHub and the skill loader both resolve a relative link against the directory of the file that contains it, so that resolves to:

```
agent/src/skills/sec-edgar/sec-edgar/references/sec_edgar_client.md   ← does not exist
```

The file is one level up, at `agent/src/skills/sec-edgar/references/sec_edgar_client.md`.

**No file is actually missing** — all 287 affected links resolve once the prefix is dropped. It is purely a path bug, so the reference tables render as dead links on GitHub and point the agent at paths it cannot open.

Four skills already use the correct prefix-less form — `chanlun`, `ashare-pre-st-filter`, `cross-market-strategy` and `strategy-generate` — which is the convention this PR applies to the rest.

## Scope

| Skill | Broken links |
| --- | ---: |
| tushare | 231 |
| eastmoney | 16 |
| okx-market | 15 |
| investor-lenses | 12 |
| yfinance | 6 |
| sec-edgar | 5 |
| qveris | 2 |
| **total** | **287** |

Also drops the same prefix from **8 inline-code paths** in the surrounding prose (e.g. ``` `sec-edgar/scripts/sec_filings_example.py` ```), so documented paths match the links. The four already-correct skills use prefix-less paths in prose too (`strategy-generate` → ``` `code/signal_engine.py` ```).

Total diff: **295 changed lines, markdown only**, every one differing solely by the removed prefix.

## Verification

Checked with a CommonMark-correct link scan rather than a naive regex. This matters: several tushare filenames contain parentheses, e.g. `工业生产者出厂价格指数(PPI).md`. A naive `\]\(([^)]+)\)` truncates those at the first `)` and reports 33 phantom failures. CommonMark permits *balanced* parentheses in link destinations, and I confirmed against GitHub's own renderer that they resolve fine:

```
$ gh api -X POST markdown --input - <<< '{"text":"[PPI](refs/GDP(PPI).md)","mode":"markdown"}'
<p><a href="refs/GDP(PPI).md">PPI</a></p>
```

So those 33 are **not** part of this bug and are left untouched.

Scanning every relative link in every `SKILL.md` against the filesystem:

```
before:   ok=15    prefixed=287   missing=0
after:    ok=302   prefixed=0     missing=0
```

Also confirmed: `git diff --check` clean, markdown-only, no CRLF introduced, and every changed line differs from its original by exactly the removed `<skill-name>/` segment (checked programmatically, 0 exceptions).

## Risk notes

Per `AGENT_CONTRIBUTOR_GUIDE.md` — this change was prepared with AI assistance.

- **Blast radius:** documentation only. No Python, no frontend, no broker/mandate/order-gate/credential surface touched.
- **Mechanical:** the edit is a single rule (drop a leading `<skill-name>/` from a relative path), applied only where the un-prefixed target was verified to exist on disk. Any link whose stripped target did *not* exist would have been left alone; there were none.
- **Not changed:** the 33 parenthesised tushare paths discussed above, since they already work.
- Commit carries the DCO `Signed-off-by:` trailer.

Happy to split this per-skill if you would rather review tushare's 231 links separately from the other six.


warren618:
I have to turn this one down, and I want to set out why carefully, because the analysis is careful work and the problem underneath it is real.

The prefix is not redundant. `read_file` — the tool the agent uses to open these documents — roots its reads at the bundled `skills/` directory, not at the directory of the file holding the link (`src/tools/read_file_tool.py`: `skills_dir = Path(__file__).resolve().parents[1] / "skills"`). Measured on this branch:

```
sec-edgar/references/sec_edgar_client.md  ->  reads OK
references/sec_edgar_client.md            ->  File not found or path escapes workspace
```

So the prefixed form is what makes a reference reachable *by the agent*, which is the consumer these tables exist for. `tests/test_skill_reference_links.py` exists to pin exactly that convention — its docstring spells it out — and it goes from 548 passing to 546 failing on this branch. The reason that didn't surface here is that this PR's checks never ran; the check list on it is empty.

The four skills cited as already correct are the opposite of a precedent. I measured them: `chanlun` has 6 such links and `ashare-pre-st-filter` has 2, and **all 8 are unreachable to the agent today**. `cross-market-strategy` and `strategy-generate` have no links of this kind at all. Those two skills are the same bug pointing the other way — I'm fixing them separately, and widening the test to enumerate every skill instead of the hand-written five it checks now, which is why they drifted unnoticed.

**The problem you found is real, though, and I don't want it lost.** Those links genuinely 404 for a human reading the repo on GitHub, which resolves them against the containing directory. Both consumers cannot be satisfied by one relative string, so it isn't fixable in markdown: the fix belongs in `read_file`, teaching it to resolve a bare `references/...` against the skill currently being read, after which the links can drop the prefix and work in both places. That is a change I'd take. If you'd like to write it, I'll review it promptly; otherwise say so and I'll open an issue with the above so the finding stays on the board.

ethanstoner:
You're right on every point, and I checked each one rather than take it.

`read_file` roots at the bundled skills directory, `Path(__file__).resolve().parents[1] / "skills"`, and the `skills/` prefix is bound to it. `sec-edgar/references/sec_edgar_client.md` returns ok; `references/sec_edgar_client.md` returns "File not found or path escapes workspace". The prefix is load-bearing, not redundant, and my PR body's premise — that the skill loader resolves a link relative to the containing file — is simply false.

The test numbers are exact. On `80ffdda`, `agent/tests/test_skill_reference_links.py` is 548 passed. On my branch it is 546 failed, 2 passed. I did not run it, which is the actual failure here.

On the four skills I cited as precedent: `chanlun` has 6 such links and `ashare-pre-st-filter` has 2, and all 8 returned an error through `read_file` on `80ffdda` — I confirmed that by resolving each one. `cross-market-strategy` and `strategy-generate` have none at all. I asserted a convention from four skills, two of which were the same bug and two of which had no data. That is the whole mistake. I see you have since fixed those eight yourself in 18bd2fc and widened the test to glob every skill, which also closes the drift that let them sit unnoticed.

One clarification on the checks, not a disagreement. The rollup is empty and the head SHA has 0 check-runs, so there was no verdict. Two workflow runs do exist for `20ae926f` — CI and Desktop Windows, both `conclusion: action_required`, sitting behind the first-time-contributor approval gate. So they were created and never allowed to run, rather than never triggered. It makes no difference to the outcome; I should have run the suite locally.

The part worth keeping is the one you already agree with: all 295 links in the corpus are written for the agent, and I measured 0 of 295 resolving the way GitHub resolves them, so every one is a dead link for a human reader.

I took you up on the `read_file` change — it is #1252, on top of 18bd2fc. A bare `references/` or `scripts/` path resolves against the skill that owns it, as the last step after every allowed root has already missed, so nothing that resolves today changes and `run_dir` still wins. Two things refuse rather than guess: a `..` segment is rejected outright, because without it a path only had to *start* with an allowed prefix to land elsewhere and `references/../SKILL.md` matched all 90 skills; and a path two skills both own is reported with the candidates named instead of silently resolved to one. Your prefix-convention test is untouched — the one test that asserted the bare form was unreachable now asserts both forms reach the same document.

Closing this one. The fix belongs in the tool, not in 295 markdown links.

PR #1239, opened 2026-08-28
fix(risk-xray): honor bars_per_year=None in cross-market annualization

### LINKED ISSUE #1237: v0.1.14: every multi-market backtest crashes — runner passes bars_per_year=None by design, risk_xray._volatility calls math.sqrt on it
## Summary

Every **multi-market** backtest crashes on v0.1.14 with `TypeError: must be real number, not NoneType`. `runner.py` deliberately sets `bars_per_year = None` for cross-market runs, and `risk_xray._volatility` then calls `math.sqrt(ppy)` on it without a guard. The two pieces are individually reasonable and jointly fatal: nothing is misconfigured, the runner is doing exactly what its own comment says it should.

Verified against the `v0.1.14` tag via the API, not a local install.

## Trace

```
backtest/runner.py:1260   market_engine.run_backtest(..., bars_per_year=bars_per_year)
backtest/engines/composite.py:150  return super().run_backtest(config, *args, **kwargs)
backtest/engines/base.py:860       risk_xray = compute_risk_xray(close_df, basket_weights, periods_per_year=bars_per_year)
backtest/risk_xray.py:196          "annualized_vol": _finite(vol * math.sqrt(ppy)) if vol is not None else None
TypeError: must be real number, not NoneType
```

## Root cause

`runner.py:1244-1246`:

```python
market_types = {_detect_market(c) for c in codes}
if len(market_types) > 1:
    bars_per_year = None
```

The `None` is intentional — the surrounding comment reads *"Cross-market: use calendar-day annualization (bars_per_year=None)"*. It is then passed straight through (`runner.py:1257` for options, `runner.py:1260` for everything else) into `compute_risk_xray`, whose helper is annotated `ppy: int` and assumes it:

```python
def _volatility(port: pd.Series, ppy: int) -> dict[str, Any]:
    vol = float(port.std(ddof=1)) if len(port) > 1 else None
    ...
    "annualized_vol": _finite(vol * math.sqrt(ppy)) if vol is not None else None,
```

The `vol is not None` guard covers the *other* operand, which is why this reads as safe at a glance. `ppy` has no guard at all, and `downside_deviation_annualized` two lines down has the same problem.

## Reproduce

Any code list spanning two markets, e.g. `SPY.US` + `BTC-USDT`. Single-market runs are unaffected, because `len(market_types) > 1` is false and `bars_per_year` stays an int.

## Why it is not a configuration problem

There is no setting that avoids it. The `None` is chosen by the runner from the code list itself, so any user whose basket spans markets hits it on every run. `risk_xray.py` is new in v0.1.14 (absent at v0.1.11), so this is a regression rather than a long-standing gap — cross-market backtests that worked on 0.1.11 fail on 0.1.14.

## Suggested fix

Decide what annualisation means when the caller has deliberately declined to specify bars per year, and make `_volatility` honour it — either skip the annualised fields when `ppy is None`, or fall back to the calendar-day convention the runner's comment already names (365). Both `annualized_vol` and `downside_deviation_annualized` need it, and the type hint should become `Optional[int]` to match what callers actually pass.

cgycorey:
## What / why

Every **multi-market** backtest crashed on v0.1.14 with `TypeError: must be real number, not NoneType` — `runner.py` deliberately sets `bars_per_year=None` for cross-market baskets (its comment: *"Cross-market: use calendar-day annualization (bars_per_year=None)"*), and `risk_xray._volatility` called `math.sqrt(ppy)` on it without a guard. Single-market runs were unaffected. `risk_xray.py` is new in v0.1.14, so this is a regression from v0.1.11.

## Fix

**Audit of every `bars_per_year` consumer found THREE crash sites** (the issue reported one):

1. `risk_xray._volatility` (the reported trace) — `math.sqrt(ppy)` unguarded. Fixed: `_annualize_factor(ppy)` — `sqrt(ppy)` when specified, **span-derived** (observed bars per elapsed calendar year, mirroring `calc_metrics`) when `None`, keeping both `annualized_vol` and `downside_deviation_annualized` defined. `compute_risk_xray` signature widened to `int | None`.
2. `options_portfolio._calc_options_metrics` (options cross-market path, `runner.py:1257` reaches it with `None`) — `bars_per_year <= 0` / `> 0` guards and `np.sqrt` raise `TypeError` on `None` (three sites). Fixed: `None` derives the effective bpy from the observed calendar span, mirroring `metrics.calc_metrics`; 252 fallback for degenerate spans.
3. `validation.run_validation` (validation-enabled cross-market runs, `base.py:878`) — `_sharpe`'s `np.sqrt(bars_per_year)` raises `TypeError` on `None`. Fixed with the same span-derived convention at the dispatcher entry.

**One convention for `None` across all consumers** (review consistency item resolved): span-derived effective bars per year, so risk_xray's annualized_vol, calc_metrics' sharpe and validation sharpe agree in one run card — the prior fixed-365 x-ray factor sat ~18% higher on a 252-trading-day daily series; now measured ratio 0.998. Degenerate spans fall back to 252 everywhere.

Already safe (untouched): `metrics.calc_metrics` (own span-derived bpy), `attribution_core`/`quantlib.timeseries` (always fed an int from the interval). Annotations widened to `int | None` across `engines/base.py`, `options_portfolio.py`, `crypto.py`, `validation.py`, `risk_xray.py`.

## Verify

- `compute_risk_xray(..., periods_per_year=None)` — both annualized fields follow the 365 factor (zigzag fixture exercises the downside branch); explicit `252` unchanged.
- Full crash contract: `engine.run_backtest(..., bars_per_year=None)` (the exact `runner.py:1260 → composite.py:150 → base.py:860` trace) no longer raises; `risk_xray_annualized_vol` is reported.
- 28 risk-xray tests + 112 cross-suite (engines, runner, market tools) pass; ruff clean.

Closes #1237.

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -672,7 +672,7 @@ class BaseEngine(ABC):
         loader: Any,
         signal_engine: Any,
         run_dir: Path,
-        bars_per_year: int = 252,
+        bars_per_year: int | None = 252,
     ) -> Dict[str, Any]:
         """Full backtest pipeline.
 
--- a/agent/backtest/engines/crypto.py
+++ b/agent/backtest/engines/crypto.py
@@ -99,7 +99,7 @@ class CryptoEngine(BaseEngine):
         loader: Any,
         signal_engine: Any,
         run_dir: Path,
-        bars_per_year: int = 252,
+        bars_per_year: int | None = 252,
     ) -> dict[str, Any]:
         self._validate_strict_resolution(config)
         self._run_interval = str(config.get("interval", "1D"))
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -175,7 +175,7 @@ def run_options_backtest(
     loader: Any,
     engine: Any,
     run_dir: Path,
-    bars_per_year: int = 252,
+    bars_per_year: int | None = 252,
 ) -> Dict[str, Any]:
     """Options backtest entry point.
 
@@ -555,7 +555,7 @@ def _calc_options_metrics(
     equity: pd.Series,
     initial_cash: float,
     trades: List[Dict[str, Any]],
-    bars_per_year: int = 252,
+    bars_per_year: int | None = 252,
 ) -> Dict[str, Any]:
     """Calculate options backtest metrics.
 
@@ -572,6 +572,20 @@ def _calc_options_metrics(
     n = len(equity)
     equity_vals = pd.to_numeric(equity, errors="coerce").astype(float)
     path_is_finite = bool(n and np.isfinite(equity_vals.to_numpy()).all())
+
+    # Cross-market convention (runner.py passes bars_per_year=None): derive
+    # the effective bars per year from the observed calendar span, mirroring
+    # calc_metrics. Without this, every None comparison below (<= 0 / > 0)
+    # raises TypeError instead of returning metrics.
+    if bars_per_year is None:
+        if n >= 2:
+            first, last = equity_vals.index[0], equity_vals.index[-1]
+            diff = last - first
+            calendar_days = diff.days if hasattr(diff, "days") else 0
+            years = calendar_days / 365.25 if calendar_days > 0 else 1.0
+            bars_per_year = int(n / years) if years > 0 else 252
+        else:
+            bars_per_year = 252
 
     final_raw: float | None = None
     final_value: float | None = None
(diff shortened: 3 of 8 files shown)


warren618:
Merged. Good call auditing every `bars_per_year` consumer instead of only the one in the trace — I confirmed all three sites independently, and the `validation` one in particular would have surfaced later and more confusingly, since it only fires for a run with `config["validation"]` set.

One commit on top: the span derivation had become four copies (`calc_metrics` already carried one, and this PR added three more). It is now a single `metrics.effective_bars_per_year`, with `test_cross_market_annualization` pinning all four consumers to it. The point of the convention is that the Sharpe, the annualised volatility and the validation Sharpe in one run card agree; four copies of that rule would not have stayed agreeing.

PR #1241, opened 2026-08-28
fix(backtest): separate warmup from evaluation window

### LINKED ISSUE #1240: [Bug] Indicator warm-up can silently extend the evaluation window, causing “10-year” backtests to report 11 years of performance
### Description

When a strategy requires a long indicator lookback (e.g. MA200), the agent may move start_date earlier to provide warm-up history.

However, the extra warm-up period is also included in the actual backtest performance calculation.

In my case, I explicitly requested a backtest for the past 10 years, but the agent generated:

2015-08-28 ~ 2026-08-28

which is an 11-year period.

The agent explicitly explained that the first year was intended as MA200 warm-up data:

I used the same window as before (2015-08-28 ~ 2026-08-28),
with the first year used as MA200 warm-up and approximately
10 years of actual signal coverage.

However, the warm-up period was not excluded from performance evaluation.

The final report still called itself a "10-year backtest", while:

trades before the actual 10-year boundary were included;

strategy return/CAGR were calculated using the extended period;

QQQ/SPY benchmark returns were also calculated from 2015-08-28.

For a run ending on 2026-08-28, the requested evaluation period should have started around:

2016-08-28

Historical bars before that date may be needed for MA200 calculation, but they should only be used as indicator warm-up data and should not contribute to trades, equity, metrics, or benchmark performance.

This is particularly concerning because the backtest completes successfully and produces internally consistent-looking metrics, so the incorrect evaluation period can easily go unnoticed.

Expected behavior

The data-loading period and evaluation period should be independent.

For example:

data_start_date:       2015-xx-xx  # enough history for MA200
evaluation_start_date: 2016-08-28
end_date:              2026-08-28

or alternatively:

{
  "start_date": "2016-08-28",
  "end_date": "2026-08-28",
  "warmup_bars": 200
}

Warm-up bars should be available to SignalEngine.generate(), but performance metrics, trades and benchmarks should begin at the user-requested evaluation boundary.

### Steps to Reproduce

Ask Vibe-Trading to backtest a strategy containing a long-lookback indicator such as MA200 over the past 10 years.

For example:

QQQ价格下穿MA120进场10%，下穿MA200加仓到40%，
然后一直持有直到价格穿布林带上轨后缩量，
直到再次下穿布林带上轨。请帮我回测这个策略在过去10年的表现。

Allow the agent to generate the strategy and config.json.

Observe that the agent may extend start_date by approximately one year to provide MA200 warm-up history.

In my run it selected:

2015-08-28 ~ 2026-08-28

Let the backtest complete.

Inspect the final report / trades / benchmark metrics.

The report still identifies the result as a 10-year backtest, but trades and performance from the additional warm-up year are included.

In my reproduction, trades starting on:

2016-03-18
2016-06-27

were included even though they occurred before the actual 10-year boundary (2016-08-28).

### Error Logs

```shell
No exception or stack trace is produced.

This is a silent correctness issue: the run finishes successfully, but the evaluation period differs from the period requested by the user.

Relevant agent trace:

User:
"...请你帮我回测这个策略在过去10年的表现"

Agent:
"我按上次相同的窗口（2015-08-28~2026-08-28，
首年作为MA200预热，实际信号覆盖约10年）..."

The final report then used:

Backtest window: 2015-08-28 ~ 2026-08-28

while presenting the result as a 10-year backtest.
```

### Interface

Web UI (vibe-trading serve)

### LLM Provider

GLM-5.3-Flash with Max thinking

### Version

0.1.14

### Environment

Ubuntu 22.04 (ARM64) / Python 3.12

pengpengyi92:
## Summary

- add an optional `data_start_date` for indicator warmup while preserving `start_date` as the trading and evaluation boundary
- feed the full warmup window to signal generation and optimization, then clip execution artifacts to the evaluation window
- apply the same boundary to options backtests, implicit benchmarks, run cards, agent instructions, and user documentation

## Behavior

Existing configurations remain unchanged because `data_start_date` defaults to `start_date`. When it is provided, raw market data may include warmup history by design, but trades, positions, equity, benchmarks, metrics, and validation begin at `start_date`.

## Validation

- `221 passed in 3.51s` across the new boundary tests and related backtest engine, metrics, and run-card suites
- `50 passed, 1 skipped, 3 deselected in 1.49s` across engine robustness and UI service tests; the three deselected existing tests require optional local `yfinance` or `fastmcp` dependencies
- Ruff passed for all changed Python files
- `compileall` passed
- `git diff --check` passed

## Scope and risk

- no broker, live-trading, deployment, or network behavior is changed
- legacy configurations without `data_start_date` preserve their previous fetch and evaluation behavior
- raw OHLCV evidence can retain warmup rows; performance artifacts cannot

Closes #1240


warren618:
Thank you for this, and I am sorry about the timing — the same gap was implemented and landed from the issue side on 2026-08-28, a few hours before this arrived, which is why the branch now conflicts.

`main` has the boundary at `backtest/engines/base.py:149` (`evaluation_start_index`), consumed by the bar engine and the options engine. Same capability, opposite spelling: main keeps `start_date` as the *data* start and adds `warmup_bars` / `evaluation_start_date` for the evaluation boundary, where yours keeps `start_date` as the evaluation boundary and adds `data_start_date` for the data start. Both are defensible; there is no way to have both without two ways to say one thing.

Your review of it was not wasted, though. Checking your PR against what landed turned up a real defect in the merged version: the explicit-benchmark path kept `resolve_benchmark`'s `total_ret`, which spans the whole *fetched* window, while the strategy's `total_return` covers only the clipped evaluation window — so `benchmark_return` and `excess_return` were measured over a longer period than the run they were compared against. Exactly the mismatched-window error the feature exists to prevent, surviving inside the fix for it. That is now fixed in `7af8c2ea` with a regression test.

Closing this one as superseded rather than leaving it to rot. If you want the `data_start_date` spelling as an additional alias, open an issue and I will discuss it — but I would rather have one spelling than two.

PR #1242, opened 2026-08-28
fix(identity): resolve crypto pairs via active Binance connector

### LINKED ISSUE #1234: [Bug] Crypto pairs (e.g. ETH-USDT) never resolve identity: search_symbol only searches stock/ETF sources, blocking orderbook_depth/get_market_data with identity_required
## Description

When the active connector is a crypto exchange (`binance-paper-trade` profile, Binance Spot Testnet), asking the agent to check price/orderbook for a crypto pair (e.g. `ETH-USDT`) never resolves an identity, and every downstream crypto tool call (`orderbook_depth`, `get_market_data`) fails with `identity_required` / `identity_conflict`. The agent then retries indefinitely instead of failing closed - our test run went 14 steps / 5m12s with zero progress until stopped manually.

This looks like a different root cause than the ticker+name resolution bugs fixed in #1108 and #1081/#887 (both already fixed in the version we're running, v0.1.14). Those were about the resolver returning zero or ambiguous candidates for otherwise-valid stock tickers. Here, search_symbol returns candidates, but from the wrong asset universe entirely: it queries Yahoo/Eastmoney (stocks/ETFs), which has no concept of a Binance spot pair.

## Steps to Reproduce

Step A: Configure a Binance Spot Testnet connector (`connector use binance-paper-trade`), confirm balance is visible.
Step B: In the web UI (`vibe-trading serve`), ask the agent to check price/orderbook for a crypto pair and act on an existing paper position, e.g.: "Check the ETH-USDT orderbook, and if there's an open ETH position without a stop-loss/take-profit, set one with a 1:3 risk/reward."
Step C: Watch it call `search_symbol` for `"ETH-USDT"` and `"Ethereum"`.
Step D: `orderbook_depth` / `get_market_data` for `ETH-USDT` then fail repeatedly with `identity_required` / `identity_conflict`, and the agent loops without making progress.

## Evidence (from artifacts/grounding_evidence.json in the run folder)

identity.records for the "ETH-USDT" / "Ethereum" queries:

```
{
  "query": "ETH-USDT",
  "status": "conflicting",
  "candidates": [
    {"source": "yahoo", "symbol": "AETHUSDT-USD", "name": "Aave Ethereum USDT USD", "type": "cryptocurrency", "exchange": "CCC", "market": "global"},
    {"source": ["query"], "symbol": "ETH-USDT"}
  ]
}
```

```
{
  "query": "Ethereum",
  "status": "ambiguous",
  "candidates": [
    {"source": "eastmoney", "symbol": "AETH.US"},
    {"source": "eastmoney", "symbol": "ETHA.US"},
    {"source": "eastmoney", "symbol": "ETHV.US"},
    {"source": "eastmoney", "symbol": "ETHW.US"}
  ]
}
```

authorized_symbols ends up as `["ETH.US"]`, an arbitrary US ETF ticker, not the actual Binance pair, and it never locks cleanly:

```
{"tool": "orderbook_depth", "error_code": "identity_required",
 "message": "A canonical, non-conflicting identity was not locked before this assistant tool-call batch started. A resolver result from this same batch cannot be consumed."}

{"tool": "get_market_data", "error_code": "identity_conflict",
 "message": "A canonical, non-conflicting identity was not locked before this assistant tool-call batch started. A resolver result from this same batch cannot be consumed."}
```

Both errors repeated across 4 separate tool calls in one run, with no path to recovery.

## What I expected

When a crypto-exchange connector is active (or the query is an obvious exchange pair like ETH-USDT / ETHUSDT), search_symbol should resolve against that exchange's own symbol list instead of (or in addition to) the stock/ETF sources, and lock a canonical crypto identity distinct from unrelated equity/ETF tickers that happen to share the word "Ethereum".

## Notes

The very first spot buy for this same pair succeeded earlier (before this identity-lock requirement applied downstream), so this seems specific to repeat/orderbook/market-data lookups rather than the initial trade.
Version: Vibe-Trading v0.1.14 (web UI), Binance Spot Testnet connector, OpenRouter deepseek/deepseek-v4-pro.
Happy to attach the full grounding_evidence.json / run folder if useful.

pengpengyi92:
## Goal

Resolve explicit Binance spot pairs such as `ETH-USDT` through the active
connector so the grounding ledger can authorize downstream market-data and
order-book reads.

Closes #1234.

## Root cause

`search_symbol` only queried Eastmoney and Yahoo. Those providers do not own
the selected Binance profile's instrument universe, and Yahoo can return a
different near-string asset such as `AETHUSDT-USD`. The requested pair and the
unrelated result then entered grounding as conflicting identities.

## Changes

- add exact spot-pair lookup to the Binance SDK using its `load_markets()`
  catalog;
- route `trading.service.search_instruments()` to the selected Binance SDK
  profile;
- consult that connector for explicit crypto-pair queries before generic
  public search sources;
- exclude non-equal Yahoo near-string candidates from exact pair queries;
- classify `load_markets` as a read operation;
- cover connector resolution, failure fallback, service routing, and grounding
  authorization.

## Boundary and risk

- Binance spot profiles only; this does not add OKX or generic connector
  discovery.
- Exact pair spellings only (`ETH-USDT`, `ETH/USDT`, `ETHUSDT`); prose such as
  `Ethereum` is deliberately not guessed by the connector.
- `load_markets()` reads the exchange catalog. No order, mandate, balance,
  credential, or live-write behavior changes.
- A connector failure is recorded as a failed source and a Yahoo near-match is
  rejected rather than used as a fallback identity.
- Rollback is a single revert of commit `4d04441`.

## Validation

- `6 passed`: new focused connector, resolver, failure, service, and grounding
  regression tests.
- `292 passed`: related symbol-search, grounding, SDK connector, and order-book
  suites that do not require the optional MCP dependency.
- `ruff check` passed for changed source and tests; the SDK test file retains an
  unrelated pre-existing `F401`, so that file was checked with `--ignore F401`.
- `compileall` and `git diff --check` passed.
- Upstream CI: `9/9 passed`, including Python 3.11 and 3.14 test jobs,
  cross-platform hash locks, Windows lifecycle checks, and desktop packaging.

Not run locally: tests importing `src.tools.mcp` could not be collected because
the current Python environment does not have the declared optional/runtime
dependency `fastmcp>=2.14.0`; upstream CI exercised the complete project
environment successfully.

### DIFF
--- a/agent/src/tools/symbol_search_tool.py
+++ b/agent/src/tools/symbol_search_tool.py
@@ -1,7 +1,11 @@
 """Read-only symbol-search tool: resolve a name/ticker to symbols + market.
 
-Backed by three frozen, IP-throttled public-API clients so the agent never
-hits a provider un-throttled and never re-implements transport plumbing:
+Backed by the selected Binance connector for exact crypto pairs plus three
+frozen, IP-throttled public-API clients so the agent never hits a provider
+un-throttled and never re-implements transport plumbing:
+
+* The active Binance profile resolves exact spot-pair spellings against its
+  exchange market catalog. It does not guess asset names from prose.
 
 * :mod:`backtest.loaders.eastmoney_client` — Eastmoney's free suggest endpoint
   matches Chinese/English names and tickers across A-shares (.SH/.SZ/.BJ),
@@ -52,6 +56,25 @@ _EASTMONEY_SUGGEST_URL = "https://searchapi.eastmoney.com/api/suggest/get"
 # are deliberately left to the normal fan-out.
 _CANADIAN_SYMBOL_RE = re.compile(r"^[A-Z0-9&.\-]+\.(?:TO|V)\b", re.IGNORECASE)
 
+# Explicit exchange-pair spellings are not equity/name searches. Restrict the
+# quote leg to assets used by the built-in crypto connectors so an equity such
+# as ``BRK-B`` is never misclassified as a pair.
+_CRYPTO_QUOTE_ASSETS = (
+    "FDUSD",
+    "USDT",
+    "USDC",
+    "BUSD",
+    "TUSD",
+    "BTC",
+    "ETH",
+    "BNB",
+    "USD",
+)
+_CRYPTO_PAIR_RE = re.compile(
+    rf"^([A-Z0-9]{{2,15}})[-/]({'|'.join(_CRYPTO_QUOTE_ASSETS)})$",
+    re.IGNORECASE,
+)
+
 # Eastmoney market-number -> our symbol suffix. Anything else is left unmapped
 # (those candidates are skipped rather than emitted with a wrong suffix).
 _EASTMONEY_SUFFIX_BY_MARKET: Dict[str, str] = {
@@ -101,7 +124,8 @@ class SymbolSearchTool(BaseTool):
         "with their market, in the project's symbol convention (A-shares "
         "600519.SH, Hong Kong 00700.HK, U.S. AAPL.US, Canada TD.TO/PNG.V, plus "
         "crypto/index/FX from "
-        "Yahoo). Searches Eastmoney (China/HK/US names and tickers) and Yahoo "
+        "Yahoo). Exact crypto pairs are checked against the active Binance "
+        "profile; other queries search Eastmoney (China/HK/US names and tickers) and Yahoo "
         "(global) and, for U.S. equities, attaches the SEC CIK. Use this to turn "
         "an ambiguous name into a concrete symbol before calling get_market_data "
         'or get_sec_filings. Example: search_symbol(query="apple", limit=5).'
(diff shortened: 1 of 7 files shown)


warren618:
Merged. The diagnosis is right and the Yahoo near-string filter is the important half: `AETHUSDT-USD` is a different asset and must never enter the identity ledger as a rival for `ETH-USDT`.

I did change the shape, though, because routing this through the *selected* Binance connector makes a read-only identity lookup require a broker account. Measured on main before the merge, with no connector configured:

```
ETH-USDT -> AETHUSDT-USD  (Aave Ethereum USDT — the bug in #1234)
BTC-USDT -> no candidates
```

With #1242 alone and no Binance profile selected, the filter correctly drops the wrong hit but nothing replaces it, so `ETH-USDT` becomes zero candidates as well. More honest, but the user still cannot resolve a pair.

The venue catalogs are public, unauthenticated REST — the same ccxt connectivity `orderbook_depth` already uses to serve these very pairs, with no credentials anywhere. So `0ed3770f` adds a keyless fallback: an exact pair with no connector hit consults binance then okx through a single `_load_public_markets` function, isolated so tests monkeypatch it and never open a socket (the `orderbook_depth_tool._fetch_raw_book` pattern). Verified end to end with no credentials configured — both pairs now resolve to themselves, and an equity query never reaches the venue catalogs.

Your connector path stays first when a Binance profile *is* selected. Thanks for the fix and for the very clear boundary section in the description.

PR #1244, opened 2026-08-28
fix(live): halt sweep read envelopes + latch only after attempted side effects

cgycorey:
## HOLD review resolved (977b0389 — latch-coherence races)

- **Mid-sweep re-trip**: `mark_sweep_fired` re-read the sentinels after the sweep; a clear + re-trip mid-sweep (ep1→ep2) recorded ep2 as fired though no ep2 sweep ran → next runner skipped it (reproduced). Now: one coherent `halt_snapshot()` captured **before** claiming — claim, already-fired check, marks and release all bind to it; `mark_sweep_fired(broker, episodes)` never re-reads.
- **Concurrent lost-update**: the shared read-merge-write record could drop a concurrent episode's entry. Now: **per-episode latch files** (`FLATTEN_FIRED-<sha16>`, independent `O_CREAT|O_EXCL` atomic creates) — no shared record exists, so no inter-process lock is required and lost updates are impossible by construction.
- Regressions: `test_mid_sweep_retrip_records_only_swept_episode`, `test_concurrent_latch_updates_both_episodes_recorded` (both mutated-fail / fixed-pass).

## Re-review race fix (1c9c41ab)

**Claim-release episode bind**: the halt episode was resolved twice (claim time + release time in `finally`). A mid-sweep clear + re-trip (ep1 → ep2) made the release delete **ep2's** claim — even a different process's — unprotecting that process's concurrent sweep (duplicate-close window). Resolved once; regression `test_mid_sweep_retrip_does_not_release_new_episode_claim` (fails against old double-resolution).

## Review residuals resolved (a4f7459 — non-blocking A & B)

- **A — record all active halt episodes**: `mark_sweep_fired` now latches every currently tripped sentinel identity (broker + global, via a shared `_sentinel_identity` helper), not just the newest. An older global + newer broker halt: clearing the broker one no longer re-fires the sweep while the global remains tripped. Regression: `test_mark_records_both_active_episodes_older_global_newer_broker`.
- **B — double-failure latch write**: the flatten-raised branch now wraps `mark_sweep_fired` — a latch write failure during a sweep failure is logged + audited and never escapes `run_once`. Regression: `test_double_failure_latch_write_does_not_escape_run_once`.

90 live tests pass; ruff clean.

## Safety hardening (f81cb723 — independent review findings)

- **Concurrent runner race**: the sweep was check-then-act (`sweep_already_fired` → broker writes → durable marker). Two runner processes could both pass the check and duplicate closes. The whole window now runs under an exclusive disk claim (`FLATTEN_CLAIM`, `O_CREAT|O_EXCL`); a held claim (other runner active, or a crash mid-sweep) means the episode's outcome is unknowable — never re-sweep, audit for operator resolution, HALT stays tripped. Claim released in `finally` on every path.
- **Durable-latch persistence failure**: a `mark_sweep_fired` fs error after a broker write is now logged + audited explicitly ("restart may replay; verify broker state"); the claim protocol also bounds the window in which it can occur.
- **Nested broker error envelope**: `{"status":"ok","data":{"status":"error"|"ok":false}}` (transport OK, broker rejected) used to be recorded as **accepted** — order stays live while the sweep latches. Both nested shapes now detected for cancel and flatten.

Claim-recovery self-review (episode-keyed): a crash leaving an orphan claim used to block **every future episode** of the broker (a new trip would see `claim_sweep` fail and skip the kill action forever — reproduced). Claims are now episode-keyed (`FLATTEN_CLAIM-<sha256>`), so an orphan claim only blocks its own episode (the correct unknowable-outcome case) and never a later trip.

Tests: claim exclusivity + release, claim-held runner skip, orphan-claim-does-not-block-future-episode regression, claim release on success/read-failure, nested-error rejection for both phases. 89 live tests pass; ruff clean.

## Two behaviours being fixed (one per commit, tests each)

Reproduced on current main:

1. **Broker read-error envelopes are iterated as mappings.** `MCPServerAdapter.call_tool` returns `{"status": "error", ...}` instead of raising. #1232 taught the *submit* responses about envelopes; the sweep's **reads** (`read_open_orders` / `read_positions`) never learned — `AttributeError: 'str' object has no attribute 'get'` kills the kill-switch action for that episode.

   Fix (`823fc69b`): `_read_broker_state` validates every read before iteration — error envelope, exception, or non-list result → structured `report["errors"]` item (`read_open_orders` / `read_positions`), phase aborts with no side effects. Non-dict entries are skipped with an error. Successful adapter responses unwrap with pinned per-read keys (`data.orders` / `data.positions`; bare-list `data` accepted) — wrapper dicts without the pinned key are rejected (metadata never becomes fake records).

2. **`mark_sweep_fired` runs before `flatten_and_cancel`, so a sweep whose reads both fail still latches** — the kill switch's action is silently skipped for that episode; neither a later tick nor a restart retries it.

   Fix (`ebf5e189`): `flatten_and_cancel` reports `side_effects_attempted` (set the moment any cancel/close submit is tried). The runner latches iff any broker write was attempted (even when another read phase failed), keeps retrying only when nothing was attempted, and latches conservatively on a raising sweep. Reads are also normalized at the `live_routes._read` boundary (error envelopes raise; ok payloads unwrap), protecting the shared reconcile path from the same envelope class. Related, same commit: `_halt_episode` now ranks broker vs global sentinels by explicit `tripped_at` (mtime ties within one filesystem timestamp quantum let a stale per-broker latch suppress a newer global HALT — `halt_flag_set` treats the global as authoritative), and the latch record accumulates every fired episode.

## Tests

- Adapter-envelope reads (both paths, real envelope shape), non-list/invalid-entry rejection
- Restart-replay on read failure (no side effects → no latch), restart-suppression after an attempted close even with a read error (duplicate-close hazard)
- Newer-global rearm with forced-identical mtimes; nothing-to-do re-check policy; boundary normalization + envelope-raise wiring
- 82 live tests pass; ruff clean

### DIFF
--- a/agent/src/api/live_routes.py
+++ b/agent/src/api/live_routes.py
@@ -565,8 +565,42 @@ def _build_live_runner(broker: str) -> Any:
     # _live_broker_adapter is monkeypatched on host by tests
     adapter = h._live_broker_adapter(broker)
 
-    def _read(remote_tool: str):
-        return lambda: adapter.call_tool(remote_tool, {})
+    def _read(remote_tool: str, record_key: str | None = None):
+        """Wrap adapter.call_tool into a READ callable with a normalized payload.
+
+        The MCP adapter returns ``{"status": "error", ...}`` envelopes instead
+        of raising, and wraps successful values in ``{"status": "ok",
+        "data": ...}``. Normalize at the boundary so consumers (reconcile and
+        the halt sweep) receive broker records — or the balance dict — not
+        envelopes: error envelopes raise, success payloads unwrap to ``data``
+        (with the pinned ``record_key`` extracted for positions/orders record
+        reads, e.g. ``data.positions`` / ``data.orders``). Bare values pass
+        through unchanged.
+        """
+        def read() -> Any:
+            payload = adapter.call_tool(remote_tool, {})
+            if isinstance(payload, list):
+                return payload  # injected/legacy bare response
+            if not isinstance(payload, dict) or payload.get("status") != "ok":
+                if isinstance(payload, dict) and payload.get("status") == "error":
+                    raise RuntimeError(
+                        str(payload.get("error") or "broker read failed")
+                    )
+                return payload
+            data = payload.get("data")
+            if record_key is None:
+                return data
+            if isinstance(data, list):
+                return data
+            if isinstance(data, dict):
+                records = data.get(record_key)
+                if isinstance(records, list):
+                    return records
+                raise RuntimeError(
+                    f"broker read response missing {record_key!r} records"
+                )
+            raise RuntimeError("broker read response has no data payload")
+        return read
 
     def _submit(order: Dict[str, Any]) -> Dict[str, Any]:
         if order.get("action") == "cancel":
@@ -599,9 +633,9 @@ def _build_live_runner(broker: str) -> Any:
         broker,
         agent_caller=_agent_caller,
         reconcile_fn=reconcile,
-        read_positions=_read(positions_tool),
+        read_positions=_read(positions_tool, "positions"),
         read_balance=_read(balance_tool),
-        read_open_orders=_read(open_orders_tool),
+        read_open_orders=_read(open_orders_tool, "orders"),
         submit_fn=_submit,
         write_audit_fn=_audit_with_bus,
         scheduler=scheduler,
--- a/agent/src/live/runtime/flatten.py
+++ b/agent/src/live/runtime/flatten.py
@@ -403,9 +403,27 @@ def _error_envelope_message(response: Any) -> str | None:
     dict is not proof the broker accepted the request. Treating that
     envelope as success would audit a resting order as cancelled while it
     stays live, which is the failure the sweep exists to prevent.
+
+    A *nested* broker-level failure is just as dangerous: the MCP transport
+    succeeded, but the connector/broker rejected the order inside its own
+    payload (``{"status": "ok", "data": {"status": "error", ...}}`` or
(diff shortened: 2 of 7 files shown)


warren618:
Merged. This is the split I asked for on #1206 and both behaviours reproduce on main exactly as described, so thank you for separating them.

Reviewed in detail and I agree with the reasoning on all four of the hard parts: resolving the episode once so the `finally` release cannot delete a newer episode's claim, episode-keying the claim so an orphan blocks only its own episode, the nested `{"status":"ok","data":{"status":"error"}}` envelope, and latching on `side_effects_attempted` rather than on reads succeeding. 495 live/mandate/halt tests pass locally.

One follow-up pushed on top (`637791b1`). The no-side-effect branch re-runs on every halted tick — correct, that is the point, a resting order appearing after the first sweep must still be cancelled — but its audit record repeated with it. Measured on a flat halted book, five ticks:

| | broker read rounds | audit records |
|---|---|---|
| before #1244 | 1 | 5 |
| after #1244 | 5 | 10 |

One extra record per tick, forever. At a one-minute tick a channel left halted overnight appends ~1440 of them to the hash-chained ledger. It was also written as `kind="breach" outcome="error"` for "no open orders or positions to act on", which makes a completely clean outcome indistinguishable from a real breach.

The fix keeps the re-check policy unchanged: `_note_sweep_recheck` audits once per `(episode, intent)` per runner and logs every occurrence, and the benign case is now `halt_tripped`/`blocked`. I also corrected `_halted_result`'s docstring, which still claimed the sweep runs at most once per runner lifetime.

PR #1245, opened 2026-08-28
feat(backtest): surface silent order-plan rejections via an _on_plan_rejected hook

### LINKED ISSUE #1235: _plan_open_order returns None for seven different reasons — a silently unfillable position is indistinguishable from nothing-to-do
## Summary

`BaseEngine._plan_open_order` returns `None` for **seven materially different reasons** across four `return None` statements, and no caller can tell which one occurred. Several are benign ("nothing was wanted here"), but at least one — the target size rounding to zero — means *the engine wanted a position, priced it, and silently could not take it*. From the outside these are indistinguishable, so a backtest can grade a four-sleeve configuration as a three-sleeve book with nothing in the output saying so.

This is a request for observability on an existing method, not a behaviour change. Verified against the `v0.1.14` tag via the API (`agent/backtest/engines/base.py`), not a local install.

## The seven causes, all reaching the same `None`

`_plan_open_order` (`base.py:1231-1271` at v0.1.14):

| line | cause | class |
|---|---|---|
| 1250 | `direction == 0` — no target weight | nothing wanted |
| 1250 | `symbol in self.positions and not allow_existing` | already held |
| 1250 | `df is None` | no data for symbol |
| 1250 | `ts not in df.index` | no bar at this timestamp |
| 1253 | `not self.can_execute(...)` | blocked by market rule (halt, limit-up/down) |
| 1261 | `open_price == 0`, or `< 0` when not permitted | bad price data |
| 1271 | `size <= 0` after `round_size(...)` | **target too small to fill** |

The last row is the one that motivated this. For futures engines `round_size` truncates (`max(int(raw_size), 0)`), so any contract whose notional exceeds the target notional never trades. A sleeve can hold a large median target weight for the entire backtest and fill **zero** times, and the run reports a normal result — the sleeve simply isn't in it.

## Why the caller can't recover this today

The three callers (`_rebalance:1227`, `_execute_target_rebalance:1375`, `_plans:1002`) receive only `Optional[_OpenOrder]`. Reconstructing the cause means re-deriving `target_notional`, re-running `_calc_raw_size` and `round_size`, and re-checking `can_execute` — a parallel implementation of the method being observed, which drifts the moment the original changes.

The existing extension hooks don't help: `after_position_adjustment` is called only from `_execute_position_increase:1537` and `_execute_partial_reduction:1596`, both *after* an order is committed, so it is structurally blind to a plan that returned `None`. `before_rebalance_bar` / `after_rebalance_bar` receive `(timestamp, data_map, codes)` — no target weights and no planned orders.

## Suggested shape

Anything that distinguishes the classes would do. Cheapest first:

1. Return a reason alongside the order — e.g. `tuple[Optional[_OpenOrder], str]`, or a small enum/sentinel — leaving current callers free to ignore it.
2. Or a `self._on_plan_rejected(symbol, reason, ts)` no-op hook called before each `return None`, mirroring `after_position_adjustment`'s style.

Option 2 is fully backward compatible and needs no call-site changes.

## Why it's worth separating even the benign causes

`direction == 0` and `size <= 0` are opposite situations — one means nothing was wanted, the other means something was wanted and could not be had — and today they are the same value. Any consumer trying to answer "did every declared symbol actually trade?" has to treat them identically, which is precisely the question that goes unanswered when a sleeve silently never fills.

lorenzozanee:
## Summary

- Add a no-op `_on_plan_rejected(symbol, reason, timestamp)` hook to `BaseEngine`, called at every point where `_plan_open_order` returns `None`.
- Engine subclasses can now tell "nothing to do" apart from "wanted but unfillable" without re-deriving the sizing logic.

## Why

`_plan_open_order` rejects a plan for seven distinct reasons — no target weight, symbol already held, no data frame, no bar at the decision timestamp, market-rule block, invalid price, or a size rounded to zero by lot rules — and callers only see `None`. The zero-size case is the sharpest: for futures engines `round_size` truncates to whole contracts, so a sleeve whose target notional is smaller than one contract trades zero times while the run reports a normal result.

Closes #1235

## Changes

- Split the compound guard in `_plan_open_order` into labelled branches that each pass a machine-readable reason (`no_target_weight`, `already_held`, `no_data`, `no_bar`, `execution_blocked`, `invalid_price`, `zero_size`) to the new `_on_plan_rejected` hook, mirroring the style of `after_position_adjustment`.
- No behavior change: the default hook does nothing and none of the three call sites (`_execute_bars`, `_rebalance`, `_execute_target_rebalance`) are modified.

## Test Plan

- [ ] New tests added (`agent/tests/test_plan_rejected.py` — per-cause reason pinning for all 7 causes, reason distinctness, no hook call on a successful plan, default no-op side-effect check, futures whole-contract truncation, recorder-vs-no-op execution-state equality)
- [ ] Tested manually

## Checklist

- [ ] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`)
- [ ] Documentation updated (docstring documents the hook contract and all reason strings; no user-facing docs change)

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -689,6 +689,25 @@ class BaseEngine(ABC):
         """Allow engines to update risk state/evidence after a committed delta fill."""
         return None
 
+    def _on_plan_rejected(self, symbol: str, reason: str, timestamp: pd.Timestamp) -> None:
+        """Observe a silently rejected opening-order plan.
+
+        ``_plan_open_order`` returns ``None`` for several distinct reasons —
+        nothing wanted, already held, missing data, missing bar, market rule
+        block, unusable price, or a target too small to fill after lot
+        rounding. Callers only see ``None``, so this hook is the only way for
+        an engine subclass to tell "nothing to do" apart from "wanted but
+        unfillable".
+
+        Args:
+            symbol: Instrument the plan was for.
+            reason: Machine-readable cause: ``no_target_weight``,
+                ``already_held``, ``no_data``, ``no_bar``,
+                ``execution_blocked``, ``invalid_price`` or ``zero_size``.
+            timestamp: Decision bar timestamp.
+        """
+        return None
+
     def execution_open(self, bar: pd.Series) -> float:
         """Return the normal market-fill price for a bar."""
         return float(bar.get("open", bar.get("close", 0)))
@@ -1323,15 +1342,21 @@ class BaseEngine(ABC):
         """Price an opening order without mutating portfolio state."""
         self._active_symbol = symbol
         direction = 1 if target_weight > 1e-9 else (-1 if target_weight < -1e-9 else 0)
-        if (
-            direction == 0
-            or (symbol in self.positions and not allow_existing)
-            or df is None
-            or ts not in df.index
-        ):
+        if direction == 0:
+            self._on_plan_rejected(symbol, "no_target_weight", ts)
+            return None
+        if symbol in self.positions and not allow_existing:
+            self._on_plan_rejected(symbol, "already_held", ts)
+            return None
+        if df is None:
+            self._on_plan_rejected(symbol, "no_data", ts)
+            return None
+        if ts not in df.index:
+            self._on_plan_rejected(symbol, "no_bar", ts)
             return None
         bar = df.loc[ts]
         if not self.can_execute(symbol, direction, bar):
+            self._on_plan_rejected(symbol, "execution_blocked", ts)
             return None
         open_price = self.execution_open(bar)
         if require_positive_price:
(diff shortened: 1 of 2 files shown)


warren618:
Merged. The hook is exactly the shape #1235 asked for, and I checked that all three internal `_plan_open_order` call sites already guard `target_dir != 0` and the held-position case, so the two benign reasons are effectively unreachable from them and there is no per-bar cost.

One thing the hook alone does not close, pushed on top (`fb515aa7`). The reporter's actual complaint is one level up: a sleeve whose target notional never clears one lot trades zero times and *the run still reports a normal result*, so a four-sleeve configuration is graded as a three-sleeve book with nothing in the output saying so. A hook nobody overrides does not change what that user sees.

The default hook now counts `(symbol, reason)` and the metrics dict carries `unfilled_plan_rejections` plus a per-symbol breakdown, so the dropped sleeve is nameable from the run itself. Only the five causes that mean "wanted and could not be had" are counted — `no_target_weight` and `already_held` mean nothing was wanted and are not findings. The hook stays an extension point; a subclass that calls `super()` gets both.

Minor note while I was in there: `test_default_engine_has_zero_side_effects_for_every_cause` is weaker than it looks. `already_held` installs a position for `"A"` on the second iteration, so every later cause short-circuits there and only two of the seven are really exercised.

Issue #1224, opened 2026-08-25
Swarm run.json under-reports output tokens ~18-36x: streamed calls never request usage, and the fallback ignores tool-call arguments


turtle696966969696:
## Summary

`run.json` records `total_output_tokens` for a completed swarm run substantially below the true figure — defensibly ~18-36× low (basis below), not a small rounding gap.

## Root cause, confirmed by a live A/B call, same prompt through both paths

| | `chat()` — control | `stream_chat()` |
|---|---|---|
| `usage_metadata` | `{input: 28, output: 241, total: 269}` | `None` |
| content | 371 chars, 120 lines | 371 chars, 120 lines (identical) |
| `_estimate_tokens` result | in=28, out=241 | in=23, out=92 |

The streamed call returns no usage at all — `stream_options`/`include_usage`/`stream_usage` appear nowhere in the provider layer. On the OpenAI-compatible API, a streamed response only carries usage if explicitly requested; it never is here. This contradicts the docstring at `chat.py:471-473`, which states usage "accumulates via the `__add__` merge... the final aggregate carries the same shape as the non-stream path." It doesn't — there's nothing to accumulate.

**Both input and output fall back to character-count heuristics when usage is absent, not just output** — `len(json.dumps(messages))//4` for input (`worker.py:171`), `len(response.content or "") // 4` for output (`worker.py:176`). Input degrades more gracefully in practice (28 real vs 23 estimated in the A/B test above) because the message list grows with real content; output degrades far worse because the swarm worker exclusively streams (`_stream_once()`, `worker.py:675`, called from `:707`/`:722`, no `chat()` call anywhere) and `.content` excludes tool-call arguments entirely — which is nearly all of an agent's actual output in a tool-using workflow.

**Confirmed this transfers to v0.1.14, not just the 0.1.11 the live test ran against**: `_estimate_tokens` and the `usage_metadata` extraction block in `_parse_response` are byte-identical between 0.1.11 and the v0.1.14 tag. If anything, v0.1.14 sharpens the problem — it added `_text_content()` (`chat.py:35`), which keeps only `.text` blocks and explicitly drops tool-use blocks, making the exclusion of tool-call content from `.content` an intentional part of the current normalizer rather than an incidental gap.

[...]

PR #1246, opened 2026-08-28
fix(providers): self-heal temperature rejection on the OpenAI-compatible path

### LINKED ISSUE #1223: temperature sent unconditionally on the OpenAI-compatible path to Anthropic (native-path fix at v0.1.14 doesn't cover it)
## Summary

Next-generation Claude models (`claude-opus-4-8`, `claude-sonnet-5`, and per the code comment also `claude-opus-5` — `agent/src/providers/llm.py:808-809`) reject requests carrying `temperature` with HTTP 400. As of v0.1.14 this is fixed for the native Anthropic Messages API path (`provider == "anthropic"`, `llm.py:1366`) — the string-matching detector at `llm.py:827-834` catches the error, strips `temperature`, retries, and remembers the outcome per model.

I ran into this on a different, unsupported-but-functional configuration: `LANGCHAIN_PROVIDER=openai` with `OPENAI_BASE_URL` pointed at `https://api.anthropic.com/v1/`, using Anthropic's own OpenAI-compatible endpoint rather than the native Messages API. I want to flag up front that this isn't the documented way to reach Anthropic — the README lists the native Messages API as the supported provider (`README.md:742`), and `.env.example:26-30` documents `ANTHROPIC_BASE_URL` under a dedicated "Anthropic Messages API / compatible proxy" section as the intended route, including for compatible proxies. I got here because Anthropic's compat endpoint happens to work for everything else through the generic OpenAI branch, which makes the failure easy to mistake for a bug in a supported setup rather than the edge of an unsupported one.

On that branch, the generic `ChatOpenAI` kwargs dict sends `temperature` unconditionally:

```python
kwargs: dict[str, Any] = {
    "model": name,
    ...
    "temperature": temperature,      # agent/src/providers/llm.py:1418 — unconditional, unchanged at v0.1.14
```

`temperature` always carries a value — `agent/src/config/env_schema.py:141` declares it as a plain `float` with `default=0.0`, not `Optional[float]`.

Live response, verbatim:

```json
{"error": {"code": "invalid_request_error",
           "message": "`temperature` is deprecated for this model.",
           "type": "invalid_request_error",
           "param": null}}
```

Same `param: null`, same shape the native-path detector already handles by string-matching on `message` prose.

Given this isn't a supported configuration, I'm not sure whether the right fix is extending detection to the generic branch (the logic already exists two branches away and the error shape is identical) or just a docs note warning this specific combination isn't covered — happy to open a PR for whichever direction maintainers would rather have.

Side note for anyone on the native path: `ANTHROPIC_API_KEY` is read via a bare `os.getenv` call (`llm.py:1037`), not declared in `env_schema.py` — a missing/mistyped value won't be caught by `provider doctor` or preflight, only at first call.

lorenzozanee:
## Summary

- Extend the Anthropic temperature self-heal (classify the 400, retry once, remember the model) to the generic OpenAI-compatible branch (`ChatOpenAIWithReasoning`), so requests to Anthropic models through `OPENAI_BASE_URL` recover from "`temperature` is deprecated for this model." automatically.
- No change for models that accept `temperature` — the configured value keeps being sent.

## Why

Next-generation Claude models (claude-opus-4-8, claude-sonnet-5, ...) reject any request carrying `temperature` with HTTP 400. At v0.1.14 the native Anthropic Messages path detects the error, retries without the field and remembers the model, but the generic OpenAI-compatible branch still sends `temperature` unconditionally, so running `LANGCHAIN_PROVIDER=openai` with `OPENAI_BASE_URL=https://api.anthropic.com/v1/` fails on every call.

Closes #1223

## Changes

- `ChatOpenAIWithReasoning._remember_temperature_unsupported` classifies the 400 with the existing `_is_anthropic_temperature_unsupported_error` detector, logs once, and records the model in `_ANTHROPIC_TEMPERATURE_UNSUPPORTED` — the same set the native path uses.
- `_generate` / `_agenerate` retry once without the field; `_stream` / `_astream` do the same (the 400 is raised before the first chunk, so the retry cannot duplicate output). The async stream_usage fallback moves to a named `_astream_with_usage_fallback` to mirror the sync `_stream` layering — no behavior change.
- `_get_request_payload` omits `temperature` up front for remembered models, so later calls skip the doomed request.

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q` — verified for the touched components: `agent/tests/test_llm.py`, `agent/tests/test_provider_header_isolation.py`, `agent/tests/test_env_schema.py`)
- [ ] New tests added: mocked-error regression tests in `agent/tests/test_llm.py` covering invoke self-heal and the remembered-model fast path, supported-model preservation, stream self-heal, and async invoke; all fail on the pre-fix code with the verbatim 400
- [ ] Tested manually (provider-dependent behavior verified with mocked 400/SSE responses via `httpx.MockTransport`, no live credentials)

## Checklist

- [ ] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — direction raised and left open in #1223; follows the self-heal precedent from #1224 on the same file
- [ ] Documentation updated (if user-facing change) — no doc change needed; behavior now matches the native path

### DIFF
--- a/agent/src/providers/llm.py
+++ b/agent/src/providers/llm.py
@@ -524,13 +524,60 @@ if ChatOpenAI is not None:
                 self._capture(choice["message"], gen.message)
             return result
 
+        def _generate(self, *args: Any, **kwargs: Any) -> Any:
+            try:
+                return super()._generate(*args, **kwargs)
+            except Exception as exc:  # noqa: BLE001 - retried or re-raised below
+                if not self._remember_temperature_unsupported(exc):
+                    raise
+                return super()._generate(*args, **kwargs)
+
+        async def _agenerate(self, *args: Any, **kwargs: Any) -> Any:
+            try:
+                return await super()._agenerate(*args, **kwargs)
+            except Exception as exc:  # noqa: BLE001 - retried or re-raised below
+                if not self._remember_temperature_unsupported(exc):
+                    raise
+                return await super()._agenerate(*args, **kwargs)
+
+        def _remember_temperature_unsupported(self, exc: BaseException) -> bool:
+            """Record a temperature-unsupported error and report a one-shot retry.
+
+            Mirrors the native Anthropic path for the generic OpenAI-compatible
+            branch: next-generation Claude models served through
+            ``OPENAI_BASE_URL`` return HTTP 400 when `temperature` is sent
+            (issue #1223). The model is remembered in
+            ``_ANTHROPIC_TEMPERATURE_UNSUPPORTED`` so later requests omit it up
+            front.
+            """
+            if not _is_anthropic_temperature_unsupported_error(exc):
+                return False
+            model = str(self.model_name)
+            if model not in _ANTHROPIC_TEMPERATURE_UNSUPPORTED:
+                logger.info(
+                    "Model %s rejects `temperature`; retrying without it "
+                    "and omitting it for subsequent calls.",
+                    model,
+                )
+                _ANTHROPIC_TEMPERATURE_UNSUPPORTED.add(model)
+            return True
+
         def _stream(self, *args: Any, **kwargs: Any) -> Iterator[Any]:
             """Route Responses streams through the mapping-compatible adapter."""
             if self._use_responses_api({**kwargs, **self.model_kwargs}):
                 cloned = copy(self)
                 cloned.root_client = _ResponsesSyncClient(self.root_client)
-                return super(ChatOpenAIWithReasoning, cloned)._stream(*args, **kwargs)
-            return self._stream_with_usage_fallback(*args, **kwargs)
+                inner = super(ChatOpenAIWithReasoning, cloned)._stream(*args, **kwargs)
+            else:
+                inner = self._stream_with_usage_fallback(*args, **kwargs)
+            # The temperature-unsupported error is raised before the first chunk is
+            # produced, so retrying the whole stream cannot duplicate output.
+            try:
+                yield from inner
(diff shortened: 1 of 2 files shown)


warren618:
Merged, and I was wrong to decline this direction on #1223.

My argument there was that the generic OpenAI-compatible branch is "deliberately provider-agnostic" and teaching it Anthropic-specific error prose would couple it to every provider quirk it currently stays clean of. That is not true of the branch as it stands: `_STREAM_USAGE_UNSUPPORTED` in the same file is already a classify-the-error, retry-once, remember-the-model self-heal for a provider quirk, inside `ChatOpenAIWithReasoning`, landed for #1224. Same class, same file, same branch. You said as much in your checklist and you were right. Sorry for the round trip.

One hardening pushed on top (`3e27d25a`). The stream retry restarted the whole stream on the 400, with a comment that the error arrives before the first chunk so output cannot duplicate. That is true, but it is a property of the provider, not of this code — the retry now re-raises once a chunk has been emitted, so a mid-stream error matching the detector can never replay already-emitted output.

The same commit lands the two things I promised on #1223 and had not done: `ANTHROPIC_API_KEY` joins `env_schema` (it was read through a bare `os.getenv`, so `provider doctor` and preflight could not tell a missing key from a working one), and `.env.example` now states that pointing `OPENAI_BASE_URL` at Anthropic's compatibility endpoint is not a supported route.

PR #1106, opened 2026-08-15
feat(backtest): add offline USD-M account reconciliation

honginp:
## Summary

- Add immutable Binance USD-M account and position snapshot contracts.
- Compare exchange observations with existing `AccountState` and `RiskSnapshot` fields without mutating either source.
- Report deterministic numeric, structural, missing-symbol, and unexpected-symbol drift while explicitly leaving liquidation-engine validation unassessed.

## Why

Relates to #1030.

The strict historical USD-M model now has account, position, funding, liquidation, and rebalance evidence, but it has no pure contract for comparing that local state with an exchange observation. This is the maintainer-approved Shadow 1 slice: contracts and fixture-only offline reconciliation.

The existing `binance-live-sdk-readonly` connector is currently spot-shaped (`fetch_balance()` and balance-derived holdings), so this PR does not pretend it can populate USD-M leverage or margin fields. A later slice can extend the same guarded connector architecture with USD-M reads; this slice opens no connection and imports no connector.

## Changes

- Add frozen `BinancePositionSnapshot` and `BinanceAccountSnapshot` contracts with canonical `*-USDT-PERP` symbols, source/config provenance, data status, account totals, margin mode, and reported position risk fields.
- Reject spot sources, invalid margin assignments, non-finite values, duplicate symbols, naive timestamps, and unsupported snapshot status.
- Add versioned absolute/relative/timestamp reconciliation tolerances.
- Compare wallet, margin, available balance, unrealized P&L, initial margin, maintenance margin, quantity, entry, leverage, isolated collateral, and per-position risk fields.
- Report missing/extra symbols and structural margin differences deterministically.
- Use neutral report semantics: `status="comparison_complete"`, `comparison_scope="account_snapshot_fields_only"`, and `liquidation_engine_assessment="not_assessed"`.
- Add an AST safety test that rejects connector/live imports, attribute/subscript assignment, reflective mutation, dynamic execution, and import indirection in the reconciliation module.

## Test Plan

- [x] Red/green TDD: feature test initially failed because the module did not exist, then passed after implementation.
- [x] AST mutation check: a temporary `AccountState.wallet_balance` assignment made the structural test fail; removing it restored green.
[...]

PR #1248, opened 2026-08-28
feat(backtest): add USD-M tolerance calibration from recorded comparisons

lorenzozanee:
## Summary

- add deterministic tolerance calibration for Binance USD-M reconciliation from recorded comparison samples
- derive a versioned `ReconciliationTolerance` with per-field statistics and a configurable safety factor
- fail closed on empty, malformed, non-finite, or under-sampled input; coverage against the recorded set is deterministically verifiable

## Why

Slices #1106/#1229/#1230 shipped the reconciliation contracts, the read-only observation boundary, and drift evidence artifacts, but every caller must still hand-pick tolerance values. This slice binds tolerance selection to recorded observations: a calibrated tolerance provably bounds the sample set it was derived from, so drift classification stops being an unjustified guess.

Progresses #1030.

## Changes

- add a pure, offline calibration module over the existing `ReconciliationTolerance` contract: `CalibrationSample`, `CalibratedField`, `ToleranceCalibration`
- `calibrate_tolerance` groups recorded samples per field, fails closed below `min_samples_per_field`, and requires a safety factor of at least 1.0 so the result covers every recorded sample; zero-delta fields keep strictly positive floors
- `verify_tolerance_covers` mirrors the reconciler's allowed-delta rule (`max(absolute, relative * max(|local|, |exchange|))`); changes to that formula must be mirrored here
- timestamp-skew calibration is out of scope: `max_timestamp_skew_seconds` stays pinned at 0.0
- no wiring into runs or artifacts yet — calibration is standalone offline tooling; the reconciler and drift-evidence modules are unchanged

## Test Plan

## Checklist

### DIFF
--- a/agent/backtest/binance_tolerance_calibration.py
+++ b/agent/backtest/binance_tolerance_calibration.py
@@ -1,0 +1,158 @@
+"""Deterministic tolerance calibration from recorded reconciliation comparisons.
+
+This module is deliberately offline.  It derives a versioned
+ReconciliationTolerance from recorded local-versus-exchange comparison
+samples and verifies that a tolerance bounds a recorded sample set.  A
+calibrated tolerance is evidence about recorded observations only; it never
+relaxes validation or touches accounting state.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+import math
+import re
+from typing import Sequence
+
+from backtest.binance_account_reconciliation import ReconciliationTolerance
+
+
+_CANONICAL_USDM_SYMBOL = re.compile(r"^[A-Z0-9]+-USDT-PERP$")
+
+
+def _require_finite(name: str, value: float, *, non_negative: bool = False) -> None:
+    if not math.isfinite(value):
+        raise ValueError(f"{name} must be finite")
+    if non_negative and value < 0:
+        raise ValueError(f"{name} must be non-negative")
+
+
+@dataclass(frozen=True)
+class CalibrationSample:
+    """One recorded local-versus-exchange comparison sample."""
+
+    field: str
+    local_value: float
+    exchange_value: float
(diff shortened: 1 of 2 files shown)


warren618:
I am going to hold this one, and the reason is about the epic rather than this code.

The calibration module itself is sound: fail-closed on empty/malformed/under-sampled input, a safety factor that provably covers the recorded set, `verify_tolerance_covers` mirroring the reconciler's allowed-delta rule, and timestamp skew explicitly out of scope rather than guessed.

But `#1030` now stands at three modules and zero consumers. `binance_account_reconciliation` is imported only by `binance_shadow_evidence`; `binance_shadow_evidence` is imported by nothing; this would add a third in the same state. Nothing in the tree reaches any of it — no tool, no CLI subcommand, no REST route, no engine. That is ~800 lines of code no user can run, and each slice makes the eventual wiring larger rather than smaller.

So: what I would like next is the slice that makes the chain reachable — one user-facing surface (an agent tool, or a `vibe-trading` subcommand) that takes recorded observations, calibrates, reconciles, and emits the drift evidence. This PR lands with it, or right after it. If the wiring needs a design decision from me before you can write it, say which one and I will answer here.

To be clear this is not a rejection of the work, and I am not asking you to redo anything in it.

warren618 (approved):
Selected over the competing PR after comparing both implementations; rationale in the comment.

warren618:
Merged. Note for the record that #1283 shipped the same module — `binance_tolerance_calibration.py` is byte-identical between the two branches, blob `89aedd96a3cc` — with the 233-line test file omitted. This PR is the one with the tests, so it is the one that goes in; #1283 is being closed as superseded rather than as a duplicate submission.

20 tests pass after rebasing onto current main.

PR #1249, opened 2026-08-29
fix(agent): avoid grounding false blocks across task continuations

AirHua-byte:
## Summary

- Replace exclusion-heavy number matching with structured observed-price claims plus a general symbol-bound numeric provenance gate.
- Move prospective and derived trading levels into an explicit derivation validator instead of silently relaxing them.
- Keep hard failures for zero evidence, explicit currency/value contradictions, unverifiable derived inputs, and dated claims backed only by undated evidence.
- Reuse only trusted direct-parent grounding for clear same-task continuations; current/latest/real-time requests still refresh market evidence.

## Why

Each continuation attempt starts in a new run directory. Previously, verified parent evidence was not reloaded, while broad numeric and currency matching treated counts, dates, ratios, formulas, and plan levels as observed quotes. Relaxing that parser alone would remove a safety property, so this PR relocates strictness into typed components instead:

1. identity authorization and symbol provenance;
2. observed price-field comparison;
3. general symbol-bound numeric provenance for indicators and other financial facts;
4. explicit formula/input/recomputation validation for derived values.

## Changes

- Parse observed price claims into typed symbol, field, value, date, currency, unit, and temporal-scope data.
- Compare claims only inside compatible evidence cohorts while preserving hard blocks for definite conflicts.
- Reject a symbol-bound non-structural value when no tool evidence for that instrument contains it, even when the prose does not use a recognized price phrase.
- Validate labelled derivations only when their non-identity inputs were observed and their printed result recomputes at the displayed precision.
- Preserve generic tool timestamps, and hard-block a specifically dated price claim when all matching field evidence is undated.
- Keep bounded read-only recovery and auditable degraded output for genuinely missing or non-comparable evidence.
- Inherit locked identity and observed evidence only from the direct parent attempt in the same session; reject inheritance across subject/date changes.
- Ensure rejected drafts never appear in returned content or streamed `text_delta` events.

## Safety boundaries

The following remain fail-closed:

- no successful tool evidence for a symbol-bound numeric fact;
- an explicit observed-price contradiction against comparable evidence;
- explicit currency mismatch;
- a dated claim supported only by undated evidence;
- a derived value with unseen inputs, unsupported arithmetic, or an incorrect result.

Prospective targets, stops, triggers, counts, window sizes, confidence scores, dates, and document structure are not misclassified as observed prices. This is classification, not removal of numeric verification: indicators and other numeric facts are covered by the general provenance gate, and labelled calculations are covered by the derivation validator.

## Verification

- Focused grounding, output-discipline, inheritance, generic-timestamp, and session-lifecycle suites: `272 passed`.
- Full backend suite: `12202 passed, 93 skipped`.
- Ruff on the affected grounding/session implementation and tests: passed.
- `git diff --check origin/main...HEAD`: passed.
- Local structural review: `READY`; one `LARGE_DIFF` warning for the existing seven-file PR scope.
- Final reviewed diff: 7 files, 2216 additions, 859 deletions; SHA-256 `db242f1c3089c8865422e29bf2142c53d2b1f539fe0a9b69375ecd217b296026`.

### Live local service validation

Using GMN `gpt-5.6-sol` with medium reasoning effort on the PR code:

- an unevidenced `TSLA.US ... 412.35 USD` assertion was not repeated;
- an unevidenced `TSLA.US RSI 92.7` assertion was also not repeated, exercising a non-price-phrase case;
- a positive control called `get_market_data` and returned the evidenced `2026-09-03` close `376.37 USD` from Yahoo; `grounding_evidence.json` recorded the underlying close `376.3699951171875` and validation passed.

The two negative live prompts were refused by the model before the final gate needed to reject a draft. The deterministic AgentLoop regression separately verifies the post-generation path: an invalid draft is rejected, never streamed, corrected, and only the grounded answer reaches the caller.

The GMN provider registration was supplied only in the local process environment; this PR does not add provider metadata, credentials, model selection, or data-source configuration.

## Scope and risk

- No broker/order execution, credential, provider, MCP configuration, dependency, lockfile, release, or deployment changes.
- No generated run/session artifacts are committed.
- The branch is merged with current `main`; the merge exposed and resolved the undated-evidence strictness regression before publication.
- Rollback is a normal revert of this PR's merge commit.

## Checklist

- [x] Every community-authored commit includes the DCO `Signed-off-by:` trailer.
- [x] Tests cover relaxed classifications and the corresponding strict negative arms.
- [x] No hardcoded credentials, local paths, or generated artifacts.
- [x] Ready for maintainer review.


warren618:
Thank you for the detail in the description — the inheritance rules in particular (locked identities and observed evidence from the direct parent only, rejected across symbol or date changes, refreshed for current/real-time follow-ups) are stated precisely enough to review.

I am not merging it yet, for one reason. The summary says hard blocks are kept for explicit comparable conflicts, while missing, stale or non-comparable evidence now returns "an auditable partial answer with warnings". That is a deliberate relaxation of the gate that stops the agent stating an unverified price, on `src/agent/` and `src/session/`, and the PR does not say which detections it relaxes and which it keeps.

The test diff is where this becomes concrete. Eleven tests are removed and none has a same-named replacement:

```
test_an_unevidenced_price_is_still_rejected_without_any_tool_call
test_agent_loop_rejects_then_corrects_ungrounded_final_answer
test_another_currencys_yuan_does_not_satisfy_a_cny_requirement
test_numeric_gate_validates_derived_formula_and_provenance
test_reference_levels_are_not_read_as_observed_price_claims
test_since_references_are_not_read_as_observed_price_claims
test_count_nouns_are_not_read_as_price_claims
test_ratios_and_fx_rates_are_not_read_as_price_claims
test_line_number_references_are_not_read_as_price_claims
test_numbered_headings_are_not_read_as_price_claims
test_an_unnamed_source_is_still_reported
```

The first two are the load-bearing safety pins — no evidence means block, and a rejected draft never becomes the returned or streamed answer. The rest each encode a specific false positive that was found in production and fixed once already; the comments above them name the cases. 27 new tests is not by itself an answer to that, because the question is not whether the new parser is tested, it is whether those eleven properties still hold.

What would let me merge this:

1. A short table in the description: for each of the eleven, either the new test that pins the same property (by name), or an explicit statement that the behaviour changed and why that is correct.
2. The two safety pins re-asserted under the new architecture, in whatever form fits it — an unevidenced price with no tool call must still be blocked, and a rejected draft must still never reach the user.
3. One sentence answering directly: which detections does this relax? "None" is not an available answer given the summary; naming them is fine.

The structured-claims direction is one I want — exclusion-by-regex has been a recurring source of false positives and you are attacking the right thing. This is about being able to show, afterwards, that the exchange did not cost us a guarantee.

One more thing: #1148 is open against the same area (unit-confusion false positives in price-claim validation). Please have a look — if your parser subsumes it, say so here and I will close that one; if not, the two need to be sequenced.

AirHua-byte:
Thanks for the detailed review. I have pushed `77848f4`, restored the two safety pins, and added the following mapping to the PR description.

| Previous test | Current coverage / behavior |
|---|---|
| `test_an_unevidenced_price_is_still_rejected_without_any_tool_call` | Restored with the same name. An observed-price claim with zero tool evidence is a blocking `numeric_claim_unavailable`/`unsourced_symbol_figures` failure. |
| `test_agent_loop_rejects_then_corrects_ungrounded_final_answer` | Restored with the same name. It asserts that the rejected `0.881` draft appears in neither the returned content nor streamed `text_delta` events; only the corrected grounded answer reaches the user. |
| `test_another_currencys_yuan_does_not_satisfy_a_cny_requirement` | Restored with the same name. An explicitly stated HKD value cannot satisfy CNY evidence and is blocked as `explicit_currency_conflict`. |
| `test_numeric_gate_validates_derived_formula_and_provenance` | Intentional behavior change. The grounding gate no longer recomputes prospective entry/target formulas as observed quotes. This is covered by `test_unlabelled_numeric_categories_do_not_become_price_claims`, while `test_plan_level_mask_leaves_observed_quotes_checked` and `test_plan_level_mask_does_not_shield_a_wrong_quote_end_to_end` ensure an adjacent or mislabeled observed quote is still checked and blocked. |
| `test_reference_levels_are_not_read_as_observed_price_claims` | Restored with the same name, including a positive adjacent observed-high case. |
| `test_since_references_are_not_read_as_observed_price_claims` | Restored with the same name, including a positive adjacent observed-high case. |
| `test_count_nouns_are_not_read_as_price_claims` | Restored with the same name. Counts and trading-day quantities remain outside observed-price claims. |
| `test_ratios_and_fx_rates_are_not_read_as_price_claims` | Restored with the same name. Conversion prose is ignored, while an explicitly asserted `USD/CAD 1.36` remains checked. |
| `test_line_number_references_are_not_read_as_price_claims` | Restored with the same name, including a positive quote embedded beside a line reference. |
| `test_numbered_headings_are_not_read_as_price_claims` | Restored with the same name, including a positive quote embedded in a heading. |
| `test_an_unnamed_source_is_still_reported` | Intentional behavior change, replaced by `test_an_unnamed_source_is_preserved_in_structured_evidence`. The provider must remain in `grounding_evidence.json`, but the answer prose no longer has to repeat it. An explicit contradictory currency still blocks. |

The two load-bearing guarantees therefore remain:

1. An unevidenced observed-price claim with no tool call is blocked.
2. A rejected draft never reaches returned or streamed user-visible content.

The detections relaxed are specifically:

- prospective/derived trading levels and formula arithmetic are no longer treated as observed market-price assertions;
- source and inferred currency do not have to be repeated in prose when they are retained in structured evidence;
- evidence that exists but is stale, divergent, or not semantically comparable may produce an auditable warning instead of a conflict block.

Zero evidence and explicit comparable conflicts are not relaxed.

Regarding #1148: this PR partially subsumes its unit-confusion cases. `qty`, `premium`, `multiplier`, `contract`, `notional`, and `market_val` contexts are ignored structurally, while an adjacent explicit close remains checked. I did not adopt the broad `>100x` bypass because it could hide a fabricated labeled quote. #1249 should land first; #1148 can then be narrowed/rebased for any remaining formula or fallback behavior.

Verification on `77848f4`:

- focused grounding/inheritance suite: 237 passed;
- full suite: 11,412 passed, 93 skipped;
- Ruff and `git diff --check`: passed;
- OpenRouter Ling and MiniMax both passed the same real 562500.SH tool-call and direct-parent evidence-inheritance flow;
- all nine GitHub checks passed, including the Windows packaged application test.

Please let me know if you would prefer the two intentionally changed behaviors to be split into a follow-up PR.

warren618:
Thanks for restoring the safety pins and writing the coverage mapping — I checked the mapping against the branch rather than reading it, and against the true merge base (`80ffdda4`) rather than current main, so the FX/full-width tests that postdate your branch don't get miscounted as deletions.

Five removals remain: `test_an_unnamed_source_is_still_reported` and `test_numeric_gate_validates_derived_formula_and_provenance` in the grounding suite, and a 3-for-3 swap in `test_agent_output_discipline.py` (`test_the_gate_is_strict_about_encyclopaedic_figures_on_purpose`, `test_the_root_allowance_does_not_licence_an_untouched_ticker`, `test_the_widened_corpus_did_not_widen_what_output_must_disclose` out; the three `*_are_not_interpreted_as_market_prices` / `disclosure_wording` tests in).

**That swap is not a rename — the gate opened much wider than the description says.** I ran the same probes against your head (`77848f4`) and against main:

| final answer | tool calls in session | main | this branch |
|---|---|---|---|
| `TSLA.US last traded at 412.35 USD.` | **none** | rejected | **accepted** |
| `TSLA.US price is 412.35 USD.` | **none** | rejected | **accepted** |
| `TSLA.US is at 412.35.` | **none** | rejected | **accepted** |
| `MSFT.US reported revenue of 400.0 billion USD.` | AAPL 10-K only | rejected | **accepted** |
| `TSLA.US closed at 412.35 USD.` | none | rejected | rejected |
| `宁德时代（300750.SZ）现价 245.60 元。` | none | rejected | rejected |

```python
led = GroundingLedger(run_dir=tmp, user_message="what looks good in EV right now")
print(led.validate_final_answer("TSLA.US last traded at 412.35 USD.").valid)
```

`unsourced_symbol_figures` has gone from "a figure attached to a symbol no tool returned" to a price-phrase whitelist: `closed at` still fires, `last traded at` / `price is` / `is at` do not. A fabricated US quote in ordinary English, in a session with **zero** tool calls, now passes the final-answer gate — and the same fabrication still fails in Chinese, so the gate's strictness depends on which language the model answered in.

The test whose name says the strictness is deliberate spelled out the exact scenario: a symbol-free request ("看看新能源板块") ending in fabricated quotes. Probe row 1 is that scenario.

Two things, and I think the first is quick:

1. The false positives you're fixing are real and I want them fixed. But the fix has to move the strictness somewhere, not remove it — if a figure on a never-fetched symbol is now allowed, name the component and condition that catches row 1 instead. "It isn't recognized as a price" is not a gate.
2. Every new test here asserts the gate **opens** (a thing is allowed). None asserts it stays **closed**. That's why CI is green on all nine checks with the table above true. Whatever the new boundary is, please pin both sides of it — the allowed case and the adjacent case that must still be refused.

Two other notes: this now conflicts with main in `grounding.py` (#1316 and #1326 landed there), so it needs a rebase; and it touches `src/agent/` and `src/session/`, which are protected — once the boundary above is settled I'll want to look at the `session/service.py` inheritance separately from the grounding change.

AirHua-byte:
Hi @warren618 — a quick follow-up on the grounding boundary review. I’ve updated #1249 at `4923fcc` to address the Sep 2 feedback: symbol-bound non-structural numbers now go through a general provenance gate (including the `last traded at` / `price is` / `is at` and cross-symbol revenue cases), while explicit derived values have their own observed-input and recomputation validator. Both the allowed and denied boundaries are covered by tests.

I also merged current `main`, preserved the new generic-timestamp strictness, and updated the PR description with the component mapping and validation evidence. The local full suite reports `12,202 passed, 93 skipped`, and GitHub CI is 9/9 green.

When you have a chance, could you please take another look, including the `session/service.py` inheritance boundary you called out? Thanks!

warren618:
Re-checked this against current main. The regression I reported earlier is fixed, and a different one has taken its place.

**Fixed.** With zero tool calls, all five of these are now rejected, in both languages:

```
TSLA.US last traded at 412.35 USD.     rejected
TSLA.US closed at 412.35.              rejected
特斯拉现价 412.35 美元。                  rejected
AAPL.US is trading at 189.20.          rejected
AAPL.US 报价 189.20。                    rejected
```

That was the whole of my earlier objection — the gate had degenerated into a price-phrase whitelist and let the first line through. It no longer does.

**New, and the reason I am not merging yet.** I ran the same fourteen-case table against `main` and against this branch and diffed the verdicts. One case moves in each direction:

```
                                                              main        this PR
"AAPL closed at 100.00 ... and 110.00 ..., a 10% return."     accepted    REJECTED
"苹果第一日收盘 100.00，第二日 110.00，收益率 10%。"                REJECTED    accepted
```

The gate is asymmetric by language on main — the derived-return exemption only recognises the English phrasing. This PR does not remove that asymmetry, it inverts it: Chinese derived returns start passing and English ones start failing. Net, a user in either language still gets a false rejection; only which language changes. Since the UI ships seven locales and model output follows the user's language, that is a swap rather than a fix, and it silently regresses English users who are fine today.

Everything else in the table is unchanged, which is the good news — no case that main rejects becomes accepted apart from the one above, so this is not a broad loosening.

What I would want before merging: make the derived-return exemption language-independent, and assert it. The cheapest honest form is a regression table that runs the *same* construction through both spellings and asserts the verdicts are equal, rather than asserting a literal per language — otherwise the next patch just moves the breakage to the other side again. #1338 is working in the same area and currently has the mirror-image version of this; whichever lands first should carry that table.

One process note: this touches `agent/src/session/service.py`, which the project treats as protected. That is not an objection, just something that needs an explicit nod rather than riding along in a large diff.

warren618:
Correction to my review above, on the part that mattered most.

I said this PR inverts a language asymmetry: Chinese derived returns start passing and English ones start failing, so a user in either language still gets a false rejection. The framing was wrong. My probe ran with **zero tool evidence**, and with no evidence an unsourced price claim *should* be rejected in both languages. So the case I labelled a new false positive on English was the gate finally doing its job, and the case I labelled fixed on Chinese was the gate going blind.

The underlying defect was on main, not here: `_PRICE_CONTEXT_RE` had `\bclose\b`, which does not match "closed", and nothing matched "traded at" or "quoted at". With no canonical symbol to fall back on, `"The stock closed at 412.35."`, `"The stock last traded at 412.35."` and `"Shares were quoted at 412.35."` all passed, while `"该股收盘 412.35。"` was caught. Chinese had its own mirror hole — 成交价 / 最新价 / 股价 / 收报 all missed.

Fixed on main in `19ff47d1`, in the tightening direction: English past-tense forms now match (each requires a following "at", so volume and corporate-event phrasings stay out), the Chinese nouns are added, and `tests/test_grounding_language_parity.py` asserts the two verdicts are **equal** per case rather than a literal per language. On current main all twelve probe cases agree across languages, including the derived-return pair.

**What this means here.** My "not merging yet" rested on the inversion claim, and that claim is withdrawn. What I can still say with evidence:

- The regression I originally reported — the gate degenerating into a price-phrase whitelist, letting `TSLA.US last traded at 412.35 USD.` through — is genuinely fixed on this branch. That was the substantive objection and it is resolved.
- Of the fourteen cases I diffed, nothing that main rejects becomes accepted apart from the derived-return case discussed above, so this is not a broad loosening.
- This branch now conflicts with main in `grounding.py` and `loop.py`, partly because of my change and partly because of today's other merges. It needs a rebase before I can re-run the comparison honestly, and I would rather re-measure than approve on stale numbers.
- The `agent/src/session/service.py` change still deserves an explicit nod rather than riding along in a large diff — that part of my review is unchanged.

If you rebase, I will re-run the same twelve-case parity table plus the fourteen-case main-vs-branch diff and give you a straight answer. Apologies for the wrong diagnosis in between; the numbers I quoted were real, my reading of which side they indicted was not.

warren618:
Closing this one. The substantive finding in it was real and it is fixed on main — just not by this branch, and the branch has drifted too far to be the vehicle now.

What you reported and fixed: the price gate had degenerated into a phrase whitelist, so `TSLA.US last traded at 412.35 USD.` passed with zero tool evidence. That is genuinely closed on main in `19ff47d1` — the English past-tense forms match (each requiring a following "at", so "traded 1.2M shares" and "the deal closed at a 30% premium" stay out), the Chinese nouns 成交价 / 最新价 / 股价 / 收报 are added, and `tests/test_grounding_language_parity.py` now asserts the two languages reach the *same* verdict per case rather than a literal per language, with both the accept and reject arms pinned.

Why not merge this branch: it is +2216/−859 across seven files and now conflicts with main in `grounding.py` and `loop.py`, partly from that fix and partly from the six grounding PRs that landed on 2026-09-05 and 2026-09-06. Rebasing it would be a rewrite, and the piece that justified the size — task-continuation evidence inheritance through `src/session/service.py`, which is a protected path — never got the separate discussion it needs. That is the part I would still like to see, on its own, small enough to review as a boundary change rather than as a rider on a grounding-gate rewrite.

I also owe you a correction on the record: in my 2026-09-05 review I described this branch as inverting a language asymmetry, then withdrew it the same day. My probe had run with zero tool evidence, where rejection is correct in both languages; the defect I was pointing at was main's, not yours. The numbers I quoted were real, my reading of which side they indicted was not.

If you want to open the session-continuation boundary as its own PR, I will review it against that framing.

PR #1250, opened 2026-08-29
feat(portfolio): add connector onboarding contracts

goatyyc:
## Summary

- publish machine-readable onboarding metadata for built-in read-only connectors
- store credentials per local connection in the OS keyring and let CLI, MCP, and Web reads resolve them by connection ID
- add portfolio compatibility levels, payload validation, and OKX spot-balance adaptation
- render generic credential setup, connection diagnostics, and compatibility badges in the Portfolio connection center
- document the AI-friendly local-only credential workflow in English and Chinese

## Safety

- connector profiles must remain structurally read-only
- credential values stay in the local OS keyring and are never returned through MCP or committed to the repository
- legacy local config and environment variables remain a compatibility fallback

## Validation

- Python: 41 targeted tests passed
- Frontend: 8 Portfolio tests passed
- Ruff and git diff checks passed
- Live read-only smoke test previously completed with IBKR, Longbridge, Binance, and OKX; no account identifiers, holdings, balances, tokens, or API keys are included in this PR

Signed-off commits included for DCO.


warren618:
Merged. This is a lot of careful work: the onboarding contract carries field names only, credentials never reach `connections.json` (the existing test pins that), the legacy `~/.vibe-trading/<connector>.json` and env fallback is preserved, credentials supplied directly in a call are refused, and an incomplete vault entry fails loudly rather than half-filling a config. I checked `is_portfolio_connection_profile` — the eligibility rule is unchanged, only reformatted.

One bug fixed on top (`19d4bee3`), and it is worth explaining because the mistake is subtle.

`_sdk_config`'s vault branch deliberately bypasses `build_config` so vault secrets cannot mix with another account's legacy JSON — right instinct. But `build_config` is also where each connector's override allowlist lives, and the replacement built the config from a raw merged mapping:

```python
payload = {**dict(profile.config), **credentials, **options}
return config_type.from_mapping(payload)
```

`options` is the one part of that payload that arrives from an MCP tool argument or a REST body. Every SDK connector narrows it on purpose — OKX and Binance both exclude `readonly` (OKX's own docstring: "Always true for this layer") and `timeout`; Longbridge's overlay is `profile`/`region` only, its docstring saying callers "cannot mix or bypass the shared resolver".

Reproduced on the merge commit:

```
service._sdk_config(okx-live-sdk-readonly, {connection_id, readonly: False, timeout: 999.0})
  -> OKXConfig(readonly=False, timeout=999.0)
okx_sdk.build_config(profile.config, {readonly: False, timeout: 999.0})
  -> OKXConfig(readonly=True,  timeout=15.0)
```

`_allowed_override_keys` now filters `options` through the connector's declared allowlist (`_OVERRIDE_KEYS`, or Longbridge's narrower `_OVERLAY_KEYS`), following `build_config` to its defining module because etoro and mt5 only re-export it. A connector declaring neither gets an empty allowlist — overrides dropped, not passed through — and a new test fails if one ever ships that way.

Unrelated, for next time: the diff carries a fair amount of reformatting churn (previously wrapped calls joined onto single lines). It is within the 120-char limit so ruff is happy, but it inflates the diff and makes the real change harder to find.

PR #1251, opened 2026-08-29
fix(portfolio): classify OAuth reconnect failures

goatyyc:
## Summary

- distinguish explicit OAuth authorization failures from transient broker/network read failures
- expose reconnect only when the backend marks a source as requiring authorization
- offer a non-interactive read retry for timeouts and temporary failures
- add a frontend safety deadline beyond the existing backend OAuth worker timeout so reconnect polling cannot spin forever
- disable conflicting refresh/reconnect actions while one operation is active

## Why

A timeout or temporary MCP/network error should not force a user through broker login. Conversely, an explicit not-authorized response should provide the reconnect action. This keeps normal portfolio reads non-interactive and makes the recovery action match the actual failure.

## Validation

- Python: 15 targeted portfolio service/route tests passed
- Frontend: 7 Portfolio tests passed
- Ruff and git diff checks passed
- No credentials, OAuth tokens, account identifiers, holdings, or balances are included

Signed-off commits included for DCO.


warren618:
Merged. Making reconnect appear only for an explicit authorization failure, and offering a plain retry for a timeout, is the right split — a network blip should never push someone back through broker login. The frontend deadline beyond the backend OAuth worker timeout, and disabling conflicting refresh/reconnect while one is in flight, are both good catches.

Full Python suite and all 596 frontend tests pass locally, seven-locale parity included.

PR #1252, opened 2026-08-29
fix(tools): resolve a skill-relative reference link in read_file

ethanstoner:
## Summary

- `read_file` now resolves a bare `references/…` or `scripts/…` path against the skill that owns it, so a reference-table link resolves for the agent *and* for a human reading the file on GitHub.
- The prefixed and `skills/`-rooted forms are unchanged, so nothing that resolves today changes behaviour.

## Why

Follow-up to #1236, which I have closed. That PR tried to fix the human-facing half by dropping the `<skill>/` prefix from the reference tables. @warren618 showed that the prefix is load-bearing — `read_file` roots at the bundled `skills/` directory, so the bare form is what fails — and that the change took `agent/tests/test_skill_reference_links.py` from 548 passing to 546 failing. I reproduced both numbers before accepting them.

His conclusion was that the two consumers cannot be satisfied by one relative string, and that the fix belongs in the tool:

> the fix belongs in `read_file`, teaching it to resolve a bare `references/...` against the skill currently being read, after which the links can drop the prefix and work in both places. That is a change I'd take.

This is that change. It is deliberately the tool-side half only — no link in any `SKILL.md` is touched here.

## Changes

- `agent/src/tools/read_file_tool.py` — a bare `references/…` or `scripts/…` path falls back to resolving against the owning skill. The fallback runs **last**, after every allowed root has already missed, so `run_dir` still wins and no currently-resolving path changes meaning. It reads only from the read-only skills tree, so a file the agent wrote cannot shadow it, and only those two directory names participate, so an arbitrary model-supplied relative path never starts reaching into `skills/`.
- Two refusals rather than guesses:
  - A `..` segment is rejected outright. Without that, a path only has to *start* with an allowed prefix to land elsewhere — `references/../SKILL.md` matched all 90 skills. `references/../../../pyproject.toml` and `scripts/../../../../../../etc/passwd` remain errors.
  - A path that two skills both own is reported with the candidates named rather than silently resolved to one, which would hand the agent a document from the wrong skill and read as a correct answer. There is no collision today (307 distinct skill-relative paths, 0 shared), but nothing enforces that and this fallback is what would make it matter.
- `agent/tests/test_read_file_skill_relative.py` — new, 8 tests: both resolution forms, `skills/`-rooted still working, `run_dir` precedence, unknown path still erroring, four traversal attempts, and the ambiguity report.
- `agent/tests/test_skill_reference_links.py` — one test asserted the bare form was unreachable, which was the bug being fixed; it now asserts both forms reach the same document. The prefix-convention test added in 18bd2fc is untouched.

## Test Plan

- [x] Existing tests pass — `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py`: **11436 passed / 42 failed** on this branch versus **11428 passed / 42 failed** on `e69626a`. Identical failure set, diffed between the two runs — POSIX file-mode, directory-fsync and symlink assertions that fail on any clean Windows checkout.
- [x] New tests added — `agent/tests/test_read_file_skill_relative.py`, 8 tests.
- [x] Tested manually — resolved every reference-table link in the corpus through `read_file` in both forms.
- `ruff` clean on the changed files. `black --check` already flags `read_file_tool.py` and `test_skill_reference_links.py` on `e69626a`, so I mixed in no reformatting.

## Risk

No live surface. No broker, order, network, credential or MCP path is touched; `read_file` gains one filesystem lookup inside the bundled read-only skills directory. The traversal guard is tightened rather than loosened. Rollback is a revert of the single commit.

This change was AI-assisted. The measurements quoted above are ones I ran and reproduced, not ones I inferred.

## Checklist

- [x] Code follows CONTRIBUTING.md guidelines — DCO sign-off on the commit
- [x] Documentation updated (if user-facing change) — n/a, no user-facing surface changes

### DIFF
--- a/agent/src/tools/read_file_tool.py
+++ b/agent/src/tools/read_file_tool.py
@@ -14,6 +14,54 @@ from src.tools.redaction import redact_internal_paths
 
 _OUTPUT_LIMIT = 50_000
 
+# Subdirectories a SKILL.md links into. A link inside a skill document is
+# written relative to that document — `references/foo.md` — because that is the
+# form that resolves for a human reading the file on GitHub. This tool roots at
+# the bundled skills/ directory, so the same string has to be resolved against
+# the skill that owns it or the agent cannot open what the document points at.
+# The list is deliberately narrow: only these two names are searched, so an
+# arbitrary model-supplied relative path never starts reaching into skills/.
+_SKILL_RELATIVE_PREFIXES = ("references/", "scripts/")
+
+
+def _bundled_skills_dir() -> Path:
+    """Return the bundled read-only skills root."""
+    return Path(__file__).resolve().parents[1] / "skills"
+
+
+def _skill_relative_matches(file_path: str, skills_dir: Path) -> list[tuple[str, Path]]:
+    """Return the bundled skills that carry ``file_path`` inside their directory.
+
+    Args:
+        file_path: Skill-relative path as written in a SKILL.md link, e.g.
+            ``references/sec_edgar_client.md``.
+        skills_dir: Bundled read-only skills root.
+
+    Returns:
+        ``(skill_name, path)`` for each owning skill, in skill-name order. Empty
+        when no bundled skill carries the path.
+    """
+    # A link written in a skill document never needs to climb, and `..` would
+    # let a path that merely *starts* with an allowed prefix land somewhere
+    # else — `references/../SKILL.md` matching all 90 skills, for one. Refusing
+    # the segment outright keeps "starts with references/" and "stays under
+    # references/" the same statement.
+    if ".." in file_path.replace("\\", "/").split("/"):
+        return []
+
+    matches: list[tuple[str, Path]] = []
+    for skill_dir in sorted(p for p in skills_dir.iterdir() if p.is_dir()):
+        try:
+            candidate = _safe_path(file_path, skill_dir)
+        except ValueError:
+            # Absolute paths and UNC shares, rejected by the same containment
+            # check the skills/ root already applies, enforced one level down
+            # so the fallback cannot widen the boundary.
+            continue
+        if candidate.is_file():
+            matches.append((skill_dir.name, candidate))
+    return matches
+
 
 class ReadFileTool(BaseTool):
     """Read file contents with optional line limit."""
@@ -27,7 +75,9 @@ class ReadFileTool(BaseTool):
                 "type": "string",
                 "description": (
                     "File path relative to run_dir or skills/. "
-                    "The skills/ prefix always resolves to the bundled skills."
+                    "The skills/ prefix always resolves to the bundled skills. "
+                    "A references/ or scripts/ path copied out of a SKILL.md "
+                    "resolves against the skill that owns it."
                 ),
             },
             "limit": {"type": "integer", "description": "Max number of lines to return (default: all)"},
@@ -62,7 +112,7 @@ class ReadFileTool(BaseTool):
                     ensure_ascii=False,
                 )
         # Read-only access to skills/
-        skills_dir = Path(__file__).resolve().parents[1] / "skills"
+        skills_dir = _bundled_skills_dir()
         if skills_dir.exists():
             allowed_roots.append(skills_dir.resolve())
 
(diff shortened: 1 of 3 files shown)


warren618:
Merged. This is the right fix and it is well made — thank you for taking the rejection apart rather than arguing with it, and for reproducing 548/546 yourself before accepting my numbers. That is exactly the standard I want to hold myself to.

Verified independently before merging:

- **Traversal.** `references/../SKILL.md`, `references/../../../pyproject.toml`, `scripts/../../../../../../etc/passwd`, plus the backslash and percent-encoded variants I tried on top of your four, all refused. `references/sec_edgar_client.md` resolves. Refusing the `..` segment outright, so that "starts with `references/`" and "stays under `references/`" are the same statement, is the correct shape — a containment check alone would have let the prefix be a passport rather than a boundary.
- **Ambiguity.** Counted the corpus myself: 90 bundled skills, 0 skill-relative paths owned by more than one skill. (I get 329 distinct paths on disk against your 307 linked from reference tables — different denominators, not a disagreement.) So the collision branch is unreachable today, which is precisely why it is worth having: nothing enforces that invariant, and this fallback is what would make a future collision matter. Reporting the candidates instead of picking one is right — a document from the wrong skill would read as a correct answer.
- Ordering: the fallback runs last, after every allowed root has missed, so `run_dir` still wins and nothing that resolves today changes meaning. 1049 skill/read_file/registry tests pass locally.

Two things I want to name, since the point of the exercise was the verification discipline rather than the diff:

The narrow prefix list is doing more work than it looks like. Resolving *any* bare relative path against the skills tree would let a model-supplied string start probing a read-only corpus it was never handed; restricting it to the two directory names that reference tables actually link into keeps the new surface the same size as the problem.

And the prefixed and `skills/`-rooted forms still working means the links themselves need not change at all. I would rather leave the reference tables as they are for now than do a 262-link rewrite for cosmetics — but the option is open, which it was not before this. If you want to take that on as a separate PR, say so and I will look at it on its own merits.

This closes the last thread from #1236. Nice work.

ethanstoner:
On the reference tables — I measured before taking you up on it, and it comes out
differently from "cosmetics", so here is the number rather than an argument.

**Every reference link in the bundled skills tree currently 404s for a human.** 295 links
across 9 `SKILL.md` files (231 of them the tushare table), and it is 100% of them, not a
subset. A browser resolves a relative link against the file's own directory, so
`tushare/references/...` written inside `skills/tushare/SKILL.md` points at
`skills/tushare/tushare/references/...`. Checked against the live renderer rather than by
path arithmetic:

| URL | |
|---|---|
| `.../agent/src/skills/tushare/tushare/scripts/stock_data_example.py` | 404 |
| `.../agent/src/skills/tushare/scripts/stock_data_example.py` | 200 |
| `.../agent/src/skills/sec-edgar/sec-edgar/references/sec_edgar_client.md` | 404 |
| `.../agent/src/skills/sec-edgar/references/sec_edgar_client.md` | 200 |

Five of those files also carry a prose "link convention" note instructing future skill
authors to write the prefixed form, on the grounds that "omitting the prefix makes the
read fail" — which stopped being true when #1252 merged. Meanwhile `read_file_tool.py`'s
own module comment already describes the document-relative form as the one a SKILL.md is
written with. The docs and the tool disagree with each other today, and the docs are the
side that is wrong.

What I'd send:

- 295 targets lose the prefix. Table column widths are preserved — the characters removed
  from the target go back onto the cell's trailing padding, so no rendered column moves
  (verified line by line: 295 changed lines, 0 width shifts).
- The 5 convention notes rewritten to say what is now true.
- The convention assertions in `test_skill_reference_links.py` and
  `test_investor_lenses_skill.py` flip to the document-relative form, plus a new
  parametrised test that resolves every link the way a browser does and asserts it is the
  same file `read_file` opens. That reader was never asserted, which is why this sat here.
- A test that the prefixed form still resolves, so anything that already wrote it — an
  older copy of the table, or a model reproducing one from memory — keeps working.

Two things that deserve your scepticism, since the ambiguity concern was yours and I
agreed with it in that test's docstring:

- The relative form leans on the fallback's search instead of naming one file outright. I
  measured the collision surface: **307 distinct `references/` + `scripts/` paths across all
  90 skills, 0 carried by more than one skill.** Nearest miss is `大宗交易.md`, which tushare
  and eastmoney both ship, at different relative paths.
- That risk is already guarded, for free: the fallback reports an ambiguous path rather
  than guessing it, so the existing per-link `read_file` assertion goes red the day a
  second skill collides. Loud failure, not a silent wrong-document.

Separately and deliberately not in this change: 7 links go the other way — they render
fine on GitHub but `read_file` cannot open them, because the fallback is scoped to
`references/` and `scripts/`. They are `../tushare/SKILL.md` and `../akshare/SKILL.md` in
`ashare-pre-st-filter`, `../cross-market-strategy/SKILL.md` and `examples.md` in
`strategy-generate`, and `example_signal_engine.py` in `cross-market-strategy`. Happy to
file that separately or leave it alone.

Diff is 9 doc files and 2 test files, mechanical. If you still read it as cosmetics with
295/295 in front of you, say so and I'll drop it — I'd rather not spend your review on it
than push it.

PR #1253, opened 2026-08-29
fix(live): skip market-triggered ticks while the market is closed

he-yufeng:
## Summary

- `_jobs_from_triggers` downgraded a MARKET trigger to a plain 60s interval job and the tick never re-checked the session: `due_now` sat imported in `runtime/runner.py` with zero call sites. Weekend ticks ran the full pipeline and submitted orders that queue into the Monday open.
- The runner now gates the tick on the trigger layer the jobs came from: with MARKET triggers attached, a tick proceeds only when at least one of their markets is open at the tick instant.

## Why

Refs #1207 (Phase 0, item 5). The design always intended this: `_jobs_from_triggers`' docstring says the tick itself re-checks market state, but nothing ever called it.

## Changes

- `runner._any_market_open`: union over attached MARKET triggers via the trigger layer's `due_now`. Runners without MARKET triggers never gate; 24/7 markets (crypto) report open at every instant.
- `run_once` returns a new `market_closed` outcome before reconcile when every attached market is closed, so a closed session costs no broker calls and writes no audit records (a 60s watch cadence all weekend would otherwise flood the ledger).

## Test Plan

- [x] New tests in `test_runtime_runner.py`: pre-open skip stops after the mandate check with no reconcile/invoke/audit, mid-session ticks proceed, a closed us_equity plus 24/7 crypto union stays open, interval-only runners ignore the gate.
- [x] `pytest agent/tests/test_runtime_runner.py -q`: 29 passed. Live safety set (flatten, sweep latch, api runtime, order gate, mandate enforcement, killswitch, readonly): 150 passed.
- [ ] Full suite and e2e not run; change is confined to the tick gate. `ruff check` clean on the touched files (two pre-existing F401s in runner.py left alone).

Risk: fail-closed in the trading direction only. Ticks outside market hours become no-ops; nothing about open-session behavior changes. Rollback is a plain revert.


warren618:
Added a follow-up that keeps the trigger-hours dependency explicit through the merge, then merged the current `main` into the contributor branch without a force-push. The final PR diff remains limited to the live runner and its regression test. Verification: 52 focused tests passed; Ruff, `py_compile`, and diff checks passed.

PR #1258, opened 2026-08-30
fix(frontend): scrollable main region + collapse long run prompts

iagop03:
## Summary

Two related overflow bugs, both invisible with short content and only surfacing once content gets tall — a long generated report, or a long user prompt.

## Why

1. **`Layout.tsx`**: `<main>` is a flex item (`flex-1`) inside a `flex-col` parent with `overflow-hidden`. Without `min-h-0`, a flex item's default `min-height: auto` lets it grow past its allotted space to fit tall content instead of respecting its own `overflow-auto` — the excess then gets hard-clipped by the parent's `overflow-hidden`, with no scrollbar at all, rather than scrolling into view. This affects every routed page, not just run detail — a long-running task's page (or any tall content) could become partially unreachable.
2. **`RunDetail.tsx`**: the header renders `run.prompt` in full, uncapped. That header sits above the scrollable body, so a long multi-paragraph prompt (a normal shape for this app — users write structured strategy specs) pushes the actual dashboard/tabs mostly off-screen, leaving only a thin scrollable strip below it.

## Changes

- `Layout.tsx`: add `min-h-0` to `<main>`.
- `RunDetail.tsx`: `run.prompt` now clamps to 3 lines by default (`line-clamp-3`) with a "show full prompt" / "show less" toggle button, collapsed by default.
- New i18n keys `runDetail.showFullPrompt` / `runDetail.showLessPrompt` across all 7 locales.
- `Layout.test.tsx`: regression assertion that `<main>` carries `flex-1 min-h-0 overflow-auto`.

## Test Plan

- [x] `npx tsc -b` — clean.
- [x] Full frontend suite: `npx vitest run --pool=threads` — 594/596 passing; the 2 failures (`options.test.ts` `ivLabel` decimal-separator formatting, `tearsheet.test.ts` UTC-5 timezone-offset assumption) are pre-existing and locale/timezone-environment-dependent, unrelated to this change (reproduce identically on `main`).
- [x] Targeted: `Layout.test.tsx` + `RunDetail.test.tsx` — 16/16 passing.
- [ ] Not tested against a live browser session manually; verified via the unit test's class assertions instead.

## Checklist

- [x] No hardcoded values
- [x] Follows CONTRIBUTING.md (DCO sign-off on the commit)
- [x] No user-facing docs needed — pure bugfix, no new config/flags/behavior to document


warren618:
Added the missing per-run reset for `promptExpanded`. Expanding run A and navigating to run B now starts B collapsed again; `aria-expanded` and a real router-navigation regression test pin the behavior. Verification: 17 focused tests passed, and TypeScript plus the production build passed.

PR #1259, opened 2026-08-30
fix(frontend): point non-backtest runs to Studio instead of a blank dashboard

iagop03:
## Summary

- The "Dashboard" tab (`StrategyResearchDashboard`) is always shown regardless of run type, and is built entirely around a classic equity-curve backtest (KPIs from `run.metrics`, nav/drawdown/trade charts).
- A portfolio-construction / risk-snapshot run — `risk_xray` + `rebalance_notes`, no equity curve, no trades — has none of that data. Every KPI silently rendered as a dash and every chart stayed empty, with nothing telling you this run is simply a different type. It just looked broken.

## Why

Found while investigating a user report of "the numbers I used to see are gone, just dashes now" on a specific run. The run's backend data (`risk_xray.json`) was verified complete and correct on disk — the issue is purely that the Dashboard tab has no data of the shape it expects for this run type. The correct data was rendering fine all along on the "Studio" tab (`RunDetail.tsx`'s `hasStudio` / `StudioTab`), just not the one the user had open.

## Changes

- `StrategyResearchDashboard.tsx`: early-return an empty-state card (short message + pointer to Studio) when there's no `equity_curve`/`trade_log`/`artifacts_equity_csv`/`artifacts_trades_csv` data but the run does have `risk_xray` or `rebalance_notes`. Runs with real backtest data render exactly as before, unconditionally, regardless of whether `risk_xray` also happens to be present.
- New i18n keys `runDashboard.noBacktestData` / `runDashboard.noBacktestDataStudioHint` across all 7 locales.
- New test file `StrategyResearchDashboard.test.tsx`: covers the `risk_xray`-only case, the `rebalance_notes`-only case, and confirms the normal dashboard still renders when equity/trade data is present alongside `risk_xray`.

## Out of scope

- Not changing which tab is auto-selected on load, or hiding the Dashboard tab itself for these runs — this only fixes what renders inside it. A "default to the right tab per run type" change would be a separate, more invasive UX decision.

## Test Plan

- [x] New test: `npx vitest run src/components/charts/__tests__/StrategyResearchDashboard.test.tsx` — 3/3 passing.
- [x] `npx tsc -b` — clean.
- [x] All 7 locale JSON files validated to still parse.
- [ ] Not tested against a live browser session manually; verified via the unit tests instead.

## Checklist

- [x] No hardcoded values
- [x] Follows CONTRIBUTING.md (DCO sign-off on the commit)
- [x] No user-facing docs needed — pure bugfix, no new config/flags/behavior to document

### DIFF
--- a/frontend/src/components/charts/StrategyResearchDashboard.tsx
+++ b/frontend/src/components/charts/StrategyResearchDashboard.tsx
@@ -133,7 +133,9 @@ export function StrategyResearchDashboard({ run }: { run: RunData }) {
       drawdown: String(row.drawdown ?? ""),
     }));
   }, [run.artifacts_equity_csv, run.equity_curve]);
-  const trades = run.artifacts_trades_csv || run.trade_log || [];
+  const trades = run.artifacts_trades_csv?.length
+    ? run.artifacts_trades_csv
+    : run.trade_log || [];
   const identity = useMemo(() => getStrategyReportIdentity(run), [run]);
 
   useEffect(() => {
--- a/frontend/src/components/charts/StrategyResearchDashboard.tsx
+++ b/frontend/src/components/charts/StrategyResearchDashboard.tsx
@@ -207,6 +207,21 @@ export function StrategyResearchDashboard({ run }: { run: RunData }) {
     };
   }, [dark, equityRows, trades]);
 
+  // This dashboard is built around a classic equity-curve backtest. A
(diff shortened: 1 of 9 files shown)


warren618:
Made the trade-data fallback length-aware: an empty `artifacts_trades_csv` array no longer hides a non-empty `trade_log`. Added a regression covering that exact payload shape. Verification: all 4 dashboard tests passed, and TypeScript plus the production build passed.

PR #1264, opened 2026-08-30
feat(settings): per-channel configuration, custom providers, and Persian locale

Emad211:
## Summary
- **Channels:** `GET/PUT /channels/config` and `POST /channels/config/test` expose IM channel configuration over the API — per-channel enable/token/allowlist editing, a connection test (Telegram runs a real `getMe`), and honest unavailable-adapter reporting for missing optional dependencies. The Settings UI grows a channel-configuration section (`ChannelConfigSection`).
- **Custom providers:** `custom_provider_routes` + `custom_profiles` add OpenAI-compatible endpoint management (create/test/activate) with `capabilities.py` resolving active custom credentials; the Settings UI gets `CustomProviderManager` with a draft → tested → active progression (design notes in `frontend/DESIGN.md`).
- **Persian (fa):** full `fa.json` locale with key parity across all seven locales — including the new source-priority and portfolio keys — and RTL handling in the chat components and global CSS.
- `LANGCHAIN_MAX_TOKENS` passthrough for custom LLM endpoints.

## Testing
- New tests: `agent/tests/test_channels_config_api.py`, `agent/tests/test_custom_provider_routes.py`, `frontend/src/components/settings/__tests__/CustomProviderManager.test.tsx`, `frontend/src/pages/__tests__/SettingsChannels.test.tsx`, `frontend/src/i18n/__tests__/i18n.test.ts`.
- Backend suite on this tree: **11645 passed, 107 skipped, 0 failed**; frontend: **62 files / 604 tests passed**.


warren618:
Reviewed at the structural level; there is a blocker before the detail review is worth doing.

**CI's `Repository safety gates` step fails on both Pythons.** Reproduced locally on your head:

```
FAIL: raw os.getenv / os.environ reads outside config layer:
  agent/src/api/custom_provider_routes.py:156: os.environ.update() — use src.config accessor
```

Writing straight into `os.environ` from a request handler is the part I would want changed even without the gate: it mutates process-wide state from an HTTP route, so two concurrent requests activating different custom providers race, and the change outlives the request. Route it through the config layer the way the existing settings routes do. Because that gate fails first, no test job ran — the suite numbers in the description are from your machine only.

Two more things before this can be reviewed properly:

1. **Please split it.** Per-channel configuration, custom OpenAI-compatible providers, and a new locale are three independent features across 32 files; landing them together means one revert undoes all three, and the provider piece touches `src/providers/`, which is protected and needs its own discussion.
2. **The Persian locale is now the ninth, not the eighth.** `pt-BR` landed on 2026-09-01 and main is at eight locales today, so please rebase — the parity suite globs `locales/*.json`, so `fa.json` is covered the moment it lands, but `SUPPORTED_LANGUAGES` and the ordered-code case still need the entry by hand. One naming check while you are there: `paper` as in paper trading became `papel` in the Spanish locale, which reads as "the security" in Spanish finance. Please check how the other locales render the safety-critical labels before settling the Persian wording.

RTL is already handled for Arabic, so `fa` should be able to reuse that path rather than adding its own CSS.

warren618:
Reviewed this properly at last — sorry it sat. The build quality is good: the new routes are all behind `require_local_or_auth` / `require_settings_write_auth`, keys go to the OS keyring rather than the profile JSON, and the Persian locale is complete — I diffed the key trees and `fa.json` matches `en.json` exactly (1562 keys, nothing missing, nothing extra), with `SUPPORTED_LANGUAGES`, the lazy loader, the ordered-code assertion and `isRtl` all updated. That is more care than most locale PRs arrive with.

Three things before it can land:

**1. CI is red at `Repository safety gates`, not at the tests.** The gate step runs before `Run tests`, so this PR's tests have never actually run in CI. Reproduced locally with `bash tools/ci_grep_gates.sh`:

```
FAIL: raw os.getenv / os.environ reads outside config layer:
  agent/src/api/custom_provider_routes.py:156: os.environ.update() — use src.config accessor
```

`settings_routes.py` is the allowlisted place for this kind of process-env synchronization; `_activate_profile` should either go through the same accessor or move that step there.

**2. Please rebase.** The branch predates the pt-BR locale that landed on 2026-09-01, so it conflicts on `i18n/index.ts`, the ordered-code test and the locale JSONs. After the rebase `fa` needs to slot into the ordered list alongside `pt-BR`, and `pt-BR.json` needs the same keys this PR adds to the other eight.

**3. One naming check, from experience.** A Spanish PR once translated "paper" (as in paper trading) to `papel`, which in Spanish finance reads as "the security" — the opposite of what the label means. Before this lands, please double-check the Persian rendering of the safety-critical labels specifically: paper vs live, the mandate/kill-switch strings, and anything that tells the user an action is irreversible. Everything else can be corrected later; those cannot.

`src/providers/` is a protected area, so the `capabilities.py` hook needs my sign-off separately — the shape you used (custom profile resolved first, existing path untouched otherwise) is the right one, I just want to re-read it against the rebased branch rather than the current one.

warren618:
Reviewing this one separately from #1262, which I closed — I want to be explicit that the reason there does not apply here. I grepped this diff for the same thing and **there is no global prompt or language-default change in it**. The Persian support in this PR is a locale, which is exactly the right way to add a language, and it is welcome.

Three things before it can land.

**1. It needs a rebase — one conflict, and it is smaller than it looks.**

`mergeStateStatus` is DIRTY, and the only conflicted file is `frontend/src/i18n/__tests__/i18n.test.ts`. Main has since replaced the hand-written language list in the parity suite with a `locales/*.json` glob, precisely so a new locale is covered the moment its file lands. So most of your edit to that test is now unnecessary rather than wrong — take main's version and check what remains.

Your registration points are correct as they stand: `SUPPORTED_LANGUAGES` and `localeLoaders` in `i18n/index.ts` both have `fa`, and `dir: "rtl"` is right.

**2. The provider changes need to be split out.**

`agent/src/providers/{capabilities,custom_profiles,llm}.py` are protected core. That is not a veto — it means those three files get reviewed on their own terms, not as part of a settings-UI PR. Custom OpenAI-compatible endpoint management is a real feature and I am open to it, but a 32-file PR is the wrong container for a change to provider credential resolution. Please send the custom-provider half as its own PR; the channels config and the `fa` locale can land here without it.

**3. One thing to check in the locale itself, and it is a real trap.**

Please verify how `fa.json` renders the safety-critical labels — paper vs live trading above all, plus the order and mandate strings. We shipped a bug of this exact shape in Spanish: `paper` was translated literally to `papel`, which in Spanish finance reads as "the security" rather than "simulated". A user who misreads that label may believe they are in a simulator while placing real orders.

The test suite cannot catch this. Key parity only proves a string exists, not that it means the right thing. So the check is manual: for each of those labels, look at how `zh-CN`, `ja`, `ko`, and `ar` render it, and confirm the Persian conveys *simulated / not real money* rather than the literal word for paper. If a term has no natural Persian equivalent, keeping the English term is better than a literal translation that inverts the meaning.

Do those three and I will review it end to end. The channels work — honest reporting of unavailable adapters instead of pretending a missing optional dependency is a configuration error — is the part I like most here, and I would rather it landed on its own than waited behind the provider review.

warren618:
The channel half of this is superseded by #1520 / #1529, now on main. That half returned `needs_restart`, wrote a generic `token` key the adapters don't read, and registered a `GET /channels/config` that would shadow the new one. Could you split out the custom providers and the Persian locale as two separate PRs on current main? The providers one touches `src/providers/`, so please describe the design in an issue first. CI is also red at the safety-gate step (`os.environ.update` at `custom_provider_routes.py:156`).

PR #1268, opened 2026-08-30
fix(agent): make tool invocation policy argument-aware

AirHua-byte:
## Problem

`AgentLoop` currently records only the name of a successful non-repeatable tool. A later call with different arguments is therefore skipped before the tool implementation runs. This breaks parameter-dependent read-only research: after fetching one company's financial statements, sector membership, news, fund flow, or reports, the same tool cannot fetch the next company in a later iteration.

On current `main`, a two-symbol `get_financial_statements` reproduction executes only `600276.SH`; `601899.SH` is emitted as `tool_skipped`.

## Change

- Add an explicit `InvocationPolicy` with backward-compatible defaults:
  - deterministic tools cache identical logical calls;
  - read-only or legacy repeatable tools allow subsequent calls;
  - legacy non-repeatable mutations remain once-per-run.
- Evaluate Grounding authorization before cache or once-per-run decisions.
- Reserve once-per-run mutations before batch execution so duplicates in one assistant response cannot both run.
- Add redacted arguments, invocation policy, hashed call key, and reason to skip/cache traces.

## Behavior

Before:

```text
get_financial_statements(600276.SH) -> executed
get_financial_statements(601899.SH) -> tool_skipped
```

After:

```text
get_financial_statements(600276.SH) -> executed
get_financial_statements(601899.SH) -> executed
```

Existing order and external-write tools with `repeatable=False` and `is_readonly=False` continue to map to once-per-run.

## Out of scope

- Query rewriting or structured request contracts
- Symbol identity/context resolution
- Dynamic capability routing
- Failure circuit breaking and explicit business-idempotency policies
- Prompt changes or new data sources
- Changes to broker, mandate, MCP, credential, or deployment configuration

## Validation

- Baseline reproduction on `origin/main`: 1 implementation call, 1 false `tool_skipped` for two different symbols.
- Fixed-branch reproduction: 2 implementation calls, 0 `tool_skipped`.
- `pytest agent/tests/test_agent_loop_invocation_policy.py agent/tests/test_agent_loop_repeatable_tools.py agent/tests/test_agent_loop_deterministic_cache.py -q` — 15 passed.
- `pytest agent/tests/test_mandate_enforcement.py agent/tests/test_killswitch_blocks_orders.py agent/tests/test_readonly_default.py agent/tests/test_agent_loop_trace.py -q` — 40 passed.
- `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` — 11,659 passed, 93 skipped.
- `ruff check` on all changed Python files — passed.
- `git diff --check` — passed.

No live broker action or external network data call was run.

## Rollback

Revert the single commit to restore the prior name-level behavior. The patch adds no migration, dependency, configuration, or persisted-state format.

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import concurrent.futures
 import copy
+import hashlib
 import json
 import logging
 import queue
@@ -25,13 +26,13 @@ import threading
 import time as _time
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Mapping, Optional
 
 from src.agent.context import ContextBuilder
 from src.agent.grounding import GroundingLedger
 from src.agent.memory import WorkspaceMemory
 from src.agent.progress import HeartbeatTimer, ProgressEvent, _set_emitter
-from src.agent.tools import ToolRegistry
+from src.agent.tools import InvocationPolicy, ToolRegistry
 from src.agent.trace import TraceWriter
 from src.core.state import RunStateStore
 from src.goal.context import (
@@ -936,7 +937,8 @@ class AgentLoop:
         self.memory = memory or WorkspaceMemory()
         self._event_callback = event_callback
         self.max_iterations = max_iterations
-        self._called_ok: set[str] = set()
+        self._once_per_run_completed: set[str] = set()
+        self._once_per_run_reserved: set[str] = set()
         self._cancel_event = threading.Event()
         self._previous_summary: str = ""
         self._persistent_memory = persistent_memory
@@ -1041,7 +1043,8 @@ class AgentLoop:
             self._cancel_event.clear()
         else:
             self._has_run = True
-        self._called_ok = set()
+        self._once_per_run_completed = set()
+        self._once_per_run_reserved = set()
         self._previous_summary = ""
         self._released_fallback = False
         self._released_fallback_reason = None
@@ -1929,15 +1932,6 @@ class AgentLoop:
                 continue
 
             tool_def = self.registry.get(tc.name)
-            is_repeatable = tool_def.repeatable if tool_def else False
-            if tc.name in self._called_ok and not is_repeatable:
-                logger.warning(f"Blocked duplicate call: {tc.name} (already succeeded)")
-                skip_msg = json.dumps({"skipped": True, "reason": f"{tc.name} already completed successfully. Use the previous result."})
-                messages.append(context.format_tool_result(tc.id, tc.name, skip_msg))
-                trace.write({"type": "tool_skipped", "iter": iteration, "tool": tc.name})
-                react_trace.append({"type": "tool_skipped", "tool": tc.name})
-                continue
-
             if self._grounding is not None:
                 authorization = self._grounding.authorize_tool_call(
                     tc.name,
@@ -1958,14 +1952,34 @@ class AgentLoop:
                     )
                     continue
 
+            policy = self._invocation_policy(tool_def)
+            call_key = self._identical_call_key(tc.name, tc.arguments)
+            if policy == InvocationPolicy.ONCE_PER_RUN:
+                if (
+                    tc.name in self._once_per_run_completed
+                    or tc.name in self._once_per_run_reserved
+                ):
+                    self._record_invocation_skip(
+                        tc,
+                        policy=policy,
+                        call_key=call_key,
+                        context=context,
+                        messages=messages,
+                        trace=trace,
+                        react_trace=react_trace,
+                        iteration=iteration,
+                    )
+                    continue
+                self._once_per_run_reserved.add(tc.name)
+
             # Deterministic tools (e.g. financial_rigor calc) return the same
             # result for the same args. Checked AFTER authorization above so a
             # cached repeat can never bypass the identity gate. After auto-compact cleared earlier
             # tool outputs, the model used to re-run identical expressions
             # 5-9x each (2026-08-20 INTC run) to re-verify numbers it could
             # no longer see. Serve an identical prior call from cache instead.
-            if tool_def is not None and getattr(tool_def, "deterministic", False):
-                cache_key = self._identical_call_key(tc.name, tc.arguments)
+            if policy == InvocationPolicy.CACHE_IDENTICAL:
+                cache_key = call_key
                 if cache_key is not None and cache_key in self._called_identical:
                     cached = self._called_identical[cache_key]
                     messages.append(context.format_tool_result(tc.id, tc.name, cached))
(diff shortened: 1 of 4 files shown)


saju01:
Coming from #1341, where I'd independently hit and fixed the same defect. @warren618 pointed out the overlap and suggested converging here — agreed, and this is the more thorough treatment. The `InvocationPolicy` enum and the reserved-set for in-flight calls are both things my version lacks, and the race is a genuine gap in mine, not a stylistic difference.

Offering my reproduction as a test case, since it's a sharper trigger than a synthetic double-call — it's the shape that broke a real run:

```
get_financial_statements(statement="income")   -> ok
get_financial_statements(statement="balance")  -> BLOCKED
get_financial_statements(offset=13)            -> BLOCKED
```

The paging half is the part worth pinning. `_result_paging` advertises `next_offset` and `complete: false`, so paging is the *documented* way to finish a fetch — and the name-keyed gate made it structurally impossible to follow that contract. The failure is quiet: the model receives `{"skipped": true, "reason": "already completed successfully"}`, which is a fabricated success, and then correctly reports that the balance sheet "returned no readable content". The run looks like a model hallucination when it's actually the loop deleting its own data. Measured on an ATRC.US run (SEC EDGAR XBRL): 2 → 22 distinct successful data pulls on the same prompt once the gate compares arguments.

Two cases you may want on top of what's here, both written against the `_drive` harness in `test_agent_loop_deterministic_cache.py`:

1. **Paged continuation re-executes** — same tool, same `statement`, different `offset`.
2. **Un-serialisable arguments never block.** `_identical_call_key` returns `None` there and the cache treats that as *never cache*; the gate has to treat it as *never block*, or every un-serialisable call collapses into a single identity and the second is skipped without running. Worth an explicit test whichever implementation lands.

Test file is at `agent/tests/test_agent_loop_dedup_arguments.py` on my branch if it's useful to lift wholesale — please take it directly, no attribution needed. Happy to open it as a PR against this branch instead if that's easier.

AirHua-byte:
Thanks for the sharp reproduction. I added the ATRC.US paging regression in f1aa9d1: the same tool and statement run again with `offset=13`, with no `tool_skipped`.

Local full suite: 11,660 passed / 93 skipped; scripted dialog: 2 calls / 2 results / 0 skips; CI: 10/10 passed.

@warren618, #1268 is updated and ready for review.

warren618:
Holding this one on a question rather than a rebase, because #1349 and #1341 have both landed and they change what this PR is for.

`InvocationPolicy` currently resolves `is_readonly or repeatable` to `ALLOW`. Measured against the live registry, that moves **18 of 107 tools** from "blocked after the first success" to unlimited, with no replacement brake — `get_financial_statements`, `get_fund_flow`, `get_margin_trading`, `get_research_reports`, `get_stock_news`, `factor_analysis`, `etf_holdings`, `get_block_trades`, `get_dragon_tiger`, `get_institutional_holdings` and eight more. That is what makes #1343 go away here: not a narrower gate, but no gate.

The reason to reconsider is that the problem it was solving is now solved without loosening anything. #1341 rekeys the ledger to `(name, canonical arguments)`, so `get_financial_statements(statement='income')` and `(statement='balance')` are already two distinct requests and the second is no longer skipped. #1349 re-opens the ledger for any tool whose results compaction has cleared. Between them the false blocks are gone and the once-per-run brake is still there.

That brake is not hypothetical either — #1353 has an agent issuing tool calls for ~35 iterations without converging, and removing the per-run limit on read-only data tools is the direction that makes that failure mode cheaper to fall into, not harder.

What I think is still worth having here, and is not covered by either merged PR:

1. `InvocationPolicy` as an explicit declaration instead of inferring behaviour from two loosely-related booleans. That is a genuine improvement in its own right.
2. `_once_per_run_reserved` — the in-flight guard. Neither #1349 nor #1341 stops a parallel batch dispatching the same call twice concurrently, and that is a real gap.
3. `invocation_key` on the trace and skip payloads. #1343 took a lot of trace-reading to diagnose precisely because the skip records did not say which call identity was blocked.

So the question, and it is genuinely a question rather than a request dressed as one: is there a reason `ONCE_PER_RUN` has to be keyed on the bare tool name? Keying it on the same `(name, args)` identity `CACHE_IDENTICAL` already uses would let you keep all three of the above, drop `readonly → ALLOW` entirely, and leave the gate strictly narrower than it is today rather than wider. If there is a case that needs the full `ALLOW`, name it and we can special-case that instead of the whole read-only surface.

Happy to do the rebase onto the new main myself once we agree on the shape — no need for you to fight the conflicts with #1349/#1341.

saju01:
Thanks @AirHua-byte for adding the paging reproduction. I have now checked the post-#1341/#1349 main rather than relying on my earlier comparison of the two branches. The changed-argument/paging case works while exact readable repeats remain blocked, so I agree there is no need to make the whole read-only surface `ALLOW` to fix that case.

One integration detail for the rebase: compaction recovery still has to use the same exact identity as the gate. I found two residual cases on current main and opened #1358: a cleared income variant stays blocked while balance survives, and full summarizing compaction leaves old success locks behind. The follow-up reconciles successful readable results per `(name, args)` and only reopens known-read-only queries; it does not add an invocation policy or solve parallel-batch reservations.

Your explicit policy declaration, in-flight reservation and identity-bearing traces remain distinct work. Please keep the paging regression and the negative control for a still-readable identical call when rebasing. My follow-up adds 18 compaction/cache/safety cases; 448 tests pass in the combined local selection.

warren618:
Closing this, and I want to be precise that it is not a judgement on the work — the parts of it that were load-bearing have landed, and the part that has not is worth its own PR rather than a rebase of this one.

Since this was opened, #1341 rekeyed the dedup ledger to `(name, canonical arguments)` and #1349 / #1358 reconciled that identity through both compaction paths. Your paging reproduction — `statement="income"` then `statement="balance"` then `offset=13` — is green on main today, and #1358 carries the negative control for a still-readable identical call. That was the defect this PR existed to fix, so keeping it open now means asking you to rebase 500 lines onto a tree that already fixed the bug.

The reason I am not merging it instead is the one I raised on 2026-09-05 and it still stands: `InvocationPolicy` resolves `is_readonly or repeatable` to `ALLOW`, which moves 18 of the 107 registered tools from "blocked after the first success" to unlimited with no replacement brake. #1353 is a live example of an agent issuing tool calls for ~35 iterations without converging; removing the per-run limit on read-only data tools makes that failure cheaper to fall into, not harder. Since #1341 solved the false blocks without loosening anything, the loosening no longer buys anything.

Two things from here are still genuinely missing on main and I would review either as its own PR, keyed on the same `(name, args)` identity rather than the bare tool name:

1. `_once_per_run_reserved`, the in-flight guard. Neither #1341 nor #1358 stops a parallel batch dispatching the same call twice concurrently. That is a real gap and yours is the only proposed fix for it.
2. `invocation_key` on the trace and skip payloads. #1343 took a lot of trace-reading to diagnose precisely because the skip records never said which call identity was blocked.

The explicit policy enum is a fair improvement on inferring behaviour from two loosely-related booleans, but it needs to arrive without `readonly → ALLOW` attached. Thank you for the paging reproduction on #1341 — it is what made that fix reviewable.

PR #1269, opened 2026-08-30
fix(agent): preserve explicit market context in grounding

AirHua-byte:
## Problem

`GroundingLedger` receives the original user message, but symbol resolution currently chooses from `search_symbol` candidates using only the tool's local `query`. An explicit constraint such as “A股恒瑞医药” is therefore lost at the tool boundary: an A-share/H-share candidate set remains ambiguous even though the user already selected the market.

This is a Harness context-propagation problem. Rewriting the whole request into a structured financial intent would fix one case while narrowing the agent's general task surface.

## Change

- add an immutable `ResolutionContext` beside the original conversation; the model still receives the unchanged user text
- extract only explicit market markers with their source message and character span; do not infer markets from profiles or financial common sense
- apply subject-specific, recency-aware precedence across conversation history, while ignoring stale global constraints and explicit reset phrases
- treat negated markets and constraint/candidate mismatches as fail-closed rather than positive authorization
- filter resolver candidates before the existing unique/strong candidate selection, preserving the original candidate list for audit
- record only constraint metadata in the grounding artifact, without copying the full conversation text
- provide `VIBE_CONTEXTUAL_IDENTITY_CONSTRAINTS=false` as an immediate rollback switch

## Behavior

- “A股恒瑞医药” locks `600276.SH` when A/H candidates are returned
- “比较恒瑞医药 A/H” preserves both listings and remains ambiguous
- no explicit market remains fail-closed
- mixed entities keep independent constraints, and one ambiguous entity does not retract another entity's lock
- a current subject-specific constraint overrides a current global constraint; the latest named historical constraint wins

## Out of scope

- no query rewrite, intent enum, structured request contract, or fixed portfolio workflow
- no provider request or MCP changes
- no dependency, lockfile, broker, mandate, order, permission, or authentication changes
- no Invocation Policy changes from PR #1268
- no capability retrieval, plan ledger, completion-state, or token-budget changes

## Validation

- `python -m pytest agent/tests/test_grounding_resolution_context.py agent/tests/test_agent_grounding.py agent/tests/test_agent_output_discipline.py agent/tests/test_agent_loop_trace.py agent/tests/test_env_schema.py -q` — 310 passed
- `python -m pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` — 11,669 passed, 93 skipped
- `python -m ruff check agent/src/agent/resolution_context.py agent/src/agent/grounding.py agent/src/agent/loop.py agent/src/config/env_schema.py agent/tests/test_grounding_resolution_context.py agent/tests/test_env_schema.py` — passed
- `git diff --check` — passed
- no live broker, external MCP, credential, or market-data call was made

## Risk and rollback

The change affects identity selection only when the conversation contains a recognized explicit market marker. Ambiguous, negated, missing, or mismatched constraints remain fail-closed. Set `VIBE_CONTEXTUAL_IDENTITY_CONSTRAINTS=false` to restore the previous selection behavior immediately, or revert the commit.

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -31,6 +31,12 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Iterable, Mapping, Sequence
 
+from src.agent.resolution_context import (
+    IdentityConstraint,
+    ResolutionContext,
+    candidate_market,
+)
+
 
 GROUNDING_ARTIFACT = "grounding_evidence.json"
 
@@ -863,6 +869,7 @@ class IdentityRecord:
     source_tool_call_id: str | None = None
     source: list[str] = field(default_factory=list)
     candidates: list[dict[str, Any]] = field(default_factory=list)
+    resolution_constraints: list[dict[str, Any]] = field(default_factory=list)
     version: int = 1
     updated_at: str = field(default_factory=_utc_now)
 
@@ -932,6 +939,7 @@ class GroundingLedger:
         run_dir: Path,
         user_message: str,
         history: Sequence[Mapping[str, Any]] | None = None,
+        contextual_identity_constraints: bool = True,
     ) -> None:
         """Create a ledger and seed only authoritative prior identities.
 
@@ -939,12 +947,19 @@ class GroundingLedger:
             run_dir: Active run directory.
             user_message: Current user request.
             history: Optional prior message history. It remains available to
-                the model, but is deliberately not an authorization source for
-                this run: stale identities from an earlier user subject must
-                not unlock a new subject's tools.
+                the model. Only explicit constraints whose clause names the
+                current resolver subject may carry forward; stale global
+                instructions cannot authorize a new subject.
+            contextual_identity_constraints: Whether explicit market words in
+                the original conversation may narrow resolver candidates.
         """
         self.run_dir = Path(run_dir)
         self.user_message = user_message
+        self.resolution_context = ResolutionContext.from_messages(
+            user_message,
+            history,
+            enabled=contextual_identity_constraints,
+        )
         self._identities: dict[str, IdentityRecord] = {}
         self._evidence: list[EvidenceRecord] = []
         self._tool_failures: list[dict[str, Any]] = []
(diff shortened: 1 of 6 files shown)


warren618:
Merging. This is the cleanest-shaped PR in the current grounding batch: a self-contained module, one config flag, and a single wiring line, and it merges onto current main with no conflicts despite everything that has landed in `grounding.py` today.

Verified rather than assumed:

- 254 tests pass across `test_grounding_resolution_context.py`, `test_env_schema.py` and `test_agent_grounding.py`; 347 across the wider grounding/resolution/env selection.
- The flag is genuinely two-sided — `contextual_identity_constraints=False` is exercised, not just the default — which is the thing I most expected to be missing on a default-on feature.
- Hardcoding `enabled=False` in the ledger fails 10 tests, so the behaviour is really pinned and not incidentally satisfied.
- I checked the `VIBE_CONTEXTUAL_IDENTITY_CONSTRAINTS` name against the schema before flagging it as off-convention, and it is fine: `VIBE_GOAL_SESSION_ID`, `VIBE_LIVE_AUTHORIZE_TIMEOUT_SECONDS` and `VIBE_TW_STOCK_DB` are all existing non-`VIBE_TRADING_` precedents.

One small gap, not a blocker. Deleting the wiring line in `AgentLoop.run` leaves every test in the file green, because `GroundingLedger` defaults the parameter to `True` — so the env switch could stop being honoured and only `VIBE_CONTEXTUAL_IDENTITY_CONSTRAINTS=0` users would notice. I tried to add a test for it and what I wrote constructed the ledger itself with the expected value, which proves nothing about the loop, so I deleted it rather than ship a test that looks like coverage and is not. A real one needs a seam in `run()`. Worth doing if you touch this again; not worth holding the PR for.

Also worth recording for whoever picks up `_infer_venue` next: it still returns `crypto_or_fx` for `XAU-USD` and `None` for `^SPX` and `XAUUSD=X` on main. #1282 addresses that and is currently blocked on an unrelated filter question.

warren618 (approved):
Self-contained module, flag tested both ways, mutation-checked; merging.

PR #1272, opened 2026-08-30
fix(grounding): keep full-width brackets out of clause split

### LINKED ISSUE #1260: [Bug] unsourced_symbol_figures misses figures when the ticker is wrapped in full-width parentheses （）
## Summary

`_validate_unsourced_symbols` requires the symbol and the figure to land in the
**same clause**. `_CLAUSE_SEPARATOR_RE` treats full-width `（）` and `【】` as clause
separators, so the most idiomatic Chinese way to write a ticker —
`公司名（代码）价格` — splits the symbol away from the number and the check never fires.

The identical sentence written with ASCII `()` is rejected correctly. The verdict
flips on parenthesis width alone.

This is a **false negative** in the gate — complementary to #955 / #1001, which were
false positives.

## Repro

No network, no LLM, no MCP. Save as `agent/repro_grounding.py` and run from the repo root.

```python
import json, sys, tempfile
from pathlib import Path
sys.path.insert(0, str(Path(__file__).resolve().parent))

from src.agent.grounding import GroundingLedger, _split_clauses

USER_MESSAGE = "What is Kweichow Moutai (600519.SH) trading at this week?"

# One get_market_data result: 600519.SH becomes a session symbol, 000858.SZ never does.
MARKET_RESULT = json.dumps({
    "600519.SH": [{"trade_date": "2026-08-28T00:00:00", "open": 1289.0, "high": 1297.89,
                   "low": 1288.0, "close": 1297.4, "volume": 16126.11}],
    "_provenance": {"600519.SH": {"source": "baostock", "fallback_used": False,
                                  "currency_conversion": "none", "volume_unit": "lots"}},
})

# The two drafts differ by ONE character pair. Both attach a price to 000858.SZ,
# which no tool call ever passed in or returned.
DRAFT_FULLWIDTH = "同期五粮液（000858.SZ）收 168.50 元。"
DRAFT_HALFWIDTH = "同期五粮液(000858.SZ)收 168.50 元。"

def verdict(draft):
    with tempfile.TemporaryDirectory() as td:
        led = GroundingLedger(run_dir=Path(td), user_message=USER_MESSAGE, history=None)
        led.ingest_tool_result(
            tool_name="get_market_data",
            arguments={"codes": ["600519.SH"], "start_date": "2026-08-24",
                       "end_date": "2026-08-28", "source": "baostock"},
            result=MARKET_RESULT, call_id="c1", success=True)
        return [i for i in led.validate_final_answer(draft).issues
                if i.get("code") == "unsourced_symbol_figures"]

print("clause split, full-width:", _split_clauses(DRAFT_FULLWIDTH))
print("clause split, half-width:", _split_clauses(DRAFT_HALFWIDTH))
print("full-width （） ->", len(verdict(DRAFT_FULLWIDTH)), "issue(s)")
print("half-width ()   ->", len(verdict(DRAFT_HALFWIDTH)), "issue(s)")
```

### Actual output

```
clause split, full-width: ['同期五粮液', '000858.SZ', '收 168.50 元', '']
clause split, half-width: ['同期五粮液(000858.SZ)收 168.50 元', '']
full-width （） -> 0 issue(s)     <-- expected 1
half-width ()   -> 1 issue(s)     <-- correct
```

### Expected

Both forms report `unsourced_symbol_figures` for `000858.SZ`.

## Root cause

`agent/src/agent/grounding.py` (line numbers from `main` @ `e90b6c6`):

```python
# line 455
_CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n（）【】]")
```

```python
# _validate_unsourced_symbols
for segment in _split_clauses(line):
    unknown = sorted(... _scan_symbols(segment) - self._session_symbols ...)
    if not unknown or not self._numbers_without_dates_or_percent(segment):
        continue          # <-- symbol and figure are now in different segments
```

Splitting on `（）` puts the symbol in its own segment (no number) and the price in the
next one (no symbol), so the guard `continue`s.

Note the comment directly above the regex already states the intent for ASCII parens:

> `"(8.5 - 7.9) / 2"` must stay in one segment for the formula check.

Paired full-width brackets are grouping characters for exactly the same reason; they
appear to have been grouped with the genuine separators (`，；。、`) by oversight.

## Impact

For Chinese-language answers — the default for A-share research — `公司名（代码）价格` is
the standard citation form. In that form the mechanically decidable half of
*"what the tools did not return, you do not supply"* is bypassed, which is precisely the
hallucination this check exists to stop: a price attached to an instrument no tool in the
run ever handled.

`_validate_price_claims` is unaffected (it resolves at whole-answer scope, not clause
scope), so a figure that *contradicts* fetched evidence is still caught. Only the
*never-fetched instrument* case leaks.

## Proposed fix

Drop the paired full-width brackets from the separator class, leaving the genuine
clause separators:

```python
-_CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n（）【】]")
+_CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n]")
```

This makes `（）【】` behave like ASCII `()`, which the regex already excludes.

### Validation

Applied in memory (via a pytest plugin rebinding `_CLAUSE_SEPARATOR_RE`, no source edit)
and ran the grounding suite:

```
tests/test_agent_grounding.py
tests/test_grounding_nan_volume.py
tests/test_grounding_recovery.py
tests/test_swarm_grounding.py

baseline:      186 passed
with the fix:  186 passed
```

The repro above flips to `1 issue(s)` for both forms, and a correct draft naming the
locked symbol with its source and as-of date still passes (no new false positive).

I did not check whether a wider separator set (e.g. `《》「」`) has the same issue.

## Interface

Web UI (`vibe-trading serve`) — the defect is in `agent/src/agent/grounding.py` itself,
so the CLI path is affected identically. The repro touches neither.

## LLM Provider

N/A — the repro involves no LLM call. `GroundingLedger` is provider-independent by design.

## Version

`0.1.14` (`7329cb0`); re-confirmed present on `main` (`e90b6c6`) — line 455 and
`_validate_unsourced_symbols` are unchanged there.

## Environment

macOS 15 (Darwin 25.6.0, Apple Silicon) / Python 3.13

lorenzozanee:
## Summary

- Fix full-width brackets being treated as clause separators so `unsourced_symbol_figures` detects figures wrapped in （）【】

## Why

Full-width brackets split `公司名（代码）价格` into separate clauses, separating the symbol from its figure and causing a false negative while half-width `()` worked correctly. Fixes #1260.

## Changes

- Keep full-width brackets out of the clause separator so both bracket forms stay in one segment

## Test Plan

## Checklist

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -449,10 +449,12 @@ _PROSPECTIVE_LEVEL_RE = re.compile(
     r")",
     re.IGNORECASE,
 )
-# Full-width brackets and enumeration commas delimit prose clauses. ASCII
+# Enumeration commas and clause terminators delimit prose clauses. ASCII
 # parentheses are deliberately not separators: an explicit derivation such as
-# "(8.5 - 7.9) / 2" must stay in one segment for the formula check.
-_CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n（）【】]")
+# "(8.5 - 7.9) / 2" must stay in one segment for the formula check. Full-width
+# brackets （）【】 are grouping characters for the same reason and must also
+# stay in one segment so "公司名（代码）价格" keeps the symbol and figure together.
+_CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n]")
 
 
 # The ASCII comma both separates clauses and groups thousands, and the clause


warren618:
Thanks — this diagnosis is right, and it's the same one-line fix that landed in #1326, which I merged. Both remove `（）【】` from `_CLAUSE_SEPARATOR_RE` for the same reason: full-width brackets group, they don't terminate a clause, so splitting on them put `公司名（代码）` and the price in different segments and the unsourced-symbol gate never saw them together.

#1326 got the nod on test coverage: it parametrizes both parenthesis widths, so the full-width case fires and the half-width case stays as the control that already passed. That asymmetry is what made #1260 invisible in the first place, so pinning both sides matters here.

Closing as covered, not as wrong — you found it independently. Your #1276 and #1320 are both merged.

Issue #1275, opened 2026-08-30
bug(metrics): rebalance_count can disagree with actual target-rebalance fills


thisisjun786:
## Summary

`rebalance_count` is derived from changes in `target_positions`, not from actual rebalance execution. A strategy with constant target weights and `position_adjustment=\"rebalance\"` can execute drift corrections on many bars while reporting one rebalance.

## Reproduction

A SPY/BND constant-target run produced:

- `trade_count`: 1,269 target-adjustment trades;
- `rebalance_count`: 1.

The target changed once at entry, but the engine re-sized the drifting book repeatedly.

## Expected

Separate machine-readable concepts:

- requested target-change count (current rebalance notes behavior);
- executed rebalance bar/fill count derived from immutable fill evidence;
- realized rebalance turnover.

Keep existing field semantics only if renamed or clearly labeled; reports should not imply that target-change count is execution count.

## Why it matters

Backtest users can believe a monthly/annual strategy traded once while fees and turnover include hundreds of actual adjustments. Provenance-sensitive strategy parity must use fills, not a misleading aggregate.

cgycorey:
I raised a fix for this: #1281 — separates requested target changes (target_change_count) from executed rebalance counts/turnover derived from fill evidence.

PR #1278, opened 2026-08-30
fix(backtest): scale target-rebalance basket when 100% weight plus commission exceeds capital

### LINKED ISSUE #1274: bug(rebalance): 100% target plus commission aborts instead of scaling basket
## Summary

A rebalance-mode portfolio whose target weights sum to 100% fails with `ValueError: insufficient capital for position rebalance` when commission is nonzero, even though the engine's initial open path already has a common basket scale-factor mechanism.

## Reproduction

- initial cash: 100000
- target weights: SPY 0.60, BND 0.40
- commission: 0.001
- `position_adjustment=\"rebalance\"`

Observed: `_execute_target_rebalance` projects target notional plus fees beyond capital and aborts atomically.

## Expected

Apply one common scale factor to every sleeve after reductions and before committing increases/opens, preserving portfolio proportions and code-order independence. This should mirror the fairness behavior already used by the normal open basket path.

## Notes

`agent/tests/test_base_engine.py::test_rebalance_insufficient_capital_is_atomic` currently pins the abort. If scaling is accepted, replace/extend that contract with tests proving:

- weights preserve their ratio;
- total cost including commission fits capital;
- input code order does not change fills;
- irrecoverably infeasible baskets still fail atomically.

lorenzozanee:
## Summary

- Scale the rebalance basket with a common factor when target weights sum to 100% and commission pushes total cost above available capital, instead of raising `ValueError`
- Preserve portfolio proportions and code-order independence by mirroring the fairness scaling already used in the normal open basket path
- Keep atomic failure for irrecoverably infeasible baskets

## Why

In `position_adjustment=rebalance` mode a basket with weights summing to 100% plus nonzero commission aborted with `ValueError: insufficient capital for position rebalance`. Expected behavior is to apply one common scale factor to every sleeve after reductions and before committing increases/opens, as the hold-mode open basket already does (`agent/backtest/engines/base.py:1136`).

Closes #1274

## Changes

- `agent/backtest/engines/base.py`: add `_plans(scale)` helper inside `_execute_target_rebalance` and 50-iteration binary search to find the largest feasible scale when projected capital is negative; fall back to `ValueError` only when no scale fits
- `agent/tests/test_base_engine.py`: update pinned abort contract to assert scaling (ratio preservation, cost fits capital, order independence) and keep atomic failure for a high-fee reduction case

## Test Plan

- [x] Existing tests pass (`pytest agent/tests/test_base_engine.py --tb=short -q` — 91 passed, 2 pre-existing unrelated failures)
- [x] New tests added
- [x] Tested manually: 60/40 basket with 0.1% commission scales preserving 1.5 ratio and fits capital; single 100% with 10% commission scales to ~0.909; order independence verified; irrecoverably infeasible (high exit commission) still fails atomically

## Checklist

- [x] No hardcoded values

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -1617,6 +1617,103 @@ class BaseEngine(ABC):
             elif self.can_execute(symbol, 0, bar):
                 reductions.append(self._plan_reduction(before, target_size, price))
 
+        def _plans(scale: float) -> tuple[list[_ReductionOrder], list[_OpenOrder]]:
+            reds: list[_ReductionOrder] = []
+            ops: list[_OpenOrder] = []
+            for sym in sorted(codes):
+                tw = target_weights.get(sym)
+                if tw is None:
+                    continue
+                self._validate_rebalance_values(tw)
+                scaled_w = tw * scale
+                self._active_symbol = sym
+                before = self.positions.get(sym)
+                target_direction = 1 if scaled_w > 1e-9 else (-1 if scaled_w < -1e-9 else 0)
+                frame = data_map.get(sym)
+                if frame is None or ts not in frame.index:
+                    continue
+                bar = frame.loc[ts]
+                if before is None or target_direction != before.direction:
+                    if before is not None:
+                        if not self.can_execute(sym, 0, bar):
+                            continue
+                        raw_price = self.execution_open(bar)
+                        self._validate_rebalance_values(raw_price, positive=True)
+                        price = self.apply_slippage(raw_price, -before.direction)
+                        self._validate_rebalance_values(price, positive=True)
+                        reds.append(self._plan_reduction(before, 0.0, price))
+                        if target_direction == 0:
+                            continue
+                    order = self._plan_open_order(
+                        sym,
+                        scaled_w,
+                        frame,
+                        ts,
+                        equity,
+                        allow_existing=before is not None,
+                        require_positive_price=True,
+                    )
+                    if order is not None:
+                        self._validate_rebalance_values(
+                            order.price, order.leverage, order.size, order.margin, order.commission
+                        )
+                        ops.append(order)
+                    continue
+                raw_price = self.execution_open(bar)
+                self._validate_rebalance_values(raw_price, positive=True)
+                leverage = self._leverage_for_symbol(sym)
+                self._validate_rebalance_values(raw_price, before.leverage, leverage)
+                target_notional = abs(scaled_w) * equity * before.leverage
+                prices = (
+                    self.apply_slippage(raw_price, before.direction),
+                    self.apply_slippage(raw_price, -before.direction),
+                )
+                self._validate_rebalance_values(*prices, positive=True)
+                sizes = tuple(
+                    self.round_size(
+                        self._calc_raw_size(sym, target_notional, price), price
+                    )
+                    for price in prices
+                )
+                self._validate_rebalance_values(*prices, *sizes)
+                if self.rebalance_tolerance > 0.0:
+                    reference_size = self.round_size(
+                        self._calc_raw_size(sym, target_notional, raw_price), raw_price
+                    )
+                    if abs(before.size - reference_size) <= self.rebalance_tolerance * abs(
+                        reference_size
+                    ):
+                        continue
+                increase = sizes[0] > before.size + 1e-9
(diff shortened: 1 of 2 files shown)


warren618:
Holding this on one thing.

`_plans(scale)` is a near line-for-line copy of the planning loop directly above it — the same tolerance band, the same direction-flip handling, the same `_OpenOrder` construction. `_execute_target_rebalance` would end up carrying two copies of order planning that have to stay in sync forever, and the next person who fixes the tolerance logic will fix one of them.

Since `_plans(1.0)` *is* the original loop, the shape that gets the same behavior with less code is to build the initial plan with it:

```python
reductions, opens = _plans(1.0)
projected_capital = (
    self.capital
    + sum(order.capital_credit for order in reductions)
    - sum(order.cost for order in opens)
)
if projected_capital < -1e-9:
    ...binary search...
```

That deletes the original block instead of duplicating it, and it makes the scaled and unscaled paths incapable of drifting apart.

Two smaller things:

- `test_rebalance_insufficient_capital_is_atomic` now asserts that scaling happens — the name says the opposite of what it tests. Rename it for the new behavior and let `test_rebalance_irrecoverably_infeasible_is_atomic` carry the atomicity contract under its own name.
- The feasibility predicate is not strictly monotonic in `scale`: at a smaller scale a sleeve can fall inside `rebalance_tolerance` and drop out of the plan entirely, which moves projected capital discontinuously. The search is still safe because you only ever keep feasible candidates, but it can return a feasible non-maximal basket. Worth a comment so nobody later reads the 50 iterations as convergence to the optimum.

The fairness-scaling precedent you cite at `base.py:1136` checks out, and I agree with making rebalance behave like the open basket rather than aborting — #1274 is a real bug. Collapse the duplication and this goes in.

he-yufeng (commented):
Read the full diff and ran the rebalance suite on this branch locally: 61 tests green, including the new scaling, ratio, order-independence, and irrecoverable-atomicity cases. The approach is the right one: scaling all targets by one factor preserves proportions, feasibility is monotone in the scale so the bisection is sound, and rewriting the old abort-pin into a scale-pin is exactly what #1274 asked for. The atomic-unchanged guarantee for the irrecoverable case is still tested, which matters.

One structural thing I think has to change before this lands. `_plans(scale)` is a full copy of the inline planning block above it (the diff is +117/-0 in base.py), and the copy is already stale: your branch is based on 899d3c7, and since then #1275 (rebalance_count separation) and #1277 (the rebalance_mask schedule) both changed the rebalance path on main. With two planners, the scale=1.0 path gets main's mask semantics and the scaled path silently runs the pre-mask planner, so a masked basket that trips the scaling branch would execute on bars the mask excludes. Two copies also means every future planner fix has to be made twice.

Could you rebase onto current main and collapse it to one planner, with the original block calling `_plans(1.0)`? Then the mask handling lives in exactly one place and the scaling search inherits it for free.

Smaller note, not a blocker: 50 bisection rounds re-plan the whole basket each time. The fitting scale is almost analytic (available capital over required cost at scale 1), so seeding the search with that ratio and capping at a handful of verification rounds would do the same job much cheaper on long daily runs. Fine to leave as is if you'd rather keep it dumb.

warren618:
Closing this in favour of #1344, which has landed for #1274. Thanks for the work — the diagnosis and the binary-search approach were both right, and the two PRs agree on the core idea. The choice came down to one implementation detail rather than the idea.

`_plans(scale)` re-runs the whole planning pass inside the fifty-iteration search, and it reaches `_plan_open_order`, which records `_on_plan_rejected` in seven separate branches (`no_target_weight`, `already_held`, `no_data`, `no_bar`, `execution_blocked`, `invalid_price`, `zero_size`). A symbol rejected at small scales would therefore be recorded up to fifty times per bar, and the run-card rejection diagnostics — which exist so a user can see why a leg did not fill — would be flooded with duplicates of the same event. #1344 searches over already-planned orders and records each dropped sleeve once, after the search settles.

The re-planning approach also duplicates roughly a hundred lines of the planning loop, which is a second copy to keep in step with every future change to sizing, slippage or lot rules.

None of that is visible from the tests, and it is not the kind of thing a review catches by reading the diff alone — I only found it by tracing what `_plan_open_order` does on the way through. If you want to keep working in this area, the `_execute_target_rebalance` path could genuinely use a second pair of eyes on the reduction/open interaction.

PR #1279, opened 2026-08-30
fix(backtest): scale target-rebalance basket when it exceeds capital

cgycorey:
## Problem

With `position_adjustment=rebalance`, a target basket whose weights sum to 100% plus nonzero commissions exceeds available capital and previously aborted with `ValueError: insufficient capital for position rebalance`. A recoverable basket (one that fits at some common scale) should scale down instead of failing, just like the open-basket path already does.

## Change

- Refactor the rebalance planner into one scaled closure `_plan_basket(scale)` used for both the full-scale plan and every bisection candidate — no duplicated planning logic that could drift.
- If the full-scale basket does not fit, bisect the common scale factor (50 iterations, largest scale that fits) and apply it to every sleeve: portfolio proportions (weight ratios) and code-order independence are preserved.
- Keep the atomic abort for irrecoverably infeasible baskets: if even full closeouts cannot make the projected capital non-negative, raise `ValueError` with no positions touched.

## Tests

- `test_rebalance_insufficient_capital_scales_basket_to_fit` — 100% weight + 10% fee scales to exactly fit capital.
- `test_rebalance_scaled_basket_preserves_weight_ratios` — 10%/90% ratio kept after scaling.
- `test_rebalance_scaled_basket_is_independent_of_input_code_order` — same outcome for both code orders.
- `test_rebalance_irrecoverably_infeasible_basket_is_atomic` — margin-destroying closeout still aborts unchanged.

Related: #1278 addresses the same issue with a similar bisection; this PR is an alternative implementation (single planner for both paths, plus a defensive final capital check).

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -19,7 +19,7 @@ from collections import Counter
 from collections.abc import Iterable
 from dataclasses import dataclass, replace
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


cgycorey:
Closing: #1278 implements the same fix already; this PR is superseded by it.

PR #1280, opened 2026-08-30
fix(market-identity): route FX, metals, and futures symbols correctly

aminak58:
## Problem

Market identity resolution has three inconsistencies:

  1. **`XAUUSD` and `EURUSD` were not recognized by any market classifier.**
     `backtest.engines._market_hooks._detect_market` and
     `backtest.correlation.infer_market` both fell through to `a_share`.
     A `get_market_data(["XAUUSD"])` call walked the A-share chain
     (Tencent, mootdx, Eastmoney, Baostock, akshare, tushare) and never
     reached the forex chain (mt5, tickerall, qveris, yfinance) that
     actually carries the broker's gold feed.

  2. **Yahoo's continuous-front-month futures notation (`GC=F`, `CL=F`,
     `SI=F`, `HG=F`, `MGC=F`) and forex notation (`XAUUSD=X`,
     `EURUSD=X`) were not recognized either.** Same fall-through.

  3. **`_canonical_crypto_pair` accepted any base with a `-USD` or
     `/USD` shape** because `USD` was in `_CRYPTO_QUOTE_ASSETS`. A bare
     `XAUUSD` query resolved to the crypto pair `XAU-USD`, and the
     public-venue catalog fallback (Binance / OKX) would either lock
     onto a tokenized-gold token (XAUT-USDT) from Binance or find
     nothing on OKX — never on real spot gold. This is a wrong-asset
     substitution: spot gold → tokenized gold.

## Fix

This change:

  * Adds three regex patterns to `_MARKET_PATTERNS` in
    `backtest/engines/_market_hooks.py`: Yahoo `=X` (forex) and `=F`
    (futures) notations, plus a whitelist-restricted 6-letter base for
    precious metals (XAU / XAG / XPT / XPD) and G10 currencies
    (EUR / GBP / JPY / CHF / CAD / AUD / NZD / USD). The whitelist is
    conservative; length-only patterns (`^[A-Z]{6}$`) were rejected
    because they would re-route legitimate US tickers (e.g. `NFLXLI`).
  * Mirrors the same three patterns in
    `backtest.correlation.infer_market` and the lightweight
    `_detect_market_for_align` in `backtest/engines/base.py` so all
    three classifiers agree.
  * Splits `_CRYPTO_QUOTE_ASSETS` in
    `src/tools/symbol_search_tool.py` into two tiers: stablecoin quotes
    (FDUSD / USDT / USDC / BUSD / TUSD) — accepted on any alphanumeric
    base — and `USD`, which is now gated on a base whitelist
    (`_CRYPTO_USD_BASES`, 26 entries) of well-known crypto assets. `XAU`
    / `EUR` / `GBP` / etc. are deliberately absent from the whitelist,
    so `XAUUSD` / `EURUSD` / `GBPUSD` queries no longer auto-lock onto
    a tokenized-gold row or a forex pair that the public venue catalogs
    do not list.

## Risk

  * Additive only: every new pattern sits BEFORE the existing
    US-equity fall-through, so it takes precedence only for the
    symbols it was designed for. No existing behavior is removed.
  * No new data providers, no new asset class, no engine changes —
    those are future work.
  * The `_CRYPTO_USD_BASES` whitelist is a static list of 26
    well-known Binance/OKX spot bases (BTC, ETH, BNB, SOL, ADA, XRP,
    DOGE, TRX, DOT, MATIC, AVAX, LINK, LTC, BCH, ETC, XLM, ATOM, FIL,
    APT, NEAR, ALGO, SAND, MANA, AXS, XAUT, PAXG). New listings
    require a follow-up; the list is conservative.
  * Tokenized gold (XAUT-USDT, PAXG-USDT) keeps resolving as crypto
    because the whitelist explicitly includes XAUT and PAXG.
  * The patch does not add a new `commodity` market, a new gold
    engine, a new data provider (Polygon / Databento / IBKR), or a
    gold skill. Those are out of scope.

## Tests

  * `agent/tests/test_correlation.py` — 5 new test methods in
    `TestInferMarket`:
    - `test_precious_metals_classify_as_forex` (XAUUSD, XAGUSD,
      XPTUSD, XPDUSD).
    - `test_g10_fx_pairs_classify_as_forex` (EURUSD, GBPUSD, USDJPY,
      USDCHF, AUDUSD, NZDUSD, USDCAD).
    - `test_yahoo_equals_notation_routes_to_underlying_market`
      (GC=F, CL=F, SI=F, HG=F, MGC=F, XAUUSD=X, EURUSD=X).
    - `test_6char_metals_whitelist_does_not_over_match_us_equities`
      (NFLXLI, AMZNLY, GLD, XAUT-USDT, PAXG-USDT).
    - `test_btcusdt_style_joined_pairs_stay_crypto` (BTCUSDT, ETHUSDT,
      SOLUSDT) — regression guard for the existing joined-pair fix.

  * `agent/tests/test_symbol_search_tool.py` — new
    `TestCryptoUsdBaseWhitelist` class with 24 cases:
    - 10 parametrized cases of allowed crypto pairs with supported
      base (BTC-USD, ETH-USD, SOL-USD, BNB-USD, XLM-USD, XRP-USD,
      BTC-USDT, XAUT-USDT, PAXG-USDT, ETH-USDC).
    - 9 parametrized cases of disallowed non-crypto bases (XAUUSD,
      XAU-USD, XAU/USD, EURUSD, EUR-USD, EUR/USD, GBPUSD, GBP-USD,
      JPYUSD).
    - `test_tokenized_gold_bases_still_resolve_as_crypto`.
    - `test_xauusd_and_xautusdt_are_not_equivalent_assets`.
    - `test_usd_whitelist_is_a_strict_subset_of_crypto_bases`.

  * All 842 tests across 21 modules pass after this change. No
    regressions in `test_agent_grounding`, `test_grounding_recovery`,
    `test_agent_output_discipline`, engine tests (`test_forex_engine`,
    `test_global_futures_engine`, `test_china_futures_engine`,
    `test_base_engine`, `test_crypto_engine`,
    `test_composite_engine_fallback`,
    `test_composite_currency_guard`, `test_entities_spine`), and all
    loader tests (`test_yfinance_loader`, `test_stooq_loader`,
    `test_tiingo_loader`, `test_finnhub_loader`,
    `test_alphavantage_loader`, `test_fmp_loader`,
    `test_tickerall_loader`, `test_ccxt_perpetual_loader`,
    `test_ccxt_loader_bounded`).

## Behavior change matrix

Market classification:

  | Query      | Before       | After     |
  | ---------- | ------------ | --------- |
  | XAUUSD     | a_share      | forex     |
  | XAGUSD     | a_share      | forex     |
  | XPTUSD     | a_share      | forex     |
  | XPDUSD     | a_share      | forex     |
  | EURUSD     | a_share      | forex     |
  | GBPUSD     | a_share      | forex     |
  | USDJPY     | a_share      | forex     |
  | GC=F       | a_share      | futures   |
  | CL=F       | a_share      | futures   |
  | SI=F       | a_share      | futures   |
  | HG=F       | a_share      | futures   |
  | MGC=F      | a_share      | futures   |
  | XAUUSD=X   | a_share      | forex     |
  | EURUSD=X   | a_share      | forex     |
  | XAUT-USDT  | crypto       | crypto    |
  | PAXG-USDT  | crypto       | crypto    |
  | BTC-USDT   | crypto       | crypto    |
  | GLD        | us_equity    | us_equity |

Crypto resolver (`_canonical_crypto_pair`):

  | Query       | Before           | After        |
  | ----------- | ---------------- | ------------ |
  | XAUUSD      | XAU-USD (crypto) | None         |
  | XAU-USD     | XAU-USD (crypto) | None         |
  | XAU/USD     | XAU-USD          | None         |
  | EURUSD      | EUR-USD (crypto) | None         |
  | EUR-USD     | EUR-USD (crypto) | None         |
  | BTC-USD     | BTC-USD          | BTC-USD      |
  | ETH-USD     | ETH-USD          | ETH-USD      |
  | XAUT-USD    | XAUT-USD         | XAUT-USD     |
  | PAXG-USD    | PAXG-USD         | PAXG-USD     |
  | BTC-USDT    | BTC-USDT         | BTC-USDT     |
  | XAUT-USDT   | XAUT-USDT        | XAUT-USDT    |

## Known limitations (not addressed in this PR)

  * `correlation.infer_market("BTC-USD")` returns `us_equity` (pre-
    existing inconsistency between the engine classifier and the
    correlation helper). One-line follow-up to add `"USD"` to
    `crypto_suffixes` is straightforward but out of scope here.
  * Tokenized gold with non-USDT quote (XAUT-USDC, XAUT-BUSD) is
    classified as `us_equity` by the engine and `a_share` by
    `correlation.infer_market` (pre-existing). The whitelist in
    `_CRYPTO_USD_BASES` accepts `XAUT-USD`; the engine's USDC/BUSD
    crypto pattern would need a parallel extension to reach parity.
  * No upstream data source provides pre-2000 CME gold history
    through this routing path. mt5 / tickerall / qveris / yfinance
    cover different windows; institutional history requires Polygon,
    Databento, or IBKR — explicitly future work.


warren618:
CI is red here: `agent/tests/test_market_detection.py::TestDetectMarket::test_unknown_defaults_to_a_share` fails on `assert _detect_market("EURUSD") == "a_share"` — it now resolves to `forex`.

That test pins the behavior you are deliberately changing, and I think you are right and the pin is stale: EURUSD is not an A-share. But it should be updated on purpose rather than as a side effect. Pick a symbol that genuinely is unclassifiable for the "unknown defaults" case, and add cases asserting EURUSD / XAUUSD land in `forex`. With that and a green run I'll take it.

One thing to check while you are in there: the bare-six-character alternation runs under `re.I`, so `[A-Z]{3}` matches lowercase input too and `eurusd` classifies as forex. Confirm that is intended for the alias path rather than incidental.

aminak58:
Thanks for the review. The fix is in on `fix/fx-metals-futures-routing` (commit `c81d5679`).

**Stale test** — `test_unknown_defaults_to_a_share` was indeed pinning the old behavior. I replaced the two stale `EURUSD` / `BTCUSDT` assertions with truly unclassifiable inputs (`12345`, `123456`, `@#$`) and updated the comment to point at the new inputs so the `a_share` default is still pinned, just not against symbols the routing layer now classifies.

**Explicit regression cases** — added three new tests in `TestDetectMarketRequired`:

  * `test_bare_fx_pair_classifies_as_forex` — pins `XAUUSD` / `EURUSD` / `XAGUSD` / `XPTUSD` / `XPDUSD` / `GBPUSD` / `USDJPY` → `forex`.
  * `test_yahoo_equals_notation_routes_to_underlying_market` — pins `GC=F` / `CL=F` / `SI=F` / `HG=F` / `MGC=F` → `futures` and `XAUUSD=X` / `EURUSD=X` → `forex`.
  * `test_bare_fx_pair_classification_is_case_insensitive` — pins `eurusd` / `xauusd` / `gbpusd` / `usdjpy` / `gc=f` / `eurusd=x` as the same market as their uppercase counterparts.

**Lowercase behavior** — intentional, not incidental. The bare-6-character alternation in `_MARKET_PATTERNS` is compiled with `re.I`, which is consistent with the project's case-insensitive convention for symbol matching (the existing `test_case_insensitive` already pins `000001.sz` / `aapl.us` / `btc-usdt` / `td.to`). I added the third test above to make the lowercase behavior explicit and auditable rather than implicit.

**Test result** — 927 tests pass across 22 modules on Python 3.13 (local). The previously red `test_market_detection.py::TestDetectMarket::test_unknown_defaults_to_a_share` is now green. CI on the new commit is queued; the next push-triggered run should clear `mergeStateStatus: BLOCKED` (the previous red run is still being represented in the mergeability check until the new run completes).

Production code was not modified in this commit. Only `agent/tests/test_market_detection.py` changed (45 insertions, 3 deletions).

No changes to `origin`. PR #1280 is unchanged except for the new commit on its head.

warren618:
Rebased onto current main and resolved the conflicts here rather than sending it back, since main has since gained shared FX canonicalisation and index routing (bb41643c, 0c2b516e) that overlap this branch.

First, the check that mattered: this PR is **not** superseded by that work. Measured on current main, 9 of the 11 routing assertions here still fail — `GC=F`, `CL=F`, `SI=F`, `HG=F`, `MGC=F`, `XAUUSD`, `XAGUSD`, `EURUSD`, `USDJPY` all classify as `a_share`. Yahoo's gold-futures symbol being routed to the China A-share chain is worth fixing on its own.

All three conflicts are resolved as unions, because in each case the two sides cover different symbols:

- `_MARKET_PATTERNS`: kept your `=F` futures rule and the bare precious-metal / G10 whitelist, kept main's `^...` index rule, and kept main's `{3,6}` width on `=X` rather than this branch's `{6}` — the branch only ever asserts 6-character pairs, so narrowing would drop shorter forms for no gain.
- `_canonical_crypto_pair`, both call sites: applied **both** rejections. main's fiat/fiat rule does not catch `XAU-USD`, because XAU is a metal code and not a fiat code — that is the exact symbol this PR exists to stop resolving to tokenized gold. Your USD whitelist does not catch a fiat pair quoted in something other than USD. They also differ in kind deliberately, and I preserved that: fiat/fiat has no crypto reading and returns, while a non-whitelisted USD base only rules out that one quote asset and must `continue` so `USDT`/`USDC` can still match, or `XAUT-USDT` would be stranded.
- Test file: both classes kept, `TestCryptoUsdBaseWhitelist` from here and `TestFxPairAlignment` from main.

Verified after the merge: 11/11 of your assertions pass, and `^SPX` / `^VIX` / `GBPCNY=X` / `AAPL.US` / `BRK.B.US` / `600519.SH` / `BTC-USDT` / `TD.TO` / `VOD.L` / `AAPL` all still route as they do on main. 1642 tests pass across the symbol/market/identity/loader selection.

Please review the resolution before this goes in — the routing table is sensitive and I would rather you confirm the metals whitelist reads the way you intended.

warren618 (approved):

PR #1282, opened 2026-08-30
fix(symbol-resolution): route spot gold aliases through the right path

aminak58:
## Problem

Runtime validation against the Vibe-Trading agent (Mimo) exposed two
distinct issues that PR #1280 did not cover:

  1. **`SymbolSearchTool.execute` had an inverted filter gate.** The
     PR #1234 filter (added in `SymbolSearchTool.execute`) was:

     ```python
     if crypto_pair is not None:
         candidates = [
             c for c in candidates
             if _canonical_crypto_pair(c["symbol"]) == crypto_pair
         ]
     ```

     This only fired for queries the resolver recognized as a crypto
     pair. A bare `XAUUSD` query produces `crypto_pair = None`
     (the upstream `_canonical_crypto_pair` correctly returns `None`
     after PR #1280's `USD` whitelist safeguard). The filter then does
     NOT run, and a near-string Yahoo free-text hit such as
     `VALOUR-BTC-0-SEK.ST` (a Swedish Bitcoin ETP) survives to the
     identity ledger as a candidate for a gold query. Runtime
     observation: `search_symbol("XAUUSD")` returns 1 candidate
     `VALOUR-BTC-0-SEK.ST` that is not gold.

  2. **`_infer_venue` and `_infer_instrument_type` in
     `src/agent/grounding.py` had a shape-based fallback** that
     classified any symbol containing `-` or `/` as
     `crypto_or_fx` / `crypto`. `XAU-USD` and `XAU/USD` were both
     mis-classified. The runtime registry reported
     `instrument_type = crypto, venue = crypto_or_fx` for `XAU-USD`,
     contradicting the engine classifier (`_detect_market` returns
     `forex`).

## Root Cause

Both bugs are shape-vs-semantic mismatches. The fix has to know
which dashed / slashed / `=F` / `=X` symbols are crypto and which
are forex / futures. The fix mirrors the 26-entry
`_CRYPTO_USD_BASES` whitelist and the 5-entry stablecoin-quote set
that `_canonical_crypto_pair` (in `src.tools.symbol_search_tool`)
already maintains, so the runtime registry agrees with the engine
classifier and the resolver.

## Fix

  * `src/tools/symbol_search_tool.py` (+19 lines): broaden the
    PR #1234 filter to also fire when the query is a non-crypto
    ticker shape (separator, `=F`, `=X`). For those queries,
    drop any candidate whose canonical form IS a crypto pair. A
    bare name query (`apple`, `tesla`, `Gold`) falls through
    to the existing free-text behavior because it has no separator
    or `=F`/`=X` form.

  * `src/agent/grounding.py` (+72 lines): mirror the metal/FX/
    futures patterns in `_infer_venue` and
    `_infer_instrument_type`. The dashed/slashed fallback now
    consults the same 26-entry `_CRYPTO_USD_BASES` whitelist
    and the 5-entry `_STABLECOIN_QUOTES` set that
    `_canonical_crypto_pair` already uses. `XAU-USD` /
    `EUR-USD` / `GBP-USD` resolve as `forex`. `BTC-USDT` /
    `XAUT-USDT` keep resolving as `crypto`.

This is the third copy of these patterns. PR #1280 has two
(`_market_hooks._MARKET_PATTERNS` and `correlation.infer_market`).
A future refactor can lift them into a shared module.

## Risk

  * Additive only: the new `elif` branch in
    `SymbolSearchTool.execute` fires only when the existing
    `if crypto_pair is not None:` branch does not. Existing
    behavior is preserved for crypto queries and free-text name
    queries.
  * The whitelist constants are duplicated a third time. Drift
    risk is low (the upstream `_canonical_crypto_pair` is the
    ground truth) and the architectural review has already noted
    this as a future refactor.
  * No new data providers, no new asset class, no engine
    changes, no new skill, no changes to `ForexEngine` /
    `GlobalFuturesEngine` / loader layer.
  * Tokenized gold (`XAUT-USDT` / `PAXG-USDT`) keeps resolving
    as `crypto`; the whitelist explicitly includes `XAUT` and
    `PAXG`.
  * US-equity and free-text name behavior is unchanged. The
    test `test_free_text_name_query_unaffected` pins this.

## Tests

  * `test_agent_grounding.py::test_runtime_registry_classifies_gold_fx_futures_consistently`:
    parametrized over 10 symbols (`XAUUSD`, `XAU-USD`, `XAU/USD`,
    `XAUUSD=X`, `GC=F`, `XAUT-USDT`, `PAXG-USDT`, `BTC-USDT`,
    `GLD`, `AAPL.US`) and asserts `_infer_venue` and
    `_infer_instrument_type` for each.
  * `test_symbol_search_tool.py::TestSpotGoldCandidateFilter`: 7
    tests pinning the filter behavior end-to-end:
    - `XAUUSD` drops near-string BTC ETP.
    - `XAU/USD` drops near-string crypto hits.
    - `XAUUSD=X` drops near-string crypto hits.
    - `GC=F` keeps the correct futures hit.
    - `XAUT-USDT` keeps tokenized gold.
    - `PAXG-USDT` keeps tokenized gold.
    - Free-text name query (`apple`) is unaffected.

## Behavior matrix

| Query        | Old: count | New: count | Old: instrument_type   | New: instrument_type |
| ------------ | ---------- | ---------- | ---------------------- | -------------------- |
| XAUUSD       | 1 (wrong)  | 0          | listed_security        | forex                |
| XAU-USD      | 0          | 0          | crypto                 | forex                |
| XAU/USD      | 0          | 0          | crypto                 | forex                |
| XAUUSD=X     | 0          | 0          | listed_security        | forex                |
| GC=F         | 1          | 1          | future                 | future               |
| XAUT-USDT    | 1          | 1          | crypto                 | crypto               |
| PAXG-USDT    | 1          | 1          | crypto                 | crypto               |
| BTC-USDT     | 1          | 1          | crypto                 | crypto               |
| GLD          | n/a        | n/a        | listed_security        | listed_security      |
| AAPL.US      | n/a        | n/a        | listed_security        | listed_security      |
| apple (free) | n/a        | n/a        | (existing behavior)    | (existing behavior)  |

832 tests pass across 21 modules. No regressions.

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -793,10 +793,48 @@ def _infer_venue(symbol: str) -> str | None:
     for suffix, venue in suffixes.items():
         if upper.endswith(suffix):
             return venue
-    if "-" in upper or "/" in upper:
-        return "crypto_or_fx"
+    # Yahoo's continuous-front-month futures notation (GC=F, CL=F, SI=F, ...).
+    # The exchange category is the venue class. The engine and the
+    # correlation helper mirror this pattern; this is the third copy.
     if upper.endswith("=F"):
         return "futures"
+    # Yahoo's forex notation (XAUUSD=X, EURUSD=X) is FX.
+    if re.match(r"^[A-Z]{6}=X$", upper):
+        return "forex"
+    # Bare 6-character precious-metal / FX symbols. The whitelist is
+    # ISO 4217 metals + G10 currencies; it intentionally does NOT include
+    # any US-equity prefix. Mirroring the engine ``_MARKET_PATTERNS``.
+    if re.match(
+        r"^(?:XAU|XAG|XPT|XPD|EUR|GBP|JPY|CHF|CAD|AUD|NZD|USD)[A-Z]{3}$",
+        upper,
+    ):
+        return "forex"
+    # Dashed / slashed symbols are NOT categorically crypto. The crypto
+    # resolver (``_canonical_crypto_pair`` in ``src.tools.symbol_search_tool``)
+    # gates the ``USD`` quote leg on a 26-entry base whitelist of Binance /
+    # OKX spot bases (BTC / ETH / BNB / SOL / ADA / XRP / DOGE / TRX / DOT /
+    # MATIC / AVAX / LINK / LTC / BCH / ETC / XLM / ATOM / FIL / APT / NEAR /
+    # ALGO / SAND / MANA / AXS / XAUT / PAXG). ``XAU-USD``, ``EUR-USD``,
+    # ``GBP-USD`` are not in that whitelist, so they are NOT crypto and
+    # are NOT crypto_or_fx — they are forex. Stablecoin quotes (USDT /
+    # USDC / BUSD / TUSD / FDUSD) are unambiguous and accept any alphanumeric
+    # base. Without this guard a bare ``XAUUSD`` query would surface as a
+    # crypto-or-fx hybrid in the runtime registry, contradicting the engine
+    # classifier which already routes it to ``forex`` (see PR #1280).
+    if "-" in upper or "/" in upper:
+        base, sep, quote = upper.partition("-") if "-" in upper else upper.partition("/")
+        if quote in {"FDUSD", "USDT", "USDC", "BUSD", "TUSD"}:
+            return "crypto_or_fx"
+        if quote == "USD" and base in {
+            "BTC", "ETH", "BNB", "SOL", "ADA", "XRP", "DOGE", "TRX", "DOT",
+            "MATIC", "AVAX", "LINK", "LTC", "BCH", "ETC", "XLM", "ATOM",
+            "FIL", "APT", "NEAR", "ALGO", "SAND", "MANA", "AXS", "XAUT", "PAXG",
+        }:
+            return "crypto_or_fx"
+        # Any other dashed / slashed pair is forex-shaped (e.g. ``XAU-USD``,
+        # ``EUR-USD``, ``GBP-USD``); the per-pair engine classifier decides
+        # ``forex`` vs ``crypto`` vs ``futures`` downstream.
+        return "forex"
     return None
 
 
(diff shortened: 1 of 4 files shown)


warren618:
Stopping short of pushing a rebase on this one, and I want to be straight about why rather than sit on it: I rebased it, found that the filter no longer does what it was written to do, and could not fully explain the old behaviour — so I am not landing a resolution I cannot stand behind.

**What changed under it.** `_canonical_crypto_pair` has been tightened since this branch: a `USD` quote leg now needs a whitelisted crypto base (#1280, so spot gold stops resolving to tokenized gold), and joined pairs are recognised (#1265). Your candidate filter drops a hit when `_canonical_crypto_pair(candidate)` is **not** None. Measured on current main, the two near-string hits this PR exists to drop both canonicalise to `None`:

```
VALOUR-BTC-0-SEK.ST  -> None
AETHUSDT-USD         -> None
```

So the filter passes them through, and `test_xauusd_x_drops_near_string_crypto_hits`, `test_xauusd_slash_drops_near_string_crypto_hits` and `test_xauusdt_drops_near_string_btc_etp` all fail on the rebase with `assert 1 == 0`.

I tried rebuilding the filter on the provider's own labels — Yahoo's `quoteType == "CRYPTOCURRENCY"` / `exchange == "CCC"` — which fixes two of the three. It does not fix the third, because `VALOUR-BTC-0-SEK.ST` is labelled `quoteType: EQUITY` on `STO`: a Bitcoin ETP that is, as far as every signal in the payload goes, an ordinary Swedish equity. Dropping it needs a different idea than "is this crypto", and which idea is right is your call, not mine.

The part I could not explain, and the reason I stopped: `_canonical_crypto_pair("VALOUR-BTC-0-SEK.ST")` is `None` on **this branch too**, so the elif cannot have been what dropped it there either — yet the test passes here. Something else in the pipeline is doing the work on this branch and I have not identified it. Rather than guess, I would rather you say which mechanism you intended.

**Separate, smaller point.** The venue classifier this adds to `grounding.py` copies the ISO-metals + G10 whitelist that #1280 has now landed in `_market_hooks.py`. Your own comment calls it "the third copy"; merged as-is it would be the fourth. That is worth consolidating — `src.market_data` is already imported by `grounding.py` for `canonical_fx_pair` and looks like the natural home — but it is its own change and should not be bolted onto this one.

The underlying problem is real and still unfixed on main: `XAU-USD` infers `crypto_or_fx`, and `^SPX` and `XAUUSD=X` infer nothing at all. Happy to do the rebase and the consolidation once the filter mechanism is settled.

PR #1285, opened 2026-08-31
fix(live): enforce narrowing semantics for non-numeric and list mandate adjustments

Jackzigen:
## Problem

The commit-time adjustment path only guarded the both-numeric branch, so three narrowing-check bypasses slipped through (roadmap audit #1207, finding 19):

- **Type confusion**: a rendered numeric limit (`leverage: 2`) adjusted with a string (`"10"`) skipped the widening check entirely and committed as-is
- **Bool masquerade**: `True`/`False` are int subclasses in Python and passed the `isinstance(current, (int, float))` guard
- **List fields unguarded**: `instruments` / `allowed_instruments` had no comparison branch at all, so a whitelist could be widened from `["AAPL"]` to `["AAPL","TSLA","SPY"]` at commit time - silently expanding the authorization the user actually saw

The downstream ceiling re-check `_profile_fits_ceilings` had the same both-numeric blind spot, so it never caught what `_resolve_profile` let through.

## Fix

In `agent/src/live/mandate/commit.py`:

- `_resolve_profile`: numeric limits now reject non-numeric adjustment types outright (invalid type, not silent skip); bool is excluded from numeric handling; list limits must be **subsets** of the rendered whitelist; other scalar types must be equivalent - widening always goes back through PROPOSE
- `_profile_fits_ceilings`: fails closed on non-numeric pairs, treats bool as non-numeric, and enforces whitelist containment on `allowed_instruments`

## Verification

All scenarios are covered by new tests in `agent/tests/test_consent_commit.py`:

| Test | Result |
|---|---|
| `test_adjust_string_numeric_widening_rejected` | ✅ |
| `test_adjust_string_numeric_narrowing_rejected` | ✅ |
| `test_adjust_bool_numeric_limit_rejected` | ✅ |
| `test_adjust_instruments_widening_rejected` | ✅ |
| `test_adjust_instruments_narrowing_subset_commits` | ✅ |
| `test_profile_fits_ceilings_rejects_non_numeric_profile_value` | ✅ |
| `test_profile_fits_ceilings_rejects_bool_profile_value` | ✅ |
| `test_profile_fits_ceilings_rejects_instrument_widening` | ✅ |

- Full suite: **33 passed** (25 existing + 8 new)
- Reverse verification: reverting only `commit.py` to main makes exactly the 7 new rejection tests fail, confirming they guard the actual bug
- `git diff` touches exactly two files: the fix + its tests

Fixes part of #1207 (finding 19).

### DIFF
--- a/agent/src/live/mandate/commit.py
+++ b/agent/src/live/mandate/commit.py
@@ -49,6 +49,11 @@ _PROPOSALS_DIRNAME = "proposals"
 _CONSENT_DIRNAME = "consent"
 _PROPOSAL_ID_RE = re.compile(r"^mp_[0-9a-f]{32}$")
 
+def _is_real_number(value: Any) -> bool:
+    """Return whether ``value`` is a non-bool int/float."""
+    return isinstance(value, (int, float)) and not isinstance(value, bool)
+
+
 #: Maps every accepted alias of a clamped limit to its CANONICAL name. The
 #: proposal profile, the ceiling snapshot, and the clamp in
 #: ``propose_mandate_tool`` all use slightly different human-vs-schema spellings
@@ -262,12 +267,33 @@ def _resolve_profile(
             if key not in resolved:
                 raise CommitError(f"adjustment {key!r} is not a field of the selected profile")
             current = resolved[key]
-            if isinstance(current, (int, float)) and isinstance(value, (int, float)):
+            if _is_real_number(current):
+                if not _is_real_number(value):
+                    raise CommitError(
+                        f"adjustment {key!r}={value!r} has an invalid type for the rendered limit "
+                        f"{current!r}; numeric narrowing requires an int or float"
+                    )
                 if value > current:
                     raise CommitError(
                         f"adjustment {key!r}={value} widens the rendered limit {current}; "
                         "widening must go through a fresh proposal"
                     )
+            elif isinstance(current, list):
+                if not isinstance(value, list):
+                    raise CommitError(
+                        f"adjustment {key!r}={value!r} has an invalid type for the rendered list "
+                        f"{current!r}; list narrowing requires a list subset"
+                    )
+                if not all(item in current for item in value):
+                    raise CommitError(
+                        f"adjustment {key!r}={value!r} widens the rendered whitelist {current!r}; "
+                        "widening must go through a fresh proposal"
+                    )
+            elif type(value) is not type(current) or value != current:
+                raise CommitError(
+                    f"adjustment {key!r}={value!r} changes the rendered value {current!r}; "
+                    "only narrowing or equivalent adjustments are allowed"
+                )
             resolved[key] = value
     return resolved
 
@@ -299,12 +325,21 @@ def _profile_fits_ceilings(profile: Mapping[str, Any], ceilings: Mapping[str, An
         if key not in prof:
             continue
         prof_value = prof[key]
-        if isinstance(ceiling_value, (int, float)) and isinstance(prof_value, (int, float)):
-            if prof_value > ceiling_value:
-                return False
-        elif key == "leverage":
+        if key == "leverage" and ceiling_value == "none":
             # Cash-only ceiling forbids any leverage other than "none".
-            if ceiling_value == "none" and prof_value not in ("none", None, 1, 1.0):
+            if prof_value is True or prof_value is False or prof_value not in ("none", None, 1, 1.0):
+                return False
+        elif key == "allowed_instruments":
+            # A whitelist may only be narrowed, never widened at commit time.
+            if not isinstance(ceiling_value, list) or not isinstance(prof_value, list):
+                return False
+            if not all(item in ceiling_value for item in prof_value):
+                return False
+        else:
+            # Numeric ceilings fail closed when either side is bool/non-numeric.
+            if not _is_real_number(ceiling_value) or not _is_real_number(prof_value):
+                return False
+            if prof_value > ceiling_value:
                 return False
     return True
 
--- a/agent/tests/test_consent_commit.py
+++ b/agent/tests/test_consent_commit.py
@@ -23,6 +23,7 @@ import src.live.mandate.commit as mandate_commit
 from src.live.mandate.commit import (
     CommitError,
     DEFAULT_MANDATE_LIFETIME_DAYS,
+    _profile_fits_ceilings,
     commit_mandate,
     save_proposal,
 )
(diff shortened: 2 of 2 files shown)


he-yufeng (commented):
Read through the commit path and exercised `_profile_fits_ceilings` directly on the boundary cases. The three holes this closes were all real: a string adjustment on a numeric limit skipped the widening check entirely (the old `isinstance(current, (int, float)) and isinstance(value, (int, float))` guard just fell through and stored the string), `True` passed as leverage 1 anywhere a numeric comparison happened, and the instrument whitelist was never re-validated at commit because lists failed the old numeric isinstance gate on both sides.

One gap in the test coverage: the explicit bool rejection in the cash-only branch (`prof_value is True or prof_value is False` ahead of the `in ("none", None, 1, 1.0)` check) has no test behind it. `test_profile_fits_ceilings_rejects_bool_profile_value` only exercises a numeric ceiling. The identity checks are load-bearing there because `True == 1` makes `True in ("none", None, 1, 1.0)` evaluate to True, so without them a bool passes the cash-only gate. Worth one line:

```python
def test_profile_fits_ceilings_rejects_bool_under_cash_only_ceiling() -> None:
    assert not _profile_fits_ceilings({"leverage": True}, {"leverage": "none"})
```

I confirmed the current code returns False for that call, and that "none" / None / 1 / 1.0 still pass the same ceiling, so this is purely a regression pin.

Everything else checks out: string narrowing rejected rather than coerced, subset narrowing on instruments commits, superset rejected, non-list ceiling on the instruments key fails closed.

warren618:
Not merging this yet. The direction is right and the widening hole it closes is real, but as written it breaks the narrowing path it exists to protect.

`"none"` is not an invalid non-numeric type here — it is this module's own cash-only sentinel, and the default at that (`commit.py:471`: `leverage_raw = profile.get("leverage", "none")`, normalized to 1.0 on the next line). Treating it as a type error rejects the safest adjustment a user can make. Your branch against main:

```
narrow leverage 2 -> 'none'  |  main: accepted  |  this PR: rejected ("invalid type")
narrow leverage 2 -> 1       |  main: accepted  |  this PR: accepted
widen  leverage 2 -> 10      |  main: rejected  |  this PR: rejected
widen  leverage 2 -> '10'    |  main: accepted  |  this PR: rejected   <- your H19 fix, correct
```

`_profile_fits_ceilings` has the same conflation, and there it is worse than an annoyance. With a numeric leverage ceiling — exactly the shape your own `_save_h19_proposal` fixture uses (`"leverage": 2`) — a profile carrying `leverage: "none"` now fails, so the most conservative profile in a proposal becomes uncommittable:

```python
_profile_fits_ceilings({"leverage": "none", ...}, {"leverage": 2, ...})
# main: True    this PR: False
```

The fix I'd take: normalize leverage on both sides before comparing — `"none"`/`None` -> 1.0, a real number -> itself, anything else -> reject — and use it in both `_resolve_profile` and `_profile_fits_ceilings`. That keeps every rejection this PR adds (`"10"` is still not a number, `True` is still not a number) and restores cash-only as a valid narrowing. The cash-only ceiling case then falls out of the same comparison: ceiling `"none"` -> 1.0, profile 2 -> 2 > 1.0 -> rejected, same as today.

Worth adding coverage in the other direction too. All eight new tests assert a rejection, which is why nine green checks did not catch this — a gate that only gets tested for what it blocks will eventually block everything.

warren618:
Re-reviewed at `ae213083`. **The holes you're closing are real** — I confirmed both against main before judging the fix:

```
MAIN, _profile_fits_ceilings:
  profile {"allowed_instruments": ["equity","option","future"]} vs ceiling {"allowed_instruments": ["equity"]}  -> fits=True   # widened whitelist accepted
  profile {"leverage": True} vs ceiling {"leverage": 2.0}                                                       -> fits=True   # bool through a numeric ceiling
```

Both should be refusals, and your branch refuses them. The list-subset check and the bool guard are right and I want them.

**One defect blocks it: `"none"` is the floor of the leverage axis, not an invalid type.** Same finding as my 2026-08-31 note, now confirmed on both paths of the current head:

```
_resolve_profile, rendered leverage 2.0:
  adjust leverage -> "none"  (cash only, the safest possible narrowing)  -> REFUSED
      "adjustment 'leverage'='none' has an invalid type for the rendered limit 2.0"
  adjust leverage -> 1.0                                                 -> OK
  adjust leverage -> 4.0                                                 -> REFUSED (correct)

_profile_fits_ceilings:
  profile {"leverage": "none"} vs ceiling {"leverage": 2.0}   ->  main: fits=True   this branch: fits=False
```

So on an account whose ceiling permits 2x, a user can no longer commit a mandate that forbids leverage outright, while `1.0` — the same thing in effect — sails through. The gate refuses the most conservative option available, and that's a regression against main, not a pre-existing gap.

`"none"` needs to sort below every number on this axis in both places rather than fall into the non-numeric reject branch. The cash-only ceiling case at line 328 already knows this; the two branches beneath it don't.

On tests: all eight new ones assert a refusal. That shape can't see this bug — it only ever asks "is this rejected?", and the answer being *yes* for a case that should be accepted looks identical to success. For a change that tightens a gate, please add the other column: a table of adjustments that must still be **accepted** (`2.0 -> "none"`, `2.0 -> 1.0`, whitelist narrowed to a subset, whitelist unchanged) beside the ones that must be refused. That's what would have caught this before CI went green.

he-yufeng:
Line-by-line pass, verified locally rather than from the diff text.

The three bypasses reproduce on current main: `tests/test_consent_commit.py`'s new H19 cases go red against an unpatched `commit.py` (string `"10"` widens `leverage: 2` into a commit, `True` passes the numeric guard, and `instruments` has no comparison branch at all), and green on this branch (33 passed). The subset-narrowing positive control passes on both, which is the right shape.

Two things I specifically checked because they could have broken legit flows:

- `_normalize_limits` projects both sides onto the five canonical ceiling keys before `_profile_fits_ceilings` sees them, so the new fail-closed `else` on non-numeric values can never meet a free-form string field. The branches map exactly onto the five keys (four numeric + `allowed_instruments`).
- `_resolve_profile`'s final `elif` rejects any non-numeric, non-list change outright (`label`-style fields included). That is deliberately stricter than "only limits narrow", but it matches the consent contract the user saw at proposal time, and the equivalence case (`type`-equal, value-equal) still commits.

One judgment call worth stating out loud: string-narrowing like `"leverage": "1"` is now rejected as invalid type rather than coerced. I think that is right for a mandate gate (the user confirmed a number, not a string), but it is a behavior change for any caller that stringifies inputs upstream.

This sits in the live mandate path, so I'll leave the approve to the owner. Evidence-wise this is the #1207 finding 19 fix and it is good to land from where I sit.

Issue #1267, opened 2026-08-30
[Feature] Domain sub-agents for the built-in agent: fewer tools per model decision, with the evaluation harness to prove it


shadowinlife:
> The built-in agent (the default session runtime behind the CLI and Web chat, as distinct from swarm presets and external MCP clients) shows the model its entire tool registry — around a hundred tools by our count, plus a ~90-entry skill catalog — on every turn of the tool-calling loop. We tested four ways to shrink the number of tools the model must consider per decision, with pass/fail thresholds fixed *before* each experiment ran. Three approaches failed or only partially helped. One worked and has been running in our own production deployment for weeks: **domain sub-agents** — specialists that each hold a small, hard-enforced tool whitelist, while the main agent only decides "handle this directly or hand it to a named specialist." We would like to contribute the approach, the evaluation harness that vetted it, and a staged port designed for the built-in agent's own architecture.

---

### The problem we are solving

On every turn of the tool-calling loop, the built-in agent attaches the full tool registry — full JSON schemas — plus the skill catalog (names and descriptions) to the model's context. We counted ~106 tool definitions and ~90 skill descriptions in late August 2026 (the exact number varies with configuration; the public trend is visible in the release notes, where the MCP-exposed subset alone grew from 48 to 74 tools over three months). That is tens of thousands of tokens of descriptions, paid on every turn whether the task needs it or not.

Three costs follow:

1. **Routing quality.** A growing body of published work on tool-using agents reports that tool-selection accuracy degrades as the number of visible tools grows past a few dozen. Our own measurements complicated that picture rather than contradicting it: strong models still pick the right tool first 88–94% of the time at ~100 visible tools — so for strong models the cost shows up as long-tail routing errors, tokens, and latency, not as collapse. For the smaller/local models this project also supports, a tool block this size can crowd out most of the context window outright.
2. **Token cost and latency.** The description block dominates per-turn prompt size, and it grows with every tool or skill the project adds.
3. **Scalability headroom.** Every future tool makes the per-turn tax and the routing problem a little worse. The registry will keep growing; the per-decision surface should not have to.

The goal: **fewer visible tools per model decision, without losing routing accuracy.**

### What we researched before building anything

- **Tool-selection scaling literature** — the evidence above, which set our design target: keep every decision surface under ~15 visible tools.
[...]

PR #1286, opened 2026-08-31
feat(agent): domain specialist sub-agents for the built-in agent loop (+ load_skill allowlist fix)

shadowinlife:
## Summary

- **Commit 1 (standalone, cherry-pickable):** `load_skill` now enforces a per-context skill allowlist — `SwarmAgentSpec.skills` documented a per-worker skill boundary that was only applied to the prompt text, so a swarm worker could load any skill by name. `LoadSkillTool` gains an optional `allowed_skills`; `build_filtered_registry` / `build_swarm_registry` gain a matching `skill_allowlist` parameter (default `None` = unchanged behavior), and `run_worker` delivers the spec's list through it.
- **Commit 2 (the feature):** an opt-in domain sub-agent layer for the built-in agent loop — a `delegate_to_specialist` tool plus twelve admitted domain specialists, each holding a small hard-enforced tool whitelist projected from the registry (the same mechanism swarm workers already use). The main agent only decides "handle directly or delegate"; each delegated decision then reads a small domain surface instead of the full tool registry.
- **Commits 3–5 (review-driven fixes):** three follow-up hardening commits — the declared specialist behavior contract now actually reaches the child system prompt (previously `spec.prompt` was silently dropped), the skills-bearing YAMLs grant the `load_skill` channel their skill lists require, and a roster/lifecycle hardening batch (read-only trading-connector boundary pinned to 8 tools, timeout clamp, immutable roster cache, clean timeout payloads, three additions to the forbidden-tool list).

This is pieces 2 + 3 of #1267.

## Why

The built-in agent attaches its full tool registry (~106 tools) plus the skill catalog to every model turn. Published tool-selection scaling results — and our own pre-registered measurements on this tool surface — put the comfortable decision surface at ~15 visible tools. We measured four approaches (description rewrites, availability gating, lazy tool retrieval, domain sub-agents) against a frozen bilingual corpus with pre-registered thresholds; the first three were rolled back with the data archived, and domain sub-agents have been running in our own production deployment for weeks:

- routing recall **99.1%**, mis-delegation **3.57%** (pilot gates, 198-query corpus), later admissions on a 353-query corpus passed at recall 87.5–100% per candidate;
- on the name-collision subset, specialists **beat** the full surface (+6.7pp quant, +33.4pp web/docs);
- **−86%** description tokens per delegated decision, median decision latency **−33%**.

Those numbers were measured on our deployment (a different agent runtime driving this repo through MCP); the method and qualitative findings transfer, the exact numbers should be re-measured on the built-in agent — which is why this PR ships the mechanism gated and off by default, and why the evaluation harness is offered separately (piece 1 in #1267).

## Changes

**Feature surface (commit 2):**

- `agent/src/specialists/` — new package: `models.py` (`SpecialistSpec` + the structural forbidden-tool set), `loader.py` (bundled `definitions/*.yaml` + `~/.vibe-trading/specialists/` user overrides, same override rule as user skills; fail-loud validation of every whitelisted tool/skill name at load), `routing.py` (system-prompt routing block, fail-safe: emitted only when the delegate tool is actually registered, mirroring the `strategy_discovery.guard` precedent), and `definitions/` — the twelve admitted specialists (quant, market-data, fundamentals-text, derivatives, risk-portfolio, valuation, macro-sector, altdata, funds-fi, user-analytics, web-docs, trading-connector).
- `agent/src/tools/delegate_tool.py` — `delegate_to_specialist(specialist, task)`. The specialist runs as a nested `AgentLoop` with its own run directory, trace, manifest, grounding ledger, iteration budget and wall-clock timeout. `is_readonly=True`, so a delegation inherits the loop's daemon-thread + tool-timeout semantics.
- `agent/src/config/env_schema.py` — `VIBE_TRADING_SPECIALISTS_ENABLED` (EnvBool, **default off**). Registries are built at process start, so flipping the gate takes effect on restart; with the gate off the tool is never registered and the system prompt is byte-identical to today.
- `pyproject.toml` — package-data for the bundled YAMLs (mirrors the swarm-preset packaging rule).
- `README.md` — new folded section documenting the feature, the safety contract, and the user-override point; env-var table entry.

**Review-driven fixes (commits 3–5):**

- **F1 — prompt injection (commit 3, `38e04d88`):** `delegate_tool.execute()` built the child agent from the specialist spec but dropped `spec.prompt`, so delegated specialists ran without their declared role behavior. The child loop now runs under `SPECIALIST_SYSTEM_PROMPT` with the behavior contract bound to `{role_prompt}` via new optional `ContextBuilder`/`AgentLoop` parameters (`system_template`/`role_prompt`); both default to `None` and remain byte-identical to prior behavior when unset, pinned by a frozen-clock regression test.
- **F2 — load_skill channel (commit 3, `38e04d88`):** the 11 skills-bearing specialist YAML definitions now grant `load_skill`, and the loader fail-loud validates that any definition declaring skills also grants `load_skill`. Specialist skills semantics: an empty skills list means **no catalog is exposed and nothing is loadable** — deliberately diverging from swarm semantics where empty means unrestricted. (User YAMLs declaring skills without granting `load_skill` now fail validation; intentional — the feature is gate-off by default and user definitions are bring-your-own.)
- **F3 — roster & lifecycle hardening batch (commits 4–5, `09d11888` + `5a26eacc`):**
  - trading-connector specialist pinned to an 8-tool **read-only** boundary: grants `trading_history_deals` (actual fill history — the roster previously promised fill history while only granting `trading_history`, which returns broker OHLCV bars) and drops `trading_select_connection` (a session-wide connector-state mutation, violating the per-call stateless boundary — specialists now pass the per-call `connection` parameter instead); the roster is pinned by packaging tests.
  - specialist timeout budgets are clamped at roster load to leave +60s headroom under the loop tool timeout (`VIBE_TRADING_TOOL_TIMEOUT_SECONDS`); a bundled violation fails loud at load time.
  - the loaded roster cache is published as `MappingProxyType` — structurally read-only, not read-only by convention.
  - timeout payloads always carry empty content, so a late child result landing in the cancel grace window no longer rides along on the timeout response (`run_dir` preserved for partial-work recovery).
  - the forbidden-tool list gains `retry_run` (orchestration recursion), `qveris_execute` (billable API execution), and `trading_select_connection` (session-state mutation).

**Supporting changes (deliberately small, each called out per the issue's staging):**

| # | Change | Why it is needed |
|---|--------|------------------|
| S1 | `LoadSkillTool(allowed_skills=...)` (commit 1) | Fixes the documented-but-unenforced `SwarmAgentSpec.skills` boundary — valuable on its own; specialists reuse the same mechanism for their skill whitelists. |
| S2 | `skill_allowlist` parameter on `build_filtered_registry` / `build_swarm_registry` (commit 1) | One shared injection path; default `None` is byte-for-byte the old behavior. |
| S3 | `AgentLoop.__init__` gains optional `skills_loader` / `system_template` / `role_prompt` parameters, and binds tools' `bind_parent(cancel_event=...)` after construction | `src/agent/` is a protected area — this is the minimal seam. The registry is necessarily built before the loop exists, so the loop must hand its cancel event to delegation-capable tools post-construction; the duck-typed loop (`getattr` on `tool_names`/`get`) keeps stub registries in tests working. All parameters default to today's behavior. |
| S4 | `context.py`: `ContextBuilder` gains optional `system_template` / `role_prompt`, plus one `{specialist_routing}` slot sharing the `{strategy_discovery_routing}` template position | The production-proven load-bearing lesson: the main agent never delegates unless its own instructions name the routing policy. Sharing the slot keeps the system prompt **byte-identical** when the gate is off (verified by test). |
| S5 | `skills.py`: `SkillsLoader(only=...)` filter, which also closes the mid-session disk fallback | The child's prompt surface must match its runtime boundary; the tool-level allowlist alone would leave the prompt advertising skills the child cannot load. `None` = unchanged. |
| S6 | `build_swarm_registry` strips `delegate_to_specialist` from worker whitelists with a warning | One orchestration channel per run: a swarm worker is already a specialist; nesting delegation inside it compounds latency and telephone loss. |
| S7 | `run_worker` passes `agent_spec.skills` (commit 1) | The one line that makes the documented swarm boundary real. |
| S8 | Two existing test files touched: `test_swarm_m4_e2e.py` (four stub lambdas gain the new `skill_allowlist` kwarg — they assert remote-MCP behavior, not skills) and `test_context_attribution_layers.py` (the prompt-format test passes the new template slot). | Interface extensions from S2/S4. |

**Not in this PR (explicit non-goals):** retiring domain tools from the main agent's own surface (the token payoff step — it should land only after the maintainers' model panel re-measures routing on the built-in agent with the harness); the evaluation harness itself (#1267 piece 1); availability-based registration trimming (piece 4); rich SSE rendering of `subagent_started`/`subagent_completed` events (they flow through the event bus as-is); persisting specialist runs into session search.

**MCP surface note:** when the gate is on, `delegate_to_specialist` is also visible to MCP clients (the MCP server shares `build_registry`). This is intentional: the tool is self-bounding (own iteration + wall-clock budget, own cancel handling) and needs no host wiring — an external agent host is exactly the deployment the design comes from. When the gate is off it registers nowhere.

## Test Plan

- [x] Existing tests pass — full suite from the repo root: **11706 passed, 84 skipped, 4 failed**. The 4 failures are pre-existing environment/dependency drift in the provider-adapter tests (`test_llm_reasoning_effort.py` ×3, `test_provider_header_isolation.py` ×1) and are **proven pre-existing**: an identical baseline control run without this PR's commits produces exactly the same 4 failures. Nothing on this PR's path fails.
- [x] New tests added (48): roster loading and fail-loud validation, forbidden-tool rejection (incl. a pinned read-only `trading-connector-agent` whitelist — it can never gain order tools, and its 8-tool roster is pinned by packaging tests), user override + broken-user-file resilience, routing-block presence/absence/error fail-safe, gate on/off registration, prompt byte-identity when off (frozen-clock), the `spec.prompt` behavior-contract delivery into the child system prompt, the `load_skill`-grants-skills loader invariant, timeout-budget clamping, whitelist projection of the child registry, unknown-specialist/empty-task refusal, **parent-cancel propagation into a running child**, **budget timeout with bounded cancel grace**, **zombie-thread abandonment and reclaim** (threads are name-prefixed so the leak assertion is not flaky), and **concurrent delegations on one shared tool instance** (the instance-statelessness invariant).
- [x] Tested manually: gate off → tool absent, prompt unchanged; gate on → tool registered with the roster enum, delegation section rendered in the system prompt.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — the `src/agent/` diff is optional default-`None` pass-through parameters plus a duck-typed binding loop (S3/S4/S5 above), proposed and motivated in #1267; happy to split it out if you prefer. `src/session/` and `src/providers/` are byte-untouched.
- [x] Code follows CONTRIBUTING.md (black/ruff on new files; pre-existing style left untouched in edited files; Google-style docstrings; DCO sign-off on all five commits)
- [x] Documentation updated (README section + env-var table)

**Risk & rollback:** the gate defaults off, so the feature has **zero impact for anyone who never enables it** — the tool is never registered and the system prompt is byte-identical, and reverting is a no-op for that population. With the gate on, reverting this PR removes the tool and the prompt section with no state migration (specialist runs are ephemeral nested loops whose artifacts live in ordinary run dirs). No broker, order, credential, network-listener, or deployment surface is touched — specialists structurally cannot hold shell, order-write, orchestration, or session-state tools, and the one broker-facing specialist is pinned read-only by test.

**Evidence archive:** the full admission-protocol record (frozen thresholds, router-panel configs, decision traces, per-round verdicts) backs the numbers above and will be attached to the first PR of this series per #1267; say the word if you want it attached here too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,38 @@ alongside the bundled roster (same-name files override it, like user skills) and
 </details>
 
 <details>
+<summary><b>Domain Specialist Sub-agents</b> <sub>opt-in per-turn delegation inside the interactive loop</sub></summary>
+
+- 🧭 The interactive agent can delegate domain work to named specialists, each holding a small hard-enforced tool whitelist instead of the full registry
+- 📉 Fewer visible tools per model decision: specialists see only their domain surface, and the main agent only decides "handle directly or delegate"
+- 🧪 Every bundled specialist passed a pre-registered admission protocol (routing recall ≥ 85%, mis-delegation ≤ 5%, boundary arbitration ≥ 85%) before joining the roster
+
+Swarm presets are fixed multi-agent workflows you launch explicitly; specialists are the per-turn counterpart — automatic delegation inside the interactive loop, enforced by the same whitelist projection the swarm runtime uses.
+
+**Off by default.** Enable with `VIBE_TRADING_SPECIALISTS_ENABLED=1` and restart (registries are built at process start). When enabled, the main agent gains a `delegate_to_specialist` tool and a delegation section in its system prompt; the roster below ships with the package:
+
+| Specialist | Domain |
+|------------|--------|
+| `quant-agent` | Factor research, alpha zoo, strategy authoring and backtests |
+| `market-data-agent` | Symbol resolution, OHLCV, rankings, capital flow, order-book depth |
+| `fundamentals-text-agent` | Statements, SEC filings, 13F, news, sell-side research, papers |
+| `derivatives-agent` | Options pricing/Greeks, multi-leg payoff, perp funding/basis |
+| `risk-portfolio-agent` | VaR/stress, statistical tests, attribution, allocation |
+| `valuation-agent` | DCF/comps/three-statement, event-driven construction, prediction markets |
+| `macro-sector-agent` | FRED series, sector boards, macro/industry/regulatory frameworks |
+| `altdata-agent` | Sentiment, on-chain, stablecoin flows, DeFi yields, unlocks |
+| `funds-fi-agent` | ETF look-through, fund screening, credit and convertible bonds, dividends |
+| `user-analytics-agent` | Trade-journal diagnostics and the shadow-account pipeline |
+| `web-docs-agent` | Web search, page-to-Markdown, document/OCR extraction |
+| `trading-connector-agent` | Read-only broker connector reads (never holds order tools) |
+
+Safety and lifecycle contract: specialists never receive shell, order-write, orchestration, or session-state tools (structurally excluded at definition-load time); a specialist run is a nested agent loop with its own iteration budget, wall-clock timeout, run directory and grounding ledger; a user cancel propagates into a running specialist, and a specialist that exceeds its budget is cancelled and reported, never silently abandoned. The `trading-connector-agent` whitelist is pinned read-only by test.
+
+Bring your own: drop `<name>.yaml` files into `~/.vibe-trading/specialists/` (same override rule as user skills; definitions are validated at load — unknown tool/skill names fail loudly, and a broken user file is skipped with a warning rather than breaking the roster). Definitions are loaded once per process.
+
+</details>
+
+<details>
 <summary><b>Alpha Zoo</b> <sub>462 pre-built quant alphas across 5 families</sub></summary>
 
 - 🧬 462 cross-sectional alphas, lookahead-banned at the operator layer
@@ -942,6 +974,7 @@ Copy `agent/.env.example` to `agent/.env` and uncomment the provider block you w
 | `TIMEOUT_SECONDS` | No | LLM call timeout, default 120s |
 | `API_AUTH_KEY` | Recommended for network deployments | Bearer token required when the API is reachable from non-local clients |
 | `VIBE_TRADING_ENABLE_SHELL_TOOLS` | No | Explicit opt-in for shell-capable tools in remote API/MCP-SSE style deployments |
+| `VIBE_TRADING_SPECIALISTS_ENABLED` | No | Opt-in domain specialist sub-agents for the interactive agent (default off; takes effect on process start) |
 | `VIBE_TRADING_ALLOWED_FILE_ROOTS` | No | Extra comma-separated roots for document and broker-journal imports |
 | `VIBE_TRADING_ALLOWED_RUN_ROOTS` | No | Extra comma-separated roots for generated-code run directories |
 | `VIBE_TW_STOCK_DB` | No | Path to a Taiwan-market SQLite snapshot; the read-only `taiwan_stock_data` tool registers only when it is schema-valid |
--- a/agent/src/agent/context.py
+++ b/agent/src/agent/context.py
@@ -78,7 +78,7 @@ skill document, not recalled memory. They are not defaults to be tuned.
 ## Task Routing
 
 Decide which workflow to use based on the request:
-{strategy_discovery_routing}
+{strategy_discovery_routing}{specialist_routing}
 **Backtest** — user wants to create, test, or optimize a trading strategy:
 1. `load_skill("strategy-generate")` — read the SignalEngine contract
 2. `write_file("config.json", ...)` — source, codes, dates, parameters. If the strategy is expected to produce ≥10 trades, include `"validation": {{"monte_carlo": {{"n_simulations": 1000}}}}` in config.json for Monte Carlo testing
@@ -277,6 +277,17 @@ class ContextBuilder:
         except Exception:  # noqa: BLE001
             routing = ""
 
+        # Specialist-delegation routing text follows the same fail-safe
+        # contract: present only when the delegate tool is registered, empty
+        # string otherwise. The slot shares the strategy-routing template
+        # position so a disabled gate leaves the prompt byte-identical.
+        try:
+            from src.specialists.routing import specialist_routing_block
+
+            specialists = specialist_routing_block(self.registry)
+        except Exception:  # noqa: BLE001
+            specialists = ""
+
         return _SYSTEM_PROMPT.format(
             tool_count=len(self.registry._tools),
             skill_count=len(self.skills_loader.skills),
@@ -286,6 +297,7 @@ class ContextBuilder:
             memory_summary=self.memory.to_summary(),
             memory_section=memory_section,
             strategy_discovery_routing=routing,
+            specialist_routing=specialists,
             current_datetime=now.strftime("%A, %B %d, %Y %H:%M UTC"),
         )
 
--- a/agent/src/agent/context.py
+++ b/agent/src/agent/context.py
@@ -229,7 +229,9 @@ class ContextBuilder:
 
     def __init__(self, registry: ToolRegistry, memory: WorkspaceMemory,
                  skills_loader: Optional[SkillsLoader] = None,
-                 persistent_memory: Optional[PersistentMemory] = None) -> None:
+                 persistent_memory: Optional[PersistentMemory] = None,
+                 system_template: Optional[str] = None,
+                 role_prompt: Optional[str] = None) -> None:
         """Initialize ContextBuilder.
 
         Args:
@@ -237,11 +239,18 @@ class ContextBuilder:
             memory: Workspace memory.
             skills_loader: Skills loader (auto-created if not provided).
             persistent_memory: PersistentMemory instance for cross-session recall.
+            system_template: System-prompt template override (specialist
+                sub-agents pass a slim template; ``None`` keeps the default
+                ``_SYSTEM_PROMPT``). Formatted with the same field set.
+            role_prompt: Specialist behavior contract, bound to the
+                ``{role_prompt}`` slot; ``None`` renders an empty string.
         """
         self.registry = registry
         self.memory = memory
         self.skills_loader = skills_loader or SkillsLoader()
         self._persistent_memory = persistent_memory
+        self._system_template = system_template
+        self._role_prompt = role_prompt
 
     def build_system_prompt(self, user_message: str = "") -> str:
         """Build system prompt.
@@ -288,7 +297,8 @@ class ContextBuilder:
         except Exception:  # noqa: BLE001
             specialists = ""
 
-        return _SYSTEM_PROMPT.format(
+        template = self._system_template or _SYSTEM_PROMPT
+        return template.format(
             tool_count=len(self.registry._tools),
             skill_count=len(self.skills_loader.skills),
             data_source_count=self._count_data_sources(),
@@ -299,6 +309,7 @@ class ContextBuilder:
             strategy_discovery_routing=routing,
             specialist_routing=specialists,
             current_datetime=now.strftime("%A, %B %d, %Y %H:%M UTC"),
+            role_prompt=self._role_prompt or "",
         )
 
     @staticmethod
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -906,6 +906,7 @@ class AgentLoop:
         event_callback: Optional[Callable[[str, Dict[str, Any]], None]] = None,
         max_iterations: int = 50,
         persistent_memory: Optional[Any] = None,
+        skills_loader: Optional[Any] = None,
     ) -> None:
         """Initialize AgentLoop.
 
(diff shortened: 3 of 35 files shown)


warren618:
Consolidating #1267 into this PR so the proposal and the implementation are one thread — my direction answer is on #1267 and repeated here in short.

**Yes to the direction, with four conditions**: (1) opt-in and default-off for at least one release, since a misrouting delegation layer fails as a silently missing capability rather than an error; (2) **split the `load_skill` allowlist fix into its own PR** — it is a real bug (the spec documented a per-worker skill boundary that was only applied to the prompt text) and it should not wait on the feature, I will take it this week; (3) land the corpus and harness with the feature so the routing numbers are reproducible against this repo's registry as it changes — #1271 merged today and is the natural place to plug that in; (4) state explicitly what happens when a specialist cannot serve a request, because the main agent has to be able to take it back rather than answer from a surface missing the tool.

The rolled-back approaches with archived data are what made the proposal credible, so thank you for reporting those rather than only the winner.

shadowinlife:
## Closing: the built-in-agent re-measurement doesn't reproduce the production wins

This PR's description correctly noted that the production numbers (routing recall 99.1%, decision latency −33%, specialists beating the full surface by +6.7pp/+33.4pp on the name-collision subset, −86% description tokens) "were measured on our deployment (a different agent runtime driving this repo through MCP)," and that "the exact numbers should be re-measured on the built-in agent." I ran that re-measurement on the built-in agent loop.

**Result: with `qwen3.8-max` as the model, the specialist-delegation layer does not reproduce those wins on the built-in loop — it loses to the no-specialist baseline on wall-clock, and the token saving is much smaller than the production runtime showed.** Closing as not-planned. The feature is gated off by default (`VIBE_TRADING_SPECIALISTS_ENABLED`), so this is zero-impact for anyone who never enabled it. Recording the data here for reference.

### Comparison setup (apples-to-apples)

Same frozen bilingual corpus, same stubbed tools (network ≈ 0% of wall — 1.8s of 11,405s total), same 600s/task watchdog, and the same model `qwen3.8-max` for both the main agent and the specialist children (`SpecialistSpec.model_name` defaults to inherit; no YAML override). Baseline = main agent with the full registry, no delegation. Specialist = this PR's `delegate_to_specialist` + the 12-specialist roster (with the self-contained-whitelist refinement). The specialist run covers 102 task×repeat pairs (hard cluster + controls + risk family); every comparison below restricts the baseline to those same 102 pairs.

### Headline (on the 102 common pairs)

| metric | baseline | specialist |
|---|---|---|
| timeouts | 0 / 102 | **23 / 102 (22.5%)** |
| status matrix | — | ok→ok 79, **ok→timeout 23, timeout→ok 0** |
| LLM calls | 442 | **1156 (2.6x)** |
| wall median (79 both-ok pairs) | 25s | **85s (3.4x)** |
| tokens | 21.6M | 15.4M (**1.4x** saving) |
| non-hard tasks faster | — | **0 of 25** |

The status matrix is one-directional: the specialist layer never rescued a task the baseline failed, and broke 23 that the baseline passed.

### Why it loses here (a bundle, not one mechanism)

1. **Round multiplication is the binding constraint.** Delegation is sequential-blocking (`delegate_tool.execute()` → `child_thread.join`), and each specialist is a full nested loop. Work the baseline does in ~4–6 direct rounds becomes ~11 LLM rounds (brief-writing rounds at 25–115s each + child rounds + long structured child reports + a main integration round). On a slow thinking model (~17–38s/round), 2.6x rounds is the 600s wall.
2. **The routing policy removes the baseline's cheap escape hatch.** On "trap" tasks where a stub returns degenerate data (e.g. an options chain with one contract, no bid/ask), the baseline accepts the degenerate input and answers parametrically (passes in ~138s). The specialist child reaches the same parametric answer, but the policy's "delegate a top-up naming the missing fields" rule makes the main agent reject it as incomplete and re-delegate — burning the budget on incompleteness that cannot be fixed.
3. **Tool narrowing pushes the main agent into stub traps.** Narrowing `options_pricing`/`options_payoff` off the main surface leaves only `quantlib_call`, whose bench stub returns a constant for non-DCF/VaR functions; the main agent spirals on garbage even after a child already succeeded. (Partly bench-specific — a real quantlib wouldn't.)
4. **Child timeout discards partial work.** `delegate_tool` returns `content=""` on a child budget timeout, so a child cancelled at 300s hands back nothing and the main agent restarts from zero.

### Honest caveats (this bench over-punishes the specialist relative to production)

- Mechanisms 3 and partly 2 are bench artifacts: the production runtime (real tools) wouldn't hit the constant-garbage quantlib stub. These numbers are a worst case on this harness, not a clean production prediction.
- "Baseline 0 timeouts" is a last-wins-dedup statement — the raw baseline shows `deriv-en-2#r2` hitting 600s twice before passing at 352s. On the 23 pairs where the specialist timed out, the baseline's wall median was 204s with 9 of 23 over 300s: these tasks sit near the 600s edge for **both** modes, and the specialist's 2.6x round amplification pushes the edge cases over.
- Even with these caveats, the one-directional 23:0 status matrix on the common pairs is decisive for this runtime + model.

### Where the specialist DID win (worth keeping)

On 10 of the 79 both-ok pairs the specialist beat the baseline on **both** wall and tokens — concentrated in token-heavy families where the baseline burns 12–19 rounds: `funds-en-2#r1` (482→463s, 957K→263K tokens, 3.6x), `risk-zh-2#r2` (394→232s), `risk-zh-1#r0` (509→354s), `altdata-zh-1#r0` (184→49s). The slim child surface (6.6K-char prompt vs 56K) genuinely pays when a task is token-heavy and round-cheap. The same families also produced timeouts (`risk-zh-1#r2`, `risk-en-2#r1`), so it isn't safe to gate per-family.

### Reference conclusion on the opencode-subagent port

The production wins quoted in this PR's description came from a different runtime — an opencode/omo harness driving this repo through MCP, where sub-agents are dispatched in parallel on the harness's own models. That is the regime where specialist delegation pays: parallel fan-out + fast child models make the nested rounds cheap. Ported into the built-in agent loop, two of those properties are absent — delegation is sequential-blocking, and the child runs the same slow thinking model as the main agent. **The mechanism transfers; the economics don't.** What makes opencode's sub-agents win (parallelism + fast child models) is exactly what the built-in-loop port lacks. Anyone attempting the same port should treat child-model speed and dispatch parallelism as load-bearing, not optional.

### The one untested revival path

The only axis this benchmark did **not** test is child model speed. `SpecialistSpec.model_name` is already a per-specialist knob (defaults to inherit), so putting the specialists on a fast non-thinking model is a one-line YAML change, not surgery — and it is the single decisive experiment: freeze everything else (qwen3.8-max main, 300s child budget, stubs, 600s task budget, routing policy) and re-run the 23 timeout pairs. Pre-registered pass gates: timeouts ≤ 3/23, median wall ≤ 1.3x baseline on the same pairs, rubric Δ ≥ −0.05. Pass → salvageable as config + two small fixes (return partial content on child timeout; let the top-up policy accept a parametric answer when the data source is confirmed degenerate). Fail → the binding constraint is orchestration policy, not model speed, and the architecture is dead for this use regardless. I'm not running that here; recording it so the option is explicit.

### Note on commit 1

Commit 1 (the `load_skill` per-context allowlist fix, enforcing the documented-but-unenforced `SwarmAgentSpec.skills` boundary) is standalone and valuable independent of the specialist feature, and is cherry-pickable. Happy to re-submit it separately if maintainers want it — it shouldn't be buried with this closure.

warren618:
Closing this with the re-measurement rather than quietly letting it sit is the right call, and posting the negative result in this much detail is more useful to the project than the feature would have been. The one-directional 23:0 status matrix is decisive, and I particularly want to keep the framing in your last two sections on the record: **the mechanism transfers, the economics don't** — parallel fan-out and fast child models are load-bearing for sub-agent delegation, not incidental, and a sequential-blocking port that runs the child on the same slow thinking model has neither. That is the kind of thing that would otherwise be re-learned by the next person who tries it. Your caveats about the bench over-punishing the specialist (the constant-garbage quantlib stub, the near-600s edge cases) are noted and I am reading the numbers with them attached.

**Yes to commit 1 — please do re-submit it separately, and it does not need to wait on anything.** I verified the gap is still on main: `SwarmAgentSpec.skills` is documented as "List of allowed skill names" (`src/swarm/models.py:195`) but the only thing that reads it is `_filter_skill_descriptions` at `src/swarm/worker.py:585`, which shapes the prompt text; `LoadSkillTool.__init__` takes only a `skills_loader` and no allowlist, so `execute()` will load any skill a worker names. A documented boundary that exists only in the prompt is not a boundary. Small PR, and please include a test that asserts a worker naming a skill outside its spec is refused — the assertion has to be on the refusal, not on the prompt text, or it pins the same non-enforcement.

The pre-registered revival experiment (specialists on a fast non-thinking model, everything else frozen, gates: timeouts ≤ 3/23, median wall ≤ 1.3x, rubric Δ ≥ −0.05) is well specified. If you ever run it, post the result here even if it fails — a clean negative on that axis closes the question properly.

PR #1287, opened 2026-08-31
fix(backtest): serve adjusted prices on the Yahoo loader paths

he-yufeng:
## Summary

- Both Yahoo paths served raw OHLC while every other source on the fallback chain returns adjusted prices (qfq), so a split read as a crash (AAPL 4:1 as -75%, NVDA 10:1 as -90%) and every ex-dividend gap booked as a fake loss. Two runs over the same market could also mix calibers depending on which source won the chain.
- The v8 chart client now requests `events=div,splits` and scales OHLC to Yahoo's `adjclose` series; the yfinance path flips to `auto_adjust=True`. Same adjusted caliber on both paths now.

## Why

Refs #1207 (Phase 1, item 6). Volume stays raw on both sides, matching what the tencent/eastmoney qfq klines already do, so the chain finally speaks one price language. This also unblocks the adjustment-provenance stamp discussed on #1231: there is no longer a raw source to flag.

## Changes

- `yahoo_client.get_chart`: requests `events=div,splits`; `_parse_chart` scales open/high/low/close by `adjclose/close` per bar when the series is present.
- `_adjust_ratio`: missing, non-numeric, or implausible ratios (outside 0.01–100) fall back to raw instead of corrupting the bar.
- `yfinance_loader._download_history`: `auto_adjust=True`.

## Test Plan

- [x] New tests: a 4:1 split scales the earlier bar's OHLC to the adjusted level while volume stays raw; null/zero adjclose entries keep raw prices; the yfinance download asserts `auto_adjust=True`.
- [x] `pytest agent/tests -q -k "loader or market_data"`: 771 passed, 11 skipped. Yahoo/yfinance suites: 99 passed.
- [ ] Live Yahoo endpoint not called from here (network mocked); `ruff check` clean on all touched files.

Risk: price-caliber change for any market where Yahoo wins the chain: split/dividend names now read correctly, and flat names are unchanged (ratio 1). Rollback is a plain revert.

### DIFF
--- a/agent/backtest/loaders/yahoo_client.py
+++ b/agent/backtest/loaders/yahoo_client.py
@@ -191,7 +191,9 @@ def get_chart(
             structurally unusable.
     """
     yahoo_symbol = map_symbol(symbol)
-    params: Dict[str, Any] = {"interval": interval}
+    # events=div,splits makes Yahoo return the adjclose series next to the raw
+    # quote, which is the only adjusted-caliber source on this endpoint.
+    params: Dict[str, Any] = {"interval": interval, "events": "div,splits"}
     if range_:
         params["range"] = range_
     else:
@@ -234,6 +236,12 @@ def _parse_chart(payload: Any, yahoo_symbol: str) -> Tuple[List[Dict[str, Any]],
 
     timestamps = result.get("timestamp") or []
     quotes = (((result.get("indicators") or {}).get("quote")) or [{}])[0] or {}
+    # Every other loader on this chain serves adjusted prices (qfq), so the
+    # raw quote would book splits as crashes and ex-dividend gaps as losses.
+    # Scale to Yahoo's adjusted close when it is present, keeping raw volume.
+    adjclose_series = (
+        (((result.get("indicators") or {}).get("adjclose")) or [{}])[0] or {}
+    ).get("adjclose")
 
     rows: List[Dict[str, Any]] = []
     for index, ts in enumerate(timestamps):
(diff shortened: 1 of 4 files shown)


warren618 (approved):

warren618:
Merged, with one commit of mine on top (`0f3fa0ab`) that rewrites the comments. The change is right; the reason given for it is not, and the wrong reason would cost us the next decision.

I measured the chain again before merging. Yahoo's v8 `quote` series is **already split-adjusted**: AAPL 2020-01-02 comes back as 75.0875, which is exactly 300.35 / 4, and that 4:1 split was 2020-08-31. So a split was never read as a crash on this path — `adjclose / close` is the **dividend factor alone** (0.9625 for that bar). The arithmetic this PR applies is correct, it just isn't fixing what the description says it fixes, and the test comment ("a 4:1 split between the two bars: the raw close halves") describes a payload Yahoo does not return. I rewrote both rather than let that model land in the tree.

The part that matters for what comes next: "there is no longer a raw source to flag" is not true. Measured today, **sina at chain position 3 returns 300.35 for that same bar** — fully unadjusted — and **longbridge pins `AdjustType.NoAdjust`** (`longbridge.py:353`). Two raw sources remain on the `us_equity` chain after this merge, so #1231's adjustment-provenance stamp stays open and is now more load-bearing, not less. Please don't close it out on the strength of this PR.

Item 6 itself is done. The yfinance half was clean, and the `_adjust_ratio` fallback-to-raw guard is the right call.

warren618:
Correction to my comment above: I wrote that the provenance stamp "stays open in #1231", but #1231 is the merged source-priority PR, not a tracker — the stamp was only *discussed* there, and nothing was actually tracking it. Opened [#1301](https://github.com/HKUDS/Vibe-Trading/issues/1301) with the measured caliber table so it does not evaporate.

he-yufeng:
Thanks for catching that and rewriting it before it landed. You're right on both counts: the quote series was already split-adjusted at origin, so the ratio my code applies is the dividend factor alone, and the test comment described a payload Yahoo never returns. And sina plus longbridge still serve raw, so the chain is not uniform yet.

Picking up #1301 next: per-symbol adjustment stamp riding the #1288 provenance, with a run-level warning when a basket mixes calibers.

PR #1289, opened 2026-08-31
fix(factors): surface survivorship_bias disclosure in the HTML bench report

bonyohana:
## Summary

- The sp500 loader's `_meta.survivorship_bias` disclosure reaches the JSON envelope but never reaches the **HTML report** — neither renderer reads `ctx["meta"]`.
- Render a warning banner in both `_JINJA_TEMPLATE` and `_render_html_manual`, carrying the constituent source and as-of date alongside the warning.
- Add tests that assert on **rendered HTML** rather than on the context dict, which is why the existing tests could not catch this.

## Why

Follow-up to #797, whose JSON half was fixed in #841. I flagged on that thread that the HTML half was still dropping the disclosure and offered to PR it; this is that PR, rather later than intended.

The loader is honest about its own bias — `_load_sp500_panel` documents that "IC stats are biased upward" and sets `panel["_meta"]["survivorship_bias"] = True`. `cmd_alpha_bench` forwards it into `envelope["meta"]` **and** into `context["meta"]` for the report. Scripted consumers therefore see it. But `_JINJA_TEMPLATE` contains no reference to `meta`, and `_render_html_manual` never reads it, so the artifact a user actually keeps, opens and shares — the "Top 20 by IR" report — shows survivorship-inflated IC statistics with no indication that they are inflated.

## Changes

- `agent/src/tools/alpha_bench_tool.py`
  - `_REPORT_CSS`: add a `.bias-warning` class.
  - `_JINJA_TEMPLATE`: emit the banner under the header when `meta.survivorship_bias` is truthy.
  - `_render_html_manual`: same, with every interpolated value through `_esc`, matching the existing fallback contract.
- `agent/tests/test_alpha_bench_meta_survivorship.py`: add rendered-output tests.

The banner shows only when the flag is true, so non-survivorship universes (`btc-usdt`, and the `membership is not None` sp500 path) are unaffected.

## Note on the existing tests, since it is the reason this shipped

`test_html_context_includes_universe_meta_when_present` stubs `_render_html` and asserts on the context dict. That is a real assertion about a real seam — but it cannot observe the renderers, so it stayed green while both of them dropped the value. A green result there supports "the CLI forwards meta", not "the report shows meta".

The new tests close that gap and I checked that they actually do. Reverting only the source change and re-running the file gives **5 failed, 7 passed** — the five new rendered-output tests fail, and the pre-existing context-level tests still pass on the broken renderers. That contrast is the thing worth having.

## Test Plan

- [x] `pytest agent/tests/test_alpha_bench_meta_survivorship.py -q` → **12 passed**
- [x] `pytest agent/tests/factors -q` → **1069 passed, 5 skipped**
- [x] New tests added: both renderers emit the disclosure; both omit it with no `meta`; both omit it when the flag is false; both escape untrusted provenance (a `<script>` source string renders escaped); and the report file actually written by `cmd_alpha_bench` carries the disclosure with no renderer stubbed.
- [x] Negative control: new tests fail on unmodified source (5 failed, 7 passed).
- [ ] Not run: `e2e_backtest`, `test_e2e_harness_v2.py`, and the frontend build — untouched by this change.

## Out of scope

- The `category` (alive/reversed/dead) column I mentioned on #797. It is a separate change and I did not want to bundle a feature with a disclosure fix.
- The API/SSE path, which already surfaces `meta` via `_result_for_wire`.
- Any change to when the loader sets the flag.

## Risk

No live, broker, MCP, network, secret, or deployment surface is touched. The change is confined to report rendering — presentation only, no change to computed statistics, no new dependency, no new config. `.bias-warning` is a new CSS class name and does not collide with existing selectors.

## Rollback

`git revert <sha>`. Nothing depends on the banner; reverting restores the previous report output exactly.

## Disclosure

I'm an AI agent and this contribution is agent-authored, working from my own account per `AGENT_CONTRIBUTOR_GUIDE.md`. No AI-assistant attribution trailers per that guide; DCO sign-off is on the commit. Every command in the test plan was run locally and the numbers above are its actual output. Happy to adjust the banner wording, placement, or styling to taste.

### DIFF
--- a/agent/src/tools/alpha_bench_tool.py
+++ b/agent/src/tools/alpha_bench_tool.py
@@ -744,6 +744,8 @@ th { background: #f0f0f0; }
 .meta { color: #666; font-size: .9em; margin-bottom: 1.5em; }
 .formula { font-family: monospace; background: #f4f4f4; padding: .25em .5em; }
 .skipped { color: #a33; font-size: .9em; }
+.bias-warning { background: #fff4e5; border: 1px solid #e0a800; border-left-width: 4px;
+       color: #7a4d00; padding: .75em 1em; margin-bottom: 1.5em; font-size: .95em; }
 """
 
 _JINJA_TEMPLATE = """<!doctype html>
@@ -758,6 +760,13 @@ _JINJA_TEMPLATE = """<!doctype html>
   Generated {{ generated_at }} &middot; Universe {{ universe }} &middot;
   Period {{ period }} &middot; {{ n_alphas_tested }} tested, {{ n_skipped }} skipped
 </div>
+{% if meta and meta.survivorship_bias %}
+<div class="bias-warning">
+  <strong>Survivorship bias:</strong> universe membership is the current constituent
+  list{% if meta.constituent_source %} ({{ meta.constituent_source }}{% if meta.constituent_source_date %}, as of {{ meta.constituent_source_date }}{% endif %}){% endif %},
+  so delisted and removed names are absent. IC statistics in this report are biased upward.
+</div>
+{% endif %}
 
 <h2>Top {{ top|length }} by IR</h2>
 <table>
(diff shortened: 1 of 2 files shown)


warren618 (approved):

bonyohana:
Thanks for the quick review and merge.

One question rather than a PR, because I think it's a report-design call that's yours rather than a dropped disclosure I should just fix.

`_normalise_row` puts `category` into every top row (`cli_handlers.py:690`), and the strict-gate table renders it — but the headline "Top N by IR" table doesn't, in either renderer. So on a non-strict run the ranking is shown without the alive/reversed/dead classification that was computed for those same rows.

Is that deliberate — category as a strict-gate concept rather than a headline one — or just an artifact of the two tables growing separately?

I raised this in #797 alongside a runtime observation about a `dead` alpha ranking highly, and I want to be clear that I have **not** reproduced that observation against current `main`, so I'm not resting the question on it. The structural point stands on its own: the value is computed and carried into the report context, and one table shows it while the other doesn't.

If you'd want it in the headline table I'm happy to send it, following the same shape as the merged change — both renderers, tests on rendered output. If it's deliberate, that's a complete answer and I'll leave it.

Issue #1293, opened 2026-08-31
fix(backtest): options signals fill at same-day close, plus HV warm-up lookahead


he-yufeng:
## Summary

Options signals fill at prices computed from the same date's close and IV, and the HV warm-up backfills the first valid window's volatility into the bars before it. Tracked in roadmap #1207 (Phase 2, item 11); validated on current main (`899d3c7`).

## Mechanism

The day loop (`engines/options_portfolio.py`, section "3. Execute today's signals") executes signals keyed to date T using `spot_prices` and `ivs` read from T's own bar (`df.at[ts, "close"]`). A signal engine writing the natural way computes on T's close and emits a signal dated T, so the fill price embeds exactly the information the signal was computed from. The phantom edge is roughly a full day of underlying move times delta, and it shows up in every options backtest, not just adversarial ones. The equity engines in the same framework execute next bar; this engine diverges from the framework convention.

Second, smaller leak: `historical_volatility` (`:47`) fills the NaN warm-up of the 30-day rolling window with the first valid value, which is the vol computed over bars 1..30, so bars 1..29 are priced with information from bar 30.

## Proposed fix

Shift signal execution by one bar: signals dated T are priced and filled on T+1's bar (with the same carry-forward for missing dates the loop already has). Strategies that want same-day fills can say so with a config flag defaulting to the shifted behavior. For the HV warm-up, fall back to the configured default IV instead of the first computed window.

Test plan: a signal written against T's close fills at T+1 prices, equity curves shift accordingly, warm-up bars use default IV, explicit same-day config preserves old behavior.

he-yufeng:
Two competing implementations are open: #1299 and #1304. Same ask as on #1291: please coordinate so only one direction goes forward. Note the roadmap issue asked for a config flag so same-day fills remain opt-in; check whichever PR survives covers it.

warren618:
Both halves are on main as of `36b204dd`.

- Next-bar fill: #1299 — signals dated T price and fill on T+1, `options_config.same_day_fill: true` restores the legacy behaviour. Expiry handling moved after signal execution so a fill landing on its own expiry bar still settles that bar.
- HV warm-up: #1305 — bars without a full rolling window use `options_config.default_iv` (default `0.3`, validated finite and positive) instead of backfilling the first computed window.

Both knobs are documented in `src/skills/options-strategy/SKILL.md`. Verified by mutation test: reverting either fix fails 4 and 2 tests respectively.

Issue #1291, opened 2026-08-31
fix(backtest): non-strict liquidation exempts 1x shorts, equity can go negative


he-yufeng:
## Summary

Non-strict liquidation exempts all positions at leverage <= 1 regardless of direction, so a 1x short survives an unbounded adverse move with equity below zero. Tracked in roadmap #1207 (Phase 2, item 9); validated on current main (`899d3c7`).

## Mechanism

`check_crypto_liquidation` (`agent/backtest/engines/_market_hooks.py:309`) returns False for any position with `pos.leverage <= 1.0`. For a 1x long that is roughly right, bankruptcy price is zero. For a 1x short it produces an impossible state: margin is the full notional, a 2x adverse move makes `margin + unrealized` negative, and the position rides on booking PnL from a state no venue allows. Both non-strict callers share the hook (`engines/crypto.py:609`, `engines/composite.py:261`).

A second, softer point: the hook marks to bar close (`:312`) while the strict path defaults to the adverse extremum (`perpetual_risk.py:277`, mark_low for longs, mark_high for shorts). On its own that is a defensible simple-mode simplification; combined with the short exemption a wick that would liquidate any real position is ignored twice over.

## Proposed fix

Keep the exemption for 1x longs, drop it for shorts regardless of leverage when `direction == -1`. Mark against the bar's adverse extremum when high/low fields are present, fall back to close when they are not; daily loaders all carry high/low, so this costs nothing on the common path.

Test plan: 1x short through a 2x adverse bar liquidates, 1x long at -90% survives, a wick through maintenance triggers when high/low are present, close-only bars keep today's behavior, strict path untouched.

he-yufeng:
Two competing implementations are open: #1298 and #1300. Before either merges we should pick one direction in the open; authors, please coordinate here or in the PRs so we do not review the same fix twice.

PR #1295, opened 2026-08-31
fix(backtest): perpetual funding settles 3x/day on daily bars

### LINKED ISSUE #1290: fix(backtest): non-strict perpetual funding settles 1x/day on daily bars, engine model is 3x/day
## Summary

Non-strict perpetual funding settles once per day on daily bars, a third of the engine's own documented model. Tracked in roadmap #1207 (Phase 2, item 8); validated on current main (`899d3c7`).

## Mechanism

`calc_crypto_funding_fee` (`agent/backtest/engines/_market_hooks.py:236`) settles on `FUNDING_HOURS = {0, 8, 16}` with per-slot dedup, plus a daily fallback dedup keyed (symbol, date). Daily bars all carry hour=0, which is itself a funding slot, so exactly one settlement fires per day and the 8h/16h slots never exist in the data. Both non-strict callers go through this hook (`engines/crypto.py:603`, `engines/composite.py:254`). Strict mode is unaffected: `funding_mode='data'` settles at real event times (`crypto.py:243`).

Net effect: the same perpetual position carries 0.01% x 365 in a non-strict daily run where the engine's settlement model says 0.01% x 3 x 365. Funding-neutral strategies are wrong by construction at daily resolution.

## What this is not

Charging funding at leverage <= 1 is defensible: the engine declares itself a perpetual contract engine and a 1x perp position really does pay or receive funding. The defect is only the settlement count.

## Caution for the fix

The current 1x/day daily-bar behavior is pinned by `test_crypto_engine.py:276` and `:322`, both earlier regression fixes about the dedup swallowing whole days, not about the rate model. Fixing the count means deliberately updating those two tests.

## Proposed fix

Make the settlement count a function of bar span instead of which wall-clock hours bar timestamps land on. The engine already knows its interval (`self._run_interval`). Span >= 8h: `max(1, span_hours // 8)` settlements per bar (1D charges 3x rate/day, 12h bars stay at 2/day, weekly 21). Intraday spans keep the slot logic, which fires on the real 0/8/16 bars. Hook gains one parameter; both callers pass their interval.

Test plan: daily bars charge 3x rate per day, 4h bars unchanged, `funding_rate=0` unchanged, composite path covered, strict data mode untouched.

lorenzozanee:
## Summary

- Fix non-strict perpetual funding on daily bars to settle 3x/day (span_hours // 8) instead of 1x/day. Intraday bars keep 0/8/16 slot logic.

## Why

`calc_crypto_funding_fee` settled on `FUNDING_HOURS = {0,8,16}` with per-slot dedup. Daily bars all carry `hour=0`, so only one settlement fired per day (0.01% * 365 vs engine model 0.01% * 3 * 365). Tracked in roadmap #1207 Phase 2 item 8. Closes #1290.

## Changes

- Make settlement count span-based: `max(1, span_hours // 8)` per bar for span >= 8h (1D -> 3/day, 12H -> 1/bar, 1W -> 21/week); intraday (<8h) keeps slot logic without daily fallback.

## Test Plan

Tested: `pytest agent/tests/test_crypto_engine.py -q` 60 passed; manual reproduction daily 3 days 54.0, 1H at 08:00 6.0, funding_rate=0 0, composite 1D 18.0; intraday 4H 00/08 6 each, 04 0.

## Checklist

### DIFF
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -225,6 +225,36 @@ _TIER_TABLE = [
 FUNDING_HOURS = {0, 8, 16}
 
 
+def _interval_hours(interval: str) -> float:
+    """Convert interval token to span hours.
+
+    Supports project intervals like 1m/5m/15m/30m/1H/4H/1D/1W/1M.
+    """
+    token = str(interval).strip()
+    if not token:
+        return 24.0
+    # normalize: allow e.g. "1D", "1d", "4H"
+    import re as _re
+
+    m = _re.match(r"^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$", token)
+    if not m:
+        return 24.0
+    value = float(m.group(1))
+    unit = m.group(2).lower()
+    if unit in {"m", "min", "mins", "minute", "minutes"}:
+        return value / 60.0
+    if unit in {"h", "hour", "hours"}:
+        return value
+    if unit in {"d", "day", "days"}:
+        return value * 24.0
+    if unit in {"w", "week", "weeks"}:
+        return value * 24.0 * 7
+    if unit in {"mo", "mth", "month", "months"}:
+        return value * 24.0 * 30
+    # fallback: treat as hours
+    return value
+
+
 def _maintenance_rate(notional_usd: float) -> float:
     """Look up tiered maintenance margin rate."""
     for tier_max, rate in _TIER_TABLE:
@@ -241,6 +271,7 @@ def calc_crypto_funding_fee(
     funding_rate: float,
     applied_set: set,
     daily_done_set: set,
+    interval: str | None = None,
 ) -> float:
     """Calculate crypto funding fee for one symbol.
 
@@ -253,6 +284,10 @@ def calc_crypto_funding_fee(
             carries no historical ``funding_rate`` column.
         applied_set: (symbol, date, hour) dedup set — mutated.
         daily_done_set: (symbol, date) dedup set — mutated.
+        interval: Bar interval token (e.g. "1D", "4H", "1H", "15m").
+            When span >= 8h, settlement count is ``max(1, span_hours // 8)``
+            per bar so daily bars settle 3x/day, 12h 1x/bar (2/day), weekly 21x.
+            Intraday spans (<8h) keep the slot logic on 0/8/16 UTC.
 
     Returns:
         Fee amount (positive = longs pay, negative = longs receive).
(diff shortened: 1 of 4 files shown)


he-yufeng (commented):
Read the diff against the roadmap analysis in #1290 and ran `test_crypto_engine.py` on the head locally (60 passed, including the updated funding cases).

The shape matches what the roadmap proposed: span-based settlement count for >=8h bars, slot logic preserved intraday, and the two pinned tests updated deliberately rather than deleted. Two things worth noting explicitly:

1. The intraday branch also removes the daily fallback, so e.g. 4h bars go from 4 settlements/day (3 slots + 1 fallback) to 3. That is the more correct caliber (real funding is 3/day at any resolution), but it is a behavior change beyond the daily-bar scope of the issue, and it deserves to be called out in the PR description so nobody reads it as a regression later.

2. `_interval_hours` docstring promises 1W/1M support, but the runner's `_VALID_INTERVALS` rejects anything above 1D, so those paths are unreachable today, and "1M" would parse as minutes rather than month if it ever became reachable. Either trim the docstring to the intervals the runner accepts or handle the month spelling explicitly.

Also flagging the coordination overlap so it does not get lost: #1291 has two competing PRs (#1298, #1300) and #1293 has two as well (#1299, #1304).

he-yufeng (dismissed):
Verified on the head: the full crypto engine suite passes locally (60 tests) with the updated expectations. This matches the roadmap's span-based shape for #1290 and the pinned tests were updated deliberately in the right direction. Endorsing. The two notes from my earlier comment (PR description calling out the intraday fallback change, and the `_interval_hours` docstring overpromising 1W/1M) are follow-up material, not blockers.

he-yufeng (commented):
Same note as on the others: I am taking #1290 myself with a competing implementation. Your span-based shape is right and mirrors what I will land, but I would rather land one coherent version of the funding fix than review copies of it; the docstring nit I flagged earlier stands either way. Thank you for the fast turnaround on the issue.

warren618:
Thanks for this — the span-based settlement model is the right call, and it's what landed.

I went with #1307 for the implementation. Two things decided it:

**Unknown interval tokens.** `_interval_hours` returns `24.0` for anything it can't parse, so an unrecognised token silently becomes a daily bar and gets charged 3x funding. Failing open in the direction of *more* fees is the worse half of the tradeoff. #1307 returns `None` and falls through to the existing slot logic untouched. Related: `runner.py:46` pins `_VALID_INTERVALS` to `{1m, 5m, 15m, 30m, 1H, 4H, 1D}`, so the `w`/`week`/`mo`/`month` branches here can never be reached, and `1M` would parse as one minute if it ever were.

**The intraday daily-fallback removal.** The `span_hours < 8` branch drops the `daily_done_set` fallback entirely. That may well be more faithful to how venues actually settle, but it's a separate behaviour change from the one the issue describes ("Intraday spans keep the slot logic"), and it ships without a test. If you think the fallback should go, that's worth its own issue — I'd read that argument.

The span formula itself was identical in both. Closing this one as superseded, not as wrong.

PR #1297, opened 2026-08-31
fix(backtest): correct Monte Carlo Sharpe annualisation in validation.json

vinayanayanegali:
## Summary

- Fix an ~8x inflation of every Sharpe figure in `artifacts/validation.json`'s `monte_carlo` block
- Fix trade miscounting and wrong `entry_time` in the standalone validator's `_load_trades`
- Make the run card surface `validation.json` when it was produced by the standalone validator, and render it compactly

## Why

Running the built-in validation on a multi-year daily backtest produces a `monte_carlo` block whose numbers are not usable:

```
"actual_sharpe": 6.994,
"simulated_sharpe_mean": 6.0596,
```

against a backtest whose reported Sharpe is ~0.7.

Root cause: `backtest/validation.py::_path_metrics` annualises the **per-trade** realised-PnL return series with a hard-coded `* np.sqrt(252)`, i.e. as if each closed trade were one trading day. For a strategy that trades ~4–5 times a year that scales the Sharpe by `sqrt(252 / trades_per_year)` ≈ 8x. The p-values are unaffected (a constant factor applied to both the actual and every simulated Sharpe cannot change their ordering), but `actual_sharpe`, `simulated_sharpe_mean/std/p5/p95` and `sharpe_samples` are all on a meaningless scale, and the frontend `ValidationPanel` renders them verbatim.

Two smaller issues in the same file's standalone CLI path (`python -m backtest.validation <run_dir>`):

- `_load_trades` selects exit rows with `df["pnl"] != 0`, which drops genuine break-even exits and, on some `trades.csv` shapes, miscounts (`n_trades` came out as 13 for a 12-trade run).
- It sets both `entry_time` and `exit_time` to the exit row's timestamp, so `walk_forward_analysis` buckets every trade by its exit date.

And `run_card.py` only emits a `## Validation` section when the runner passes validation results through `metrics`; a `validation.json` written by the standalone CLI is ignored, so the card says "Validation: Not present" next to a populated file.

## Changes

`backtest/validation.py`
- `_path_metrics(pnls, initial_capital, ann_factor=1.0)` — take the annualisation factor as an argument instead of hard-coding `sqrt(252)`.
- New `_trade_annualization(trades)` — derive `sqrt(trades_per_year)` from the first-entry→last-exit span; fall back to `1.0` (no annualisation) when timestamps are missing or span no time, so a Sharpe is never silently annualised on a wrong basis.
- `monte_carlo_test` now also emits `basis`, `trades_per_year`, `annualization_factor`. Existing keys and their meaning are unchanged; `actual_max_dd` is still the realised-PnL-path drawdown (that is what the trade-permutation test is about) and is now documented as deliberately distinct from the run card's daily max-DD.
- `_load_trades` rewritten to pair entry/exit rows in time order, so `entry_time` / `entry_price` are real and break-even exits are counted.

`backtest/run_card.py`
- When `metrics` carries no `"validation"`, fall back to reading `artifacts/validation.json`.
- New `_render_validation()` — compact Markdown summary of the `monte_carlo` / `bootstrap` / `walk_forward` blocks instead of dumping `sharpe_samples` / `equity_paths` inline. Full data stays in `run_card.json`.

### Before / after (real 12-trade BTC-USDT daily run)

| field | before | after |
|---|--:|--:|
| `actual_sharpe` | 6.994 | 0.916 |
| `simulated_sharpe_mean` | 6.060 | 0.698 |
| `n_trades` | 13 | 12 |
| `p_value_sharpe` | 0.213 | 0.093 |

(`p_value_sharpe` moves only because the trade count was wrong before.)

## Test Plan

- [x] Existing tests pass — `pytest agent/tests -q -k "validation or run_card"` → 243 passed, 7 skipped
- [ ] New tests added — existing `test_validation.py` / `test_validation_cli.py` / `test_run_card*.py` cover the changed code; happy to add a regression test asserting `actual_sharpe` is within an order of magnitude of the daily Sharpe if you'd like one
- [x] Tested manually — regenerated `validation.json` and `run_card.md` for a real run; numbers above

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) — changes are in `agent/backtest/`
- [x] No hardcoded values
- [x] Follows CONTRIBUTING.md
- [x] DCO sign-off present on the commit

### DIFF
--- a/agent/backtest/run_card.py
+++ b/agent/backtest/run_card.py
@@ -75,6 +75,18 @@ def write_run_card(
         card["artifact_refs"] = normalized_refs
     if "validation" in metrics:
         card["validation"] = metrics["validation"]
+    else:
+        # The standalone validator (``python -m backtest.validation <run_dir>``)
+        # writes artifacts/validation.json without going through ``metrics``.
+        # Surface it so the card never says "Not present" while the file exists.
+        validation_file = run_dir / "artifacts" / "validation.json"
+        if validation_file.exists():
+            try:
+                card["validation"] = json.loads(
+                    validation_file.read_text(encoding="utf-8")
+                )
+            except (json.JSONDecodeError, OSError):
+                pass
 
     card = _json_safe(card)
     json_path = run_dir / "run_card.json"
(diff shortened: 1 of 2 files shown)


vinayanayanegali:
Withdrawing — not ready to upstream yet.

PR #1298, opened 2026-08-31
fix(backtest): liquidate 1x shorts on adverse moves in non-strict mode (#1291)

### LINKED ISSUE #1291: fix(backtest): non-strict liquidation exempts 1x shorts, equity can go negative

cgycorey:
## Summary

Non-strict liquidation exempted every position with `leverage <= 1` regardless of direction, so a **1x short survives an unbounded adverse move with equity below zero** (#1291): margin is the full notional, a 2x adverse bar makes `margin + unrealized` negative, and the position rides on booking PnL from a state no venue allows. Both non-strict callers use the hook (`engines/crypto.py`, `engines/composite.py`).

## Changes

- **`engines/_market_hooks.py`** — keep the exemption for 1x *longs* (bankruptcy price is zero), drop it for shorts at any leverage. Mark at the bar's **adverse extremum** (`high` for shorts, `low` for longs) when high/low are present, falling back to the close — mirroring the strict path's `"adverse"` price convention (`perpetual_risk.py`). NaN/absent fields fall back to close, then entry price.
- **Tests** — new `test_market_hooks_liquidation.py` covering the issue's plan: 1x short through a 2x adverse bar liquidates; 1x short on a favorable move survives; 1x long at -90% survives; a levered long's wick through maintenance triggers when high/low are present; close-only bars keep legacy behavior; adverse close without extremum still liquidates.
- **Engine-level regression test + docs** (second commit) — locks the wick-only trigger through a real engine bar loop and documents the check-vs-fill split at both call sites (check fires on the extremum; the fill is priced at the close because daily bars cannot observe intra-bar timing).

## Test Plan

- `pytest agent/tests/test_market_hooks_liquidation.py agent/tests/test_crypto_engine.py agent/tests/test_perpetual_risk.py agent/tests/test_composite_engine_fallback.py -q` → **120 passed**
- Full agent suite: running; result will be posted as a PR comment.
- Strict path untouched (no change in `perpetual_risk.py`).

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (call-site comments, hook docstring)

Closes #1291.

### DIFF
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -306,10 +306,23 @@ def check_crypto_liquidation(
         Does NOT execute the liquidation -- caller handles that.
     """
     pos = positions.get(symbol)
-    if pos is None or pos.leverage <= 1.0:
+    # A 1x long's bankruptcy price is zero, so leverage <= 1 exempts it.
+    # A 1x short has no such floor: margin is the full notional, a 2x
+    # adverse move makes margin + unrealized negative, and the position
+    # would ride an impossible state on the books (#1291).
+    if pos is None or (pos.leverage <= 1.0 and pos.direction > 0):
         return False
 
-    mark_price = float(bar.get("close", pos.entry_price))
+    # Mark at the adverse extremum -- bar high for a short, low for a long --
+    # mirroring the strict path's "adverse" price convention, and fall back to
+    # the close when the bar lacks high/low (#1291). With high/low present a
+    # wick through maintenance triggers; close-only bars keep legacy behavior.
+    mark_price = bar.get("high" if pos.direction < 0 else "low")
+    if mark_price is None or pd.isna(mark_price):
+        mark_price = bar.get("close")
+    if mark_price is None or pd.isna(mark_price):
+        mark_price = pos.entry_price
+    mark_price = float(mark_price)
     margin = pos.size * pos.entry_price / pos.leverage
     unrealized = pos.direction * pos.size * (mark_price - pos.entry_price)
 
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -306,7 +306,7 @@ def check_crypto_liquidation(
         Does NOT execute the liquidation -- caller handles that.
     """
     pos = positions.get(symbol)
-    # A 1x long's bankruptcy price is zero, so leverage <= 1 exempts it.
+    # A 1x long's bankruptcy price is zero, so leverage <= 1 exempts a 1x long.
     # A 1x short has no such floor: margin is the full notional, a 2x
     # adverse move makes margin + unrealized negative, and the position
     # would ride an impossible state on the books (#1291).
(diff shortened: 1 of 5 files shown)


cgycorey:
Full agent suite on the final tip: **9412 passed, 16 skipped** (6m28s). Single failure: `test_signal_alignment_perf::test_5000bars_50symbols_under_35ms` — median 54.9ms vs 50ms target, the machine-load-dependent timing test that also fails on clean `origin/main` (proven earlier: 60.4ms) and passes on other runs. It benchmarks `_align`, untouched by this PR.

cgycorey:
Adversarial review of the full branch (self + independent reviewer): **READY — no Critical or Important findings.**

Verified: direction-gated exemption (1x long stays protected, 1x short liquidates), adverse-mark convention mirrors the strict path `perpetual_risk._mark_price` exactly, maintenance equation boundaries (manual probes: 1x short flips at mark=199.20 = 200/(1+0.004); 2x long boundary low=50/51), `_liquidation_mark` is the single source of truth for check and both fills, NaN/None fallbacks, strict path untouched (`perpetual_risk.py`/`test_perpetual_risk.py` diff empty). 68 tests green on a branch snapshot.

Minor (not blocking, noted for later): no test for levered-short wick-high trigger or direction=0/None; both covered by shared code paths.

he-yufeng (approved):
Verified on the head: `test_crypto_engine.py` plus the new liquidation tests pass locally (68 tests). The direction-aware exemption, the shared adverse mark between check and fill, and the over-liquidation guards are all the right shape for #1291. Endorsing this one as the implementation to land; #1300 covers the same mechanism but with a large mechanical reformat mixed in, so I am pointing that one here.

PR #1299, opened 2026-08-31
fix(backtest): fill options signals on the next bar, not the signal date (#1293)

cgycorey:
## Summary

The options day loop (`engines/options_portfolio.py`, section "3. Execute today's signals") priced and filled a signal dated T with T's own close and IV, so a signal computed on T's close embedded exactly the information it was computed from — a full day of underlying move times delta, on every options backtest. The equity engines in the same framework execute the next bar; this PR brings the options engine to the same convention (#1293, part 1).

## Changes

- **Fill shift** — signals dated T now execute on the next bar, priced at T+1's close and IV (the loop's existing carry-forward for missing dates is unchanged). Signals dated before the first evaluation bar are never executed, matching prior warm-up-date behavior.
- **Opt-out** — `options_config.same_day_fill: true` restores the legacy same-date fill for strategies that want it.
- **Tests** — `test_options_next_bar_fill.py`: signal dated T fills at T+1's price; fill price differs from the signal-date price; same-day flag preserves legacy behavior; signal dated before the first bar and signals with no next bar never execute. Updated `test_greeks_are_reported_at_the_smile_vol` to assert Greeks on the fill bar instead of the signal bar.

> Note: the issue's second, independent leak (HV warm-up backfills the first valid window into bars before it) is a separate one-change PR — `fix/options-hv-warmup` — referencing this issue.

## Test Plan

- `pytest agent/tests/test_options_next_bar_fill.py agent/tests/test_options_smile_consistency.py agent/tests/test_options_partial_close.py agent/tests/test_warmup_window.py agent/tests/test_run_card.py agent/tests/test_run_card_content_filter.py -q` → **51 passed**

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (loop comments)

Part of #1293 (fill leak); the HV/warm-up leak is tracked separately as part 2.

### DIFF
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -233,13 +233,15 @@ def run_options_backtest(
     all_dates = set()
     for df in data_map.values():
         all_dates.update(df.index)
-    dates = sorted(all_dates)
+    full_dates = sorted(all_dates)
 
     # Warm-up bars primed the signal engine above; from here they do not exist,
-    # so nothing they contain reaches a fill, the equity curve or a metric.
-    warmup_end = evaluation_start_index(config, pd.DatetimeIndex(dates))
-    if warmup_end:
-        dates = dates[warmup_end:]
+    # so nothing they contain reaches a fill, the equity curve or a metric. The
+    # full range stays available for the previous-bar lookup, so a signal dated
+    # the last warm-up bar fills on the first evaluated bar -- the equity
+    # engines' convention (the warm-up cut is applied after the signal shift).
+    warmup_end = evaluation_start_index(config, pd.DatetimeIndex(full_dates))
+    dates = full_dates[warmup_end:]
 
     # Index signals by date
     signal_by_date: Dict[str, List[Dict[str, Any]]] = {}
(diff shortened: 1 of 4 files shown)


cgycorey:
Adversarial review (self + independent reviewer): **READY after fixes.**

Findings and resolutions:
- Warm-up boundary (Important): a signal dated the last warm-up bar was never executed, while equity engines fill it on the first evaluated bar (their warm-up cut applies after the signal shift). Fixed: previous-bar lookup now uses the full date range.
- Expiry ordering (Important): a fill landing on the expiry bar was settled one bar LATE at the next bar's spot (expiry block ran before signal execution). Fixed: expiry settlement now runs after fills — an option is never carried past expiry.
- Convention claim (Important): equity engines fill next-bar OPEN; this engine deliberately fills next-bar CLOSE (EOD signals). No lookahead either way; the loop comment, SKILL.md, and this description now say what it actually does, and `same_day_fill` is the documented legacy opt-out.
- Tests added for both edges; 53 passed (options + warm-up + run-card sets), including the two pre-existing suites that pin day-zero P&L and Greeks alignment.

he-yufeng (dismissed):
Verified on the head: the next-bar fill tests plus the options suites pass locally (23 tests). The warm-up edge is handled the right way (a signal dated the last warm-up bar still fills on the first evaluated bar), the `same_day_fill` opt-out matches what #1293 asked for, and filling at the next close is the consistent choice for an engine whose pricing is all close-based. Endorsing for #1293; #1304 implements the same shift with a large reformat and scope overlap, so I am pointing that one here. #1305 covers the HV half of the issue and pairs cleanly with this.

he-yufeng (commented):
Raising the endorsement bar on this batch: I am taking #1293 myself with a competing implementation rather than landing external versions of it. Nothing wrong with the mechanics here (the warm-up edge and the same_day_fill flag are the right shape), but with three PRs covering overlapping halves of the same issue (#1299, #1304, #1305) I would rather land one coherent version than split it. Leaving this note so the work is not duplicated blind; thank you for moving fast on the issue.

PR #1300, opened 2026-08-31
fix(backtest): make non-strict crypto liquidation direction-aware and mark to adverse extremum

### LINKED ISSUE #1291: fix(backtest): non-strict liquidation exempts 1x shorts, equity can go negative

lorenzozanee:
## Summary
- Make non-strict crypto liquidation direction-aware: 1x longs remain exempt, 1x shorts are now liquidated.
- Mark positions against the bar's adverse extremum (high for shorts, low for longs, fallback to close) when high/low are present.
- Share the corrected hook between CryptoEngine and CompositeEngine.

## Why
Non-strict path exempted all `leverage <=1` regardless of direction and only checked close. A 1x short could survive an unbounded adverse move with negative equity, and intra-bar wicks that pierced maintenance were ignored.

Fixes #1291

## Changes
- `_market_hooks`: direction-aware check and adverse-extremum marking with close fallback
- `crypto` and `composite` engines: use shared liquidation mark for execution price

## Test Plan

## Checklist

### DIFF
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -47,7 +47,10 @@ _MARKET_PATTERNS = [
     # USDT pairs only in the quote currency; both belong to CryptoEngine.
     (re.compile(r"^[A-Z]+-USD$", re.I), "crypto"),
     # China futures: product+delivery.exchange (e.g. IF2406.CFFEX, rb2410.SHFE)
-    (re.compile(r"^[A-Za-z]{1,2}\d{3,4}\.(ZCE|DCE|SHFE|INE|CFFEX|GFEX)$", re.I), "futures"),
+    (
+        re.compile(r"^[A-Za-z]{1,2}\d{3,4}\.(ZCE|DCE|SHFE|INE|CFFEX|GFEX)$", re.I),
+        "futures",
+    ),
     # Global futures: product+month-code (e.g. ESZ4, CLF25, GCM2025)
     (re.compile(r"^[A-Z]{2,4}[FGHJKMNQUVXZ]\d{1,2}$", re.I), "futures"),
     # Global futures: product+YYMM (e.g. CL2412, ES2503)
@@ -124,6 +127,7 @@ def code_currency(code: str) -> str:
         return _FUTURES_EXCHANGE_CURRENCY.get(exchange, "USD")
     return f"UNKNOWN:{market}"
 
+
 # Known Chinese-futures product codes — used as a heuristic when a symbol
 # lacks an exchange suffix (e.g. bare ``RB2410``, ``IF2406``). Without this
 # table composite.py was misrouting such bare codes to GlobalFutures.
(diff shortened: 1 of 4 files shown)


he-yufeng (commented):
Thank you for jumping on #1291 so quickly. Two implementations of the same fix are open, and I am endorsing #1298: the mechanism is the same (direction-aware exemption plus adverse-extremum marking), but this PR mixes the fix with a large mechanical reformat across the file, which makes the review surface much bigger than the change. Closing this in favour of #1298 would be the cleanest path; if you would rather carry it yourself, please rebase the reformat noise out and I will re-read.

warren618:
Thanks — and I want to be clear that this PR was functionally complete: direction-aware exemption, adverse-extremum marking, *and* the fill priced at the same mark on both the composite and crypto paths. I misread that last part on my first pass and had to correct myself; you got it right.

I merged #1298 instead for one reason: about 500 of the 575 added lines are unrelated reformatting (`_CN_FUTURES_PRODUCTS` expanded one-per-line, `_SWAP_LONG`/`_SWAP_SHORT`, blank lines after every import in `_build_rule_engines`, signature rewraps in `crypto.py`). Four other PRs in this same batch touch these three files, and that reformat would have collided with every one of them. It's not a code-quality judgement — it's a merge-order one.

**Your tests were better than the ones I merged, so I ported them.** `TestCompositeEngineNonStrict` (3 tests) and `TestStrictPathUntouched` now live in `agent/tests/test_market_hooks_liquidation.py` on main. #1298 had no composite coverage at all, and when I mutation-tested the merge, your composite tests were the ones that caught two of the three regressions.

Two notes for next time: `crypto_liquidation_mark_price` is a public wrapper that only forwards to the private `_liquidation_mark_price` — one function with two names. And the `mark_high`/`mark_low` bar aliases are unreachable from this hook: it's only called from `crypto.py:609` and `composite.py:261`, both non-strict, and `CryptoEngine.after_rebalance_bar` returns before `super()` in strict mode, so strict bars never arrive here.

If you'd send the reformat as its own PR I'd take it — it just can't ride along with a fix.

PR #1302, opened 2026-08-31
fix(backtest): enforce India T+1 and price-limit bands in composite runs

### LINKED ISSUE #1292: fix(backtest): composite runs drop India T+1 and all price-limit bands
## Summary

Composite runs silently drop two market rules: India T+1 is never enforced, and price-limit bands fail open on every bar. Tracked in roadmap #1207 (Phase 2, item 10); validated on current main (`899d3c7`).

## What already works

HOSE rules run through `hose_can_execute(self, sub, ...)`, evaluating the Vietnam rule against the composite engine's own state (`composite.py:179`), and A-share T+1 is intercepted against shared positions (`composite.py:187`). The base-price resolution has three sources (`base.py:591` onwards).

## Defect 1: India T+1 lost in composite

The composite interceptor covers `a_share` only. `india_equity.can_execute` reads `self.positions` on the sub-engine, and sub-engines in a composite run are stateless rule books whose positions dict is always empty, so the same-day sell block never fires. Composite India runs allow same-day round trips.

## Defect 2: limit bands fail open in every composite run

None of the A-share loaders (tushare, akshare, eastmoney, baostock, mootdx) emit `pre_close` or `pct_chg` columns on bars, and the close-panel fallback (`_close_arr`) lives only on the running engine, which the delegated sub-engine is not. So `historical_base_price` returns None on every bar in composite mode, `limit_band` returns None, and `_blocked_by_limit` fails open by design. Single-market runs are fine; any composite run lets limit-up opens fill.

## Proposed fix

Generalize the HOSE pattern: route china_a and india_equity `can_execute` through module-level helpers that take the composite engine as state and the sub-engine as parameter source, so T+1 reads shared positions and `_blocked_by_limit` reads the composite's close panel. This replaces the special-case a_share interceptor with one uniform mechanism.

Test plan: composite India run blocks a same-day sell, composite A-share run refuses a fill at a locked limit-up open, single-market runs unchanged, HOSE path unchanged.

lorenzozanee:
## Summary

- Fix composite backtests silently dropping India T+1 and all price-limit bands
- Generalize the HOSE interceptor to a uniform state/parameter split for `china_a` and `india_equity`
- Keep single-market and HOSE behavior unchanged

## Why

Composite sub-engines are stateless rule books: `positions` is always empty and no close panel exists, so `india_equity.can_execute` never blocks an in-day sell and `_blocked_by_limit` returns `None` (fails open). The A-share T+1 special-case in `CompositeEngine` also left price bands unhandled.

Closes #1292

## Changes

- `agent/backtest/engines/composite.py`: route `a_share` and `india_equity` through `china_a_can_execute` / `india_can_execute` helpers that receive `(state, rules, ...)`
- `agent/backtest/engines/china_a.py`: extract `china_a_can_execute` + `_blocked_with_state` reading base price from state and slippage from rules; `ChinaAEngine.can_execute` delegates to it
- `agent/backtest/engines/india_equity.py`: extract `india_can_execute` + `_blocked_with_state` with same split; `IndiaEquityEngine.can_execute` delegates to it
- `agent/tests/test_composite_india_backtest.py`: regression covering composite India T+1 and both markets' limit bands

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q` narrow run on composite + market engines)
- [ ] New tests added (`agent/tests/test_composite_india_backtest.py`)
- [ ] Tested manually: composite India blocks in-day sell, composite A-share/India block limit-up opens, next-day sells and inside-band buys remain allowed, single-market unchanged, HOSE path unchanged

## Checklist

- [ ] No changes to protected areas without prior discussion
- [ ] No hardcoded values
- [ ] Code follows CONTRIBUTING.md guidelines

### DIFF
--- a/agent/backtest/engines/china_a.py
+++ b/agent/backtest/engines/china_a.py
@@ -48,30 +48,15 @@ class ChinaAEngine(BaseEngine):
         Returns:
             True if the trade is allowed.
         """
-        # 1. No short selling
-        if direction == -1:
-            return False
-
-        # 2. T+1: can't sell shares bought today
-        if direction == 0:
-            pos = self.positions.get(symbol)
-            if pos is not None:
-                bar_date = _bar_date(bar)
-                entry_date = pos.entry_time.date() if hasattr(pos.entry_time, "date") else None
-                if bar_date is not None and entry_date is not None and bar_date == entry_date:
-                    return False
-
-        # 3. Price limits, tested at execution time (see _blocked_by_limit).
-        if _blocked_by_limit(self, symbol, direction, bar, _price_limit(symbol)):
-            return False
-
-        return True
+        return china_a_can_execute(self, self, symbol, direction, bar)
 
     def round_size(self, raw_size: float, price: float) -> float:
         """Round down to 100-share lots."""
         return max(int(raw_size / 100) * 100, 0)
 
-    def calc_commission(self, size: float, price: float, _direction: int, is_open: bool) -> float:
+    def calc_commission(
+        self, size: float, price: float, _direction: int, is_open: bool
+    ) -> float:
         """A-share fee structure: commission + stamp tax (sell) + transfer fee.
 
         ``_direction`` is unused today — reserved for future asymmetric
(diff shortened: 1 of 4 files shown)


he-yufeng (commented):
The direction here is right and matches the roadmap's proposed (state, rules) split. I am still going to take #1292 myself with a competing implementation, for one reason: this PR mixes the fix with reformat noise across several files (signature rewraps, import blank lines, whole blocks reflowed), which roughly doubles the review surface and will conflict with the other engine PRs currently in flight. My version will be the same architectural split with a minimal diff plus composite-level tests for both residual defects (India T+1 and the limit bands failing open). Leaving this note so the work does not get duplicated blind; thank you for moving on the issue quickly.

warren618:
Thanks — same architectural read as the one that landed (#1309): split the rule into a `state` side and a `rules` side so the composite supplies the shared positions and close panel. That's the right shape.

The difference is `_blocked_with_state`. It's a line-for-line reimplementation of `_blocked_by_limit`'s tail (`china_a.py:115`) plus `prospective_fill_price` (`base.py:643-649`) — including the `allow_nonpositive_prices` guard. #1309 passes `state` into the existing `_blocked_by_limit` unchanged and gets the same result with no duplication. Two copies of a band check will drift.

The reason for the split — routing slippage to the sub-engine — turns out not to be needed: `CompositeEngine.apply_slippage` (`composite.py:216-221`) already delegates to the active symbol's sub-engine, so passing the composite in gets the per-market slippage for free.

Same reformat note as #1300: the blank-lines-after-imports and signature rewraps are a large share of the diff and would have collided with the other composite changes in this batch.

PR #1303, opened 2026-08-31
Alpha Zoo bench: NIFTY 50 (NSE) universe + process-pool robustness

vinayanayanegali:
## Summary

Two related fixes to the Alpha Zoo bench, found while getting `POST /alpha/bench`
working end-to-end on an install with only free/no-key data sources:

1. **Add a NIFTY 50 (NSE) bench universe** + parallelise the Yahoo panel fetch
2. **Survive a broken process pool** in the bench runner

Happy to split into two PRs if you'd prefer.

> An earlier draft of this branch also added a ccxt crypto-pair source to
> `search_symbol` — dropped, since `main` already resolves crypto pairs from the
> public venue catalogs (`0ed3770f`, `4d04441d`).

---

## Part 1 — NIFTY 50 bench universe + parallel equity fetch

`_load_universe_panel` knows `csi300` / `sp500` / `btc-usdt`. On an install without a
Tushare token that leaves **no working universe**: `csi300` needs the token, `btc-usdt`
is single-asset (cross-sectional factors reject it), and `sp500` fetched its ~500
Yahoo symbols **one HTTP call at a time** — a cold panel took >8 min and never
finished inside the SSE stream timeout.

* `_fetch_equity_panel_concurrent` — fetch a wide equity panel one symbol per worker
  (`ThreadPoolExecutor`, 8) instead of the loader's sequential loop. Reuses the loader's
  own per-symbol cache + normalisation. Cold `sp500` drops to ~1–2 min, then pickle-cached.
  `_load_sp500_panel` now uses it.
* New `nifty50` universe → `equity_in`:
  * `_fetch_nifty50_constituents` — current NIFTY 50 roster + NSE sector labels from
    Wikipedia (requests + UA, then `read_html`), same pattern as `_fetch_sp500_constituents`;
    `_NIFTY50_FALLBACK` static 50-name roster for the offline path.
  * `_load_nifty_panel` — Yahoo `.NS` symbols (no auth), synthetic vwap, sector tag when
    coverage ≥ 90 % so the industry-neutralised alpha101 factors run. Same
    survivorship-bias disclosure in `_meta` as `sp500`.
  * wired through `_UNIVERSE_TAG`, `_BENCH_UNIVERSES` (`alpha_routes`),
    `_UNIVERSE_CHOICES` + `_LIST_UNIVERSE_ALIASES` (`cli_handlers`).
* Frontend: `nifty50` in `UNIVERSE_OPTIONS`, `equity_in -> nifty50` in
  `BENCH_UNIVERSE_FOR_METADATA` (so an India factor's "Run bench" button stops silently
  benching on the S&P 500), `universeOption.nifty50` in all 8 locales.

Verified end-to-end on a live box: `POST /alpha/compare` and `POST /alpha/bench` with
`universe: nifty50` — 50/50 names from Yahoo (~40 s cold), 99/101 alpha101 scored,
sector tag drives the industry-neutral factors.

## Part 2 — bench falls back to sequential when the process pool breaks

`run_bench` parallelises across a `ProcessPoolExecutor` that **forks the current
process**. Driven from the uvicorn worker for `POST /alpha/bench`, the forked children
die on startup (fork of a multi-threaded host) and every future raises
`BrokenProcessPool` — which the per-future handler turned into a *skip*, so the endpoint
returned `n_alphas_tested: 0, n_skipped: N` with `status: ok`.

Latent until now: it needs a universe whose panel actually loads to reach the pool, and
without a Tushare token there was none (Part 1 adds one).

* re-raise `BrokenProcessPool` out of the per-future handler
* wrap the parallel block; on `BrokenProcessPool`, discard partials and run the batch
  through the existing in-process path

Verified: with the pool forced to break, an `alpha101 × nifty50` bench re-runs
in-process — 99 tested / 2 skipped / 22 alive.

---

## Test Plan

- [x] `pytest agent/tests -q -k "alpha or factor or bench or universe or run_card or cli_handler"` → **1458 passed**, 10 skipped
- [x] New tests:
  - `test_alpha_bench_nifty50.py` — constituent provenance, fallback, sector tag
  - `test_alpha_compare_api.py` — parametrised over every bench universe
  - `test_bench_parallel.py::test_broken_process_pool_falls_back_to_sequential`
- [x] Manual end-to-end on a live deployment for both parts (numbers above)
- [ ] Frontend not rebuilt in this PR — Part 1's `AlphaZoo.tsx` + locale edits are trivial
      and type-safe; verified by hand-building on the test box

## Checklist

- [x] No hardcoded secrets / paths / magic numbers
- [x] Follows CONTRIBUTING.md
- [x] DCO `Signed-off-by:` on every commit

### DIFF
--- a/agent/src/api/alpha_routes.py
+++ b/agent/src/api/alpha_routes.py
@@ -113,7 +113,7 @@ _VALID_UNIVERSES = {
 # Ranking metrics for /alpha/compare — keep in sync with
 # ``src.factors.compare_runner.SORT_KEYS`` (kept local to avoid a heavy import).
 _VALID_SORTS = {"ir", "ic_mean", "ic_positive_ratio", "ic_count"}
-_BENCH_UNIVERSES = {"csi300", "sp500", "btc-usdt"}
+_BENCH_UNIVERSES = {"csi300", "sp500", "btc-usdt", "nifty50"}
 
 
 def _now_iso() -> str:
@@ -399,7 +399,10 @@ def register_alpha_routes(
                 detail=f"unknown theme {theme!r}; expected one of {sorted(_VALID_THEMES)}",
             )
         if universe is not None:
-            _ALIAS = {"csi300": "equity_cn", "sp500": "equity_us", "btc-usdt": "crypto"}
+            _ALIAS = {
+                "csi300": "equity_cn", "sp500": "equity_us",
+                "btc-usdt": "crypto", "nifty50": "equity_in",
+            }
             universe = _ALIAS.get(universe, universe)
         if universe is not None and universe not in _VALID_UNIVERSES:
             raise HTTPException(
--- a/agent/src/factors/bench_runner.py
+++ b/agent/src/factors/bench_runner.py
@@ -20,6 +20,7 @@ import math
 import os
 import time
 from concurrent.futures import ProcessPoolExecutor, as_completed
+from concurrent.futures.process import BrokenProcessPool
 from typing import Any, Callable, Iterable
 
 import numpy as np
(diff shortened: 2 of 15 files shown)


vinayanayanegali:
Withdrawing — not ready to upstream yet.

PR #1304, opened 2026-08-31
fix(backtest): delay options signals by one bar and fix HV warm-up lookahead

### LINKED ISSUE #1293: fix(backtest): options signals fill at same-day close, plus HV warm-up lookahead

lorenzozanee:
## Summary

- Delay options signal execution by one bar (T fills at T+1) to match equity engines and remove same-day lookahead; add `same_day_fill` opt-in to preserve legacy behavior.
- Fix HV warm-up: early bars now use the configured `default_iv` instead of backfilling the first computed 30-day window.

## Why

The options engine priced fills from the signal date's close and IV, leaking the day's move into the price, while the HV helper filled early bars with a future rolling window. Equity engines already execute on the next bar.

Fixes #1293

## Changes

- Next-bar execution for options signals and HV warm-up now uses the explicit default.
- Tests updated to T+1 semantics and a small regression added for the one-bar delay.

## Test Plan

## Checklist

### DIFF
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -34,26 +34,32 @@ from src.quantlib.options import bs_greeks, bs_price, normalise_option_type
 # --- Historical volatility ---
 
 
-def historical_volatility(close: pd.Series, window: int = 30) -> pd.Series:
+def historical_volatility(
+    close: pd.Series, window: int = 30, default_iv: float = 0.3
+) -> pd.Series:
     """Calculate annualised historical volatility from a close price series.
 
     Args:
         close: Close price Series.
         window: Rolling window in days.
+        default_iv: Fallback IV for warm-up bars before the rolling window
+            is valid. The previous implementation backfilled the first
+            computed window into the early bars, leaking future information.
 
     Returns:
         Annualised historical volatility Series.
     """
     log_ret = np.log(close / close.shift(1))
     hv = log_ret.rolling(window=window).std() * np.sqrt(252)
-    return hv.fillna(hv.dropna().iloc[0] if len(hv.dropna()) > 0 else 0.3)
+    return hv.fillna(default_iv)
 
 
 # --- IV Smile model (v2) ---
 
 
-def iv_smile_adjustment(S: float, K: float, base_iv: float,
-                        skew: float = -0.15, curvature: float = 0.05) -> float:
+def iv_smile_adjustment(
+    S: float, K: float, base_iv: float, skew: float = -0.15, curvature: float = 0.05
+) -> float:
     """Adjust IV for moneyness using a quadratic smile model.
 
     IV(K) = base_iv + skew * log(K/S) + curvature * log(K/S)^2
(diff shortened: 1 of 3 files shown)


he-yufeng (commented):
Thank you for the quick work on #1293. I am endorsing #1299 for the signal-shift half: it implements the same next-bar fill and `same_day_fill` opt-out with a much tighter diff, and it handles the warm-up boundary explicitly. This PR also folds in the HV warm-up fix, which is separately covered by #1305, so there is scope overlap on both fronts. Closing this in favour of the other two would avoid three copies of the same machinery; the HV part here is good work and is preserved in #1305.

warren618:
Thanks. Both leaks are correctly identified and correctly fixed here.

I went with #1299 + #1305 because between them they cover two things this PR doesn't:

1. **The ordering consequence of the shift.** The day loop runs expiry (`options_portfolio.py:309`) *before* signal execution (`:335`). Once fills move to T+1, a signal dated the day before expiry lands on the expiry bar itself — after that bar's expiry sweep has already run — so the position settles a bar late against the wrong spot, or never settles if it's the last bar. #1299 moves expiry to after signal execution and adds a regression test for it.
2. **`default_iv` validation.** #1305 rejects a non-finite or non-positive value instead of letting it reach Black-Scholes.

Also, `same_day_fill` and `default_iv` are user-facing knobs, so they need to appear in `src/skills/options-strategy/SKILL.md` or the agent can't discover them.

One process note: the signature rewraps (`historical_volatility`, `iv_smile_adjustment`, `OptionPosition.__init__`) are most of the 324-line engine diff. Keeping fixes separate from formatting makes them much easier to review under time pressure.

PR #1305, opened 2026-08-31
fix(backtest): use configured default IV for HV warm-up bars (#1293)

cgycorey:
## Summary

`historical_volatility` (`engines/options_portfolio.py`) filled the orphan bars of the 30-day rolling window with the first computed value — the vol over bars 1..30 — so bars 1..29 were priced with information from bar 30: a small lookahead leak on every options backtest that uses historical vol (#1293, part 2).

## Changes

- **Warm-up bars use the default IV** — `historical_volatility(close, window=30, default_iv=0.3)` returns `hv.fillna(default_iv)` instead of backfilling the first valid window. Bars with a full window keep the real rolling volatility.
- **Config threaded through** — new `options_config.default_iv` (default `0.3`) feeds the warm-up fill and replaces the engine's other hardcoded `0.3` IV fallbacks, so one config governs every "no IV available" case.
- **Tests** — `test_options_hv_warmup.py`: constant closes prove the warm-up is the default IV and not the (zero) first window; configurable default; full-window bars unchanged vs the raw rolling volatility.

> Independent, one-change PR split from #1293 — the fill-shift half is #1299.

## Test Plan

- `pytest agent/tests/test_options_hv_warmup.py agent/tests/test_options_smile_consistency.py agent/tests/test_options_partial_close.py agent/tests/test_warmup_window.py -q` → **32 passed**

## Checklist

- [x] No hardcoded values
- [x] Documentation updated (function docstring)

Part of #1293 (HV warm-up leak); the fill leak is #1299.

### DIFF
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -19,6 +19,7 @@ the per-leg vol every pricing site must agree on.
 """
 
 import json
+import math
 import sys
 from pathlib import Path
 from typing import Any, Dict, List, Optional
@@ -42,9 +43,10 @@ def historical_volatility(
     Args:
         close: Close price Series.
         window: Rolling window in days.
-        default_iv: Volatility used for bars without a full window. Backfilling
-            the first computed window here would price bars before it with
-            information from the window's own end (#1293).
+        default_iv: Volatility used for any bar without a full rolling window
+            (the leading warm-up and NaN gaps). Backfilling the first computed
+            window here would price bars before it with information from the
+            window's own end (#1293).
 
     Returns:
         Annualised historical volatility Series.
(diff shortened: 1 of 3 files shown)


cgycorey:
Adversarial review (self + independent reviewer): **READY after fixes.**

Findings and resolutions:
- default_iv unvalidated (Minor→fixed): a NaN `default_iv` is a no-op `fillna` → NaN legs → `json.dumps(allow_nan=False)` raises mid-run; 0/negative silently prices forward-intrinsic. Now rejected up front (finite, > 0).
- Docstring/scope overstatement (Minor→fixed): `fillna` covers any NaN bar, not only the leading warm-up (no loader currently emits NaN rows, but the docstring and now the PR text say what it does).
- SKILL.md (Minor→fixed): `default_iv` added to the config sample and key list.

Reviewer also verified the math (warm-up is exactly 30 bars for window=30; `iloc[30:]` untouched by `fillna`) and that the `0.3 → default_iv` fallback substitutions are unreachable for any bar with data. 23 tests green on this branch.

he-yufeng (dismissed):
Verified on the head: the options suites pass locally (31 tests). Small, targeted, validated config, and it kills the warm-up lookahead exactly as #1293 described without touching anything else. Endorsing; it pairs with #1299 for the full issue.

he-yufeng (commented):
Same note as on #1299: I am taking #1293 myself with a competing implementation that covers both halves (signal shift and the HV warm-up) in one coherent diff. Your HV approach here (configured default IV with validation) is exactly the right shape and I will credit it in my version; closing in favour of that would be cleanest.

PR #1306, opened 2026-08-31
fix(backtest): hold margin on short option legs and gate opens on buying power

### LINKED ISSUE #1294: fix(backtest): options engine has no margin or buying-power constraint
## Summary

The options engine credits short premium without margin and has no buying-power constraint, so cash can go arbitrarily negative and naked-selling strategies produce the classic free-money curve. Tracked in roadmap #1207 (Phase 2, item 12); validated on current main (`899d3c7`).

## Mechanism

On an open with `qty < 0` the engine credits `opt_price * |qty| * multiplier * (1 - commission)` to cash (`engines/options_portfolio.py`, "3. Execute today's signals") and records the position. There is no margin hold, no buying-power check, and no rejection path anywhere in the file (no `margin` / `buying_power` / insufficient-cash logic exists). A strategy can sell unlimited naked calls or puts, book the premium as cash from bar one, and the equity curve only meets reality at expiry or never.

## Proposed fix

Introduce a margin model for short legs (SPAN-like approximation is fine: max of a fixed notional fraction and premium plus out-of-the-money amount, per leg), hold it against cash, reject opens that would push available buying power below zero, and mark margin daily as the leg is re-priced. Long legs keep premium-paid accounting. Config flag to opt out for research runs that intentionally model unconstrained leverage.

Test plan: naked short beyond buying power rejects with a recorded reason, margin hold releases on close/expiry, equity never books premium as free cash, long-only strategies unchanged.

he-yufeng:
Fixes #1294

## What

The options engine now holds margin on short legs (CBOE style: premium + max(rate x spot - OTM, floor x base), re-marked daily) and checks buying power on every open. Longs need the premium covered; shorts need the new leg's margin covered net of the premium it brings in. Opens that would push either below zero are rejected and recorded with the reason instead of silently booking a position. `options_config.margin_enabled: false` restores the legacy unconstrained behavior for research runs that intentionally model unlimited leverage.

Cash can no longer go arbitrarily negative, so the classic naked-selling free-money curve is gone; equity records carry `margin_hold` daily and metrics report the terminal hold plus a rejected-open count.

## Tests

New `agent/tests/test_options_margin.py`: naked short beyond buying power rejects with reason, short within power posts a daily-marked hold that grows when the underlying moves against it, hold releases on close, long beyond power rejects, and `margin_enabled: false` restores the old path. Full agent suite: 11701 passed.

## Risk / rollback

Backtest-only accounting change inside one engine; no live-trading surface. Strategies that relied on uncollateralized shorts will see rejected opens and lower reported returns, which is the intended correction. Rollback is the config flag above, or reverting this one commit.

Prepared with AI assistance (Kimi K3); mechanism validated against the engine's cash/position accounting and the test suite above.

### DIFF
--- a/agent/backtest/engines/options_portfolio.py
+++ b/agent/backtest/engines/options_portfolio.py
@@ -213,6 +213,11 @@ def run_options_backtest(
     exercise_style = options_cfg.get("exercise_style", "european")  # v2: "european" or "american"
     iv_skew = options_cfg.get("iv_skew", 0.0)         # v2: smile skew param (0 = flat)
     iv_curvature = options_cfg.get("iv_curvature", 0.0)  # v2: smile curvature
+    # Short legs hold margin and every open checks buying power; opt out for
+    # research runs that intentionally model unconstrained leverage.
+    margin_enabled = bool(options_cfg.get("margin_enabled", True))
+    margin_rate = float(options_cfg.get("margin_rate", 0.20))
+    margin_floor_rate = float(options_cfg.get("margin_floor_rate", 0.10))
 
     # Load underlying data
     data_map = loader.fetch(codes, start_date, end_date)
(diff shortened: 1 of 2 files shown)


warren618:
Merged. The margin model is right and #1294 is closed. Two things on top:

**Your two failing tests were fragile, not wrong.** With `commission: 0.0` and `contract_multiplier: 1.0`, the affordability check reduces to `cash + 500p >= 500p + 0.20 × 100 × 500`, i.e. `10_000 >= 10_000` — the premium cancels completely and only float noise in the last bit decides it. That noise moves with days-to-expiry, which the next-bar fill (#1299, merged just before yours) shifted by one day. I dropped the cash to 5,000 so the assertion is about the margin rule instead of a tie, and keyed the `margin_hold` assertion to the first bar with a hold rather than bar 0 — bar 0 is now flat by design.

**`trade_count` counted rejects.** `_calc_options_metrics` returns `"trade_count": len(trades)`, and the reject rows land in `trade_records`, so a refused order booked as executed volume. `win_rate` was safe (rejects carry `pnl == 0` and the aggregation skips zeros), but the count was not. Rejects are now excluded; `options_rejected_opens` already reports them separately.

PR #1308, opened 2026-08-31
fix(backtest): fill options signals on the next bar and de-leak the HV warm-up

### LINKED ISSUE #1293: fix(backtest): options signals fill at same-day close, plus HV warm-up lookahead

he-yufeng:
Fixes #1293

## What

Options signals are now filled on the bar after the decision date instead of the date itself. A signal computed on T's close used to be priced and filled with T's own close and IV, embedding exactly the information it was computed from (about a day of underlying move times delta of phantom edge on every options backtest). The shift is resolved against the full date range, warm-up included, so a signal dated the last warm-up bar still fills on the first evaluated bar and a signal past the last bar simply never fills. `options_config.same_day_fill: true` restores the legacy timing for reproduction runs.

The HV warm-up no longer backfills the first computed 30-day window into the bars before it; those bars now read the configured `options_config.default_iv` (0.3), so the 30th bar's volatility stops leaking into bars 1..29.

## Tests

New `agent/tests/test_options_next_bar_fill.py`: a signal dated T fills on the next trading bar at a price computed from that bar's close, the opt-out restores same-date timing, and the HV warm-up reads the default until the first full window. `test_options_smile_consistency.py` is pinned to the legacy fill because its subject is vol consistency, not timing. Options suite: 270 passed; the two new timing/HV tests are red on current main and green here.

## Risk / rollback

Reported returns of options strategies change (generally down, since the phantom edge disappears), which is the intended correction. No live-trading surface. Rollback: revert this commit or set `same_day_fill: true`.

Prepared with AI assistance (Kimi K3); both defects were validated on current main before the fix, and the new tests are verified red on main and green here.


warren618:
Thanks — this is the cleanest single implementation of the four, and the `bisect_right` remap against the full date range (warm-up included) is exactly right.

I merged #1299 + #1305 because together they cover two gaps:

1. **Expiry ordering.** The loop runs expiry (`:309`) before signal execution (`:335`). After the T→T+1 shift, a signal dated the day before expiry fills on the expiry bar *after* that bar's expiry sweep, so it settles a bar late — or not at all if it's the last bar. #1299 reorders and tests it.
2. **`default_iv` is unvalidated here.** `options_cfg.get("default_iv", 0.3)` will accept `0` or `NaN` straight into Black-Scholes. #1305 raises instead, which is what `contracts.py` asks for.

Separately: this adds `same_day_fill` and `default_iv` as user-facing config, but neither reaches `src/skills/options-strategy/SKILL.md`, so the agent has no way to learn they exist. Worth adding to the checklist for config-surface changes.

Closing as superseded. Your #1307 landed for #1290 and #1309 for #1292.

Issue #17, opened 2026-04-10
Add Docker Compose profile for frontend-only development


warren618:
## Description

Add a Docker Compose profile so frontend contributors can run just the Vite dev server + API backend without needing to install Python locally.

## Current situation

`docker compose up` builds the full image. Frontend contributors who only want to work on React/TypeScript still need a full Python environment.

## What to do

1. Add a `frontend` profile to `docker-compose.yaml` (or create `docker-compose.frontend.yaml`)
2. Frontend container: Node 20 + Vite dev server with hot reload, mounting `frontend/` as volume
3. Backend container: the existing Python image serving the API on port 8899
4. Usage: `docker compose --profile frontend up`

## Difficulty

🟡 **Medium** — Docker Compose configuration, volume mounts, network setup

## Files involved

- `docker-compose.yaml` (modify)
- `frontend/Dockerfile` (new, optional)

Matheus083:
Hi! I'm a CS student and I'd like to work on this issue. Can you assign it to me?

warren618:
Hey @Matheus083, welcome! You're assigned — go for it 🙌

A few pointers to get started:
- Check out the existing `docker-compose.yaml` in the repo root
- The frontend lives in `frontend/` (Vite + React)
- The backend API serves on port 8899

Feel free to open a draft PR early if you want feedback along the way. Happy to help if you hit any blockers!

warren618:
Resolved via #18 — thanks @Matheus083!

PR #1310, opened 2026-08-31
fix(shadow): scope attribution to the pool currency and make buckets exclusive

### LINKED ISSUE #14: Add `vibe-trading init` command for interactive .env bootstrap
## Description

Add a `vibe-trading init` CLI command that guides new users through setting up their `.env` file interactively.

## Expected behavior

```bash
$ vibe-trading init

Welcome to Vibe-Trading setup!

Select your LLM provider:
  1. OpenRouter (recommended - multiple models)
  2. OpenAI
  3. Anthropic
  4. DeepSeek
  5. Ollama (local, free)
> 1

Enter your OpenRouter API key: sk-or-...
Select default model [deepseek/deepseek-chat]: 

(Optional) Enter Tushare token for China A-share data: 

✅ Created agent/.env — you're ready to go!
   Run: vibe-trading
```

## What to do

1. Add a `init` subcommand to `agent/cli.py`
2. Use `click.prompt()` or `rich.prompt` for interactive input
3. Write the selected values to `agent/.env`
4. Validate API key format where possible (e.g., starts with `sk-`)

## Difficulty

🟡 **Medium** — CLI development, input validation, file writing

## Files involved

- `agent/cli.py` — add `init` subcommand
- `agent/.env.example` — reference for all possible variables

he-yufeng:
Fixes #14 and #17 from the trust-layer roadmap (#1207)

## What

Two fixes in the shadow-account attribution math:

**#14 — cross-currency journals.** A mixed HK+US journal used to have every roundtrip summed into `real_pnl` and compared against a single-currency shadow pool, so HKD amounts leaked into a USD delta. Attribution is now scoped to the headline pool's currency: other-currency roundtrips are excluded from the comparison and reported in `AttributionBreakdown.excluded_currencies` (currency to count) instead of being silently summed. Single-currency journals are unchanged.

**#17 — double-counted buckets.** The early/late conditions were subsets of not-within-rule, so one trade landed in both `noise` and `early`/`late` and `explained` summed it twice. The buckets are now mutually exclusive: a too-short winner is early-exit, a too-long loser is late-exit, and only the remaining out-of-range trades count as rule-violation noise.

## Tests

New `agent/tests/test_shadow_attribution.py`: a mixed journal sums only same-currency PnL and records the excluded HKD trades, the no-pool-currency path keeps the legacy sum, and each bucket case (short winner, long loser, short loser) lands in exactly one bucket. The exclusion and exclusivity tests are red on current main and green here; the existing shadow suite passes (89 tests).

## Risk / rollback

Delta attribution numbers change for mixed-currency journals (now meaningful instead of summed across currencies) and for trades that used to double count (explained totals shrink to the correct sum). No live-trading surface. Rollback: revert this commit.

Prepared with AI assistance (Kimi K3); both defects were validated on current main before the fix.

### DIFF
--- a/agent/src/shadow_account/backtester.py
+++ b/agent/src/shadow_account/backtester.py
@@ -271,6 +271,7 @@ def run_shadow_backtest(
         journal_path=journal_path,
         combined=combined,
         initial_capital=initial_capital,
+        pool_currency=headline_currency,
     )
 
     result = ShadowBacktestResult(
@@ -494,6 +495,7 @@ def _attribution_or_zero(
     journal_path: str | Path | None,
     combined: dict[str, float],
     initial_capital: float,
+    pool_currency: str | None = None,
 ) -> tuple[AttributionBreakdown, float | None, float]:
     """Compute attribution if the journal is available, else return zeros."""
     shadow_pnl = _shadow_pnl_from_metrics(combined, initial_capital)
@@ -516,7 +518,12 @@ def _attribution_or_zero(
     if not roundtrips:
         return _zero_attribution(), shadow_pnl, 0.0
 
-    return _compute_attribution(profile=profile, roundtrips=roundtrips, shadow_pnl=shadow_pnl)
+    return _compute_attribution(
+        profile=profile,
+        roundtrips=roundtrips,
+        shadow_pnl=shadow_pnl,
+        pool_currency=pool_currency,
+    )
 
 
 def _zero_attribution() -> AttributionBreakdown:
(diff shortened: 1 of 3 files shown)


warren618:
Merged. Both halves are correct — the pool is single-currency and there is no FX translation layer, so summing across currencies was meaningless, and the noise/early/late buckets really were double-counting into `explained`.

One thing added on top: `excluded_currencies` was computed and stored but never rendered. The report kept saying "your real PnL X" while quietly dropping roundtrips, which is a subtler kind of wrong than the number it replaced — a user with an HK+US journal would just see a smaller figure with no reason given. Surfaced next to the attribution table, in the same spirit as the survivorship-bias disclosure, with tests both ways (present when something was excluded, absent when nothing was).

Issue #16, opened 2026-04-10
Add benchmark comparison to CLI backtest output


warren618:
## Description

When running a backtest via CLI, automatically show the benchmark (e.g., SPY, CSI 300) performance alongside the strategy for easy comparison.

## Current behavior

CLI shows strategy metrics only. User must manually compare with benchmark.

## Expected behavior

```
Strategy vs Benchmark (SPY)
───────────────────────────────────
              Strategy    SPY
Total Return   +23.4%   +15.2%
Sharpe Ratio     1.42     0.98
Max Drawdown   -12.3%   -18.7%
───────────────────────────────────
Alpha: +8.2%  |  Beta: 0.74
```

## What to do

1. Auto-detect appropriate benchmark based on market (US → SPY, China A → 000300, Crypto → BTC-USDT)
2. Fetch benchmark data alongside strategy data
3. Calculate alpha/beta vs benchmark
4. Display side-by-side comparison in CLI output

## Difficulty

🟡 **Medium** — involves data fetching, metrics calculation, and CLI formatting

## Files involved

- `agent/backtest/` — metrics and runner
- `agent/cli.py` — output formatting

warren618:
Moved to the [Roadmap](https://github.com/HKUDS/Vibe-Trading#-roadmap) in README. Will revisit when we reach that milestone.

PR #1311, opened 2026-08-31
fix(shadow): model short lots in FIFO pairing and restate legs across corporate actions

### LINKED ISSUE #15: Create a new skill: Dividend Analysis
## Description

Add a new skill for dividend analysis — useful for income-focused investors.

## What to do

Create `agent/src/skills/dividend-analysis/SKILL.md` following the existing skill format.

### Skill should cover

- Dividend yield, payout ratio, dividend growth rate
- Dividend coverage (earnings / dividends)
- Ex-dividend date tracking
- Dividend sustainability assessment
- Sector comparison of dividend metrics
- Template for a dividend screen strategy (e.g., "high yield + low payout ratio + 5yr growth")

### Reference

Look at existing skills for the markdown structure:
- `agent/src/skills/fundamental-filter/SKILL.md`
- `agent/src/skills/earnings-forecast/SKILL.md`

## Difficulty

🟢 **Easy** — single markdown file, no code changes. Requires basic knowledge of dividend investing concepts.

## Files involved

- `agent/src/skills/dividend-analysis/SKILL.md` (new)

he-yufeng:
Fixes #15 and #16 from the trust-layer roadmap (#1207)

## What

**#16 — short sales are roundtrips now.** The FIFO queue only modelled buy-first: a sell-short vanished silently and the later cover queued as a phantom long lot that poisoned subsequent matches. Both lot types are now modelled: a sell with no open long opens a short lot, a buy first covers open shorts (PnL = entry - cover) before queueing as a long. Each roundtrip carries a `side` field.

**#15 — legs are restated to one price caliber.** A split between buy and sell used to fabricate a large fake loss and drop the leftover shares; cash dividends never entered fills. `pair_trades_fifo` accepts an optional caliber callback and restates each matched leg (price and quantity) to the sell date's caliber: a 1:2 split turns a pre-split lot of 10@100 into 20@50, so the post-split sell matches the whole position at zero fake loss. The shadow path builds the callback from the run's own adjusted OHLCV artifacts (`build_frame_adjust`), and uncovered symbols stay raw. Standalone journal use passes no callback and keeps the legacy raw behavior byte for byte.

## Tests

New `agent/tests/test_trade_journal_fifo.py`: short open plus cover is a signed roundtrip, long and short books stay separate across a mixed sequence, the split case shows the raw fake loss then zero after restatement, and an uncovered symbol stays raw. The existing shadow and journal suites pass (89 tests).

## Risk / rollback

Attribution PnL changes for journals with splits, dividends, or short trades, in the direction of reality. No live-trading surface. Rollback: revert this commit.

Prepared with AI assistance (Kimi K3); both defects were validated on current main before the fix.

### DIFF
--- a/agent/src/shadow_account/backtester.py
+++ b/agent/src/shadow_account/backtester.py
@@ -33,7 +33,7 @@ from src.shadow_account.models import (
 )
 from src.shadow_account.storage import runs_dir
 from src.tools.trade_journal_parsers import parse_file, records_to_dataframe
-from src.tools.trade_journal_tool import pair_trades_fifo
+from src.tools.trade_journal_tool import build_frame_adjust, pair_trades_fifo
 
 logger = logging.getLogger(__name__)
 
@@ -266,11 +266,22 @@ def run_shadow_backtest(
     if headline_points:
         equity_curves["combined"] = headline_points
 
+    # Restate journal legs to one price caliber where the run's own adjusted
+    # frames cover the symbol; uncovered symbols stay raw (see #15).
+    adjust_frames: dict[str, pd.DataFrame] = {}
+    for currency, _, _, _ in group_results:
+        artifacts_dir = base_dir / currency / "artifacts"
+        if artifacts_dir.is_dir():
+            for ohlcv_csv in artifacts_dir.glob("ohlcv_*.csv"):
+                symbol = ohlcv_csv.stem.removeprefix("ohlcv_")
+                frame = pd.read_csv(ohlcv_csv, index_col=0, parse_dates=True)
+                adjust_frames.setdefault(symbol, frame)
     attribution, shadow_pnl, real_pnl = _attribution_or_zero(
         profile=profile,
         journal_path=journal_path,
         combined=combined,
         initial_capital=initial_capital,
+        adjust=build_frame_adjust(adjust_frames) if adjust_frames else None,
     )
 
     result = ShadowBacktestResult(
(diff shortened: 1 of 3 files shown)


warren618:
Merged, with one correction pushed on top — the short branch had the caliber factor inverted.

`factor` is `close(to) / close(from)`, so it turns an entry-date price *into* the cover date's caliber: the price scales **by** it and the share count by its inverse. The long branch does exactly that (`lot_price_here = lot["price"] * factor`, `lot_qty_here = lot["qty"] / factor`). The short branch used `lot["price"] * (1.0 / factor)` and never restated the quantity at all.

Measured on a 1:2 split, short 10 @100 → cover 20 @50, which is economically flat:

```
LONG  rt pnl = 0.0     (correct)
SHORT rt pnl = 1500.0  (should be 0.0)
```

`100 / 0.5 = 200`, then `(200 − 50) × 10 = 1500`, and the cover consumed only 10 of the 20 restated shares. Now `entry_price = lot["price"] * factor`, `lot_qty_here = lot["qty"] / factor`, fees allocated in original-caliber shares via `consumed = take * factor` — symmetric with the long branch. Short PnL is 0.0 at qty 20.0.

The reason this got through: your four tests are `short/cover` (no adjust), `long+short separation` (no adjust), `split between legs` (long only), and `uncovered symbol stays raw`. Nothing crossed short **with** a corporate action, which is the only cell where the inversion shows. I added `test_split_between_legs_of_a_short_mirrors_the_long_case`, which asserts the short and long results are equal, plus `test_short_without_adjust_is_unchanged` for the legacy path.

The rest — short-lot FIFO, `build_frame_adjust`, the artifact wiring — is good, and I verified the premise: `base.py:2020` really does write `artifacts/ohlcv_{code}.csv`, and the shadow run's `base_dir/<currency>` is that run_dir, so the restatement is live, not dead code.

PR #1312, opened 2026-08-31
fix(live): size buy-limit orders at the worse of quote and limit

he-yufeng:
Fixes #18 from the trust-layer roadmap (#1207)

## What

A buy-limit order's notional is now sized at the worse of the live quote and the limit price, instead of the quote alone. Previously a buy limit at 2x the market passed a notional cap sized for the quote while being fillable at twice the authorized amount. The limit price now rides the `OrderIntent` from the service entry point through the gate, and the sizing rule applies only on the buy side: a sell limit creates no new exposure, so the quote stands there. The order_guard normalize path preserves the field instead of dropping it.

## Tests

New `agent/tests/test_limit_price_notional.py`: a buy limit above the quote is priced at the limit, a buy limit below the quote keeps the quote, a market order ignores the path, and a sell limit keeps the quote. The cap case is red on current main and green here; the existing gate/mandate/kill-switch suites pass (95 tests).

## Risk / rollback

Buy-limit orders that used to sneak past `max_order_notional_usd` are now correctly denied, which is the intended hardening of the mandate gate. No broker API or order-shape changes. Rollback: revert this commit.

Prepared with AI assistance (Kimi K3); the defect was validated on current main before the fix.

### DIFF
--- a/agent/src/live/enforcement.py
+++ b/agent/src/live/enforcement.py
@@ -132,6 +132,9 @@ class OrderIntent:
     quantity: float | None
     instrument_type: InstrumentType
     asset_class: AssetClass | None = None
+    # Buy-limit worst-case fill price for notional math (#18); None on market
+    # orders and on connectors that do not pass one through.
+    limit_price: float | None = None
 
 
 @dataclass(frozen=True)
--- a/agent/src/live/order_guard.py
+++ b/agent/src/live/order_guard.py
@@ -258,6 +258,7 @@ class LiveOrderGuardTool(MCPRemoteTool):
             notional_usd=enforced,
             quantity=intent.quantity,
             instrument_type=intent.instrument_type,
+            limit_price=intent.limit_price,
         )
 
     def _quote_price(self, intent: OrderIntent) -> float | None:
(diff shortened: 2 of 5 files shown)


warren618:
Merged — sizing a buy limit at `max(quote, limit)` is the right conservative call.

While reviewing the rebuild you touched, I found `_normalize_intent_notional` was already dropping `asset_class`. `check_mandate` prefers an explicit `intent.asset_class` over the instrument-type default (`enforcement.py:524`) precisely so a multi-market connector's HK/A-share order buckets correctly — so an explicit HK class silently became `us_equity` and would pass a mandate permitting only `us_equity`. Latent today, since the Robinhood MCP extractor never sets the field, but the rebuild is now field-complete and tested both ways.

One gap left open, which I'll track separately: the `max(quote, limit)` logic lives only in `sdk_order_gate._implied_notional`. `order_guard.py:246-249` still computes `quantity * quote` with no limit term, so the MCP transport keeps the exposure gap the red line says both transports must close. Your PR passes `limit_price` through `order_guard`'s rebuild but nothing reads it there. It's inert right now because the extractor doesn't parse a limit price at all — whether Robinhood's `place_equity_order` accepts one is something I can't verify without OAuth against their server.

Issue #1315, opened 2026-08-31
bug(market-data): stooq serves a JS proof-of-work challenge, silently dead at us_equity chain position 2


he-yufeng:
## What

stooq currently serves a JavaScript proof-of-work challenge to non-browser clients. The loader's plain HTTP GET gets the challenge page instead of CSV, treats it as "no data", and the us_equity chain silently falls through to position 3. Measured by @warren618 on 2026-08-31 (chain table in #1301): position 2 is dead weight today.

The silent part is the problem. An operator reordering the chain via VIBE_TRADING_MARKET_DATA_ORDER_* has no signal that this source never serves.

## Options

1. Detect the challenge page (it is HTML, not CSV, and carries known markers) and log a clear "stooq is serving an anti-bot challenge" warning instead of a generic no-data skip, so the failure is visible in run logs.
2. Try to satisfy the challenge (headers, cookies). Fragile by design; the challenge exists to keep non-browser clients out.
3. Demote or drop stooq from the default chain until it recovers.

My lean is 1 now, keep 3 on the table if the challenge persists. Detection also gives #1301's caliber stamp something honest to say: a source that never serves never stamps.

PR #1317, opened 2026-08-31
feat(market-data): stamp the price-adjustment caliber on served frames (#1301)

### LINKED ISSUE #1301: bug(market-data): the us_equity chain still mixes adjusted and unadjusted price calibers
## What

After [#1287](https://github.com/HKUDS/Vibe-Trading/pull/1287) the `us_equity` fallback chain still mixes price calibers: which source serves a symbol changes what its historical prices mean, and nothing in the response says which caliber you got.

Measured 2026-08-31, AAPL close for 2020-01-02 (raw close 300.35; the 4:1 split was 2020-08-31), fetched through each source as the chain would:

| chain pos | source | AAPL 2020-01-02 close | caliber |
|---|---|---|---|
| 1 | yahoo | 75.0875 → **72.27** after #1287 | split + dividend adjusted |
| 2 | stooq | no data | serving a JS proof-of-work challenge to non-browser clients |
| 3 | sina | **300.35** | **fully unadjusted** |
| 4 | eastmoney | adjusted | splits + dividends |
| 5 | yfinance | adjusted after #1287 | `auto_adjust=True` |
| 10 | longbridge | unadjusted | pins `AdjustType.NoAdjust` (`longbridge.py:353`) |

Two sources on the chain still return raw prices. A backtest that falls through to sina or longbridge books every split as a crash and every ex-dividend gap as a loss, and the run is indistinguishable from one served by yahoo.

Note for anyone picking this up: #1287's description says Yahoo served raw OHLC and that after it "there is no longer a raw source to flag". Neither is accurate — Yahoo's `quote` series was always split-adjusted, and sina/longbridge remain raw. The corrected reading is in the comment on that PR.

## Why it is not just a documentation gap

`VIBE_TRADING_MARKET_DATA_ORDER_*` (shipped in #1231) lets an operator reorder the chain, so the source that serves a symbol is now user-configurable — which makes silently varying calibers easier to hit, not harder.

## Proposal

Stamp the adjustment caliber as provenance on every returned frame — `adjustment: "raw" | "split" | "split_dividend"` alongside the existing `source` field — and surface it wherever provenance already surfaces. That is enough to let a backtest refuse, or at least warn about, a mixed-caliber basket. Normalizing the raw sources is a separate and larger question; labeling comes first.

he-yufeng:
## Summary

- Every served frame now carries an `adjustment` stamp in `_provenance`: `raw` / `split` / `split_dividend` / `na` / `unknown`, resolved per (source, market) from one verified table.
- A backtest whose basket mixes comparable calibers logs a warning naming the symbols and calibers, and the warning lands in the run card.

## Why

Closes #1301. After #1287 the us_equity chain still mixes calibers (sina and longbridge serve raw next to adjusted sources), and the MARKET_DATA_ORDER_* overrides make which source serves a symbol operator-configurable, so a mixed basket is a configuration accident away. Labeling comes first; refusing mixed runs stays available as a later policy flag. Normalizing the raw sources is out of scope here.

## Changes

- `backtest/loaders/registry.py`: the caliber table plus `price_caliber()` and `mixed_caliber_warning()`. Only measured or endpoint-pinned sources get a real caliber; the rest stamp `unknown` rather than guess, since origin-side adjustment is invisible from loader code (yahoo serves split-adjusted quotes with zero adjustment logic in this repo). The table is per (source, market) because tushare applies adj_factor on A-share but publishes no HK factor series.
- `src/market_data.py`: `_provenance` entries gain `adjustment`, stamped from the source that actually served the symbol, so fallback batches are labeled by who really served.
- `backtest/runner.py`: both fetch paths track per-symbol calibers; a mixed run warns once, and the warning reaches the run card through `_run_card_caliber_warning`.
- Tool and MCP docstrings now tell agents to read the stamp before comparing price levels across symbols.

stooq stamps `unknown` while it is down (#1315): its CSV convention cannot be measured until someone can fetch a split-day bar from it.

## Test Plan

- [x] Existing tests pass: full suite 11722 passed, 92 skipped (`pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py -q`)
- [x] New tests added: 26 in `agent/tests/test_price_caliber.py` covering table values (including the deliberate `unknown` entries), the tushare HK override, na markets, warning fire/silent/ignored cases, provenance stamps, and mixed vs single-caliber `fetch_data_map` runs
- [x] One existing assertion updated: `test_market_data_json_can_include_actual_source_provenance` pins the provenance dict exactly and now expects the new field
- [x] Verified the new tests fail without the patch (collection error on the missing registry symbols) and pass with it

## Checklist

- [x] Documentation updated (tool description and MCP docstring)

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -1079,6 +1079,9 @@ class BaseEngine(ABC):
         )
 
         # 9. Trust Layer run card
+        card_warnings = list(config.get("content_filter_warnings") or [])
+        if config.get("_run_card_caliber_warning"):
+            card_warnings.append(config["_run_card_caliber_warning"])
         from backtest.run_card import write_run_card
         write_run_card(
             run_dir,
@@ -1086,7 +1089,7 @@ class BaseEngine(ABC):
             m,
             data_sources=_run_card_data_sources(config, loader),
             strategy_path=run_dir / "code" / "signal_engine.py",
-            warnings=config.get("content_filter_warnings") or None,
+            warnings=card_warnings or None,
         )
 
         # Print scalar metrics (skip nested dicts for JSON compat).
--- a/agent/backtest/loaders/registry.py
+++ b/agent/backtest/loaders/registry.py
@@ -168,6 +168,90 @@ FALLBACK_CHAINS: dict[str, list[str]] = {
 
 
 # ---------------------------------------------------------------------------
+# Price caliber: what the served prices actually mean, per source (#1301)
+# ---------------------------------------------------------------------------
+
+#: A value lands here only when it is measured against live payloads (the
+#: #1301 chain table) or pinned by the loader's own endpoint/parameter
+#: choice. Anything unverified resolves to "unknown" on purpose: origin-side
+#: adjustment is invisible from loader code (yahoo serves split-adjusted
+#: quotes with zero adjustment logic in this repo), so an unmeasured source
+#: must say "unknown" rather than a guessed caliber.
+PRICE_CALIBER_BY_SOURCE: dict[str, str] = {
+    # Split- and dividend-adjusted.
+    "yahoo": "split_dividend",  # quote series split-adjusted at origin, scaled to adjclose
+    "yfinance": "split_dividend",  # auto_adjust=True
+    "eastmoney": "split_dividend",  # fqt=1 (forward-adjusted) on every kline call
+    "tencent": "split_dividend",  # fqkline qfq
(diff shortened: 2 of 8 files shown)


he-yufeng:
The two red test lanes are the fastmcp 4.0.0 breakage, not this diff: MCPError constructor, session snapshot, init-timeout and auth-argument failures identical to the ones #1313 caps, and they reproduce on main with any fresh install. Once #1313 lands this just needs a rerun; the caliber suite itself (11722 tests locally, including the 26 new ones) is green.

PR #1320, opened 2026-09-01
fix(loaders): serve adjusted prices on FMP and Tiingo paths

### LINKED ISSUE #1296: [Bug] Backtests silently use dividend-unadjusted prices: yfinance loader hardcodes auto_adjust=False, FMP loader uses unadjusted endpoint
## Description

**What happened:**
`agent/backtest/loaders/yfinance_loader.py` calls `yf.download(...)` with `auto_adjust=False` hardcoded and no way to override it from `config.json` (the loader never receives config). The FMP loader (`agent/backtest/loaders/fmp_loader.py`) only uses the unadjusted `historical-price-full` endpoint, and `tiingo_loader.py` explicitly drops Tiingo's adjusted columns. As a result every backtest is computed on dividend-unadjusted prices.

**What I expected:**
Backtests of dividend-paying instruments (bond/dividend ETFs, utilities, sectors) to use total-return prices, or at least an opt-in. Today the error is silent: the backtest completes and looks plausible, but dividends are missing from both the signal inputs (momentum scores) and the P&L.

## Impact measurement (replication of a published strategy)

I re-implemented a published monthly tactical ETF strategy (sector rotation, offensive months in sector ETFs, defensive months in AGG/HYG/IEF/BIL) that the source reports as CAGR 16.8%, Sharpe 1.14, max drawdown -14.0% (2013-02 to 2026-07):

- With VT's default yfinance loader (unadjusted): **CAGR 11.9%, Sharpe 0.90, MDD -16.7%**
- Same code, same engine, but with yfinance `auto_adjust=True` data fed through the `local` data-bridge loader: **CAGR 17.1%, Sharpe 1.28, MDD -14.9%** — i.e. within noise of the source.

The entire gap is dividend treatment. Bond ETFs make it obvious: HYG appears to fall 89→79 over 14 years unadjusted while its total return is 40→79 (Jan 2012→Jul 2026, yfinance). Defensive months in dividend payers therefore record ~0 return, and momentum scores computed on unadjusted closes mis-rank dividend payers.

## Steps to Reproduce

1. `vibe-trading run` a monthly rebalance strategy holding AGG or HYG through any multi-year window, source `yfinance`.
2. Compare `artifacts/ohlcv_AGG.US.csv` close column against yfinance `Adj Close` for the same dates — the CSV contains raw `Close`.
3. Or grep: `auto_adjust=False` is hardcoded in `yfinance_loader.py` and is unreachable from `config.json`.

## Error Logs

(none — the defect is silent)

## Suggested fix (optional)

- Add an `adjusted`/`auto_adjust` boolean to the backtest config surface and thread it to the yfinance loader; consider `True` as the default for equity ETFs (price-only crypto/forex can stay unadjusted).
- FMP loader: use FMP's adjusted historical-price endpoint (`historical-price-full/adjusted`), or accept a config choice.
- If a default change is too disruptive, a documented `config.json` key alone fixes the silent-correctness problem.

Happy to open a PR if you can point me at the preferred config plumbing.

lorenzozanee:
## Summary

Fixes #1296. FMP and Tiingo returned raw OHLC so dividends booked as losses; this brings them to qfq like Yahoo/yfinance after #1287.

## Why

Tiingo ignored adjOpen/adjClose and FMP used the retired v3 endpoint without adjClose, so the fallback chain could mix adjusted and raw calibers across runs.

## Changes

- Tiingo: prefer adjOpen/adjHigh/adjLow/adjClose when present, otherwise scale raw OHLC by adjClose/close (0.01-100 guard, volume stays raw).
- FMP: switch to stable/historical-price-eod/full?symbol=, accept both Stable array and legacy dict shapes, scale OHLC by adjClose/close.
- Tests: update Stable URL/param expectations.

## Test Plan

67 loader tests pass (FMP/Tiingo/Yahoo/yfinance); manual mock with adjClose=50 vs close=100 emits scaled OHLC and keeps volume raw; black --check and ruff check pass.

## Checklist

### DIFF
--- a/agent/backtest/loaders/fmp_loader.py
+++ b/agent/backtest/loaders/fmp_loader.py
@@ -6,12 +6,13 @@ routes through :mod:`backtest.loaders._http` so calls share one process-wide
 minimum-spacing gate and a reused session.
 
 API format (public, documented):
-  https://financialmodelingprep.com/api/v3/historical-price-full/{SYMBOL}
-    ?from=YYYY-MM-DD&to=YYYY-MM-DD&apikey=KEY
+  https://financialmodelingprep.com/stable/historical-price-eod/full
+    ?symbol=SYMBOL&from=YYYY-MM-DD&to=YYYY-MM-DD&apikey=KEY
 
-The JSON body is ``{"symbol": "AAPL", "historical": [{date, open, high, low,
-close, volume}, ...]}`` in descending date order; an unknown symbol or empty
-window yields an empty/absent ``historical`` array.
+The JSON body is a top-level array ``[{date, open, high, low, close,
+adjClose, volume}, ...]`` in descending date order (legacy
+``{"symbol": "AAPL", "historical": [...]}`` shape is still accepted for
+compatibility); an unknown symbol or empty window yields an empty array.
 
 Auth: set ``FMP_API_KEY`` in the environment. Covers US equities only.
 """
@@ -30,7 +31,7 @@ from backtest.loaders.registry import register
 logger = logging.getLogger(__name__)
 
 _API_KEY_ENV = "FMP_API_KEY"
-_BASE_URL = "https://financialmodelingprep.com/api/v3/historical-price-full"
+_BASE_URL = "https://financialmodelingprep.com/stable/historical-price-eod/full"
 
 # Shared throttle/session bucket for every FMP request in this process.
 _HOST_KEY = "fmp"
@@ -145,7 +146,10 @@ class DataLoader:
         return result
 
     def _fetch_one(
-        self, code: str, start_date: str, end_date: str,
+        self,
+        code: str,
+        start_date: str,
+        end_date: str,
     ) -> Optional[pd.DataFrame]:
         """Fetch and parse one symbol's daily bars; ``None`` on no data.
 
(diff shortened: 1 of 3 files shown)


he-yufeng (commented):
Read the full diff and ran both parsers against synthetic payloads on this branch. The semantics are right: adjusted OHLC preferred when Tiingo provides all four, the adjClose/close ratio scaling with the 0.01..100 guard, volume kept raw per qfq convention, FMP legacy dict shape still accepted, and every degenerate case (zero close, out-of-guard ratio, missing adjClose) falls back to raw. All 13 invariant cases I probed pass, and the repo's own `test_fmp_loader.py` is 20/20 green on the branch.

Two things before this can land:

1. The new behavior has no direct unit coverage. The only test edits here are the endpoint URL/params assertions in `test_fmp_loader.py`. Please add a focused parse-level test for each loader: `_parse_historical` with a Stable array carrying `adjClose != close` (scaled OHLC, raw volume) plus the legacy dict shape, and `_rows_to_frame` with full `adj*` fields and with adjClose-only (ratio path). Both are cheap synthetic payloads, no network.

2. The two red test lanes are the pre-existing fastmcp 4.0 breakage on main (fixed by #1313), unrelated to this diff. Rebasing onto current main picks up the cap and should turn CI green.

Also double-checking my own blind spot: the FMP switch to `stable/historical-price-eod/full` drops the v3 path entirely for new keys, which is the right call since v3 historical-price-full is being retired. If anyone still on a legacy grandfathered key reports a regression we can revisit, but the code keeps accepting the legacy response shape, so the parse side is compatible either way.

PR #1338, opened 2026-09-02
fix(grounding): gate analysis metrics on a completed analysis result (#1336)

### LINKED ISSUE #1336: [Bug] Block unsupported quantitative claims after analysis tool failures
### Description

**What happened:** After market-data requests were deduplicated and no backtest completed, the agent still presented specific historical-return claims, volatility and maximum-drawdown figures, target probabilities, and a categorical statement about all historical 12-month windows as measured facts.

**What I expected:** Quantitative claims attributed to a backtest or market-data analysis should require a successful supporting tool result. When the tool fails or returns no evidence, the response should explicitly mark the analysis incomplete and omit unsupported figures. A later audit prompt correctly caused the agent to retract these claims.

### Steps to Reproduce

1. Start the Web UI.
2. Ask the agent to compare several instruments, backtest them across market regimes, and report volatility, drawdown, and target probabilities.
3. Trigger or encounter failed/deduplicated market-data calls so no backtest result is produced.
4. Observe whether the final answer nevertheless provides precise backtest-style metrics and probabilities.

### Error Logs

```shell
The dedup layer is collapsing my market_data calls.
...
Strategy ... Return vol / MaxDD (backtest regimes) ... Prob. of hitting target
```

### Interface

Web UI (vibe-trading serve)

### LLM Provider

OpenAI-compatible provider

### Version

0.1.10

### Environment

Windows / Docker / Python 3.11

cgycorey:
## What

Fixes #1336. After failed or deduplicated market-data calls the agent still presented backtest/analysis metrics (return volatility, max drawdown, target probabilities) as measured facts. Those figures can only originate in an analysis tool, so `validate_final_answer` now rejects measurement-shaped analysis figures when the run has no supporting tool result.

## How

`agent/src/agent/grounding.py`:

- Records completed analysis tools (`backtest`, `factor_analysis`, `run_shadow_backtest`, `quantlib_call`) in `_analysis_completed` on ingest. Dedup is handled by the ingest loop: skipped/deduplicated calls never count.
- `_validate_analysis_claims` gate: a measurement-shaped figure (decimal or percent, with an analysis-context marker — return vol, max drawdown, sharpe, probability-of, volatility, drawdown, annualized, windows, regimes, or Chinese equivalents) is rejected with `analysis_claim_unavailable` unless the same *kind* of figure is in the run's evidence. Two support paths, avoiding a hand-maintained tool list:
  1. kind-scoped match against recorded analysis metrics — parsed from a completed backtest's run-dir `metrics.csv`/`metrics.json`, or nested numeric leaves from `factor_analysis`/`run_shadow_backtest`/`quantlib_call` whose key names a metric kind (token-split resolution, so `reported_annualized_return`, `strategy_max_drawdown`, `benchmark_return_vol` all resolve without a table);
  2. value-level match against `_evidence` — any successful tool's numeric output grounds its own kind. Percent/fraction scaling is normalized (`0.182` ↔ `18.2%`); drawdown magnitude is compared because sign conventions differ.
- The completed-analysis path is *not* a blanket bypass: a kind-scoped value check runs against the recorded metrics, so inventing a sibling metric after a real backtest is still refused (`bd24201`).
- A return figure may be arithmetic on sourced inputs rather than an invented backtest metric: `_is_explicit_derivation` (formula-shaped lines) and a from/to-frame check (growth between two observed endpoints, ±0.5%) are exempted.
- Externally attributed figures ("The paper reports a Sharpe of 1.8") are citations, not invented measurements: `_ATTRIBUTION_RE` keys on who the source is (paper/study/analysts/filings/…; Chinese 论文/文献/研究/分析师/… + reporting verb), not on citation shape — so `data`/`回测`/`strategy` subjects (how a model dresses up its own numbers) still fail. Clause-scoped, so a citation in one clause can't launder an invented sibling in the next.
- Forward-looking forecasts, definitional prose ("夏普比率大于 1.0 通常被认为较好"), and categorical historical-window facts without a window-producing analysis are handled separately.

## Test Plan

- [x] Regression table with an **accept column and a reject column**: 
  - reject: failed backtest → reject; no analysis → reject; partial support → reject (one observed metric can't launder an invented sibling); SKIPPED analysis → reject; invented sibling after a real backtest → reject; bare `收益率为 8.6%` → reject; unanchored return → reject; wrong-arithmetic return → reject; `数据/回测/strategy` subjects → reject; TSLA/特斯拉 unsourced pair → reject; cross-clause laundering → reject.
  - accept: completed backtest (values present) → allow; real-fraction `portfolio_risk_xray` + percent answer → allow; derived returns EN + ZH → allow; paper/analyst/filing restatements EN + ZH → allow; forecast → allow; refusal prose → allow; definitional prose → allow.
- [x] `python -m pytest tests/test_agent_grounding.py tests/test_agent_output_discipline.py tests/test_grounding_language_parity.py -q` → 276 passed
- [x] Full suite: `python -m pytest --ignore=tests/e2e_backtest -q` → 12192 passed (branch); CI green across all 10 jobs.

## Known limitations

- A figure's *exact* value is never checked against a paper's real text — attribution exempts the number, it does not verify the quote. (The gate's job is to distinguish citation from invention, not to fact-check the cited source.)
- The `_ATTRIBUTION_RE` subject list is deliberately English/Chinese; a citation phrased via a subject outside that list (e.g. "the vendor's whitepaper states") is still treated as a measurement claim. Extend the list if a locale or phrasing shows up.

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -43,12 +43,17 @@ GROUNDING_ARTIFACT = "grounding_evidence.json"
 
 _RESOLVER_TOOL = "search_symbol"
 
-# Tools whose successful completion grounds backtest/analysis metric claims
-# (#1336). A deduplicated ("skipped") call also counts: the loop only skips a
-# call that already completed successfully, and its result lives in the run.
+# Tools whose successful completion can ground backtest/analysis metric claims
+# (#1336). Success alone is not authority: only results that actually carried
+# metric output (parsed from the result or its run-dir artifacts) raise
+# ``analysis_claim_unavailable``. A deduplicated ("skipped") call is never a
+# completion.
 _ANALYSIS_TOOLS = frozenset(
     {"backtest", "factor_analysis", "run_shadow_backtest", "quantlib_call"}
 )
+# Tools whose run produces the equity/regime windows behind a categorical
+# "all N-month windows were profitable" statement.
+_ANALYSIS_WINDOW_TOOLS = frozenset({"backtest", "run_shadow_backtest"})
 _PRIVATE_COMPANY_SKILL_NAMES = {
     "private-company",
     "private-company-analysis",
@@ -243,7 +248,8 @@ _ANALYSIS_METRIC_RE = re.compile(
     r"\bsharpe(?: ratio)?\b|\bwin rate\b|\bhit rate\b|"
     r"\bprob(?:ability)?\.?\s+of\b|\bvolatility\b|\bdrawdown\b|"
     r"\bannualiz\w*\b|\bwindow(?:s)?\b|\bregime(?:s)?\b|"
-    r"夏普|回撤|波动率|胜率|命中率|概率|年化|回测|窗口)",
+    r"\b(?:annual|cumulative|total)\s+return\b|"
+    r"夏普|回撤|波动率|胜率|命中率|概率|年化|回测|窗口|收益(?:率)?|回报(?:率)?)",
     re.IGNORECASE,
 )
 # Measurement-shaped numbers for analysis claims: keeps the % sign and sign
@@ -261,6 +267,75 @@ _MEASURE_NUMBER_RE = re.compile(
 # facts" (#1336), forecasts stay under the "analysis, not advice" prompt rule.
 _FORECAST_FRAME_RE = re.compile(
     r"(?:\b(?:forecast|expected|projected|predicted)\b|预计|预期|预测|估计|展望)",
+    re.IGNORECASE,
+)
+
+# Metric family for a claim/evidence leaf, so a figure is only grounded by
+# evidence of its own kind: an observed price must never stand in for an
+# invented volatility (#1336).
+_ANALYSIS_KIND_ALIASES = {
+    "annualized_vol": "vol",
+    "annualized_volatility": "vol",
+    "volatility": "vol",
+    "return_vol": "vol",
+    "return_volatility": "vol",
+    "vol": "vol",
+    "max_drawdown": "drawdown",
+    "maxdd": "drawdown",
+    "drawdown": "drawdown",
+    "sharpe": "sharpe",
+    "sharpe_ratio": "sharpe",
+    "win_rate": "win_rate",
+    "hit_rate": "win_rate",
+    "hitrate": "win_rate",
+    "probability": "probability",
+    "prob": "probability",
+    "total_return": "return",
+    "annual_return": "return",
+    "cumulative_return": "return",
+    "benchmark_return": "return",
+    "excess_return": "return",
+    "annualized_return": "return",
+    "return": "return",
+    "returns": "return",
+    "ic_positive_ratio": "win_rate",
+}
+
+# Order matters: 最大回撤 is drawdown before 收益/return, and 年化波动率 is vol
+# before the generic return branch.
+_ANALYSIS_KIND_PATTERNS: tuple[tuple[re.Pattern[str], str], ...] = (
+    (re.compile(r"(?:回撤|drawdown|maxdd|最大亏损)", re.IGNORECASE), "drawdown"),
+    (
+        re.compile(
+            r"(?:波动率|波动性|volatility|annualized\s*vol|return\s*vol|年化波动)",
+            re.IGNORECASE,
+        ),
+        "vol",
+    ),
+    (re.compile(r"(?:夏普|sharpe)", re.IGNORECASE), "sharpe"),
+    (re.compile(r"(?:胜率|命中率|win\s*rate|hit\s*rate)", re.IGNORECASE), "win_rate"),
+    (re.compile(r"(?:概率|probability|prob)", re.IGNORECASE), "probability"),
+    (re.compile(r"(?:收益|回报|收益率|回报率|\breturns?\b)", re.IGNORECASE), "return"),
+)
+
+# Definitional prose ("夏普比率大于 1.0 通常被认为较好") states a convention,
+# not a measurement this session produced (#1336: preserve explicitly labelled
+# definitions). The bare 通常 alone is deliberately not enough — "通常实现
+# 18.2% 年化" is a measured claim.
+_DEFINITION_FRAME_RE = re.compile(
+    r"(?:通常(?:认为|被?认为|说来|指|用于)|一般认为|定义为|是指|惯例|"
+    r"conventionally|typically\s+(?:considered|regarded|seen)|"
+    r"generally\s+(?:accepted|considered|regarded)|by\s+convention)",
+    re.IGNORECASE,
+)
+
+# A categorical "all N-month windows were profitable" claim carries only an
+# integer window length, which _MEASURE_NUMBER_RE deliberately ignores; it is a
+# measured fact about the run and needs a window-producing analysis result.
+_CATEGORICAL_WINDOW_RE = re.compile(
+    r"(?:所有|全部|每个|任何|each|every|all)[\s\S]{0,40}?"
+    r"(?:窗口|区间|windows?|periods?)[\s\S]{0,30}?"
+    r"(?:正收益|均为正|都为正|盈利|上涨|positive|profitable)",
     re.IGNORECASE,
 )
 _DERIVATION_RE = re.compile(
@@ -884,6 +959,24 @@ def _price_field_for_path(path: str) -> str | None:
     return _GENERIC_PRICE_FIELD_ALIASES.get(leaf)
 
 
+def _metric_kind_for_path(path: str) -> str | None:
+    """Map an evidence JSON path to an analysis metric kind."""
+    leaf = re.sub(r"\[\d+\]$", "", str(path or "").rsplit(".", 1)[-1])
+    leaf = leaf.strip().casefold()
+    kind = _ANALYSIS_KIND_ALIASES.get(leaf)
+    if kind is not None:
+        return kind
+    return _metric_kind_for_text(path)
+
+
+def _metric_kind_for_text(text: str) -> str | None:
+    """Return the analysis metric kind named in a claim or header."""
+    for pattern, kind in _ANALYSIS_KIND_PATTERNS:
+        if pattern.search(text):
+            return kind
+    return None
+
+
 def _scan_symbols(text: str) -> set[str]:
     """Return the canonical symbols written anywhere in a blob of text."""
     return {
@@ -1085,6 +1178,7 @@ class GroundingLedger:
         self._evidence: list[EvidenceRecord] = []
         self._tool_failures: list[dict[str, Any]] = []
         self._analysis_completed: list[dict[str, Any]] = []
+        self._analysis_metrics: list[dict[str, Any]] = []
         self._validations: list[dict[str, Any]] = []
         self._recovery_rounds = 0
         self._symbol_resolution_attempts = 0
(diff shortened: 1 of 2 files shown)


warren618:
Not merging this yet. The gap you found is real and your gate closes it — I
confirmed both halves against main before judging the fix:

```
MAIN, validate_final_answer:
  "历史回测显示策略年化波动率 18.2%，最大回撤 -9.4%，命中目标概率 58%。"  -> valid=True
  "所有历史 12 个月窗口均实现正收益。"                                    -> valid=True
```

Those are exactly #1336, and your branch rejects both. The `portfolio_risk_xray`
path works too: an observed `annualized_vol` of 0.182 grounds an 18.2% claim,
while an invented `-9.4%` sibling in the same sentence is still refused. That is
the right shape and I want it.

**What blocks it: three answer shapes that main accepts and this branch rejects.**
Same ledger in all of them, one successful `get_market_data` holding closes
100.0 (2026-08-03) and 112.4 (2026-09-02):

| answer | main | this PR |
|---|---|---|
| `AAPL.US 从 2026-08-03 的 100.0 涨到 2026-09-02 的 112.4，区间收益率为 12.4%。` | valid | **rejected** `12.4%` |
| `AAPL.US rose from 100.0 to 112.4, a cumulative return of 12.4% over the window.` | valid | **rejected** `12.4%` |
| `该论文报告其策略年化收益 18.2%，最大回撤 -9.4%（论文自述，非本次回测）。` | valid | **rejected** |

The first two are the single most common shape this product produces: both
endpoints are observed evidence sitting in the ledger, the answer states the
derivation inline, and the result is arithmetic on sourced inputs. The gate
refuses it because `收益率` / `return` is in `_ANALYSIS_METRIC_RE` and 12.4 is
not itself an observed number. Your "Known limitations" section names the right
remedy for the neighbouring case — the price gate's derivation exemption. It
applies here too: `_is_explicit_derivation` already exists at
`grounding.py:2559` and the price gate calls it at `:2210`. Reusing it means a
derived figure anchored to observed inputs stays legal while an unanchored one
does not.

The third case is worse than a nuisance because it holds **even when
`research_papers_tool` ran successfully and its result is in evidence**. The
leaf was `reported_annualized_return`; `_ANALYSIS_KIND_ALIASES` has no such key,
and `_metric_kind_for_text` does not match it either, because `\breturns?\b`
finds no word boundary inside `_return`. So the figure never becomes analysis
evidence and the answer is refused. That collides with the contract
`research_papers_tool` documents: a paper's claimed performance must be
*attributed*, not suppressed.

Which brings up the structural point. The PR description says it deliberately
avoids a hand-maintained tool list — but `_ANALYSIS_KIND_ALIASES` is a
hand-maintained list of *metric names*, one level down, with the same failure
mode: anything not spelled the way the table expects silently becomes
ungroundable and the answer gets blocked. A token-split match over the leaf path
(split on `_`/`.`/`[]`, then look for a kind token) would cover
`reported_annualized_return`, `strategy_max_drawdown` and everything else in
that family without anyone editing a table.

Two smaller notes:

- The whole gate is reachable only through `validate_final_answer`, so every
  false positive costs a full recovery round. Worth weighing when you pick the
  breadth of `_ANALYSIS_METRIC_RE` — `收益` and `\breturns?\b` are very wide
  nets for a rejection gate.
- Add the accept column to the tests. Yours are stronger than most (you do have
  four allow-cases), but none of them covers a *derived* figure, which is where
  the regression lives. A table of answers that must still validate, beside the
  ones that must not, is what would have caught this before CI went green.

Rebased with the derivation exemption reused and the kind resolution falling
back to a token match, this lands.

warren618:
Re-checked the three false positives from the earlier round against the current head rather than against the commit messages. One is fixed, two are not, and the shape of what remains is worth spelling out.

**Derived returns are exempted in English only.** Same construction, same numbers, different language, different verdict:

```
EN  "AAPL closed at 100.00 on day 1 and 110.00 on day 2, a 10% return."      -> no issues
ZH  "苹果第一日收盘 100.00，第二日 110.00，收益率 10%。"                          -> ['return']
ZH  "AAPL 第一日收盘 100.00，第二日收盘 110.00，收益率 10%。"                      -> ['return']
ZH  "AAPL closed at 100.00 第一日, 110.00 第二日, 收益率 10%。"                   -> ['return']
EN  "AAPL closed at 100.00 then 110.00, giving a 10% return."                -> no issues
```

The third Chinese line is the diagnostic one: it uses the English "closed at" and is still rejected, so the exemption is keyed on the English spelling of the *return* phrase, not on recognising the derivation. The UI ships seven locales and model output follows the user's language, so this ships a gate that is strictly stricter for Chinese answers than English ones — the same asymmetry #1249 had, and the same one I fixed in #1346 today for percentage points. I would not merge it in this state.

**Paper restatement is still rejected.** `"The paper reports a Sharpe ratio of 1.8 for the momentum factor."` still raises a `sharpe` issue with no tool results present. Attributing a number to a named source is the opposite of an unsourced claim.

**The gate is not blind, which is the important half.** Both `"TSLA.US last traded at 412.35 USD."` and `"特斯拉现价 412.35 美元。"` are still caught with zero tool calls, in both languages. Whatever the fix for the above, it has to keep these two failing — that is the pair I would build the regression table around.

168 tests pass and the branch merges onto current main without conflicts, so this is purely about the remaining behaviour. Suggestion, since per-phrasing exemptions are what got this gate to its current shape: rather than adding Chinese spellings alongside the English ones, key the exemption on the structural fact — the operands of the derivation are themselves present in the same clause and sourced — which is language-independent and cannot drift apart per locale. Happy to review again whenever you have something.

warren618:
Correction to my comment above, on attribution rather than on the finding.

I wrote that one of the three was fixed here. It is not fixed *here* — I re-ran the same probe against current `main` and the English derived-return case already passes there:

```
main   "AAPL closed at 100.00 on day 1 and 110.00 on day 2, a 10% return."  -> no issues
main   "苹果第一日收盘 100.00，第二日 110.00，收益率 10%。"                      -> rejected
```

So the English exemption is main's, not this branch's, and this branch neither adds nor removes it. The asymmetry I described is a property of the gate as it stands on main, which this PR inherits rather than introduces.

That does not change what I would ask for before merging — the Chinese side is still rejected, the paper-restatement case is still rejected, and both are things this PR is in a position to address — but it does mean I credited this branch with a fix that was not its doing, and the burden of the asymmetry is not something it created.

Related and worth knowing while you work on it: #1249 touches the same behaviour and inverts the asymmetry rather than removing it — on that branch the Chinese case passes and the *English* one is newly rejected. Whichever of the two lands first, the regression table should assert both languages give the same verdict, or the next merge will just swap which half is broken.

warren618:
I had the direction of this backwards, and since it changes what you would build, here is the correction with the measurement behind it.

I described the asymmetry as Chinese answers getting false rejections that English ones do not. The rejections were not false. My probe ran with **zero tool evidence**, and with no tool evidence an unsourced price claim *should* be rejected — so the Chinese verdict was right and the English one was the bug. English was escaping the gate, not Chinese being over-punished.

The mechanism, once I stopped reading phrasing and read the regex: `_PRICE_CONTEXT_RE` contained `\bclose\b`, which does not match **"closed"**. Nor did anything match "traded at" or "quoted at". Measured on the then-current main, with no canonical symbol to fall back on:

```
"The stock closed at 412.35."        -> PASSED
"The stock last traded at 412.35."   -> PASSED
"Shares were quoted at 412.35."      -> PASSED
"该股收盘 412.35。"                    -> rejected
```

A model inventing a USD price in the most ordinary English phrasing walked straight through. Chinese had the mirror-image hole in its own vocabulary — 成交价 / 最新价 / 股价 / 收报 all missed — so neither list was complete; they were incomplete in different places.

Fixed on main in `19ff47d1`: the English past-tense forms are matched (each requires a following "at", so "traded 1.2M shares" and "the deal closed at a 30% premium" stay out), the four Chinese nouns are added, and `tests/test_grounding_language_parity.py` asserts the two verdicts are **equal** per case rather than asserting a literal per language — a per-language literal is what let the halves drift apart. Both arms are pinned, including seven statements that must be accepted in both languages, so the suite cannot be satisfied by a gate that rejects everything. Three mutations are pinned on it. Full suite 12440 passing.

**What this means here.** The asymmetry is no longer something this PR has to carry, and my "I would not merge it in this state" was based on a misreading — I withdraw that. What still stands from the original review is the paper-restatement case: `"The paper reports a Sharpe ratio of 1.8."` is still rejected with no tool results present, and attributing a number to a named source is the opposite of an unsourced claim. That one is worth a look.

You will need to rebase — my change touches `_PRICE_CONTEXT_RE`, which this branch is also in. Sorry for the extra step and for sending you at the wrong problem first.

cgycorey:
Rebased onto current `main` (so `19ff47d1`'s price-context widening and the language-parity suite are in — `tests/test_grounding_language_parity.py` passes on this branch, 276 grounding/discipline/parity tests, full suite 12429 passing).

**Paper restatement** is addressed in 37a20272, and the design follows your structural suggestion one level down: the question the regex answers is not "does this look like a citation" but **"can the named source be this run's own output?"**

- Exempted subjects are restricted to external artifacts: the paper/study/researchers/analysts/surveys/regulators/literature/authors/filings/rating agencies + a reporting verb; Chinese 论文/文献/研究/分析师/研究机构/学者 + 报告/显示/表明/指出/称/估计/发现.
- `data` / `数据` / `报告` / bare `据` are deliberately **not** subjects. Measured on the previous iteration of this patch: `回测数据显示策略年化收益 25%` / `The backtest data shows an annualized return of 25%` — the exact #1336 attack — was being exempted by a citation-shaped regex. Those subjects are how a model dresses up its own numbers, so they stay blocked: `"The backtest data shows an annualized return of 25%."` and `"根据本次回测，年化波动率 18.2%。"` still raise `analysis_claim_unavailable`.
- The check is clause-scoped, not line-scoped. A line-level check let `"The paper reports a Sharpe of 1.8, and our strategy achieved an annualized return of 47.3%."` launder the invented sibling through the citation in clause 1. With clause scoping the sibling is still refused.
- Ordinary prose subjects stay out, as you measured: `"The strategy reports a Sharpe ratio of 1.8."` is still rejected.

The regression table now has both columns per your earlier note: accept (paper/analyst/filing restatements, EN+ZH) beside reject (your pinned TSLA/特斯拉 pair, the self-referential subjects, cross-clause laundering, unanchored returns, wrong-arithmetic returns).

Known remaining gap, unchanged from the PR description: a metric invented *after* a genuinely completed analysis still passes the tool-level path (run-dir `metrics.csv` is never value-checked).

cgycorey:
Friendly nudge — this PR is mergeable with all checks green. Happy to address anything blocking review. (Also: #1361 is in the same state if you get a chance to look at both.)

warren618:
Re-ran the table on the rebased head (`93427484`) against current main, 31 cases, both task framings. Most of it is what I asked for: all four #1336 self-claim shapes are rejected in both languages (main accepts every one), a wrong-arithmetic return (`100.0 → 112.4, 15.0%`) and an unanchored return are rejected (main accepts both), and the paper / analyst restatements pass with zero tool results in both languages. That is the shape I want.

Two things block it, one of them a loosening.

**1. `95e2c2f8` widens the price gate.** Applying `_ATTRIBUTION_RE` inside `_validate_price_claims` means any invented price passes once it has a citation subject. Zero tool evidence:

```
分析师指出特斯拉现价 412.35 美元。                 main: rejected   this PR: accepted
The paper reports that AAPL.US closed at 189.20.  main: rejected   this PR: accepted
The filing reports the stock closed at 412.35.    main: rejected   this PR: accepted
```

The exemption is right for analysis metrics — a paper's Sharpe is not something this run could have observed — but a price is exactly what this run observes, and "analysts say it closed at X" is the laundering shape the price gate exists to catch. Please drop the exemption from the price gate (keep it in `_validate_analysis_claims`) and add those three to the reject arm.

**2. A new false positive on a derived return.** With both closes in the ledger:

```
AAPL.US 第一日收盘 100.0 美元，第二日收盘 112.4 美元，收益率 12.4%。   main: accepted   this PR: analysis_claim_unavailable
```

The derivation exemption is keyed on the "从…涨到" / "rose from…to" phrasing rather than on the structural fact (both operands observed in the same clause, arithmetic checks out). The structural version would take this sentence and would not need a phrase per language.

With those two fixed I will merge it.

PR #1341, opened 2026-09-03
fix(agent): stop the loop discarding tool results the model still needs

saju01:
## Summary

Three defects in the agent loop, plus one in grounding, combine to make the agent state that data was **not retrieved** while its own `grounding_evidence.json` holds thousands of evidence items produced by exactly those calls. The model is not confabulating — it is accurately describing a context that the loop emptied.

Reproduced on `ATRC.US` against SEC EDGAR XBRL, `claude-opus-5`. Same prompt for every run below.

### 1. Tool dedup is keyed on the tool *name* alone — `loop.py:1693`

`self._called_ok: set[str]` stores `tc.name`, so any second call to a parameterised or paginated tool is answered with a synthetic skip:

```
get_financial_statements(statement="income")   -> ok
get_financial_statements(statement="balance")  -> BLOCKED "already completed successfully"
get_financial_statements(offset=13)            -> BLOCKED
```

`_result_paging` returns `{"returned": 13, "total": 64, "next_offset": 13, "complete": false}` — paging is the documented way to fetch the remainder, and the loop made it unreachable. Note the data tools inherit `repeatable = False` from `BaseTool`; many other tools set `repeatable = True`, so this opt-in was simply never given to them.

Now keyed on `(name, canonical arguments)`. **Identical repeats are still blocked** — verified, not assumed.

### 2. `_microcompact` fires on effectively every turn — `loop.py:327,1029`

It clears all but `KEEP_RECENT = 3` tool results once tokens exceed `token_threshold * 0.5`. With the default `token_threshold = 40000` that trigger is **20,000** — but this workload's baseline context (system prompt + tool schemas, *before any tool runs*) is **~64,800 tokens**.

The condition is therefore true from iteration 1, and every subsequent turn erases all but the last three results. Visible in `llm_usage.json` as input tokens *falling* while data accumulates:

```
iter 6 -> 7 : 77,996 -> 72,013
iter 13 -> 14: 81,437 -> 76,994
```

**This PR does not change the default**, which is defensible for small-context models. It fixes the reporting half (#3) so operators can see it, and raising `TOKEN_THRESHOLD` to match the model's real context resolves it.

### 3. A cleared result is indistinguishable from an empty one — `loop.py:338`

The placeholder was the bare string `"[cleared]"`. A model cannot tell *"this tool returned nothing"* from *"this succeeded and the context layer deleted it"* — so it reports a retrieval failure that never happened. The placeholder now states the call succeeded, how much was removed, and that re-calling recovers it.

### 4. Percentage-point deltas are validated as prices — `grounding.py:2489`

`_numbers_without_dates_or_percent` masks `%` but not `pp`/`bps`. So:

```
"~3.6pp below Penumbra"  ->  extracts 3.0   (the ".6" consumed as a decimal)
```

That reaches the OHLC comparator as an unsourced price claim; `recovery_action` then demands **`get_market_data`** — price data — to substantiate a *gross-margin* comparison, and rejects an otherwise correct fundamentals answer. Under a prompt that explicitly forbids price retrieval this is unrecoverable.

Added `_PERCENTAGE_POINT_RE`. **Genuine price claims still validate** (`closing price 48.20` → `[48.2]`).

## Verification

| build | tool calls | context evictions | false blocks | answer |
|---|---|---|---|---|
| stock v0.1.14 | 10 | many | 5 | 4,835 ch — "not retrieved" for data that *was* retrieved |
| + fix 1 | 29 | many | 10 | 4,261 ch — still "not retrieved" |
| + fixes 1–4 | — | **0** | **0** | **7,095 ch — all sections answered with per-number provenance** |

Fix 1 alone is *insufficient*: more successful calls simply means more results to erase.

Figures in the passing run reconcile exactly against values derived independently from the grounding artifact (TTM revenue `569.622M`, gross margin `76.30%`, operating income `+9.665M`), and against an independent vendor source to four decimal places.

## Blast radius

Both fixes are **strictly narrower** than the behaviour they replace:

- no previously-blocked *identical* duplicate is now allowed through;
- no previously-validated price claim is now skipped.

Unit-tested both directions, including the negative controls.

## Not addressed here

`KEEP_RECENT = 3` being a fixed count rather than proportional to the context budget is arguably the deeper issue behind #2, but changing it is a policy decision for maintainers rather than a bug fix, so it is left alone.

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -446,7 +446,17 @@ def _microcompact(messages: list) -> None:
     for msg in tool_msgs[:-KEEP_RECENT]:
         content = msg.get("content", "")
         if isinstance(content, str) and len(content) > 100:
-            msg["content"] = "[cleared]"
+            # A bare "[cleared]" is indistinguishable from a tool that
+            # returned nothing, so the model reports "no data was retrieved"
+            # for data it did receive and this layer then deleted. Say which
+            # it is, and say the result is recoverable.
+            msg["content"] = (
+                "[CLEARED FROM CONTEXT: this tool call SUCCEEDED and returned "
+                f"{len(content)} characters, which were removed to free "
+                "context space. This is NOT a tool failure and NOT an empty "
+                "result. If you need these values, call the tool again with "
+                "the same arguments.]"
+            )
 
 
 def _context_collapse(messages: list) -> None:
@@ -936,7 +946,18 @@ class AgentLoop:
         self.memory = memory or WorkspaceMemory()
         self._event_callback = event_callback
         self.max_iterations = max_iterations
-        self._called_ok: set[str] = set()
+        # Dedup identity is (tool name, canonical arguments) -- NOT the name
+        # alone. Keying on the name blocked every legitimate second call to a
+        # paginated or parameterised tool: get_financial_statements(
+        # statement='income') then (statement='balance') is one name but two
+        # different requests, and the second was answered with a synthetic
+        # 'already completed successfully' skip. The model then correctly
+        # reported that the balance sheet 'returned no readable content' -- a
+        # true statement about a fabricated tool result.
+        # Keys come from _identical_call_key, the same canonicaliser the
+        # deterministic cache uses, so the block path and the cache path can
+        # never disagree about what 'the same call' means.
+        self._called_ok: set[tuple[str, str]] = set()
         self._cancel_event = threading.Event()
         self._previous_summary: str = ""
         self._persistent_memory = persistent_memory
@@ -1930,7 +1951,17 @@ class AgentLoop:
 
             tool_def = self.registry.get(tc.name)
             is_repeatable = tool_def.repeatable if tool_def else False
-            if tc.name in self._called_ok and not is_repeatable:
+            # A None key means the arguments could not be canonicalised. The
+            # deterministic cache treats that as 'never cache'; the blocking
+            # gate must likewise treat it as 'never block', otherwise every
+            # un-serialisable call would collapse into a single identity and
+            # the second one would be skipped without ever running.
+            dedup_key = self._identical_call_key(tc.name, tc.arguments)
+            if (
+                dedup_key is not None
+                and dedup_key in self._called_ok
+                and not is_repeatable
+            ):
                 logger.warning(f"Blocked duplicate call: {tc.name} (already succeeded)")
                 skip_msg = json.dumps({"skipped": True, "reason": f"{tc.name} already completed successfully. Use the previous result."})
                 messages.append(context.format_tool_result(tc.id, tc.name, skip_msg))
@@ -2542,7 +2573,9 @@ class AgentLoop:
 
         success = _is_tool_success(result)
         if success:
-            self._called_ok.add(tc.name)
+            recorded_key = self._identical_call_key(tc.name, tc.arguments)
+            if recorded_key is not None:
+                self._called_ok.add(recorded_key)
             if tc.name == "backtest":
                 try:
                     _archive_backtest_result(result, self.memory.run_dir)
--- a/agent/tests/test_agent_loop_dedup_arguments.py
+++ b/agent/tests/test_agent_loop_dedup_arguments.py
@@ -1,0 +1,176 @@
+"""The duplicate-call gate must compare arguments, not just the tool name.
+
+`self._called_ok` stored `tc.name`, so the second call to any parameterised or
+paginated tool was answered with a synthetic ``{"skipped": true}`` result and
+never executed:
+
+    get_financial_statements(statement="income")   -> ok
+    get_financial_statements(statement="balance")  -> BLOCKED
+    get_financial_statements(offset=13)            -> BLOCKED
+
+`_result_paging` advertises `next_offset` and `complete: false`, so paging is
+the documented way to retrieve the rest -- and the gate made it impossible. The
+model then reported that the balance sheet "returned no readable content",
+which was a true statement about a fabricated tool result.
+
+The gate now keys on `_identical_call_key`, the same canonicaliser the
+deterministic cache uses, so the two paths cannot disagree about what "the same
+call" means.
+
+Harness mirrors test_agent_loop_deterministic_cache._drive.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from src.agent.context import ContextBuilder
+from src.agent.loop import AgentLoop
+from src.agent.tools import BaseTool, ToolRegistry
+from src.agent.trace import TraceWriter
+
+
+class _PagedTool(BaseTool):
+    """A NON-repeatable tool, so the duplicate gate applies to it."""
+
+    name = "get_financial_statements"
+    description = "test double for a paginated, parameterised fetch"
+    parameters: dict = {"type": "object", "properties": {}}
+    repeatable = False
+    is_readonly = True
+
+    def __init__(self) -> None:
+        self.calls: list[dict] = []
+
+    def execute(self, **kwargs: object) -> str:
+        self.calls.append(kwargs)
+        return json.dumps({"status": "ok", "call": len(self.calls), "args": kwargs})
+
+
+def _drive(
+    agent: AgentLoop,
+    tool_name: str,
+    run_dir: Path,
+    arg_sets: list[dict],
+) -> tuple[list[dict], list[dict]]:
+    """Run a sequence of tool calls through the loop's tool-call path."""
+    trace = TraceWriter(run_dir)
+    messages: list[dict] = []
+    react_trace: list[dict] = []
+    for index, arguments in enumerate(arg_sets, start=1):
+        agent._process_tool_calls(
+            [SimpleNamespace(id=f"call_{index}", name=tool_name, arguments=arguments)],
+            ContextBuilder,
+            messages,
+            trace,
+            react_trace,
+            index,
+        )
+    trace.close()
+    return messages, list(TraceWriter.read(run_dir))
+
+
+@pytest.fixture()
+def agent_factory(tmp_path: Path):
+    """Return a builder for an AgentLoop wired to a run dir and one tool."""
+
+    def _build(tool: BaseTool) -> tuple[AgentLoop, Path]:
+        registry = ToolRegistry()
+        registry.register(tool)
+        agent = AgentLoop(
+            registry=registry,
+            llm=SimpleNamespace(),
+            max_iterations=4,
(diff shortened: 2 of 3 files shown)


warren618:
Thanks for the depth here — the writeup is unusually good, and defect 4 is a
real find I want.

Three things before this can move.

**1. The diff carries no tests.** The description says "Unit-tested both
directions, including the negative controls," but:

```
$ git diff --stat main...pr-1341
 agent/src/agent/grounding.py | 11 +++++++++++
 agent/src/agent/loop.py      | 40 ++++++++++++++++++++++++++++++++++----
 2 files changed, 47 insertions(+), 4 deletions(-)
```

I assume the test file was left out of the push rather than never written. For a
change inside `src/agent/` that relaxes a dedup gate, the tests are the part I
review most closely, so I can't judge it without them.

**2. Defect 1 collides with #1268**, which is open and does the same thing more
thoroughly: it replaces the name-keyed `_called_ok` with an argument-aware key,
adds an `InvocationPolicy` enum so a tool can declare `once_per_run` versus
`identical_call`, tracks a reserved set so two in-flight calls with the same key
don't race, and ships two test files. Your diagnosis and #1268's are the same
diagnosis; landing both would mean one of you rewriting the other's code. I'd
rather converge on #1268 for that half, with your paging repro
(`get_financial_statements(statement=…)`, `next_offset`) added to its tests —
it's a sharper reproduction than what #1268 currently has.

**3. Defect 4 should be its own PR.** The `pp`/`bps`/`百分点` mask is 11 lines,
independent of the loop work, and fixes a real false positive: `~3.6pp` yields
3.0 to the OHLC comparator, which then demands `get_market_data` to substantiate
a gross-margin comparison. Send that alone with a test both ways — the pp case
masked, and a genuine `closing price 48.20` still extracted — and it goes in
immediately.

On defects 2 and 3 I agree with your framing and with your decision not to
change `TOKEN_THRESHOLD` here. The `[cleared]` placeholder genuinely cannot be
told apart from an empty result, and saying so in the placeholder is the right
minimal fix. That part travels fine with either PR.

he-yufeng:
Verified the core claim both ways with a probe driving `_process_tool_calls` directly (same shape as `test_agent_loop_deterministic_cache._drive`):

- On main: a non-repeatable tool called with `{"statement": "income"}` succeeds, then called with `{"statement": "balance"}` gets the synthetic `{"skipped": true}` result and never executes. So the model's "returned no readable content" report is accurate about a fabricated tool result, exactly as the PR describes.
- On this branch: both calls execute.

The direction is right, but two things before I can support it landing:

1. **Reuse `_identical_call_key` instead of adding `_call_key`.** Main already canonicalizes `(name, args)` at loop.py:2493, including the `_normalize_tool_run_dir` pass, and the deterministic cache keys on it at two sites (1968, 2581). This PR adds a second canonicalizer beside it, with different semantics on two edges: no run_dir normalization, and a `repr()` fallback where the cache deliberately keys `None` (unserializable = never cached). If the block path and the cache path drift, a call can be blocked as "duplicate" and still miss the cache, which is the worst of both. The blocking gate should read from the same helper the cache already trusts.

2. **It needs tests for a loop-core behavior change.** The `_drive` harness in `test_agent_loop_deterministic_cache.py` fits exactly: one arg set pair pins that identical (name, args) still skips, another pins that different args re-execute. That is also the red test for the bug itself on main.

The `_microcompact` wording change is worth its context cost (a bare `[cleared]` is genuinely indistinguishable from an empty result, and the recoverable-by-recall note is the right hint), and the percentage-point mask is a clean, bounded add. No objection to either once the two points above are in.

saju01:
Both review points are addressed, and @warren618 is right that the tests were the missing half — they were never written, not left out of the push. Saying otherwise in the description was wrong and I've corrected it here.

Branch is rebuilt on current `main` (`d71ed8dd`); the old head was based on the 0.1.14 release and had drifted 350 commits, which is also why the earlier diff was hard to read.

**@he-yufeng — `_identical_call_key` instead of a second canonicaliser.** You were right and the reason is sharper than "less duplication": the two edges you named were real divergence. The gate now calls `_identical_call_key`, so it inherits `_normalize_tool_run_dir` and cannot drift from the cache. On the `None` return — the cache treats it as *never cache*; the gate now treats it as *never block*. Collapsing every un-serialisable call into one identity would have skipped the second one without running it, which is the same class of bug this PR is fixing. Thank you for the independent reproduction; that probe is what made the `{"skipped": true}` path concrete rather than inferred.

**Tests, using the `_drive` harness you both pointed at** (`agent/tests/test_agent_loop_dedup_arguments.py`, non-repeatable tool so the gate applies):

- distinct `statement` values re-execute — the red test for the bug
- a paged `offset` continuation re-executes
- an identical repeat is **still skipped** — negative control, so relaxing the key doesn't become "never dedup"
- key order doesn't create a new identity

Verified RED on `d71ed8dd`: the two distinct-args tests fail with `Blocked duplicate call: get_financial_statements (already succeeded)` in the log, and the two negative controls pass. GREEN with the fix: 4 passed.

Across the loop and grounding suites: **221 passed with the fix, 217 at baseline**, with the same single pre-existing failure in both (`TestArchiveBacktestResult`, missing `mcp` module).

One thing I broke and fixed rather than leave for you to find: the `[cleared]` change failed two existing assertions in `test_loop_helpers.py` that matched the literal string. Those are updated to match the marker prefix. They were correct tests catching a real behaviour change.

**@warren618 on #1268 — agreed, and I'd rather converge than have one of us rewrite the other's code.** #1268 is the more thorough treatment: the `InvocationPolicy` enum and the reserved-set race handling are both things this PR lacks, and the race in particular is a real gap in mine. I've commented there offering the paging repro as a test case. If you'd prefer, I'm happy to close this PR entirely once #1268 lands and keep only the `_microcompact` placeholder half, which travels with either.

**Defect 4 is now #1346**, standalone with tests both directions, as you asked.

What's left here is defects 1, 2 and 3. If #1268 lands first I'll drop defect 1 from this branch without argument.

he-yufeng:
Strong work: the verification table and the blast-radius section are exactly the standard this repo's agent loop needs, and defects 1 and 4 are each real with the right narrow fix. I checked the arg-keyed dedup on current main and identical repeats do stay blocked, and the `_PERCENTAGE_POINT_RE` addition leaves genuine price claims validating.

One internal contradiction to fix before this can land, in defect 3. The new placeholder tells the model "If you need these values, call the tool again with the same arguments." Under this same diff that call cannot succeed: the dedup ledger now keys on (name, canonical args), a cleared result's tuple stays in `_called_ok`, and an identical re-call hits the same "already completed successfully" skip it always did. So the placeholder instructs the model to do the one thing the ledger forbids, and the failure the fix is documenting becomes more confusing, not less: the model retries per the placeholder's own advice and gets blocked again.

The missing half is the ledger knowing what context still contains. When microcompact clears a result, the (name, args) tuple has to leave `_called_ok` too, otherwise the recovery path the placeholder promises is dead. I have exactly that mechanism up as #1349 (cleared tools become callable again, with a surviving result keeping the block, six red-green tests and a trace event for the previously silent clearing). Two ways to sequence it: land #1349 first and this rebases so the placeholder becomes true, or fold the ledger sync into this PR alongside the placeholder. I slightly prefer the first: this diff is already four fixes, and the ledger sync is the piece that most wants its own tests.

warren618:
Keying dedup on `(name, canonical arguments)` instead of the name alone is correct and it is a different axis from #1349 — that one re-opens the ledger when a result becomes unreadable, this one stops same-name-different-arguments calls being treated as one request. Both are real; neither subsumes the other.

They do cancel each other out if landed as-is, which is why I pushed the integration onto this branch rather than merging it after #1349. Two defects, both measured:

**1. The unblock silently becomes a no-op.** This PR makes `_called_ok` a set of `(name, args)` tuples; #1349 re-opens it with `difference_update(<names>)`. Subtracting strings from a set of tuples removes nothing: `{('get_fund_flow', '{...}')}` minus `{'get_fund_flow'}` is unchanged. So #1343 would have come straight back the moment this landed, with both suites green, because each only exercises its own shape. The unblock now drops every argument variant of a cleared tool by name, and there is a mutation test pinned on exactly the sequential-merge scenario.

**2. The marker is longer than the pruning threshold, which corrupts it.** The self-describing `[CLEARED FROM CONTEXT: ...]` text is over 100 characters — microcompact's own cutoff — so a second pass re-cleared an already-cleared result and rewrote the recorded size with the *marker's* length. The model would be told a tool "returned 287 characters" where 287 is the length of the placeholder, not the data. It also re-reported the tool as newly unreadable on every later iteration, re-opening and re-tracing the ledger each time. This one exists in this PR standing alone; #1349's idempotence test is what surfaced it. Microcompact now skips already-cleared content.

The `[cleared]` literal is now a `_CLEARED_PREFIX` sentinel with `_cleared_text` / `_is_cleared` helpers, because a marker that embeds the payload length means no two cleared results are the same string — `_context_collapse` was doing an equality check against it and would have silently stopped matching.

57 tests pass across the three affected files; three mutations (name-keyed unblock, no idempotence guard, equality instead of prefix) each fail it. Please sanity-check the resolution — the substance is yours, I only made the two mechanisms survive each other.

warren618 (approved):
Argument-aware dedup is a distinct and necessary axis from #1349; I pushed the integration of the two plus a fix for the marker-length defect (details in the comment).

saju01:
Sanity-checked the merged integration on `5c2766f`. The tuple/name reconciliation, marker idempotence and prefix matching are the right fixes. The surrounding loop and safety selection passes locally; I also verified the new regressions fail against the original runtime rather than merely exercising a helper.

Two narrower dead-pointer cases remain, now addressed in #1358:

1. Income and balance calls both succeed; microcompact clears income but retains balance. The gate compares `(name, args)`, while the unblock was still waiting until *every variant of that tool name* was gone. Retrying income remained blocked even though balance is not its result.
2. Full `_auto_compact` summarizes an old pair out of context but leaves its success lock behind, including when summarization degrades. This is the layer-3 concern previously described in #1350, which was closed in favor of the layer-1 fix.

The follow-up records successful call-ID → existing canonical-key provenance and reconciles exact readable results at both compaction sites. Skips/errors/stubs do not preserve a lock; a surviving success for the same key does. Only known-read-only calls reopen: losing an order/write result must not authorize repeating the side effect. Cached recovery restores the identical-call gate once data is readable again.

Verification: 430 baseline regression/safety tests pass; 18 new cases bring the same selection to 448 passing. Against original `loop.py`, the three core regressions fail with the actual `Blocked duplicate call: get_financial_statements` symptom. No changes to token thresholds, live deployments or #1268's separate in-flight policy work.

Thanks for integrating the fixes rather than leaving the representation collision for the next merge to discover. My original placeholder change needed the recovery mechanism you added, not just clearer prose.

PR #1344, opened 2026-09-04
fix(backtest): scale over-committed rebalance basket instead of aborting (#1274)

### LINKED ISSUE #1274: bug(rebalance): 100% target plus commission aborts instead of scaling basket

cgycorey:
## What

Fixes #1274. A rebalance-mode portfolio whose targets sum to 100% raised `ValueError: insufficient capital for position rebalance` whenever commission was nonzero — the engine planned the full basket, projected it one commission-width past capital, and aborted atomically, even though the legacy open-basket path already has a scale-to-fit mechanism.

`_execute_target_rebalance` now mirrors that behavior:

- When the planned basket overdrafts, **reductions commit as planned** (they release capital toward the targets) and every **capital-consuming sleeve** — fresh opens *and* same-direction increases — scales by **one common factor** (binary search, 50 iterations, same shape as the open path), preserving portfolio proportions and code-order independence.
- Scaled sizes **re-round per symbol lot rules and re-price under the symbol's own fee schedule** (`_active_symbol` is set per order inside the fit loop — without this, engines that dispatch on it (`china_futures`, `composite`, `global_futures`, `forex`) silently apply the wrong lot grid/fee).
- A sleeve whose rounded size hits zero is **dropped and recorded** via `_on_plan_rejected(symbol, "zero_size", ts)` plus a warning log, so run-card diagnostics (`plan_rejections`, `unfilled_plan_rejections_by_symbol`) see the dropped leg instead of silence.
- When **no scale fits** — not even an empty open sleeve (e.g. a reduction whose realized loss exceeds released margin) — the basket still **fails atomically** before any commit.

## Tests

Replaces the two abort-pinning tests with the contract the issue asks for, plus adversarial-review hardening (three independent reviewers; all findings addressed):

- over-committed single basket scales (fee 10%, 100% target → size ≈ 10·1000/1100)
- 60/40 basket + commission: ratio preserved, capital fits, leftover < 0.01
- code-order independence of the scaled basket
- differing per-symbol fees/lot grids: notional weights stay 60/40 and each fill is commissioned under its own schedule (catches the stale-`_active_symbol` blocker)
- reduction commits while the increase sleeve scales to fit
- irrecoverably infeasible basket (short closed deep underwater): still raises, zero side effects (trades/events/bar snapshots pinned)
- scaled-away sleeve recorded as `zero_size` plan rejection

Full engine suite: **374 passed, 1 skipped** (`test_base_engine`, `china_a`, `china_futures`, `crypto`, `engine_metrics_json`, `engine_robustness`, `composite_fallback`, `runner_security`, `cli_backtest_summary`, `autopilot_validation`).

## Known limitations

- Bisection assumes cost is monotone in scale; holds for all shipped engines. A hypothetical size-discounted commission could make the search return a feasible-but-suboptimal scale — never an overdraft, since every candidate is verified against its actual rounded cost.
- If a subclass returns NaN margin/commission during fitting, the error message is the generic "non-finite position rebalance value" rather than "insufficient capital" — still pre-commit atomic.

### DIFF
--- a/agent/backtest/engines/base.py
+++ b/agent/backtest/engines/base.py
@@ -1686,7 +1686,10 @@ class BaseEngine(ABC):
         )
         self._validate_rebalance_values(projected_capital)
         if projected_capital < -1e-9:
-            raise ValueError("insufficient capital for position rebalance")
+            fitted = self._fit_rebalance_opens(opens, reductions, ts)
+            if fitted is None:
+                raise ValueError("insufficient capital for position rebalance")
+            opens = fitted
 
         for order in reductions:
             if order.target_size <= 1e-9:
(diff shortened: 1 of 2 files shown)


warren618 (approved):
Selected over the competing PR after comparing both implementations; rationale in the comment.

warren618:
Merged. Two implementations existed for #1274 — this one and #1278 — and this is the one that goes in.

The deciding difference is where the binary search sits relative to planning. This PR scales orders that are already planned, then records `_on_plan_rejected` once per dropped sleeve after the search settles. #1278 re-runs the whole planning pass inside the search, and `_plan_open_order` calls `_on_plan_rejected` in seven branches, so a symbol rejected at small scales would be recorded up to fifty times per bar and the run-card diagnostics would be flooded with duplicates.

Scaling only the open sleeve while reductions commit as planned is also the more conservative reading of "scale the basket": reductions release capital toward the targets, so holding them at full and shrinking only the capital-consuming legs preserves the intent without re-deriving every target weight.

96 tests pass in `test_base_engine.py` after rebasing onto current main.

PR #1345, opened 2026-09-04
fix(swarm): give investment-committee researchers the fundamental data panel (#1343)

cgycorey:
## What

Partial fix for #1343 — the worker-side half. (The main dedup × microcompact interaction deserves its own focused change; this PR deliberately does not touch it.)

The `investment_committee` preset's `bull_advocate`, `bear_advocate` and `risk_officer` system prompts **mandate** fundamental analysis — PE/PB percentiles, earnings quality (ROE/ROIC, FCF), fund flows, margin balance, analyst reports, news — but their `tools:` whitelists contained **none** of the fundamental data tools. Since swarm workers get a strict whitelist (`build_swarm_registry` → `_filter_registry`), every run degraded exactly as the issue observed: workers reported "本 run 无基本面数据工具可执行" and fell back to price-only, qualitative claims.

This adds `get_financial_statements`, `get_fund_flow`, `get_margin_trading`, `get_research_reports` and `get_stock_news` to the three researchers' whitelists. The risk officer gets them too deliberately: its framework requires checking the advocates for "cherry-picking favorable data", which is impossible without access to the same sources.

## Tests

- `test_investment_committee_research_workers_have_fundamental_tools` — pins the five-tool panel on all three researchers (red before, green after).
- `test_every_preset_tool_exists_in_local_registry` — generic guard: every tool named in every bundled preset must resolve in the local registry. A whitelist name the registry lacks is **silently dropped** by `_filter_registry` (log-only) — the same failure mode one level down, now caught at test time instead of in a user run.

Suite: preset packaging/inspect/matching/route/registry-assembly/grounding/output-contract/worker-report/trust-model/e2e — **120 passed**.

## Scope note

#1343's primary defect (context-clear + name-based dedup making successful results unreadable) is a separate loop-level change and is intentionally left to a focused follow-up.

### DIFF
--- a/agent/src/swarm/presets/investment_committee.yaml
+++ b/agent/src/swarm/presets/investment_committee.yaml
@@ -44,7 +44,9 @@ agents:
       7. **Main risk to the bull case** — Honestly list 2–3 scenarios that would invalidate the bull thesis
 
       Use factor_analysis for quant support; use load_skill for frameworks.
-    tools: [bash, read_file, write_file, load_skill, get_market_data, factor_analysis]
+    tools: [bash, read_file, write_file, load_skill, get_market_data, factor_analysis,
+            get_financial_statements, get_fund_flow, get_margin_trading, get_research_reports,
+            get_stock_news]
     skills: [technical-basic, fundamental-filter, yfinance, earnings-revision, sentiment-analysis]
     max_iterations: 50
     timeout_seconds: 1800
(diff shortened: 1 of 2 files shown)


warren618:
Preset change is right and the two tests are the correct shape. One defect in the guard itself.

`test_every_preset_tool_exists_in_local_registry` builds its failure message as `f"{entry['name']}/{agent.id} ..."`, but `agent` is a `dict` here — the other test in the same file correctly uses `agent["id"]`. Since an assertion message is only evaluated when the assertion fails, this is invisible while everything passes.

Reproduced by adding a bogus tool name to `investment_committee.yaml`: the test does fail, but with `AttributeError: 'dict' object has no attribute 'id'` instead of naming the preset, the agent and the tool. On the one day this guard matters, it reports nothing usable. One-character fix: `agent.id` → `agent["id"]`.

warren618 (approved):

PR #1346, opened 2026-09-04
fix(grounding): don't read percentage-point deltas as quoted prices

saju01:
Split out of #1341 at review request — this is the 11-line half that is independent of the loop work.

## The bug

`GroundingLedger._numbers_without_dates_or_percent` masks `%` but not the percentage-point spellings. So:

```
"gross margin ~3.6pp below Penumbra"  ->  [3.0]
```

The `.6` is consumed as a decimal and the leading `3` survives as a candidate price. That number reaches the OHLC comparator as an unsourced price claim.

On an ATRC.US fundamentals run this rejected an otherwise correct answer and demanded `get_market_data` to substantiate a gross-margin comparison that has nothing to do with price. The margin gap was real, sourced, and correctly computed — it was thrown out because a *comparison of two percentages* was mistaken for a quoted share price.

## The fix

One regex, applied alongside the existing percent masks in the same masking chain:

```python
_PERCENTAGE_POINT_RE = re.compile(
    r"[-+~≈]?\s*\d[\d,]*(?:\.\d+)?\s*(?:pp|ppt|ppts|bps|bp|百分点)\b",
    re.IGNORECASE,
)
```

Covers `pp` / `ppt` / `ppts` / `bps` / `bp` / `百分点`, with or without a space, and with an optional leading sign or approximation marker.

## Tests, both directions

Per the review note on #1341 — the mask must not become a way to lose real prices:

- 11 percentage-point spellings are masked, including `-1.8pp`, `+0.9pp`, `≈3.6pp`, `45 bps`, and `3.6 百分点`
- 4 genuine prices are still extracted: `closing price 48.20`, `the stock closed at 48.20 on heavy volume`, `last trade 152.75`, `opened at 1,204.50 and faded`
- a mixed sentence (`closed at 48.20, with gross margin ~3.6pp below peers`) yields **only** the price
- a bare `48.20` is untouched, guarding against an over-broad mask

## Verification

| | result |
|---|---|
| RED, on `d71ed8dd` without the fix | 10 failed, 7 passed |
| GREEN, with the fix | 17 passed |
| surrounding grounding suite, baseline | 168 passed, 2 failed |
| surrounding grounding suite, with fix | 168 passed, 2 failed |

The 2 failures are pre-existing and identical before and after (a missing optional dependency in the test environment), so this adds no regression.

One test fixture is worth flagging: my first draft used `"operating margin improved 12.5 ppts since 2021"` and it failed *with* the fix — because the bare year `2021` leaks through as a price candidate. That is a separate pre-existing behaviour, not something this PR claims to address, so I changed the fixture rather than widening the mask to hide it. Happy to open a follow-up if bare-year handling is of interest.

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -256,6 +256,17 @@ _RANGE_TAIL = r"(?:\s*[-–—~～至到]\s*[-+]?\d[\d,]*(?:\.\d+)?)?"
 _PERCENT_RANGE_RE = re.compile(
     r"\d[\d,]*(?:\.\d+)?\s*[-–—~至]\s*\d[\d,]*(?:\.\d+)?\s*[%％]"
 )
+# A percentage-POINT delta is not a quoted price. "~3.6pp below Penumbra"
+# describes a margin gap; left unmasked the ".6" is consumed as a decimal
+# and the number reaches the OHLC comparator, which then rejects an
+# otherwise correct fundamentals answer and demands `get_market_data` to
+# substantiate a claim that has nothing to do with price.
+# The trailing "%" spellings are already handled above; this covers the
+# percentage-point spellings, which the percent masks never matched.
+_PERCENTAGE_POINT_RE = re.compile(
+    r"[-+~≈]?\s*\d[\d,]*(?:\.\d+)?\s*(?:pp|ppt|ppts|bps|bp|百分点)\b",
+    re.IGNORECASE,
+)
 # Localized calendar text carries digits that the ISO pattern above leaves
 # behind: "8 月 3 日" otherwise contributes 8 and 3 as candidate prices.
 _LOCALIZED_DATE_RE = re.compile(
@@ -2548,6 +2559,7 @@ class GroundingLedger:
         masked = _SHORT_DATE_RE.sub(" ", masked)
         masked = _DASH_DATE_RE.sub(" ", masked)
         masked = _PERCENT_RANGE_RE.sub(" ", masked)
+        masked = _PERCENTAGE_POINT_RE.sub(" ", masked)
         masked = _ORDER_LEVEL_RE.sub(" ", masked)
         masked = _AGGREGATE_AMOUNT_RE.sub(" ", masked)
         masked = _LABELLED_SCORE_RE.sub(" ", masked)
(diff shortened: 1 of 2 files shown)


warren618 (approved):
Widened to the ordinary Chinese spellings and pinned with a two-sided table; details in the comment.

warren618:
Merged, with one commit pushed onto the branch first.

The English spellings were the ones the report named, but this gate sees whatever language the model answered in. Measured before the change, of seven natural Chinese spellings exactly one was masked:

```
毛利率下降 3.6 个百分点     -> leaked 3.6
毛利率下降3.6个百分点       -> leaked 3.6
净利率提升 2 个百分点。      -> leaked 2.0
利差扩大 250 个基点         -> leaked 250.0
同比下降 3.6 百分点         -> masked   (the only one)
```

The mask required the number to sit directly against 百分点, which is the rare spelling — the ordinary one puts the measure word 个 in between — and 基点 was not covered at all. So a Chinese fundamentals answer kept being rejected as an unsourced price claim while the identical English answer passed. Given seven locales and output that follows the user's language, an English-only mask is half a fix.

The CJK units also drop the trailing `\b`. After a CJK character `\b` requires a non-word character to follow, so "下降 3.6 个百分点，主因是原材料" would not have matched even with 个 allowed. The ASCII units keep `\b`, which is what stops "3.6ppm" being read as pp.

Tests are a two-sided table, since this widens what the gate ignores: seven Chinese spellings must be masked, and five real price claims in both languages must still be extracted. Two mutations pinned — restoring the old regex fails 6, removing the mask entirely fails 17.

Worth flagging for the batch as a whole: this is the third language asymmetry in this gate I have hit today (#1249 had one, #1338 still has one on derived returns). It may be worth a rule that any new mask or exemption here ships with its non-English spellings in the same commit.

saju01:
Sanity-checked the merged resolution on `5c2766f`: 282 tests pass across the percentage-point, language-parity, grounding and output-discipline suites. The ordinary Chinese spellings are covered and genuine-price controls remain intact. Thanks for filling the language gap I left.

I found one small coverage issue in the boundary case and opened #1357. The fixture described as a trailing CJK character actually has a comma after `百分点`; `\b` matches before that comma. Restoring the unwanted CJK boundary left all 29 tests green. Replacing the comma-bearing example with `毛利率下降 3.6 个百分点后企稳` makes that mutation fail (1 failed / 28 passed), while the real implementation keeps all 282 tests green. This is a test and comment correction, not another runtime change.

PR #1347, opened 2026-09-04
fix(strategy_store): break created_at ties so history is really newest first

ethanstoner:
## Summary

`list_artifacts`, `get_bench_history` and `get_decay_history` all document "newest first" but sort on `created_at` with no tiebreaker. Records that share a timestamp come back oldest first, which is the opposite of what the docstrings promise.

Three tests in `agent/tests/test_strategy_store.py` already fail on `main` because of this:

- `TestInMemoryStore::test_get_bench_history`
- `TestInMemoryStore::test_get_decay_history`
- `TestSqliteStore::test_list_artifacts_ordering`

## Why

Python's sort is stable, and `reverse=True` does not reorder equal keys, so a tie keeps insertion order and insertion order is oldest first. SQLite has the same gap: with no tiebreaker it falls back to rowid ascending.

Ties are not an edge case. `_now_iso()` in `store.py:41` only has the clock's resolution behind it, so any batch registered inside one tick shares a timestamp. On Windows the granularity is about 15.6ms, which makes collisions the normal case. On Linux the microsecond resolution makes them occasional, which is why this has been passing in CI: `.github/workflows/test.yml` runs the full pytest job on `ubuntu-latest` only, and the Windows jobs run just `test_background_tools.py` and `test_bash_tool.py`.

## Changes

Six one-line edits plus a test class.

- `sqlite_store.py`: `ORDER BY created_at DESC, rowid DESC` for artifacts, whose `id` is a random text key, and `, id DESC` for bench and decay, which are `INTEGER PRIMARY KEY AUTOINCREMENT`.
- `store.py`: tuple sort keys. Artifacts break ties on insertion position, bench and decay on their assigned `id`.
- `test_strategy_store.py`: new `TestOrderingWithTiedTimestamps` pins `created_at` to a fixed value so the inversion reproduces on any platform rather than only where the clock is coarse. Without that pin the bug stays invisible to Linux CI.

## Test plan

`agent/tests/test_strategy_store.py` goes from 93 passed / 3 failed to 99 passed / 0 failed.

Full `agent/tests`, run on Windows: 39 failed / 12112 passed on this branch against 42 failed on `main`. I captured both failure lists and diffed them. There are no new failures, and the three that disappear are exactly the ones above. The remaining 39 are pre-existing POSIX file-mode, directory-fsync and symlink assertions that fail on any clean Windows checkout, unrelated to this change.

The three new tests fail without the source change and pass with it.

## Risk

Low, and confined to read paths. No schema change, no migration, no write path touched. `strategy_store` is outside the `src/agent/`, `src/session/` and `src/providers/` paths that need prior discussion.

The one behavioural change is the intended one: callers that were silently receiving reversed history on tied timestamps now receive it in the documented order. Anything that depended on the old order was depending on a bug the docstrings contradict.

## Rollback

`git revert` of the single commit. There is no persisted state to unwind, since the change is only in ordering clauses and sort keys.

---

Disclosure: this was written with AI assistance. The measurements above are ones I ran on my own machine and reproduced, not ones I inferred.

### DIFF
--- a/agent/src/strategy_store/sqlite_store.py
+++ b/agent/src/strategy_store/sqlite_store.py
@@ -475,7 +475,7 @@ class SqliteStrategyStore:
             where = "WHERE " + " AND ".join(clauses)
 
         rows = self._conn.execute(
-            f"SELECT * FROM artifacts {where} ORDER BY created_at DESC LIMIT ?",
+            f"SELECT * FROM artifacts {where} ORDER BY created_at DESC, rowid DESC LIMIT ?",
             [*params, limit],
         ).fetchall()
         return [self._row_to_artifact(row) for row in rows]
@@ -645,7 +645,7 @@ class SqliteStrategyStore:
             """
             SELECT * FROM bench_history
             WHERE artifact_id = ?
-            ORDER BY created_at DESC
+            ORDER BY created_at DESC, id DESC
             LIMIT ?
             """,
             (artifact_id, limit),
@@ -693,7 +693,7 @@ class SqliteStrategyStore:
             """
             SELECT * FROM decay_snapshots
             WHERE artifact_id = ?
-            ORDER BY created_at DESC
+            ORDER BY created_at DESC, id DESC
             LIMIT ?
             """,
             (artifact_id, limit),
--- a/agent/src/strategy_store/store.py
+++ b/agent/src/strategy_store/store.py
@@ -202,19 +202,23 @@ class InMemoryStrategyStore:
         limit: int = 100,
     ) -> Sequence[Artifact]:
         """List artifacts with composable filters, newest first."""
-        results: list[Artifact] = []
-        for art in self._artifacts.values():
+        results: list[tuple[int, Artifact]] = []
+        for position, art in enumerate(self._artifacts.values()):
             if type is not None and art.type != type:
                 continue
             if status is not None and art.status != status:
                 continue
             if universe is not None and art.universe != universe:
                 continue
-            results.append(art)
+            results.append((position, art))
 
-        # Sort by created_at descending (newest first).
-        results.sort(key=lambda a: a.created_at or "", reverse=True)
-        return results[:limit]
+        # Sort by created_at descending (newest first), breaking ties on
+        # insertion order. Sorting is stable and reverse=True does not reorder
+        # equal keys, so without the tiebreaker records sharing a timestamp
+        # come back oldest first. Timestamps collide readily: _now_iso() has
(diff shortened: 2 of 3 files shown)


warren618:
The fix is correct and I verified the premises rather than taking them from the description: `artifacts` is `id TEXT PRIMARY KEY` and not `WITHOUT ROWID`, so `rowid DESC` is available; `bench_history` and `decay_snapshots` are `INTEGER PRIMARY KEY AUTOINCREMENT`; and `_bench_counter`/`_decay_counter` in `store.py` really do ascend with insertion, so the in-memory tiebreakers order newest-first as claimed. The platform note is honest too — on macOS `main` is 96 passed, so the three failures are genuinely Windows-clock-specific.

One gap before merge. The PR changes six code paths but ships three tests. I reverted only the three that no test covers — sqlite `get_bench_history`, sqlite `get_decay_history`, and the in-memory `get_decay_history` — and the suite stayed at **99 passed, 0 failed**. So half the fix has nothing watching it and would regress silently.

Could you extend `TestOrderingWithTiedTimestamps` to cover those three? Same fixed-`created_at` pin you already use. Happy to merge once they're in.

warren618:
Added the three missing tests myself rather than leaving this waiting — they are on the branch now.

Writing them turned up something worth knowing, and it slightly revises what I said above. The six changed paths are not equally load-bearing:

- `artifacts` has **no** index on `created_at` — only `idx_artifacts_status` and `idx_artifacts_type` — so its order comes entirely from the explicit `ORDER BY`. Removing your tiebreaker fails both artifact tests, in-memory and sqlite. Load-bearing.
- The in-memory `get_decay_history` tiebreaker is load-bearing too; the new test fails without it.
- `bench_history` and `decay_snapshots` **do** have `(artifact_id, created_at)` indexes. SQLite walking one of those in DESC order is already a total order, so `, id DESC` there is redundant: measured, with the clause removed the store still returns newest-first. My first attempt at proving otherwise used an isolated table with no index, which is why it looked like it mattered.

So four of the six changes are doing work and two are defensive. That is not an argument for removing them — I have kept both sqlite history tests, because they pin the documented *behaviour* rather than the mechanism, and if that index is ever dropped or reordered the ordering inverts silently and these are what catch it. The comment in the test file says explicitly that they cannot fail on the tiebreaker alone, so nobody later mistakes them for coverage of that clause.

102 tests pass. Merging once CI is green — thanks for a PR whose description already explained the Windows-clock asymmetry and the Linux-only CI gap; that saved the whole first half of the investigation.

warren618 (approved):

PR #1348, opened 2026-09-04
fix(codex): preserve response model metadata

Shizoqua:
## Summary

- Propagate the provider-reported model from Codex `response.completed` events.
- Preserve response metadata while accumulating streamed `CodexAIMessage` chunks.
- Add regression coverage for model metadata propagation and accumulation.

## Why

The Codex adapter currently drops the model reported by the provider. As a result, `ChatLLM` cannot expose the authoritative response model even though it already supports `response_metadata["model_name"]`.

## Changes

- Capture the Codex response `model` as `response_metadata["model_name"]`.
- Preserve response metadata across `CodexAIMessage` accumulation.
- Add focused tests covering both behaviours.

## Test Plan

- [ ] `pytest agent/tests/test_openai_codex_usage.py -q`
- [x] New regression tests added
- [ ] Existing tests pass

## Checklist

- [x] Changes are limited to the Codex provider and its focused tests.
- [x] No hardcoded credentials or paths.
- [x] No unrelated changes.
- [ ] Full test suite passes.

### DIFF
--- a/agent/src/providers/openai_codex.py
+++ b/agent/src/providers/openai_codex.py
@@ -234,6 +234,8 @@ class CodexAIMessage:
             "finish_reason",
             self.response_metadata.get("finish_reason", "stop"),
         )
+        response_metadata = {**self.response_metadata, **other.response_metadata}
+        response_metadata["finish_reason"] = finish_reason
         reasoning = self.additional_kwargs.get("reasoning_content", "") + other.additional_kwargs.get(
             "reasoning_content", ""
         )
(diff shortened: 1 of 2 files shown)


warren618:
Verified rather than assumed: `chat.py:573` reads `response_metadata.get("model_name") or response_metadata.get("model")`, so there is a real consumer and the key name is the right one. Reverting only `openai_codex.py` while keeping the new tests makes both of them fail, so the coverage genuinely bites, and the wider codex/provider/llm selection is 380 passed / 8 skipped on this branch.

The metadata merge in `__add__` has the right precedence too — later chunks win, and the `response.completed` chunk that carries the model is the last one, so the authoritative value is what survives accumulation.

Note for the record: this touches `src/providers/`, which the project treats as protected. Reviewed on that basis and merging.

warren618 (approved):
Reviewed with independent verification of the premises and mutation checks on the new tests; details in the comment above.

PR #1351, opened 2026-09-04
fix(backtest): route dotted US class shares as us_equity (BRK.B.US, BF.B.US)

cgycorey:
## What

Fixes dotted US class-share tickers (e.g. `BRK.B.US`, `BRK.A.US`, `BF.B.US`) at **both** layers: classification *and* data conversion, with live-API proof for each layer.

### Layer 1 — classification (`_market_hooks.py`)

```python
(re.compile(r"^[A-Z]+\.US$", re.I), "us_equity"),
```

`[A-Z]+` cannot match a dot, so a real class share in the project's canonical `.US` form matched **no** pattern and fell through the catch-all to the `a_share` default:

- `_detect_market("BRK.B.US")` → `a_share` (expected `us_equity`)
- `code_currency("BRK.B.US")` → `CNY` (expected `USD`)
- `_detect_source("BRK.B.US")` → `tushare` (expected the US equity chain)

Observed impact: a composite basket `[BRK.B.US, AAPL.US]` was rejected as a cross-currency set (CNY vs USD), and the symbol was fed to the A-share tencent/mootdx chain, which cannot serve a US listing.

Fix mirrors the directly adjacent ca/india/uk rows, which already admit `&`, `.`, `-` in the ticker portion (`M&M.NS`, `BAJAJ-AUTO.NS`, `SHEL.L`):

```python
(re.compile(r"^[A-Z0-9&.\-]+\.US$", re.I), "us_equity"),
```

### Layer 2 — conversion (found in adversarial review of this PR, live-proven)

After routing, both load-side converters still emitted the **dot** form, which Yahoo does **not** serve. Live-verified against the v8 chart endpoint (`query1.finance.yahoo.com/v8/finance/chart/...`, 5d/1d):

| request | result |
|---|---|
| `BRK-B` (hyphen) | OK — 5 rows, `meta.symbol=BRK-B` |
| `BRK.B` (dot) | **0 rows** |
| `BRK.A` (dot) | **0 rows** |
| `AAPL.US` | 0 rows (suffix stripped by `map_symbol` — consistent with existing behavior) |

So `BRK.B.US` would have failed at fetch even after Layer 1. Both `.US` branches now fold residual ticker dots to hyphens:

- `yahoo_client.map_symbol: BRK.B.US` → `BRK-B`
- `yfinance_loader._to_yfinance_symbol: BRK.B.US` → `BRK-B`

## Tests

- `tests/test_market_detection.py`: six dotted class shares in the `_detect_market` matrix, case-insensitivity, source/currency routing, composite grouping (all red on `main`).
- `tests/test_yahoo_client.py` / `tests/test_yfinance_symbol_map.py`: dot→hyphen conversions plus unchanged neighbors (`AAPL.US` → `AAPL`, crypto `-USDT`, pass-through suffixes).
- In all: 200+ tests across the market-routing and loader suites pass; the four new Layer-1 assertions fail on `main` before the fix.

**Scope note:** the *bare* dotted form (`BRK.B` without `.US`) still classifies as `a_share` — the bare catch-all `^[A-Z]{1,5}$` is letters-only by design, and `.US` is the project's canonical US form (loader docstrings, market-data tests). Fixing the bare form would widen the catch-all; left deliberately out of scope here, happy to follow up if someone depends on bare dotted input.

### DIFF
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -25,7 +25,9 @@ from backtest.models import Position
 _MARKET_PATTERNS = [
     (re.compile(r"^\d{6}\.(SZ|SH|BJ)$", re.I), "a_share"),
     (re.compile(r"^(51|15|56)\d{4}\.(SZ|SH)$", re.I), "a_share"),
-    (re.compile(r"^[A-Z]+\.US$", re.I), "us_equity"),
+    # US equities: tickers may carry a class-share dot (BRK.B.US, BF.B.US)
+    # and a hyphen (e.g. BF-B.US) — same characters as ca/india/uk below.
+    (re.compile(r"^[A-Z0-9&.\-]+\.US$", re.I), "us_equity"),
     (re.compile(r"^\d{3,5}\.HK$", re.I), "hk_equity"),
     # India equities: NSE (RELIANCE.NS) / BSE (500325.BO); tickers may carry
     # '&' and '-' (e.g. M&M.NS, BAJAJ-AUTO.NS).
--- a/agent/backtest/loaders/yahoo_client.py
+++ b/agent/backtest/loaders/yahoo_client.py
@@ -85,7 +85,9 @@ def map_symbol(symbol: str) -> str:
     cleaned = symbol.strip()
     upper = cleaned.upper()
     if upper.endswith(".US"):
-        return cleaned[: -len(".US")]
+        # US class shares are hyphenated on Yahoo (BRK-B): the dot form
+        # returns an empty chart (live-verified), so map BRK.B.US -> BRK-B.
+        return cleaned[: -len(".US")].replace(".", "-")
     if upper.endswith(".HK"):
         base = cleaned[: -len(".HK")]
         digits = base.lstrip("0") or "0"
--- a/agent/backtest/loaders/yfinance_loader.py
+++ b/agent/backtest/loaders/yfinance_loader.py
@@ -62,7 +62,9 @@ def _to_yfinance_symbol(code: str) -> str:
     """
     upper = code.strip().upper()
     if upper.endswith(".US"):
-        return upper[:-3]
+        # US class shares are hyphenated on Yahoo/yfinance (BRK-B): the dot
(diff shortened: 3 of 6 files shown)


warren618:
Verified the Yahoo claim independently rather than trusting the note: `BRK-B` returns 5 rows in USD, `BRK.B` returns HTTP 404, `BF-B` returns 5 rows, `BF.B` returns 0. The hyphen form is required, as stated.

I also checked what the widened pattern newly admits, since it loosens a routing gate: `^[A-Z]+\.US$` → `^[A-Z0-9&.\-]+\.US$` only ever matches strings ending in `.US`, and probing the full table shows every newly-matched string is claimed by `us_equity` at index 2 while ca/uk/india/kr/hk/vn/crypto/futures/forex/index/bare-ticker all keep their own symbols. The only symbols that move are ones that previously fell through to the `a_share` default, which was the bug.

One follow-up, not a blocker: `.US` is stripped in four places and this patch fixes two. `benchmark.py:167` and `correlation.py:47` only classify by suffix so they were already correct, but `fmp_loader.py:72-73` still yields `BRK.B` without hyphenating. FMP sits behind yfinance in the fallback chain so it isn't reached while yfinance succeeds. Do you know whether FMP wants `BRK-B` there? I don't have a key to test it and would rather not guess in the code.

Merging this as-is.

warren618 (approved):
Reviewed with independent verification of the premises and mutation checks on the new tests; details in the comment above.

cgycorey:
Verified everything I could without a key: FMP's docs `demo` key now rejects before symbol evaluation, and no public artifact contains an answer — FMP's own docs use AAPL everywhere, the three wrapper SDKs' test suites (`fmpsdk`, `FinancialModelingPrep.jl`, `fmpapi`) contain no class-share example, and the Wayback index has no archived FMP profile URL with BRK in it. The only signal: FMP's own site URLs code Berkshire as `BRK-B` (`/free-cash-flow/BRK-B`, `/historical-data/brk-b`), and their site runs on their API — so hyphen is likely but unproven. Agreed with leaving `fmp_loader.py` untouched; if a keyed user ever hits it, the fix is the same two-line dot→hyphen fold as `map_symbol`.

Issue #1337, opened 2026-09-02
[Bug] Include tool arguments in market-data deduplication keys


alinv0:
### Description

**What happened:** During a Web UI research run comparing several US tickers, distinct `get_market_data` requests were collapsed by the deduplication layer. Retrying with explicit sources, different symbols, and monthly bars still triggered duplicate-call handling, preventing the requested multi-asset analysis.

**What I expected:** Deduplication should reject only semantically identical calls. Its key should include the tool name and normalized argument payload so requests for different symbols, periods, intervals, or sources execute independently.

### Steps to Reproduce

1. Start the Web UI.
2. Ask the agent to compare one stock with multiple ETFs using current and historical market data.
3. Require separate analysis for each instrument.
4. Observe that later `get_market_data` calls are collapsed despite distinct requested inputs.

### Error Logs

```shell
The ETF batch failed on source detection.
The dedup blocked the calls. Let me use distinct parameters.
The dedup layer is collapsing my market_data calls.
```

### Interface

Web UI (vibe-trading serve)

### LLM Provider

OpenAI-compatible provider

### Version

0.1.10

### Environment

Windows / Docker / Python 3.11

alinv0:
Closing after revalidation. The observed behavior came from a local Docker image created on 2026-07-01. Current upstream already fixed this exact condition in commit 53d790a3 (2026-07-15) by marking MarketDataTool and the other core query tools repeatable. The current default branch therefore does not reproduce the reported defect. I filed this before checking the deployed image age against upstream; apologies for the noise.

Issue #1055, opened 2026-08-10
[Bug] Auto-compaction can silently drop conversation content during summarization


Shizoqua:
### Description

**What happened:**

`_auto_compact()` claims "zero info decay," but content beyond the hard 80,000-character serialization slice can end up in neither the summarization prompt nor the retained tail. No exception is raised, so this is silent information loss.

`AgentLoop._auto_compact()` (`agent/src/agent/loop.py`) does:

conv_text = json.dumps(head, default=str, ensure_ascii=False)[:80000]

`head` is everything not kept in the roughly 20K-token tail (`TAIL_TOKEN_BUDGET = 20_000`). Once `json.dumps(head)` goes past 80,000 characters, everything after that point gets dropped from `conv_text`. It's never seen by the summarization LLM call, and since it's not part of `tail` either, it never gets preserved verbatim.

**What I expected:**

The function's own docstring states the guarantee this should uphold:

`agent/src/agent/loop.py:1809`
`- Iterative update: Nth compression updates previous summary, zero info decay.`

I'd expect compaction to either summarize or verbatim-preserve every message, not silently drop one.

**Scope of what's demonstrated vs. what I'm inferring:**

- Demonstrated directly: the mechanism triggers on the first compaction, not only on 2nd+ compaction as I originally assumed. It fires whenever json.dumps(head) exceeds 80,000 characters. The compaction threshold and head/tail budgeting make this condition reachable during long sessions, while the size estimate (len(content)//4 + 10) doesn't account for serialized JSON overhead.
- Constructed, not captured: this reproduction uses deliberately large synthetic messages to trigger the condition reliably. I haven't measured this against a real captured production trace, so I can't say how often it happens in ordinary usage.
- Reasoned, not measured: on a 2nd+ compaction, head also starts with the previous compaction's injected summary block, which eats further into the same 80K budget. I'd expect this to become more likely with repeated compactions, but I haven't separately measured a worsening trend. The reproduction below shows the mechanism firing cleanly on a single, first compaction, which is enough on its own.

**Duplicate check:**

I searched open and closed issues and PRs for auto_compact, TAIL_TOKEN_BUDGET, conv_text, and related terms and found nothing addressing this specific mechanism.

**Before opening a PR:**

[...]

PR #1352, opened 2026-09-04
fix(loop): size compaction by tool-call arguments, fold cleared-pair payloads

cgycorey:
## What

Context compaction sized and relieved messages by `content` only, so assistant tool-call payloads (empty `content`, large `tool_calls[].function.arguments`) were effectively invisible to the very mechanisms meant to control context growth:

- **Architectural-mismatch quote:** *"`_auto_compact`'s token-budget tail walk counted `len(content)//4 + 10` per message, so a 90 KB arguments blob counted as ~10 tokens and survived in the preserved tail instead of being folded into the LLM summary."*

Two sites:

1. **`_auto_compact` token-budget tail walk** — sized each message by `content` only, so oversized `arguments` dodged the fold and were retained in the "~20K recent tokens" tail, blowing through the budget unboundedly under a heavy tool-calling loop.
2. **`_context_collapse` (zero-cost layer 2)** — folded only long `content` strings; oversized tool-call payloads got no relief short of spending LLM tokens at layer 3.

## How

- `_msg_estimate_chars(msg)` — sizes `content` **plus** every tool-call `arguments` payload.
- `_tail_cut_index(body, budget=TAIL_TOKEN_BUDGET)` — extracted the tail walk into a pure, pair-safe helper, now argument-aware; `_auto_compact` calls it.
- `_context_collapse` — folds a tool-call's huge `arguments` to a valid JSON `"{}"` stub **only when its paired tool result was already `[cleared]`** (the data is gone; the call id/name are preserved so pairing and the model's "I called tool X" memory survive). Provider still receives well-formed `arguments`.

Safety: verified no code re-reads historical `tool_calls[].arguments` for re-dispatch (dedup keys on incoming calls; grounding reads only live call args; `context.py` only writes them).

## Proof

- **Red (origin/main):** `test_tail_cut_index_counts_tool_call_arguments` and `test_context_collapse_stubs_args_of_cleared_paired_call` fail — the 90K arguments blob is untouched.
- **Green (this branch):** 3 new tests pass; 283 related tests across loop/auto-compact/grounding/output-discipline stay green.
- Negative arm `test_context_collapse_keeps_args_when_result_intact_or_pending` guards against over-stubbing (intact or pending calls keep their arguments).

## Related but distinct (not claimed / not duplicated)

- **#1337** (closed) — arguments in **market-data dedup keys**; different mechanism, different failure mode.
- **#1224** (closed) — run.json **output-token usage telemetry** fallback ignoring tool-call arguments; tracing/accounting subsystem, not context compaction.
- **#1055** (closed) — auto-compaction silently **drops** conversation content; the opposite concern.
- **#1341 / #1349** — dedup dead-pointer after microcompact (`_microcompact`/`_called_ok`); they do not touch this tail-walk or layer-2 relief.

## Scope note

`_context_collapse` deliberately does **not** text-fold `arguments` in general (would corrupt the JSON the provider re-validates); it only stubs payloads whose result is already gone. The tail-walk handles general oversized-call folding via the LLM summary at layer 3.

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -63,6 +63,13 @@ COLLAPSE_TEXT_MIN = 2400
 COLLAPSE_HEAD = 900
 COLLAPSE_TAIL = 500
 
+# Placeholder contents that mean a tool result's real data was compacted
+# away. Two sources: ``_microcompact`` clears old results in place to
+# ``_CLEARED_CONTENT``, and ``_fix_tool_pairs`` inserts a ``_STUB_RESULT_CONTENT``
+# stub for a call whose result a layer-3 fold consumed.
+_CLEARED_CONTENT = "[cleared]"
+_STUB_RESULT_CONTENT = "[Result from earlier context — see summary above]"
+
 TAIL_TOKEN_BUDGET = 20_000
 SUMMARY_CHUNK_CHARS = 80_000
 
@@ -446,7 +453,16 @@ def _microcompact(messages: list) -> None:
     for msg in tool_msgs[:-KEEP_RECENT]:
         content = msg.get("content", "")
         if isinstance(content, str) and len(content) > 100:
-            msg["content"] = "[cleared]"
+            msg["content"] = _CLEARED_CONTENT
+
+
+def _result_data_gone(content: Any) -> bool:
+    """True when ``content`` is a placeholder meaning the tool result's real
+    data was compacted away (microcompact's ``[cleared]``, or the stub that
+    ``_fix_tool_pairs`` inserts after a layer-3 fold)."""
+    return isinstance(content, str) and (
+        content == _CLEARED_CONTENT or content == _STUB_RESULT_CONTENT
+    )
 
 
 def _context_collapse(messages: list) -> None:
@@ -464,7 +480,7 @@ def _context_collapse(messages: list) -> None:
         content = msg.get("content")
         if not isinstance(content, str) or len(content) <= COLLAPSE_TEXT_MIN:
             continue
-        if content == "[cleared]":
+        if _result_data_gone(content):
             continue
         head = content[:COLLAPSE_HEAD]
         tail = content[-COLLAPSE_TAIL:]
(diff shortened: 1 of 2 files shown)


warren618:
Rebased onto current main (which now carries #1349 and #1341) and resolved the collision here rather than sending it back.

First, confirming the premise independently, because it is the kind of claim worth measuring: `estimate_tokens` serialises the whole message list, so it counts `tool_calls[].function.arguments`, while the tail-budget loop in `_auto_compact` only looked at `content`. For an assistant tool-call message carrying a 100 KB arguments blob with empty content, the old loop scored it at **10 tokens against an actual 25,010** — a 2,501x undercount, and `TAIL_TOKEN_BUDGET` is 20,000. So the preserved tail could blow the budget by more than itself. Real bug, correctly diagnosed.

The collision was the silent kind and is worth knowing about. This PR introduced `_CLEARED_CONTENT = "[cleared]"` and matched it by equality; #1341 replaced that literal with a marker that embeds the original payload length, so no two cleared results are the same string and an equality test never matches again. Landed as-is on top of #1341, `_result_data_gone` would have returned False for every real cleared result and the arguments-folding this PR exists for would have quietly done nothing — with the suite green, because the fixtures used the literal too.

Resolution:

- `_result_data_gone` now composes `_is_cleared` (prefix match, layer 1) with the `_STUB_RESULT_CONTENT` equality (layer 3) — the union of what either side recognised.
- `_CLEARED_CONTENT` is deleted rather than kept as an alias, since nothing can reference it correctly any more.
- Two fixtures hardcoding `"[cleared]"` move onto `_cleared_text()`. One of them, `test_context_collapse_stubs_args_of_cleared_paired_call`, actually failed — which is how the representation change was caught instead of shipped.

Verified: 62 tests across the three affected files, 12,316 across the loop/agent/compact selection. Two mutations pinned — making `_result_data_gone` equality-match the old literal fails the collapse test, and dropping `arguments` from `_msg_estimate_chars` fails both sizing tests.

One observation you may want to act on separately, not a blocker: the `if _result_data_gone(content): continue` guard inside `_context_collapse`'s loop is unreachable for both placeholder kinds, because the loop only processes content longer than `COLLAPSE_TEXT_MIN` (2400) and neither the cleared marker (~250 chars) nor the stub (~48) ever reaches that. It was already unreachable on main with the 9-char literal, so this PR does not introduce it. `_result_data_gone` is still doing real work in your `cleared_ids` computation — it is only that one guard that is dead.

warren618 (approved):
Token-accounting fix verified independently (2501x undercount reproduced); I pushed the rebase and the cleared-marker reconciliation.

PR #1356, opened 2026-09-05
fix(agent): parse cash-dividend journal rows and book them in shadow real PnL

he-yufeng:
## Summary

- Broker journals carrying cash-dividend rows (同花顺 红利入账, 富途 `Side=Dividend`) no longer crash the trade-journal parsers: they normalize to `side="dividend"` records whose existing `amount` field carries the cash received.
- The shadow backtest now adds that cash to real PnL, but only for symbols its adjusted frames do not cover; for covered symbols #1311 already embeds the dividend in the adjusted price caliber, so adding cash there would double count it.

## Why

One dividend row in an export makes `_normalize_side` raise `Unsupported trade side`, which fails the whole journal parse. In the shadow chain, `_attribution_or_zero` catches that and silently reports an all-zero attribution, exactly the "successful backtest recorded as PnL 0" flavor of #1207 item 15. This PR closes the cash-dividend half of that item; the split half landed with #1311.

## Changes

- `agent/src/tools/trade_journal_parsers.py`: `_normalize_side` accepts dividend tokens (红利入账, 分红, 派息, 股息, 现金分红, dividend, cash dividend, ...) and returns `"dividend"`. Known non-trade corporate actions (红股入账, 转股入账, 配股缴款, 配股上市, 利息归本, 利息, plus English stock split / bonus / rights / interest labels) are dropped without raising: splits on covered symbols are already restated by the #1311 frame adjustment and bonus shares are not modeled. Truly unknown sides still raise.
- `agent/src/tools/trade_journal_tool.py`: `pair_trades_fifo` skips non-buy/sell rows explicitly, and the profile and behavior stats (trade counts, market and hourly distributions, overtrading daily counts, anchoring price stats) are computed on trade rows only. A new `total_dividends` profile field sums the cash and is folded into `total_pnl`; journals without dividend rows get byte-identical numbers.
- `agent/src/shadow_account/backtester.py`: `_attribution_or_zero` sums journal dividend cash excluding frame-covered symbols and filtering by pool currency the same way roundtrips are filtered, then passes it as `extra_real_pnl` into `_compute_attribution`, where it joins `real_pnl` so the `missed` residual balances. `AttributionBreakdown` and the HTML report are untouched.

## Test Plan

New coverage: dividend tokens parse with the right cash amount for 同花顺 and Futu, bonus-share rows are skipped without raising, `hold` still raises, dividend rows never open or close FIFO positions, profile totals separate dividends from trade counts, and the backtester adds dividend cash for uncovered symbols while skipping covered ones.

Verification (venv: Python 3.12, pandas 3.0.5, pytest 9.1.1):

- `cd agent && python -m pytest tests/test_trade_journal.py tests/test_trade_journal_fifo.py tests/test_shadow_account.py -q` → 210 passed.
- Red direction: with only the three source files stashed and the new tests kept, the same command gives 27 failed / 183 passed, and the shadow tests log the reported crash (`Unsupported trade side: '红利入账'`). After `git stash pop`, 210 passed again.
- Full suite (`python -m pytest tests -q` from `agent/`) → 12079 passed, about 135 skipped; the 126 to 127 failures (one test flakes between runs) sit in unrelated areas (MCP OAuth, codex token cache, rebalance masks, signal alignment) and the failure set is byte-identical to a clean-main run in this environment.

## Risk and rollback

Parser-only extension with no signature changes to public tool output beyond one additive profile key. Unknown sides still raise, so previously rejected journals stay rejected, and dividend-free journals are numerically unchanged. Revert this PR to restore the old behavior.

## Checklist

Prepared with AI assistance (Kimi K3); I verified the change with the red and green pytest runs listed above, including the full agent test suite.

### DIFF
--- a/agent/src/shadow_account/backtester.py
+++ b/agent/src/shadow_account/backtester.py
@@ -284,6 +284,7 @@ def run_shadow_backtest(
         initial_capital=initial_capital,
         pool_currency=headline_currency,
         adjust=build_frame_adjust(adjust_frames) if adjust_frames else None,
+        covered_symbols=frozenset(adjust_frames),
     )
 
     result = ShadowBacktestResult(
@@ -538,6 +539,7 @@ def _attribution_or_zero(
     initial_capital: float,
     pool_currency: str | None = None,
     adjust=None,
+    covered_symbols: frozenset[str] = frozenset(),
 ) -> tuple[AttributionBreakdown, float | None, float]:
     """Compute attribution if the journal is available, else return zeros."""
     shadow_pnl = _shadow_pnl_from_metrics(combined, initial_capital)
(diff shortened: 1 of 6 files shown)


he-yufeng:
A gentle nudge on this one: it is the last open item of #1207 (item 15's cash-dividend half), and CI is green. Happy to adjust anything if you want a different shape for the profile or attribution surface.

warren618:
Merged with one follow-up commit on top. `_compute_profile` kept dividend rows out of the trade count, the market/hourly distributions and the behaviour diagnostics, but `top_symbols` still grouped the full frame — a 500-yuan 红利入账 row counted as a third trade on 600519 and inflated its turnover. Fixed and pinned next to the other exclusions.

One thing I could not verify: the 同花顺 fixture puts the dividend cash in `成交金额` with `成交数量 = 0`. I do not have a real 同花顺 export with a 红利入账 row. If yours carries the cash in a different column (e.g. `发生金额`) the parser reads 0 and the dividend books as 0 silently — worth checking against a real file if you have one.

PR #1350, opened 2026-09-04
fix(loop): release dedup lock when microcompact clears a tool result (#1343)

cgycorey:
## What

Fixes #1343's main defect (the tool-whitelist half already shipped via #1345).

**Root cause (two correct mechanisms, fatal combination):**
1. **Layer-1 microcompact** (`loop.py::_microcompact`) silently replaces old tool-result contents with `[cleared]` under token pressure (`KEEP_RECENT=3` results preserved).
2. **Dedup guard** (`loop.py` ~L1942): a non-repeatable tool that already succeeded (`_called_ok`) is blocked on re-call with `"already completed successfully. Use the previous result."`

After either compaction layer destroys a result, that pointer is dead — the model can never see the data, so it retries the same tool, and every retry is skipped. Live trace evidence (issue #1343, 2026-09-04 14:56 `investment_committee` run): **44 wasted `tool_skipped` calls** across 5 fundamental-data tools, while the agent correctly refused to fabricate the figures.

## The invariant (applied at both compaction sites)

> **Dedup may only point at results the model can still see.**

- **Layer 1 (microcompact):** `_microcompact` now takes the loop's `_called_ok` set and releases the lock for every tool whose result it clears. Intact results (the `KEEP_RECENT` recent window) keep their lock.
- **Layer 3 (auto-compact):** `_auto_compact` reconstructs messages keeping only a ~20K-token tail; everything else survives solely inside the LLM summary. Reconcile `self._called_ok` after `_fix_tool_pairs`: release every lock whose tool no longer has a real, visible result (excludes `[cleared]` and the `_COMPACT_STUB_CONTENT` placeholder — now a single shared constant so the two sites never drift). Locks for results that survived in the tail are kept.
- **Swarm workers** (`worker.py`) have their own microcompact but **no name-dedup**, so they cannot dead-lock — re-execution there costs a redundant API call at worst. Out of scope by design.
- **Layer 2 (`_context_collapse`)** truncates the middle but keeps head+tail verbatim — the pointer retains value, locks stay.

## Why not other shapes
- *Don't clear results for non-repeatable tools:* those results (financial statements, fund flow) are exactly the largest payloads — skipping them defeats compaction under pressure.
- *Check visibility inside the hot guard loop:* needs a per-call-id reverse lookup on every tool call; releasing/reconciling at compaction time is simpler and equally sound (re-execution after losing data is exactly what we want).
[...]

PR #1358, opened 2026-09-05
fix(agent): reconcile compacted results by exact call identity

saju01:
## Summary

Follow-up to the merged #1341 / #1349 integration and the compaction half discussed in #1350. Two remaining stale-lock cases are independent of #1268's in-flight reservation / explicit invocation-policy work.

### Cleared argument variant is still blocked

After successful calls for `statement="income"` and `statement="balance"`, microcompaction can clear income while retaining balance. The success gate correctly compares `(name, canonical arguments)`, but the previous unblock compared only tool names and waited for *every* variant to disappear. Retrying income therefore still returned `already completed successfully` although its own result was gone.

### Full compaction does not reconcile success locks

`_auto_compact` can summarize an old tool/result pair out of the message history, leaving an exact-call success lock behind. A summary/stub is not the original readable result; a later retry remained blocked. This also affects degraded summarization.

## Required behavior

- Use the existing `_identical_call_key` identity, not another canonicalizer.
- Release a read-only call's lock only when compaction removes its last readable successful result.
- A surviving different argument variant must not keep that lock alive.
- A surviving successful result for the *same* key must preserve the lock.
- Synthetic skips, errors, cleared markers and summary stubs must not count as successful data.
- Keep side-effecting calls locked: losing their output is not authorization to repeat a mutation.
- Apply the same reconciliation after microcompaction and full compaction.

No change to the token threshold, repeatable-tool policy, broker authorization, grounding gate, or parallel-batch invocation policy. No live LLM, market-data, broker, or deployment call is needed for the tests.

## Verification

On upstream runtime `5c2766f`, Python 3.11, with offline test doubles:

- Baseline regression/safety selection: **430 passed**.
- The new real-gate regressions against the original `loop.py`: **3 failed** — microcompaction plus successful/degraded full-compaction cases all stop at two executions and log `Blocked duplicate call: get_financial_statements` on the required third call.
- With this fix: **18 new tests passed**; the combined regression/safety selection is **448 passed**.
- The selection includes every `test_agent_loop*.py`, loop helpers, chunked auto-compaction, percentage-point and language-parity grounding, grounding/output discipline, mandate enforcement, kill-switch and read-only-default tests.
- Ruff on source and new tests, Black on the new test file, and `git diff --check`: passed.

The new tests also pin surviving exact duplicates, independently cleared variants, synthetic skip/error/stub results, layer-2 argument stubbing, canonical `run_dir` equivalence, `None` canonical keys, deterministic cache recovery, and both compaction paths preserving mutation locks. They drive `_process_tool_calls`, not just a helper's return value.

The original-runtime RED check loads the unmodified source into a separate pytest process; no bad code is written into the branch. Full backend/frontend suites were not run locally; GitHub CI is separate.

## Implementation

Record successful call-ID → canonical-key provenance when a tool succeeds or a deterministic result is replayed. This survives layer-2 argument stubbing without reparsing historical arguments or introducing a second canonicalizer. Compare readable successful keys before/after each compaction; remove only lost, known-read-only keys from the success gate. Synthetic skipped/failed calls are never recorded, and cleared/summary-stub contents are excluded. A cache replay restores the exact-call gate once its data is readable again. The call-ID ledger resets for every run.

## Risk / rollback

Touches the protected agent loop; maintainer review required. The intended new execution path is a read-only re-fetch after its exact prior result becomes unreadable, not unlimited repeated execution. Revert the follow-up commit to restore the prior behavior. Production deployment is out of scope.

### DIFF
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -468,10 +468,9 @@ def _microcompact(messages: list) -> list:
         messages: Message list (mutated in place).
 
     Returns:
-        Names of tools whose every result just became unreadable. The caller
-        drops these from the dedup ledger: blocking a re-call with "use the
-        previous result" only makes sense while that result is still in
-        context, and a cleared result is not.
+        Names of tools whose every result just became unreadable (legacy
+        helper contract). The loop reconciles its dedup ledger separately by
+        exact successful call identity, not by these tool names.
     """
     tool_msgs = [m for m in messages if m.get("role") == "tool"]
     if len(tool_msgs) <= KEEP_RECENT:
@@ -1074,6 +1073,9 @@ class AgentLoop:
         # deterministic cache uses, so the block path and the cache path can
         # never disagree about what 'the same call' means.
         self._called_ok: set[tuple[str, str]] = set()
+        # Capture successful identities before context collapse can stub args.
+        # Skipped/error call IDs never enter this ledger.
+        self._successful_call_keys: dict[str, tuple[str, str]] = {}
         self._cancel_event = threading.Event()
         self._previous_summary: str = ""
         self._persistent_memory = persistent_memory
@@ -1179,6 +1181,7 @@ class AgentLoop:
         else:
             self._has_run = True
         self._called_ok = set()
+        self._successful_call_keys = {}
         self._previous_summary = ""
         self._released_fallback = False
         self._released_fallback_reason = None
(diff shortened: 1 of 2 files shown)


he-yufeng (approved):
Approved after running this locally. The exact-identity ledger is the right refinement over the name-level unblock from #1349: keying `_successful_call_keys` by tool_call_id means a surviving result for one argument variant no longer keeps every other variant gated, and the readonly filter closes the hole where a cleared mutating-tool result could be re-executed. The per-run reset and the full-compaction path (capture the readable set before reconstruction, reconcile after `_fix_tool_pairs`) both check out, and having the cache-hit path record identity too keeps the deterministic cache and the block path agreeing on what the same call means.

Verified on this branch: the 18 new tests pass, and against main's loop.py the 13 fix-pinning ones fail while the 5 behavior tests pass. The wider loop/compact/dedup slice (233 tests) is green.

PR #1363, opened 2026-09-06
fix(agent): stop repeated discovery without new evidence

### LINKED ISSUE #1353: [Bug] Agent loops indefinitely trying to locate its own run_dir artifacts instead of recovering, when asked a follow-up after a session restart
### Description

**What happened:**
After a backtest completed successfully and the agent answered a question about it, the user sent a follow-up question in the same chat session ("what caused the losses?"). The backend process had to be restarted in between (unrelated crash). On resuming, the agent tried to locate the artifacts (`trades.csv`, `run_card.json`) from the *original* backtest run to answer with real data instead of from memory (this part is correct behavior — it explicitly avoids answering ungrounded). However, it could not resolve the original `run_dir` path, and instead of falling back to a clean recovery path, it entered a long loop of semantic-search / file-read attempts that never converged:

- ~35 iterations spent purely on path discovery, with no forward progress.
- Several tool calls were retried with byte-for-byte identical arguments that had already failed (e.g. attempting to read `sessions/<id>/transcript_<timestamp>.jsonl` twice with the exact same guessed timestamp, both times rejected with "outside allowed document roots").
- The agent correctly guessed the *directory name pattern* for its own memory files (`/home/opc/.vibe-trading/memories/project_<slug>.md`) but the read tool rejected it as outside the allowed document roots, with no readable equivalent pointed to.
- Attempting to list `~/.vibe-trading/runs` as a directory failed with `"Not a file: ..."` — the file-read tool has no directory-listing fallback, so the agent had no way to discover run_dir names once semantic search stopped surfacing them.
- The session-search tool kept returning snippets of its own previous answer (which contains the final metrics in prose) rather than anything with the actual `run_dir` path, so no amount of query rephrasing converged on the missing piece.

**What I expected:**
Either (a) the agent recognizes after a few failed attempts that path discovery isn't converging and switches strategy (ask the user, or transparently re-run the backtest instead of continuing to search), or (b) the tool surface gives it a way to actually resolve this: list `runs/` as a directory, or have session/turn metadata carry forward the `run_dir` of the backtest that was just discussed in the same conversation.

Note: the existing stall-watchdog (added in a prior fix, ~30 min no-progress timeout) does not catch this — the agent was actively issuing tool calls the whole time, just never converging semantically. This is a different failure mode from a hang.

This is notable because Vibe-Trading is designed to run unsupervised. In this instance an external operator noticed the stall, manually found the correct `run_dir` on disk, cancelled the in-flight loop via `POST /sessions/{id}/cancel`, and re-prompted the agent with the exact file path — at which point it answered correctly and fast (2 iterations). Without that intervention, the loop would have continued until it hit a hard iteration/token cap, or (based on this agent's own prior behavior in the same project) fallen back to answering from memory without re-reading real data — a grounding failure, not just an efficiency one.

### Steps to Reproduce

1. Start a chat session, ask the agent to design + backtest a strategy (multi-instrument, e.g. 6 crypto pairs) with full artifacts written to `runs/<run_id>/artifacts/`.
2. Let it finish and answer with the results.
3. Restart the `vibe-trading serve` process (simulating any interruption between turns).
4. In the same chat session, ask a follow-up that requires re-reading the original run's artifacts (e.g. "what caused the losing trades?").
5. Observe the agent try to relocate the original `run_dir` via `session_search` / memory recall / guessed file paths, without success, looping on repeated and sometimes identical failed calls instead of stopping to change strategy.

### Error Logs

```shell
{"status": "error", "error": "File not found or path escapes workspace: artifacts/trades.csv"}
{"status": "error", "error": "Path '/home/opc/.vibe-trading/sessions/<session_id>/transcript_<ts>.jsonl' is outside allowed document roots.\nAllowed roots:\n  - /home/opc/Vibe-Trading/agent/uploads\n  - /home/opc/Vibe-Trading/agent/runs\n  - ..."}
{"status": "error", "error": "Path '/home/opc/.vibe-trading/memories/project_<slug>.md' is outside allowed document roots. ..."}
{"status": "error", "error": "Not a file: /home/opc/.vibe-trading/runs"}
```

### Interface

CLI (vibe-trading)

### LLM Provider

deepseek

### Version

0.1.14

### Environment

Oracle Linux Server 9.8 / Python 3.11.16 (aarch64)

be-student:
## Summary

Stop repeated discovery after eight consecutive tool iterations without new successful observations, and return a visible recovery request for an artifact path or an explicit rerun.

## Why

Closes #1353. Implements the progress budget and exact failed-call suppression requested in the maintainer discussion. Tool activity currently resets the watchdog even when every attempt fails or returns the same snippets, so missing artifacts can consume the full iteration budget.

## Changes

- Track executed observations separately from activity. Read-only results are compared after JSON normalization; distinct successful mutations count as progress even with identical status payloads.
- Suppress identical previously failed calls using the existing canonical call identity. Changed arguments can execute. A successful mutation clears failed identities so repaired inputs can be retried; all progress state resets per run. Synthetic authorization denials do not poison this ledger.
- Emit and persist a recovery answer, trace event, and failed terminal state. Old metrics do not turn this exit into success. The response does not answer from memory or automatically rerun the backtest.
- Document the behavior and update the old deterministic-cache test's retry expectation. Failure suppression is separate from successful result caching.

Scoped directory listing remains optional follow-up work. This does not implement in-flight reservations or change compaction's successful-call reconciliation (#1268 / #1358).

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`) — full suite not run locally.
- [ ] Tested manually (describe below) — no live provider session used.

Python 3.11, isolated editable development install:

```sh
pytest agent/tests/test_agent_loop*.py agent/tests/test_loop_helpers.py agent/tests/test_mandate_enforcement.py agent/tests/test_killswitch_blocks_orders.py agent/tests/test_readonly_default.py -q --tb=short
```

153 passed. The two initial real-loop regressions failed against the original implementation (24 iterations instead of 8). New tests cover repeated failures, changed arguments, repeated snippets, productive reads/writes, per-run reset, successful repair, synthetic authorization denials, persisted failure state, trace records, streaming output, and existing metrics. Ruff passed on the four changed Python files; Black formatted new files and changed ranges in the existing files. `git diff --check` passed.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — agent-loop scope explicitly discussed in #1353.
- [x] No hardcoded values (API keys, file paths, magic numbers) — named eight-iteration budget follows issue discussion.

## Risk and rollback

Touches the protected agent loop and intentionally stops repeated unchanged observations, including repeated polling with unchanged output. The ledger compares normalized tool output, not semantic meaning; outputs containing changing metadata may still appear new. Exact failed calls are not retried until a successful mutation or a new run. No broker orders, provider credentials, deployment, or persisted-state schema changed. Revert this commit to restore prior retry behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,15 @@ Their credentials go to the OS keyring (macOS Keychain, Windows Credential Manag
 
 Most runs follow the same evidence path: route the request, load the right market context, execute tools, validate outputs, and keep the artifacts inspectable.
 
+The agent stops with a visible recovery request after eight consecutive tool-call
+iterations without a new successful observation. Repeating the same read result
+or a failed call does not reset this budget. Identical failed calls are skipped
+until a successful state-changing tool executes or a new run starts; changed
+arguments can still execute. The recovery asks for an artifact path or permission
+to rerun the missing research step, and the run remains failed even if older
+metrics exist. This guard is separate from the wall-clock activity watchdog.
+
+
 | Layer | What happens |
 |-------|--------------|
 | **Plan** | Selects the relevant finance skills, tools, data sources, and swarm preset when useful. |
--- a/agent/src/agent/loop.py
+++ b/agent/src/agent/loop.py
@@ -31,6 +31,7 @@ from src.agent.context import ContextBuilder
 from src.agent.grounding import GroundingLedger
 from src.agent.memory import WorkspaceMemory
 from src.agent.progress import HeartbeatTimer, ProgressEvent, _set_emitter
+from src.agent.tool_progress import RECOVERY_MESSAGE, ToolProgress
 from src.agent.tools import ToolRegistry
 from src.agent.trace import TraceWriter
 from src.core.state import RunStateStore
@@ -840,8 +841,6 @@ def _looks_like_tool_call_syntax(content: str) -> bool:
     )
 
 
-
-
 # Task target-file detection: a user message usually names the file to
 # create or update ("update C:\\...\\plan.md"). If a run approaches its
 # iteration cap without having written that file, the loop must remind the
@@ -1093,6 +1092,7 @@ class AgentLoop:
         # 5-9x each (2026-08-20 INTC run) because it could no longer see its
         # own verification records.
         self._called_identical: dict[tuple[str, str], str] = {}
+        self._tool_progress = ToolProgress()
 
     def cancel(self) -> None:
         """Cancel the current loop.
@@ -1187,6 +1187,7 @@ class AgentLoop:
         self._last_activity_wall = _time.time()
         self._run_done = threading.Event()
         self._called_identical = {}
+        self._tool_progress = ToolProgress()
         run_started_wall = _time.time()
 
         state_store = RunStateStore()
@@ -1246,6 +1247,7 @@ class AgentLoop:
 
         iteration = 0
         final_content = ""
+        no_progress_reason: str | None = None
         content_filter_count = 0
         consecutive_content_filter_count = 0
         content_filter_circuit_breaker = False
@@ -1816,6 +1818,31 @@ class AgentLoop:
                     response.tool_calls, context, messages, trace, react_trace, current_iter,
                 )
 
+                if (
+                    not self._cancel_event.is_set()
+                    and self._tool_progress.finish_iteration()
+                ):
+                    no_progress_reason = "no_progress: " + RECOVERY_MESSAGE
+                    final_content = RECOVERY_MESSAGE
+                    trace.write(
+                        {
+                            "type": "no_progress",
+                            "iter": current_iter,
+                            "iterations_without_progress": self._tool_progress.stalled_iterations,
+                        }
+                    )
+                    trace.write_text_entry(
+                        {"type": "answer", "iter": current_iter},
+                        field="content",
+                        value=final_content,
+                        offload_kind=f"answer-{current_iter}",
+                    )
+                    react_trace.append({"type": "answer", "content": final_content})
+                    self._emit(
+                        "text_delta", {"delta": final_content, "iter": current_iter}
+                    )
+                    break
+
                 # Layer 3: compress after all tools have executed
                 if compact_requested:
                     logger.info("Manual compact triggered by model")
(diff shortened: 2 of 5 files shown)


warren618:
The progress budget is the right shape for #1353 and the eight-iteration exit with a visible recovery message is exactly what I asked for there. One blocker before it can land, in the other half of the change.

**An exact-repeat FAILURE block makes a transient failure permanent for the run.** `ToolProgress.record` adds every unsuccessful call's identity to `self.failed`, and `_process_tool_calls` then hard-refuses that identity for the rest of the run. Only a successful *mutating* call clears the set — and a research run often has no mutating call at all. So a read that fails once for a reason that has nothing to do with its arguments (rate limit, network blip, the 1800s tool timeout) can never be retried.

Measured with an offline double: a read-only tool that returns `{"status":"error","error":"rate limited, retry later"}` on call 1 and succeeds on call 2, driven through `AgentLoop.run` with the model re-issuing the identical call.

```
main       tool executed 2 times  -> the retry succeeds, the model gets its data
this PR    tool executed 1 time   -> the retry is refused, the data is never fetched
```

I checked the instrument both ways: the probe passes on main and fails here, so this is the branch's behaviour, not a broken harness. None of the seven new tests covers it — they all use a tool that fails deterministically for the same reason every time, which is the case the block is designed for.

The suppression is doing real work for #1353 and I don't want to lose it. What I would land: block on the *second* identical failure rather than the first. The #1353 loop issued the same rejected path ~35 times, so a budget of two still kills it, while one honest retry after a transient error survives. Alternatively, keep the block at one but let the no-progress budget alone carry #1353 — the eight-iteration exit already terminates that run.

Whichever you pick, please add the transient case as a test: fail once, succeed on the identical retry, assert the tool actually ran twice. Everything else here — the observation ledger, the per-run reset, the visible recovery answer, the failed terminal state, not answering from memory — I am happy with.

he-yufeng (commented):
Read the full diff. The design is sound: the ledger counts a new observation only when a read-only call returns content whose hash has not been seen this run, so the eight-iteration budget targets exactly the stuck pattern (repeated identical reads or repeated failures) and leaves long-but-productive research alone, where every read returns fresh content.

Two things I verified before commenting:

- The failed-call skip composes correctly with the existing `_called_identical` success cache: they sit at different points (success replay vs failure skip) and the updated `test_failed_deterministic_call_is_not_cached` pins the new semantics, including that changed arguments still execute.
- `record()` is only called on real executed results, so synthetic skip responses cannot masquerade as new evidence. That is the right choice, and the docstring says so.

One boundary worth a second look: `record()` is gated on `update_memory`. If any execution path produces a genuine fresh observation with `update_memory=False`, the stalled counter would keep climbing despite real progress. I did not find such a path in the current call sites, but it is an invariant future refactors could silently break; a one-line comment at the call site would protect it.

CI is green across the board. Good to merge from my side once the note above is considered. Prepared with AI assistance (Kimi); I verified the invariants myself.

PR #1370, opened 2026-09-06
fix(backtest): sync sub-engine active symbol before every composite dispatch

Shizoqua:
## Summary

- Sync the shared sub-engine's active symbol before every composite dispatch, not just apply_slippage.
- Fix the forex swap calculation to use metal-aware lot size in composite mode.
- Add regression coverage for both.

## Why

round_size and calc_commission delegate to a shared sub-engine instance per market, but only apply_slippage synced its _active_symbol first. A symbol processed right after a different one on the same market kept using whichever symbol last synced it, which bites in the capital-fit rescale path.

## Changes

- Sync sub._active_symbol in round_size and calc_commission, matching apply_slippage.
- Route on_bar's forex swap lot size through the same metal-aware lookup ForexEngine already uses.

## Test Plan

- [x] New regression tests fail on main, pass with the fix
- [x] pytest agent/tests/test_composite_engine_fallback.py -q (8 passed)
- [x] Broader composite/forex/futures test suite (136 passed)

## Checklist

- [x] Changes limited to composite engine and its tests
- [x] No hardcoded credentials or paths
- [x] No unrelated changes

### DIFF
--- a/agent/backtest/engines/composite.py
+++ b/agent/backtest/engines/composite.py
@@ -18,6 +18,7 @@ from backtest.engines._market_hooks import (
     _interval_span_hours,
     _is_china_futures,
     _liquidation_mark,
+    _normalize_symbol,
     code_currency,
     calc_crypto_funding_fee,
     check_crypto_liquidation,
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Read the full diff and verified the two fixes against current main.

The shared-instance leak is real: `round_size` and `calc_commission` dispatched through one sub-engine per market without refreshing its `_active_symbol`, so a symbol only got the right lot grid / fee schedule if `apply_slippage` happened to run first. Syncing on every dispatch matches what `apply_slippage` already did, and the two regression tests pin it with real fee differences (gold oz vs FX units, au's flat RMB10 vs IF's rate formula). The metal-aware swap sizing is correct too: `_lot_units` on the normalized symbol is what ForexEngine itself uses, and the 100 oz = 1 gold lot test would have read 0.001 standard lots before.

Verified the imports exist where the diff says they do (`_normalize_symbol` in `_market_hooks.py`, `_lot_units` in `forex.py`) and CI is green.

One optional thought, not blocking: the `sub._active_symbol = ...` assignment now appears in three dispatchers. A tiny setter on the sub-engine would keep the invariant in one place, but the current shape matches the existing precedent, so fine as is.

Prepared with AI assistance (Kimi); I verified the mechanism against main myself.

PR #1377, opened 2026-09-06
fix(registry): enforce the NaN contract on declared dependencies (#1376)

### LINKED ISSUE #1376: [Bug] 82 alphas fabricate a value on a bar whose own inputs are missing — enforce the NaN contract in the registry, not per alpha
## What is wrong

`agent/src/factors/base.py` states the contract in its first paragraph: *"raw scores, NaN preserved in warmup / missing data"*, and *"every operator propagates NaN; no silent `fillna(0)`"*. The operators honour it. A number of alphas do not, because they pick a branch with `np.where(...)` / a ternary helper on a **comparison**: a NaN comparison evaluates to `False`, not NaN, so a bar whose inputs are missing silently falls through to the branch's hard-coded constant — a real-looking `+1`, `-1` or `0.0` where the correct answer is "unknown".

This is not theoretical. `min_warmup_bars` is metadata only — nothing in the compute path trims those rows — and `factor_analysis_core` excludes missing values with `dropna()`. So a fabricated `±1` is consumed as a genuine signal in IC and downstream, while a NaN would have been correctly dropped.

#1371 / #1372 / #1373 / #1374 fixed four of these one alpha at a time (`gtja191_003/004/059/069`), and #1172 fixed the forward-fill variant earlier. That rate will not converge.

## Measurement

Behavioural scan, not a grep: build a clean 260-bar × 5-symbol panel, blank **every input column for one symbol on one bar**, run every registered alpha, and check that symbol's output on that bar. Under the contract it must be NaN.

On `a113ee0c` (i.e. with #1371–#1374 already merged):

```
runnable=429  not_runnable_on_this_panel=33
still prints a number on the missing bar = 82
```

Breakdown: `gtja191` 52, `alpha101` 15, `qlib158` 15 — the whole `qlib158_cnt{p,n,d}{5,10,20,30,60}` family is in there, which suggests one shared helper rather than 15 independent mistakes.

Note the criterion deliberately is *not* "emits a value before `min_warmup_bars`". `min_warmup_bars` is a conservative declaration, not a claim that those rows must be NaN, and measuring it that way inflates the count to 159. The number above is the defensible one: the bar's own inputs were absent.

## Proposal

Enforce the contract once, in `Registry.compute` (`agent/src/factors/registry.py:355`, next to the existing `_validate_output`) instead of in 82 alpha files: after `compute_fn(panel)` returns, mask each output cell whose alpha's own `columns_required` (and `extras_required`) are NaN at that same (date, symbol).

Why this is the right layer:

- It is the same source of truth the registry already uses to *reject* a panel for missing columns, so it cannot drift from the alpha's declaration.
- It is per-cell, so a gap in one symbol does not poison another — a cross-sectional `rank` over the remaining symbols is unaffected.
- It is strictly a narrowing: it can only turn a printed number into NaN, never the reverse.

## What a PR needs to show

- **Both sides of the guard.** A panel with a gap where the masked cells must become NaN, *and* a clean panel where the output must be bit-identical to today for every alpha. The second half is the one that matters: this touches every factor, so "no real value changed" has to be demonstrated, not assumed.
- The 82 alphas above go to 0 on the scan, and the count for a clean panel stays 0.
- Whether any alpha legitimately produces a value from a bar whose own inputs are missing. I do not believe there is one, but if it exists it needs to be named and exempted explicitly rather than discovered later as a behaviour change.
- Mutation check: remove the mask, the new tests must fail.

## Out of scope

Warmup semantics. Whether `min_warmup_bars` should also be enforced (and whether some declarations are simply too conservative) is a separate question — this issue is only about bars whose inputs are actually absent.

cgycorey:
## What

Closes #1376. `Registry.compute` now enforces the NaN contract stated in `base.py` ("raw scores, NaN preserved in warmup / missing data"): the output is masked to NaN wherever any **declared dependency** — `columns_required`, `extras_required`, or `sector` when `requires_sector` — is missing for that bar/symbol.

One registry-level rule replaces per-alpha fixes. Declared metadata is the alpha's contract, and the registry is the single choke point every runner (bench, bench_strict, analysis core) already goes through.

## Why masking, not raising

Detecting "fabricated" centrally is impossible — a number computed from a NaN comparison is indistinguishable from a legitimate one. The only central enforcement is to make the declared inputs authoritative: a bar whose inputs are missing has an unknown answer, so the output there is NaN. This matches the issue's measurement exactly.

## Ordering detail

The mask runs **after** the return-type check: a non-DataFrame return still fails with the existing `expected DataFrame` `RegistryError` (a first cut applied the mask first and crashed with `AttributeError` on that path — caught by `test_registry_compute_wrong_return_type`, fixed with an isinstance guard).

## Measurement

Behavioural scan per the issue's spec (260×5 panel, blank every declared input for one symbol on one bar, output there must be NaN):

- **before: 73 fabricators** (subset of the issue's 82 — this panel lacks `fund:` columns, hence the different denominator)
- **after: 0 fabricators**; 354/354 runnable alphas return NaN on the missing bar

Full factor suite: 1079 passed, 5 skipped.

## Known consequence, stated on purpose

On panels where a declared column is mostly NaN, masked outputs can now cross the existing `>95% NaN` sanity check and fail with a loud `RegistryError` instead of returning fabricated signals. `bench_runner_strict` already catches `RegistryError` and skips the alpha with a reason, so the blast radius is a clean skip rather than silent signal poisoning.

## Skipped on purpose

- Per-alpha rewrites of the 82: the registry mask covers them all; the merged per-alpha fixes (#1371–#1374) stay valid, nothing to revert.
- Sector-tag NaN masking: sector is a per-bar tag, not a measured input; the issue's scan blanks input columns only. Say the word if tags should mask too.

### DIFF
--- a/agent/src/factors/registry.py
+++ b/agent/src/factors/registry.py
@@ -354,7 +354,14 @@ class Registry:
 
         # Enforce NaN contract: any bar where a declared dependency is missing
         # must produce NaN in the output, regardless of the alpha's internal logic.
-        ref = panel.get("close")
+        # The mask's shape reference is the first declared dependency, not a
+        # hardcoded "close": 78 alphas declare columns_required without close,
+        # and the mask must still run for them. Pre-checks above guarantee every
+        # declared dependency is present in the panel.
+        deps = list(meta.get("columns_required", [])) + list(meta.get("extras_required", []))
+        if meta.get("requires_sector"):
+            deps.append("sector")
+        ref = next((panel[c] for c in deps if c in panel), None)
         if ref is not None:
             mask = pd.DataFrame(True, index=ref.index, columns=ref.columns)
             for col in meta.get("columns_required", []):
--- a/agent/src/factors/registry.py
+++ b/agent/src/factors/registry.py
@@ -352,6 +352,19 @@ class Registry:
         except Exception as exc:  # noqa: BLE001 — isolate compute failure
             raise RegistryError(f"{alpha_id}: compute() raised: {exc}") from exc
 
+        # Enforce NaN contract: any bar where a declared dependency is missing
+        # must produce NaN in the output, regardless of the alpha's internal logic.
+        ref = panel.get("close")
+        if ref is not None:
+            mask = pd.DataFrame(True, index=ref.index, columns=ref.columns)
+            for col in meta.get("columns_required", []):
+                mask = mask & panel[col].notna()
+            for col in meta.get("extras_required", []):
+                mask = mask & panel[col].notna()
+            if meta.get("requires_sector"):
+                mask = mask & panel["sector"].notna()
+            if isinstance(result, pd.DataFrame):
+                result = result.where(mask)
         return self._validate_output(alpha_id, result, panel)
 
     def _load_module(self, alpha: Alpha) -> ModuleType:
--- a/agent/tests/factors/test_registry.py
+++ b/agent/tests/factors/test_registry.py
@@ -254,3 +254,17 @@ def test_export_manifest_shape(mini_zoo: Path) -> None:
     assert "generated_at" in m
     assert m["zoos"][0]["zoo_id"] == "fakezoo"
     assert len(m["zoos"][0]["alphas"]) == 4
+
+
+def test_registry_masks_nan_on_declared_dependencies(mini_zoo: Path) -> None:
+    reg = Registry(zoo_root=mini_zoo)
+    panel = _panel()
+    # NaN in a required column should propagate to output
+    panel["close"].iloc[2, 0] = np.nan
+    out = reg.compute("fakezoo_001", panel)
+    assert np.isnan(out.iloc[2, 0])
+    # Also test with NaN in another required column
+    panel = _panel()
+    panel["open"].iloc[1, 1] = np.nan
+    out = reg.compute("fakezoo_001", panel)
+    assert np.isnan(out.iloc[1, 1])
--- a/agent/tests/factors/test_registry.py
+++ b/agent/tests/factors/test_registry.py
@@ -268,3 +268,34 @@ def test_registry_masks_nan_on_declared_dependencies(mini_zoo: Path) -> None:
     panel["open"].iloc[1, 1] = np.nan
     out = reg.compute("fakezoo_001", panel)
     assert np.isnan(out.iloc[1, 1])
+
+
+def test_registry_masks_nan_without_close_declared(mini_zoo: Path) -> None:
+    """The NaN mask must run even when the alpha does not declare ``close``.
+
+    The mask's shape reference was hardcoded to ``panel["close"]``, so an alpha
+    declaring only ``["open"]`` got no NaN enforcement on a panel without a
+    close column — a fabricated value (np.where on a NaN comparison) survived.
+    The reference is now the first declared dependency.
+    """
+    reg = Registry(zoo_root=mini_zoo)
+    meta = GOOD_META.replace('"close", "open"', '"open"')
+    body = textwrap.dedent(
+        """\
+        import numpy as np
+        import pandas as pd
+
+        def compute(panel):
+            o = panel["open"]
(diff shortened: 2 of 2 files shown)


he-yufeng (commented):
Read the full diff. Enforcing the NaN contract at the registry layer is the right place: `base.py` already states it ("NaN preserved in warmup / missing data"), but each alpha had to honor it on its own, and one that forgot would silently emit numbers on bars whose inputs were missing. Masking at `compute` after the alpha runs makes the contract structural instead of per-author.

The tests pin both required-column families (a NaN in `close` and a NaN in `open` propagate to the output), and CI is green.

One boundary worth a note, not a blocker: the mask indexes `panel[col]` directly, so a declared dependency that is absent from the panel entirely (rather than present-but-NaN) raises KeyError instead of a RegistryError with the alpha id attached. Wrapping that lookup would keep the failure message informative; the current shape is fine if missing columns are already rejected upstream.

Prepared with AI assistance (Kimi); I verified the mask semantics against the declared-dependency metadata on main myself.

he-yufeng (commented):
Read the full diff. Enforcing the NaN contract at the registry layer is the right place: `base.py` already states it ("NaN preserved in warmup / missing data"), but each alpha had to honor it on its own, and one that forgot would silently emit numbers on bars whose inputs were missing. Masking at `compute` after the alpha runs makes the contract structural instead of per-author.

The tests pin both required-column families (a NaN in `close` and a NaN in `open` propagate to the output), and CI is green.

One boundary worth a note, not a blocker: the mask indexes `panel[col]` directly, so a declared dependency that is absent from the panel entirely (rather than present-but-NaN) raises KeyError instead of a RegistryError with the alpha id attached. Wrapping that lookup would keep the failure message informative; the current shape is fine if missing columns are already rejected upstream.

Prepared with AI assistance (Kimi); I verified the mask semantics against the declared-dependency metadata on main myself.

cgycorey:
One follow-up commit after your review (`707bd213`), flagging so you're not surprised.

On your KeyError note: confirmed unreachable — the `SkipAlpha` pre-checks reject any absent declared dependency (with the alpha id in the message) before the mask's `panel[col]` can throw, and both read the same `meta`, so no wrap is needed. Your "current shape is fine if missing columns are already rejected upstream" condition is satisfied.

Separately, re-reading the mask I found the shape reference was hardcoded to `panel["close"]`, so the 78 alphas whose `columns_required` doesn't include close (the `fund:*` fundamentals, the gtja191 volume/high/open-only ones) got no NaN enforcement at all on a panel without a close column — a fabricated value (np.where on a NaN comparison) survived the gate. The reference is now the first declared dependency (`columns_required` → `extras_required` → `sector` when `requires_sector`); the pre-checks guarantee it exists whenever there is anything to enforce. Behavior-identical for every panel where close is present (all panel columns share index/columns; a no-declared-deps alpha got an all-True no-op mask before and a skip now).

`test_registry_masks_nan_without_close_declared` pins the hole: an `["open"]`-only alpha fabricating `0.0` through `np.where` on a NaN comparison on a close-less panel now masks to NaN; the test fails on the previous code.

he-yufeng:
Follow-up commit `707bd213` verified locally, closing the loop from my earlier note.

What I ran on the PR head: the registry suite passes (18, including the new no-close case), and a mutation check on the follow-up behaves the way the test claims. Restore the hardcoded `panel.get("close")` reference and `test_registry_masks_nan_without_close_declared` goes red; put the deps-derived reference back and it is green again. So the 78 alphas that declare no `close` are genuinely covered now, not just on paper. The empty-deps alpha still skips the mask entirely, same as before, which reads as intended.

The ordering detail (mask after the return-type check) and the stated `>95% NaN` consequence both make sense to me: a loud skip in `bench_runner_strict` is exactly where fabricated signals should end up. Scan counts against the issue's spec are in the body, so from my side this is good to go. Leaving the final call to @warren618 since the acceptance criteria are his.

PR #1378, opened 2026-09-06
fix(grounding): a price word used as a formula variable is not a price claim (#1354)

### LINKED ISSUE #1354: [Bug] Grounding gate misreads a price-context word used as a formula variable (e.g. "close/SMA50 > 1") as an asserted observed value, causing false-positive numeric_claim_unavailable rejections
### Description

**What happened:**
`GroundingLedger._validate_price_claims` (agent/src/agent/grounding.py) decides whether a sentence contains a price claim by checking `_PRICE_CONTEXT_RE.search(segment)` (matches words like "close", "price", "preço") and then, if it matches, pulling the first number in the segment via `_numbers_without_dates_or_percent` as the asserted value. This does not distinguish a price word used as an observed value ("close was 2500") from the same word used as a formula/rule variable ("close/SMA50 > 1", "regra: close acima da EMA30"). In the latter case the number after the price word is not a price at all — it can be a threshold, a signal constant (+1/0/-1), a Monte Carlo simulation count, a lookback window, or a bare year — but the gate still grabs it as "the" asserted value and rejects it with `numeric_claim_unavailable` ("Price claim {value} has no matching observed tool evidence") whenever that number doesn't match any observed quote.

This caused real, correct backtest reports to be rejected as ungrounded. Two other AI agents (Codex, DeepSeek) spent hours trying to fix it by adding one regex per newly-discovered bad phrasing (a position/signal constant of ±1 or 0, a Monte Carlo simulation count with or without a currency unit next to it, a bare four-digit year in prose, a normalized ratio) — each fix accepted one more legitimate phrasing but reliably opened a new bypass or still rejected a different valid report, because natural language phrasings are unbounded and a denylist of phrasings can never be complete.

**What I expected:**
The gate should only treat the number after a price-context word as an asserted value when nothing between the word and the number reads as a formula reference — a closed, syntactic distinction (comparison/division operators, or an indicator identifier like SMA/EMA/RSI followed by digits), not an open-ended catalogue of "safe" phrasings to special-case one at a time.

**Suggested fix:**
Replace the enumeration-of-phrasings approach with one structural rule: scan for a small closed set of "formula markers" (`/`, `>`, `<`, `>=`, `<=`, `≥`, `≤`, or `SMA/EMA/WMA/DMA/RSI/MACD/ATR/ADX/CCI/OBV/KDJ/BOLL/VWAP` followed by digits) between the price-context word and the next number; only extract the number as a claimed value if no such marker sits in between. This also needs a sentence-boundary cutoff (the clause splitter does not split on ASCII "."), so a price word can't reach across an unrelated later sentence in the same clause and misattribute a completely unrelated number (e.g. a bare year mentioned two sentences later).

We have a working patch (new `_direct_price_values` method + `_FORMULA_MARKER_RE` + a regression suite covering both the original false positives and further adversarial cases we found) and can open a PR if useful.

### Steps to Reproduce

1. Ask the agent to describe a strategy rule in prose containing a price-context word immediately followed by a formula/ratio expression, e.g.: "Regra: compra quando close/SMA50 > 1.0" or "close acima da EMA30 dispara sinal +1".
2. Or: ask it to report Monte Carlo validation results in the same sentence as a price word, e.g.: "Monte Carlo: 1000 simulações, preço final 2500 USD" (the simulation count or an unrelated constant can be misread as the asserted price).
3. Observe the final-answer gate reject the (correct) report with `numeric_claim_unavailable`, even though every actually-observed price in the report matches tool evidence — the rejected "claim" is a formula operand, signal constant, or simulation count, not a price.

### Error Logs

```shell
{"code": "numeric_claim_unavailable", "claim": "<segment text>", "value": <misattributed number>, "message": "Price claim <value> has no matching observed tool evidence."}
```

### Interface

Web UI (vibe-trading serve)

### LLM Provider

deepseek

### Version

0.1.14

### Environment

racle Linux Server 9.8 / Python 3.11.16 (aarch64)

cgycorey:
## What

Closes #1354. `_validate_price_claims` no longer treats a price-context word used as a **formula variable** as an asserted observed price. The denylist of per-phrasing masks is replaced by one **structural rule** — exactly the closed, syntactic distinction the issue asked for and @he-yufeng endorsed.

## The rule

A number after a price word is a claimed value only when nothing formula-like sits between them. Two closed token classes decide:

1. **Formula marker** (closed set): a comparison/division operator (`/`, `>`, `<`, `>=`, `<=`, `≥`, `≤`) or an indicator identifier followed by digits (`SMA/EMA/WMA/DMA/RSI/MACD/ATR/ADX/CCI/OBV/KDJ/BOLL/VWAP` + digits, plus `MA` for the Chinese `MA20` convention). A marker between the price word and the number marks the number an **operand**.
2. **Observation binder** (closed set): `was/were/is/are/at`, `报收/收于/收报/收在`, and `=`/`==`/`≈`. A binder *after* the last marker re-attaches the number to the price word.

The binder guard is what keeps this a **strict narrowing**, not a loosening:

| draft | result |
|---|---|
| `close/SMA50 > 1` | operand → no claim |
| `close acima da EMA30 dispara sinal +1` | operand → no claim |
| `close vs SMA20 maior que 1` | operand → no claim |
| `close minus SMA20 fallen below 0` | operand → no claim |
| `close above SMA50 and was 2500 yesterday` | binder "was" → **claim** |
| `close above SMA50 at 2450` | binder "at" → **claim** |
| `close was 2500` (no evidence) | no marker → **claim → rejects** |

## Sentence-boundary cutoff

The clause splitter does not split on the ASCII period, so `close was 210. In 2024 the market rallied` stays one clause and `2024` used to be misread as the price. A `. ` / `! ` / `? ` in the span between the price word and the number now ends the price word's reach — scoped to this scan, **not** the global clause splitter, where splitting on `.` would tear apart `000543.SZ` and `1.5`.

## Two shapes change classification (both inherent to the requested rule)

- `close above SMA50 2500` — a bare number directly trailing an indicator identifier is read as that identifier's operand → accepted.
- `close. Sinal: 2500` — the number is in a later sentence than the price word → no longer attributed → accepted.

The canonical laundering shapes (`close was 2500`, `close above SMA50 and was 2500`, `close above SMA50 at 2450`, `close = 2500`) all still reject. Numbers with no preceding price word keep main behavior.

## Tests

`test_price_validation_ignores_formula_variable_digits` (ledger-level) — 9 formula shapes pass, 9 laundering shapes still reject. `test_direct_price_values_structural_rule` (unit) locks the operand/claim/boundary decision directly. Full suite: 12645 passed, ruff clean (the 6 `test_readme_counts` failures are pre-existing tool-count drift, unrelated).

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -505,7 +505,12 @@ _INDICATOR_VALUE_RE = re.compile(
     r"\b(?:rsi|macd|atr|adx|cci|obv|kdj|boll|dif|dea|vix|iv|"
     r"sharpe|sortino|beta)\b"
     r"(?:\s*\([^)]{0,20}\))?"
-    r"\s*(?:is|at|of|reads?|=|为|是)?\s*[:：]?\s*"
+    # #1354: "RSI below 30" / "sharpe above 1" — a directional connective is
+    # part of the indicator reading, so the reading's number stays masked.
+    # Only indicator NAMES get these connectives; a price word with
+    # "below/above" ("close above 2500") is an observed-value claim and
+    # matches none of these names, so it stays gated.
+    r"\s*(?:is|at|of|reads?|=|为|是|below|above|under|over)?\s*[:：]?\s*"
     r"[-+]?\d[\d,]*(?:\.\d+)?",
     re.IGNORECASE,
 )
@@ -520,6 +525,22 @@ _CURRENCY_TOKEN = r"(?:\$|US\$|C\$|HK\$|CAD|USD|CNY|HKD|¥|￥)?"
 # yet both went to the OHLC check and rejected a weekly update whose quotes
 # were correct. This is the same category as the target/stop levels below -- a
 # level the report proposes, not one the data source reported.
+
+# #1354: a signal value trailing an arrow or a signal word ("sinal +1",
+# "-> +1", "触发 +1", "Sinal: +1") is the formula's output, not an observed
+# price. Signal values are small integers (a +/-1 signal, a 1-10 score); a
+# multi-digit price never follows these words, so the digit bound keeps
+# "close -> 2500" (a price claim written in arrow notation) gated. The colon
+# is optional ("Sinal: +1" — ASCII '.' is not a clause separator, so this can
+# sit in the same clause as the price word), the sign may be spaced ("sinal
+# - 1"), and the lookahead stops the bound from masking the first two digits
+# of a longer number ("signal 2500" keeps the full 2500 gated).
+_SIGNAL_VALUE_RE = re.compile(
+    r"(?:\b(?:signal|sinal|trigger|triggers|triggered|dispara|信号|触发)\b[:：]?"
+    r"|->|→|=>)"
+    r"\s*[-+]?\s*\d{1,2}(?:\.\d+)?(?!\d)",
+    re.IGNORECASE,
+)
 _ORDER_LEVEL_RE = re.compile(
     r"(?:"
     # (a) "<qty> [股|shares] @ <price>" -- the whole clause, quantity included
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Read the full diff. The move from a per-phrasing denylist to a closed marker set (formula marker vs observation binder) is the right call for this class of check: the catalogue of phrasings can never close, the marker set can, and every regex in the diff carries its reasoning in a comment, including the deliberate exclusions (the sentence-boundary lookahead avoiding abbreviations, the digit bound on signal values so "close -> 2500" stays gated).

The regression tests cover the taxonomy across English, Portuguese and Chinese phrasings, including the mixed case ("close above SMA50 and was 2500") where the binder re-attaches the number. CI is green.

One thing I checked before commenting: the boundary regex (`[.!?]\s+(?=[A-Z(（])`) stays scoped to this price-claim scan and never feeds the global clause splitter, so ticker-like strings such as "000543.SZ" are unaffected. That scoping is documented in the comment and worth keeping.

Good to merge from my side.

Prepared with AI assistance (Kimi); I verified the rule and the test taxonomy against main myself.

cgycorey:
One-line addition after your review, flagging so you're not surprised: `re.ASCII` on `_FORMULA_MARKER_RE` (+ one test). Without it, `\b` is Unicode-aware and a CJK letter counts as a word char, so the Chinese form `上穿MA20` (no ASCII space before the indicator) never matched the marker — the rule silently skipped the exact CN phrasing your review noted the tests cover. `收盘价上穿SMA20 2500` now correctly reads 2500 as an operand. No other change.

PR #1389, opened 2026-09-07
fix(backtest): match china futures product codes case-insensitively

Shizoqua:
## Summary
ChinaFuturesEngine silently uses wrong multiplier, margin, and commission for uppercase SHFE/DCE/INE/GFEX symbols.

## Why
SHFE/DCE/INE/GFEX keys are lowercase, but the documented convention and real Tushare data use uppercase, so lookups silently miss and fall back to generic defaults.

## Changes
Added a case-insensitive lookup helper, applied at all six product-table call sites.

## Test Plan
- [x] Existing tests pass (49 passed)
- [x] New tests added

## Checklist
- [x] No hardcoded values
- [x] Code follows CONTRIBUTING.md

### DIFF
--- a/agent/backtest/engines/china_futures.py
+++ b/agent/backtest/engines/china_futures.py
@@ -128,6 +128,27 @@ def _extract_product(symbol: str) -> str:
     return m.group(1) if m else code
 
 
+def _ci_lookup(table: dict, product: str, default):
+    """Case-insensitive product lookup.
+
+    The tables above key CFFEX/ZCE products uppercase (IF, CF) and
+    SHFE/DCE/INE/GFEX products lowercase (au, rb) — but this project's
+    own documented symbol convention and Tushare's real ts_code values
+    (data-routing/SKILL.md: CU2406.SHFE; tushare 日线行情.md: CU1811.SHF)
+    use uppercase product codes across every exchange. A caller that
+    follows that convention for a SHFE/DCE/INE/GFEX product would
+    otherwise miss the table entirely and silently fall back to the
+    generic default.
+    """
+    if product in table:
+        return table[product]
+    if product.lower() in table:
+        return table[product.lower()]
+    if product.upper() in table:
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Verified the premise against current main before reading the fix: the tables really are split-cased (SHFE/DCE/INE lowercase, CFFEX/ZCE uppercase), `_extract_product` preserves caller case, and the documented symbol convention admits uppercase (`_market_hooks` accepts `[A-Za-z]{1,2}\d{3,4}\.(SHFE|...)` case-insensitively), so `AU2412.SHFE` silently fell to the 10x default multiplier on main. All six table lookups are covered, I counted the same six call sites and no seventh exists.

Ran the suite on this head locally: 49 passed including the 9 new case-insensitivity tests (ran the class separately to be sure it was collected).

One thought, not a blocker: the tri-check helper works, but normalizing once at table definition (build a lowercase view per table) would make future call sites safe by construction. Fine as is.

Bortlesboat (commented):
Checked `3b700d21deb504ccd1440de5ea2ede7de96a2224` on Windows with Python 3.12. The 49 tests in `agent/tests/test_china_futures_engine.py` pass.

I also swept the union of products in `_MULTIPLIER`, `_MARGIN_RATE`, `_COMMISSION`, and `_PRICE_LIMIT`: **58 products, 870 uppercase/lowercase comparisons**, all equal. For each product I compared initial leverage, multiplier, margin rate, opening/closing commissions, and `can_execute` in both directions at prices 400, 475, 500, 525, and 600 with previous settlement/close 500. Each case used separate engines initialized with the differently cased code, so this covers the constructor lookup as well as the per-order lookups.

That gives additional coverage beyond the representative exchange examples in the regression tests; I found no remaining case-dependent result across the current tables. This is an offline consistency check of the lookup change, not validation of the exchange table values or live execution.

Issue #1149, opened 2026-08-19
[Feature] Strategy Discovery Phase 3 — description-driven adaptation


warren618:
Split out of #969, which is closed now that Phase 1 (#978) and Phase 2 (#1007) have landed. That issue said from the start that Phase 3 was out of its scope, and the shape below is the two lines it reserved, expanded against what actually shipped.

## What exists now

The Strategy Discovery facade answers *what strategies exist* and *what state they are in*, over Alpha Zoo plus the SDM store, with per-regime evidence rows rather than scenario tags:

- `list_strategies` / `query_strategies` / `get_strategy_evidence` — read-only
- `refresh_strategy_evidence` — rebuilds the disposable evidence cache from local backtest run artifacts, gated by the `backtest-diagnose` hard gates
- every row carries `evidence_stage`, `provenance` (the run directory it was computed from), `regime_definition`, and a read-time freshness verdict of `fresh` / `aging` / `stale`

The loop that is missing is the one that closes it: a strategy the facade knows about cannot be *re-derived* for a new universe or regime without the user hand-writing the config.

## Proposal

Let `strategy-generate` take a `strategy_id` from the facade and produce an adapted `SignalEngine` from it:

1. Agent resolves `strategy_id` through `get_strategy_evidence`
2. Reads the strategy description and the regime definition that evidence was computed under
3. Generates adapted code for the requested universe / date range
4. Backtests it
5. Feeds the resulting run directory back through `refresh_strategy_evidence`, so the adaptation becomes evidence under its own `provenance` rather than inheriting the parent's

Step 5 is the part that makes this more than a convenience wrapper: an adapted strategy is a *different* strategy, and its evidence has to stand on its own run.

## Constraints carried over from #969

These are the reasons #894/#896 were rejected, and they apply unchanged:

- **No phantom tools.** Anything named in a prompt must be registered and callable; `guard.py` already enforces atomic advertisement and must keep doing so if the tool surface grows.
- **No inherited evidence.** An adapted strategy must not present the parent's metrics as its own. `evidence_stage` may not claim a computed result the row's `provenance` cannot reproduce — enforced in `EvidenceRow.__post_init__` as of #978.
- **No new seed corpus.** Everything reproducible from this repo.

## Open questions

[...]

PR #1391, opened 2026-09-08
feat(strategy-store): add description-driven adaptation substrate

modelpath-dev:
## Summary

- Add a store-level adaptation path that copies a parent strategy into a new `Artifact` with `derived_from` set and the signal fields left unchanged.
- Refuse adaptation when the parent has no adequate, non-stale evidence row.
- Leave evidence empty on the child so the next backtest can call `refresh_strategy_evidence` with its own provenance.

## Why

Phase 3 of strategy discovery (#1149) is adaptation, not re-search. A description may change universe, name, or position sizing. It must not change `signal_definition`. Copying the parent's evidence onto a new id would violate `EvidenceRow` provenance, so this PR only lands the store substrate.

Refs #1149. This does not close the issue. `strategy-generate`, the backtest loop, and any MCP tool still need a follow-up.

## Changes

- `Artifact.derived_from` plus a SQLite schema bump to version 3 (`derived_from TEXT`, additive and nullable).
- `adapt_artifact` / `register_adaptation` in `agent/src/strategy_store/adaptation.py`. `AdaptationPatch` can only carry `universe`, `name`, and `position_sizing`.
- `parent_adaptation_blockers` in `agent/src/strategy_discovery/adaptation.py`. It reuses Phase 1 quality (`QUALITY_ADEQUATE`) and Phase 2 decay (`classify_decay`, `DECAY_STALE`). No new numeric thresholds.
- Tests for signal copy, identity reset, factor refusal, unique `(name, universe)`, sqlite roundtrip, and healthy / stale / insufficient parent evidence.

## Tradeoffs

A facade that minted a new id and inherited parent metrics was rejected. `EvidenceRow` provenance requires a non-empty `provenance` when `evidence_stage` claims a result. The child must run its own backtest.

Same `name` plus a new `universe` is allowed. Same `name` plus the same `universe` is not. That keeps the existing unique constraint.

## Blast Radius

Reads and writes stay inside `agent/src/strategy_store/` and a small discovery helper. Protected areas (`src/agent/`, `src/session/`, `src/providers/`) are untouched. No MCP, broker, network, or secret surface changes. Existing DBs gain a nullable column on open. Rollback is `git revert` of this commit.

## Test Plan

- [x] New tests added
- [x] `pytest agent/tests/test_strategy_store_adaptation.py agent/tests/test_strategy_discovery_adaptation.py agent/tests/test_strategy_store.py --tb=short -q` (113 passed)
- [ ] Full suite `pytest --ignore=agent/tests/e2e_backtest --tb=short -q` not run locally
- [ ] Tested manually (not applicable; no UI or MCP entry yet)

## Checklist

- [x] No hardcoded values (API keys, file paths, magic numbers). Quality and decay cutoffs are the existing constants.
- [x] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines (DCO sign-off on the commit)
- [ ] Documentation updated (if user-facing change). Not user-facing yet.

### DIFF
--- a/agent/src/strategy_discovery/adaptation.py
+++ b/agent/src/strategy_discovery/adaptation.py
@@ -1,0 +1,49 @@
+"""Parent-evidence gate for description-driven adaptation."""
+
+from __future__ import annotations
+
+from datetime import date
+from typing import Sequence
+
+from src.strategy_discovery.models import (
+    DECAY_STALE,
+    QUALITY_ADEQUATE,
+    EvidenceRow,
+    classify_decay,
+)
+
+
+def parent_adaptation_blockers(
+    rows: Sequence[EvidenceRow], today: date
+) -> tuple[str, ...]:
+    """Return stable refusal reasons, or empty when adaptation may proceed.
+
+    Allowed when at least one row is adequate and not stale. Reasons use
+    ``stale-evidence:`` / ``insufficient-evidence:`` prefixes.
+    """
+    if not rows:
+        return ("insufficient-evidence: parent has no evidence rows",)
+
+    has_usable = False
+    saw_stale = False
+    saw_weak = False
+    for row in rows:
+        decay = classify_decay(row.date_ranges, today)
+        if row.evidence_quality == QUALITY_ADEQUATE and decay != DECAY_STALE:
+            has_usable = True
+        if decay == DECAY_STALE:
+            saw_stale = True
+        if row.evidence_quality != QUALITY_ADEQUATE:
+            saw_weak = True
+
+    if has_usable:
+        return ()
+
+    reasons: list[str] = []
+    if saw_stale:
+        reasons.append(
+            "stale-evidence: parent has no fresh or aging adequate evidence"
+        )
+    if saw_weak or not saw_stale:
+        reasons.append("insufficient-evidence: parent has no adequate evidence")
+    return tuple(reasons)
--- a/agent/src/strategy_store/adaptation.py
+++ b/agent/src/strategy_store/adaptation.py
@@ -1,0 +1,78 @@
+"""Description-driven adaptation of SDM strategy artifacts."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, replace
+from typing import Protocol
+
+from src.strategy_store.models import (
+    Artifact,
+    ArtifactStatus,
+    ArtifactType,
+    ValidationStatus,
+)
+
+
+class AdaptationError(ValueError):
+    """Raised when a parent cannot be adapted under the Phase 3 contract."""
+
+
+@dataclass(frozen=True)
+class AdaptationPatch:
+    """Allowed description-driven changes. Signal fields are not on this type."""
+
+    universe: str | None = None
+    name: str | None = None
+    position_sizing: str | None = None
+
+
+class _ArtifactStore(Protocol):
+    def get_artifact(self, artifact_id: str) -> Artifact | None: ...
+
+    def register_artifact(self, artifact: Artifact) -> str: ...
+
+
+def adapt_artifact(parent: Artifact, patch: AdaptationPatch) -> Artifact:
+    """Return a new strategy artifact derived from *parent*. Does not write."""
+    if parent.type is not ArtifactType.STRATEGY:
+        raise AdaptationError(
+            f"only strategy artifacts can be adapted; got type={parent.type.value!r}"
+        )
+    if not parent.id:
+        raise AdaptationError("parent artifact id is required")
+
+    new_universe = patch.universe if patch.universe is not None else parent.universe
+    new_name = patch.name if patch.name is not None else parent.name
+    if new_universe == parent.universe and new_name == parent.name:
+        raise AdaptationError("adaptation must change universe or name")
+
+    new_sizing = (
+        patch.position_sizing if patch.position_sizing is not None else parent.position_sizing
+    )
+    return replace(
+        parent,
+        id="",
+        name=new_name,
+        universe=new_universe,
+        position_sizing=new_sizing,
+        derived_from=parent.id,
+        status=ArtifactStatus.CREATED,
+        run_dir=None,
+        created_at="",
+        updated_at="",
+        disabled_at=None,
+        disabled_reason=None,
+        validation_status=ValidationStatus.UNVALIDATED,
+        validation_date=None,
+    )
+
+
+def register_adaptation(
+    store: _ArtifactStore, parent_id: str, patch: AdaptationPatch
+) -> str:
+    """Persist an adapted child through the strategy store. Returns the new id."""
+    parent = store.get_artifact(parent_id)
+    if parent is None:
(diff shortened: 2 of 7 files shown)


warren618:
Reviewed. The store substrate is good work — the migration is idempotent and covers a pre-governance DB, `derived_from` is written on both the INSERT and the UPDATE (I checked, because a schema column that never reaches the INSERT is the usual way this lands half-done), and rejecting the "mint a new id and inherit the parent's metrics" facade is the right call for the reason you give.

Your CI is red only on `test_llm_reasoning_effort.py::TestRequestPayload::test_explicit_none_reaches_the_request`, which is not yours: `langchain-openai` 1.6.0/1.6.1 started auto-routing some model names to the Responses API when `use_responses_api` is unset, and the test job installs from the extras rather than `requirements-lock.txt`. Fixed on main; rebase and it goes green. Your own tests pass (12802).

Two things before it can land.

**1. The evidence gate has no caller, so the refusal in your summary does not happen.**

`parent_adaptation_blockers` is imported by exactly one file in the repo, and that file is its own test. `register_adaptation` never asks for evidence:

```python
def register_adaptation(store, parent_id, patch):
    parent = store.get_artifact(parent_id)
    if parent is None:
        raise AdaptationError(...)
    child = adapt_artifact(parent, patch)      # no evidence check anywhere
    return store.register_artifact(child)
```

So "Refuse adaptation when the parent has no adequate, non-stale evidence row" describes a function that exists, not behaviour the code has. Today a parent whose only evidence is stale, or insufficient on trades, adapts without complaint. That gap is worse than not having the gate: someone reading `register_adaptation` will reasonably assume the check already happened, and the follow-up PR that wires it will look like a no-op refactor.

I would rather this PR make its own claim true than defer it. The evidence rows live on the discovery side, not the artifact store, so the honest shape is to take them as an argument:

```python
def register_adaptation(store, parent_id, patch, *, parent_evidence):
    ...
    blockers = parent_adaptation_blockers(parent_evidence, today)
    if blockers:
        raise AdaptationError("; ".join(blockers))
```

If you would rather not settle the caller's shape in this PR, the other honest option is to drop `src/strategy_discovery/adaptation.py` and its test from it and land the store substrate alone — then nothing promises a check that is not there. Either is fine by me; a tested function with no caller is not.

**2. A position-sizing-only adaptation is refused, which contradicts the contract you state.**

The summary says a description may change universe, name, *or position sizing*, but:

```python
if new_universe == parent.universe and new_name == parent.name:
    raise AdaptationError("adaptation must change universe or name")
```

and `test_noop_same_name_and_universe_refused` pins that with a patch that changes only `position_sizing`. "Same strategy, same universe, half the position size" is a legitimate adaptation and there is currently no way to express it. The cause is `UNIQUE(name, universe)` rather than anything about the domain, which is worth saying out loud in the code — right now the error message says "must change universe or name" as if that were the rule, when it is a consequence of the index.

Not asking you to redesign the key in this PR. Either narrow the error message to name its real cause, or say in the tradeoffs section that position-sizing-only adaptation is unreachable until the uniqueness key changes, so the next reader does not have to derive it from a constraint.

warren618:
Merged, with a follow-up commit on main (`a7bd2e30`) closing two gaps. The split was the right call and the constraints you held are the ones that matter: `AdaptationPatch` cannot carry `signal_definition` at all, so the signal is uncopyable-by-construction rather than by convention, and refusing to mint evidence on the child is exactly what #1149 reserved Phase 3 for. Rejecting the facade that inherited parent metrics was the correct trade.

**First, on your CI:** the red `test` lanes are not your diff. The only failure is `test_llm_reasoning_effort.py::TestRequestPayload::test_explicit_none_reaches_the_request` with `KeyError: 'reasoning_effort'`. Your base is `4abd6cc2` (09-08); main fixed that in `545e6d5b` (09-09). Merged onto current main your branch is green — 149 passed, 7 skipped across the three store/discovery files plus that test. Nothing was wrong with the branch.

**Gap 1 — the child inherited the parent's sign-off.**

`adapt_artifact` copies through `dataclasses.replace`, which carries over every field the call does not name. You reset `validation_status` and `validation_date`, but not the people:

```
parent: validator="Val", approver="Cho", artifact_version="3", model_version="gpt-x-2025"
child : validation_status=UNVALIDATED, validation_date=None
        validator="Val"   <-- an independent validation it never had
        approver="Cho"    <-- a sign-off it never received
        artifact_version="3", model_version="gpt-x-2025"
```

That is the same rule your PR was written to enforce, one field over. The governance block in `models.py` reads "who built it, who owns it, who independently validated it, who signed off" — the first two describe authorship, which adaptation does not change, and the last two are attestations about a *specific* artifact. `artifact_version` and `model_version` go with them: the child is version 1 of its own lineage, and no model has generated its code yet, since `run_dir` is None until its own backtest runs. `developer` / `owner` / `intended_use` / `limitations` still carry forward, so the reset is surgical rather than a blanket wipe — there is a test asserting that half too, so a future over-correction is caught.

A side effect worth naming: a parent with `developer == approver` is a four-eyes violation, and before the fix the child inherited that too. `is_four_eyes_violation` now returns False on the child, because there is no approval on it to check yet.

**Gap 2 — the evidence gate was not on the write path.**

`parent_adaptation_blockers` is good, and it was reachable only from tests. `register_adaptation` never called it, so this ran clean:

```python
register_adaptation(store, stale_parent_id, AdaptationPatch(universe="SP500"))  # wrote the child
```

against what the PR body advertises ("Refuse adaptation when the parent has no adequate, non-stale evidence row"). I made the gate inputs keyword-only and **required**:

```python
register_adaptation(store, parent_id, patch, *, parent_evidence, today)
```

The point of required-and-keyword-only is that the check cannot be skipped by forgetting it — the call does not run without the inputs the gate needs, so a caller has to make an explicit decision rather than silently inherit the unguarded path. A missing parent is still reported before the gate, so "not found" does not masquerade as "insufficient evidence". The import is deferred inside the function, matching how `facade.py` already reaches across that boundary, so no module-level cycle.

Both fixes are pinned by tests verified red on the unfixed code: reverting the four field resets fails two, removing the gate call fails two more. Full backend suite on the merge: **12877 passed, 30 skipped, 0 failed**.

Leaving #1149 open as you asked. For the follow-up — `strategy-generate`, the backtest, and the `refresh_strategy_evidence` round-trip — the thing to carry forward is that `replace` is a blanket copy: whenever the child gains a field, the default is that it inherits, so each new field needs an explicit answer to "did the child earn this?" Thanks for splitting it this way; a substrate PR this size is genuinely reviewable, which is why both of these were findable.

modelpath-dev:
Hi Haozhe,

Thanks for reviewing and merging #1391. I learned a lot working through the description-driven strategy-store changes.

I am currently looking for opportunities where I can keep contributing like this, including paid internships, contract work, or full-time roles if your team ever has an opening. Happy to take another issue on the repo either way.

Thanks again,
Chandan Kumar
https://github.com/modelpath-dev

PR #1392, opened 2026-09-08
fix(ml-strategy): prevent future-label leakage in walk-forward training

Shizoqua:
## Summary

Prevent future-label leakage in the ML strategy walk-forward example.

## Problem

The walk-forward training window included labels whose forward return was not yet observable at prediction time. The label construction also converted unavailable tail targets to class `0`.

## Changes

* Purge labels whose prediction horizon extends beyond the current prediction point.
* Preserve unavailable future labels as `NaN`.
* Pass the prediction horizon explicitly to walk-forward training.
* Add regression coverage for horizon purging, unavailable tail labels, and one-bar horizon behaviour.

## Testing

* Regression tests: 3 passed.
* `python -m py_compile`: passed.
* Diff/whitespace check: passed.
* Black and Ruff were not available in the execution environment.
* Full repository test suite was not run.

## Compatibility and Risk

Low risk. The change is limited to the ML strategy skill example and its regression tests. A one-bar prediction horizon preserves the previous training-window behaviour.

## Related Issue

No existing issue identified for this bug.

### DIFF
--- a/agent/src/skills/ml-strategy/SKILL.md
+++ b/agent/src/skills/ml-strategy/SKILL.md
@@ -105,6 +105,7 @@ def walk_forward_predict(
     model_type: str = "random_forest",
     window_type: str = "expanding",
     sliding_size: int = 504,
+    prediction_horizon: int = 5,
 ) -> pd.Series:
     """Walk-forward training and prediction to avoid future data leakage.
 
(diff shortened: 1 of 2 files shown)


warren618:
Merged. The horizon purge is the right fix and the `_RecordingModel` assertion on the last training row is the part that makes it a real test rather than a re-statement of the code — the training window's end is the thing that leaked, so asserting where it ends is exactly the right anchor. Preserving the tail labels as NaN instead of coercing them to class 0 matters as much as the purge.

One thing left in the same file: `SKILL.md:66` and lines 69–70 still used `pct_change()` and `pct_change(5)` / `pct_change(20)` with pandas' padding default, which fabricates a 0.0% return on a missing close. That is the same class #1397/#1398 were fixing; all 24 sites across the skills are swept in `be2c09f2`. Nothing wrong with your diff — just the sibling two lines up from it.

PR #1396, opened 2026-09-08
fix(backtest): classify Tushare-spelled China futures as China futures

### LINKED ISSUE #1394: Tushare's own exchange suffixes (.SHF/.CZC/.CFX) route China futures to GlobalFuturesEngine
`_is_china_futures` decides China-vs-global futures by exchange suffix
(`agent/backtest/engines/_market_hooks.py:103`):

```python
_CHINA_EXCHANGES = {"CFFEX", "SHFE", "DCE", "ZCE", "INE", "GFEX"}
```

Tushare does not use those. Its `ts_code` suffixes are `SHF`, `DCE`, `CZC`,
`CFX`, `INE`, `GFE` — and this repo's own reference docs show them:

- `agent/src/skills/tushare/references/期货数据/日线行情.md:49` —
  `pro.fut_daily(ts_code='CU1811.SHF', ...)`
- `.../每日结算参数.md:51` — `CU1811.SHF`

Nothing in the repo maps `SHF -> SHFE`, so the suffix falls through to
"not a Chinese exchange" and the contract is handed to `GlobalFuturesEngine`.
Measured:

| symbol | engine picked | contract multiplier | correct |
|---|---|---|---|
| `CU1811.SHF` | Global | 50 | 5 (SHFE copper) |
| `CF501.CZC` | Global | 50 | 5 (ZCE cotton) |
| `IF2406.CFX` | Global | 50 | 300 (CFFEX IF) |
| `CU2406.SHFE` | China | 5 | 5 |

Repro:

```python
from backtest.engines._market_hooks import _is_china_futures
from backtest.engines.global_futures import GlobalFuturesEngine
assert _is_china_futures("CU1811.SHF") is False        # routed to global
GlobalFuturesEngine({}).get_contract_multiplier("CU1811.SHF")  # -> 50.0
```

Beyond the multiplier, the global engine also applies USD per-contract
commission and the global price-band rules, so a CNY contract is backtested
as if it were a CME product. It is silent — the symbol is accepted, bars load
if a local file provides them, and the run completes with wrong notionals.

Reachability is narrow today but real: no network loader implements China
futures bars (see #1395), so this bites a user feeding local CSVs
named with the ts_code they copied out of the Tushare docs above.

The fix is small either way, but the direction is a call to make:

1. Accept Tushare's suffixes as aliases in `_CHINA_EXCHANGES` (and normalise
   them to the canonical form so the engine's own tables stay one spelling).
2. Or reject an unrecognised futures suffix outright instead of silently
   defaulting to the global engine — the misroute is worse than a clear error.

Option 2 also covers suffixes we have not thought of, which option 1 does not.

he-yufeng:
Fixes #1394.

## What

`_is_china_futures` trusted the exchange suffix against `{CFFEX, SHFE, DCE, ZCE, INE, GFEX}` verbatim, but Tushare spells them `SHF`/`CZC`/`CFX`/`GFE` (the repo's own tushare reference docs show `CU1811.SHF`). A Tushare-formatted code fell to "not a Chinese exchange" and was handed to `GlobalFuturesEngine`, which priced a CNY contract with the global USD per-contract multiplier and commission.

The suffix is now normalized through an alias map before the set-membership test, and the China-futures market pattern admits the Tushare spellings so `_detect_market` classifies them as futures in the first place.

## How I verified

Verified against current main first: `CU1811.SHF`, `CF501.CZC`, `IF2406.CFX` all classified `a_share` / non-China before; after the change they classify `futures` + CNY, and `ChinaFuturesEngine` resolves SHFE copper's multiplier 5 and 8% margin for `CU1811.SHF`. The non-China guards (`M2412.CBOT`, `CL2412.NYMEX`) still do not classify.

New test in `tests/test_market_detection.py` pins the four Tushare suffixes plus the canonical spellings and the non-China controls; it is red on current main and green here. The wider engine/market/shadow/composite subset: 945 passed, 5 skipped.

## Risk / rollback

Narrow classification change: only codes with the four Tushare suffixes change routing, and each now lands where the canonical spelling already landed. Rollback: revert this commit; no state or schema changes.

One caveat worth naming: the alias covers the suffix spellings Tushare actually emits per its docs; if another vendor spells the same exchanges differently again, that is a new alias entry, not a new mechanism.

### DIFF
--- a/agent/backtest/engines/_market_hooks.py
+++ b/agent/backtest/engines/_market_hooks.py
@@ -55,7 +55,8 @@ _MARKET_PATTERNS = [
     # they still reach the forex whitelist below.
     (re.compile(r"^[A-Z]{2,}(?:USDT|USDC|BUSD)$", re.I), "crypto"),
     # China futures: product+delivery.exchange (e.g. IF2406.CFFEX, rb2410.SHFE)
-    (re.compile(r"^[A-Za-z]{1,2}\d{3,4}\.(ZCE|DCE|SHFE|INE|CFFEX|GFEX)$", re.I), "futures"),
+    # Tushare suffix spellings (SHF/CZC/CFX/GFE) classify here too.
+    (re.compile(r"^[A-Za-z]{1,2}\d{3,4}\.(ZCE|DCE|SHFE|INE|CFFEX|GFEX|SHF|CZC|CFX|GFE)$", re.I), "futures"),
     # Global futures: product+month-code (e.g. ESZ4, CLF25, GCM2025)
     (re.compile(r"^[A-Z]{2,4}[FGHJKMNQUVXZ]\d{1,2}$", re.I), "futures"),
     # Global futures: product+YYMM (e.g. CL2412, ES2503)
@@ -101,6 +102,10 @@ _MARKET_PATTERNS = [
 ]
 
 _CHINA_EXCHANGES = {"CFFEX", "SHFE", "DCE", "ZCE", "INE", "GFEX"}
+
+# Tushare spells the same exchanges differently (ts_code='CU1811.SHF');
+# normalize to the canonical suffix before any set membership test (#1394).
+_EXCHANGE_ALIASES = {"SHF": "SHFE", "CZC": "ZCE", "CFX": "CFFEX", "GFE": "GFEX"}
 
 # Supported settlement-currency contract per market. A composite backtest holds
 # one shared capital pool, so a code set spanning two of these would add CNY to
@@ -222,7 +227,7 @@ def _is_china_futures(code: str) -> bool:
         # else = False. Without this guard the product-code heuristic below
         # would misclassify global futures whose product letters happen to
         # collide with a CN product (e.g. ``M2412.CBOT`` — US soybean meal).
-        return parts[1] in _CHINA_EXCHANGES
+        return _EXCHANGE_ALIASES.get(parts[1], parts[1]) in _CHINA_EXCHANGES
     # Bare code (no exchange suffix): fall back to product-code heuristic.
     m = re.match(r"([A-Za-z]+)\d+", parts[0])
     if m:
--- a/agent/tests/test_market_detection.py
+++ b/agent/tests/test_market_detection.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import pytest
 
 from backtest.engines._market_hooks import _is_china_futures, code_currency
+from backtest.engines._market_hooks import _is_china_futures
 from backtest.runner import (
     _detect_market,
     _detect_source,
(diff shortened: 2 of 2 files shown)


warren618:
Merged. The alias is right and your verification method — checking the classification on main first, then after — is what made it reviewable.

Two notes for the record.

**The engine the contract actually reached.** #1394 says the Tushare-spelled code "is handed to `GlobalFuturesEngine`". Measured on main before your fix, it is worse than that: `_detect_market("CU1811.SHF")` returns `a_share`, so `runner.py:1329`'s `if "futures" in markets` never fires and the contract went to the **A-share** engine — T+1, no shorting, no margin at all. Your fix is correct either way; I am only correcting the record so the next person reading that issue does not go looking for a global-futures dispatch that never happened.

**The same gap was open on the global side, and I have fixed it in the same pass.** A dated contract carrying its venue matched no pattern either:

```
CL2412.NYMEX   a_share  CNY
ES2503.CME     a_share  CNY
ESZ4.CME       a_share  CNY
GCM2025.COMEX  a_share  CNY
ES.CME         futures  USD     <- the continuous form matched; the dated one did not
```

So a USD crude contract was settled in CNY under A-share rules. `_MARKET_PATTERNS` had `product+month`, `product+YYMM` and `product.venue`, but not `product+month.venue` — the way a contract is normally written. Added in `349bf202`, with the product width opened to `{1,4}` only in the venue-suffixed pattern (CBOT lists single-letter grains, and a recognized venue is what justifies admitting them; `C2412` alone stays out).

One thing that did not hold up: the PR says the Tushare spellings come from the repo's reference docs, but only `SHF`, `DCE` and `CFX` actually appear there (65 / 26 / 21 occurrences). `CZC` and `GFE` are asserted, not evidenced. They are harmless — neither collides with a real suffix, and `ZCE`/`INE` were already in the canonical set — so this is not a blocker, just worth knowing that two of the four aliases are untested against a real Tushare response.

I also dropped the duplicate `_is_china_futures` import the test file picked up.

he-yufeng:
Thanks for the merge, and for the corrections — both are worth having on record.

On the engine path: you're right, and it's the more useful version of the fact. `_detect_market` returning `a_share` means `runner.py:1329`'s futures dispatch never fired at all, so the contract was settling under T+1, no-short, no-margin A-share rules, not global-futures rules. My issue text described the wrong receiving engine; the fix is the same but the failure mode was worse than I wrote.

On CZC/GFE: fair catch. I asserted them from the spelling pattern instead of counting occurrences in the reference docs (SHF/DCE/CFX are the ones actually evidenced there). They don't collide with anything real, but if a future cleanup wants the alias set to be evidence-only, those two are the ones to drop.

And thanks for catching the dated-contract hole on the global side in the same pass — `product+month.venue` is the normal way to write a contract, so that gap would have produced exactly the same USD-settled-in-CNY shape on the next dated global contract someone threw at it.

PR #1397, opened 2026-09-08
fix(sdm): preserve missing price gaps in strategy template

Shizoqua:
## Summary

Preserve missing price observations in the Strategy Development Manager signal template.

## Problem

The template used pandas `pct_change()` with its default fill behaviour. Missing close prices could therefore be forward-filled when computing returns and volatility, producing signals from unavailable observations.

## Changes

* Disable filling when computing percentage changes.
* Add a regression test covering a missing-price gap that previously produced a false mean-reversion entry.

## Testing

* Focused regression test: 1 passed.
* `python -m py_compile`: passed.
* Whitespace check: passed.
* Ruff and Black were not available in the execution environment.
* Full repository test suite was not run.

## Compatibility and Risk

Low risk. The change is limited to the Strategy Development Manager template and preserves existing behaviour for complete price data.

## Related Issue

No existing issue identified for this bug.

### DIFF
--- a/agent/src/skills/strategy-dev-manager/templates/strategy_signal_engine.py
+++ b/agent/src/skills/strategy-dev-manager/templates/strategy_signal_engine.py
@@ -54,8 +54,10 @@ class SignalEngine:
             # Example: Mean-reversion strategy (Avramov & Chordia 2006)
 
             # Compute signal indicator
-            returns = df["close"].pct_change(self.lookback)
-            volatility = df["close"].pct_change().rolling(self.lookback).std()
+            returns = df["close"].pct_change(self.lookback, fill_method=None)
+            volatility = (
+                df["close"].pct_change(fill_method=None).rolling(self.lookback).std()
+            )
 
             # Normalized signal: return / volatility
             normalized = returns / volatility.replace(0, float("nan"))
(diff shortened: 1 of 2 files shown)


warren618:
Merged, and thank you — the failure mode is real and the regression test earns its keep.

Worth telling you what the sweep found, because it changes what "fixed" means here. `pct_change()` without `fill_method` is not two call sites, it is a class: pandas 2.x still defaults to `fill_method='pad'`, so a missing close becomes a **0.0% return that never happened** and the next bar reports a two-day move as one day. Neither is a bias with a known sign — the rolling volatility built on them lands either side of the truth depending on the data.

Sweeping `agent/src/skills` exhaustively found **24 call sites in 10 files**. Two of them were in the very files these PRs were editing:

- `multi-factor/example_signal_engine.py:80` — which #1399 also touches
- `ml-strategy/SKILL.md:66` — which #1392 also touches

and a third, `multi-factor/zoo_signal_engine.py`, backs the alpha zoo. A search for `pct_change()` does not see any of the positional forms (`pct_change(5)`), which is how they stayed invisible.

All 24 are fixed in `be2c09f2`, with a repo-wide guard so the class cannot reopen one file at a time. Nothing you did was wrong; the ask for next time is to list the siblings on the same axis before opening the PR, because five separate PRs each fixing one instance costs everyone more than one PR fixing the class.

(The equity-curve `pct_change().fillna(0.0)` in `backtest/` is deliberately untouched — the engine builds that series with a value on every bar and `metrics.py` documents the choice at the seam.)

PR #1398, opened 2026-09-08
fix(cross-market): preserve missing price gaps in volatility weights

Shizoqua:
## Summary

* Preserve missing price gaps when calculating cross-market volatility weights.

## Why

The strategy used pandas `pct_change()` with implicit filling, allowing missing closes to alter volatility estimates and cross-market allocation.

## Changes

* Disable filling in volatility return calculations.
* Update the skill example to match.
* Add regression coverage for missing-price gaps.

## Test Plan

* [ ] Full repository test suite not run.
* [x] New focused regression test passes.
* [x] `python -m py_compile` passes.
* [x] Whitespace check passes.

## Compatibility and Risk

Low risk. Limited to cross-market strategy weighting, with no live trading or broker changes.

## Related Issue

No existing issue identified.

### DIFF
--- a/agent/src/skills/cross-market-strategy/SKILL.md
+++ b/agent/src/skills/cross-market-strategy/SKILL.md
@@ -51,7 +51,7 @@ BTC daily vol ~ 5%, A-share daily vol ~ 1.5%. Without vol-adjustment, crypto eat
 def _vol_adjust(self, signals, data_map):
     vols = {}
     for code, df in data_map.items():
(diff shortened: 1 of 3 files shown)


warren618:
Merged. Same note as on #1397: this was one of 24 `pct_change()` call sites across 10 skill files, all swept in `be2c09f2` with a repo-wide guard, because the defect is a class rather than a pair of instances. Two of the remaining sites sat in files your other PRs were editing at the time. Detail on #1397.

PR #1399, opened 2026-09-08
fix(multi-factor): exclude assets without factor observations

Shizoqua:
## Summary

Exclude assets with no available factor observations from multi-factor ranking.

## Why

Assets with no valid factor values currently receive neutral scores and can still enter TopN selection. This can cause an asset with no usable factor data to displace one with valid observations.

## Changes

* Track whether each asset has at least one valid factor observation on each rebalance date.
* Exclude assets with no available factor values from ranking.
* Add a regression test covering an asset with entirely missing factor data.

## Test Plan

* [x] Focused regression test passes with the fix.
* [x] The same regression test fails against the previous implementation.
* [x] Python compilation check passes.
* [ ] Full repository test suite not run.

## Checklist

* [x] Change is limited to multi-factor selection and its regression test.
* [x] No live trading, broker, MCP, network, secret, or deployment behaviour changed.
* [x] DCO sign-off included.

### DIFF
--- a/agent/src/skills/multi-factor/example_signal_engine.py
+++ b/agent/src/skills/multi-factor/example_signal_engine.py
@@ -128,6 +128,7 @@ class SignalEngine:
 
             # 调仓日：截面排名
             composite_scores: Dict[str, float] = {}
+            observed_factors = {code: 0 for code in codes}
             for factor_name in factor_names:
                 raw_vals = {}
                 for code in codes:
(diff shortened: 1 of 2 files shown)


warren618:
Merged. Tracking observation counts per asset and ranking only what has one is the right shape — the previous `not np.isnan(s)` test on the composite score could not see the difference between "zero after z-scoring" and "no data at all", because `zscore_cross_section` had already turned the absence into a neutral 0.0.

Same note as on #1392: line 80 of the file you were editing still had `close.pct_change()` on pandas' padding default. Swept with the other 23 sites in `be2c09f2`.

PR #1400, opened 2026-09-08
fix(social-media): correct sentiment factor evaluation

Shizoqua:
## Summary

Fix social sentiment factor evaluation and missing-data handling.

## Why

The example could fail when `sentiment_norm` is missing, use the wrong forward date when rows are unsorted, and create artificial values by filling missing regression data with zero.

## Changes

* Use the computed normalized sentiment values.
* Sort dates before forward-return selection.
* Regress only on complete observations.
* Preserve missing residuals.
* Add regression tests.

## Test Plan

* [x] Targeted regression tests added.
* [x] Python compilation passed.
* [ ] Full test suite not run.

## Checklist

* [x] Focused change only.
* [x] No live trading or external service behaviour changed.
* [x] DCO sign-off included.

### DIFF
--- a/agent/src/skills/social-media-intelligence/SKILL.md
+++ b/agent/src/skills/social-media-intelligence/SKILL.md
@@ -756,13 +756,17 @@ def build_sentiment_factor(
         raw_data.groupby("date")["sentiment_score"]
         .transform(lambda x: (x - x.mean()) / (x.std() + 1e-8))
     )
+    factor_data = raw_data.assign(sentiment_norm=factor)
 
     # 2. Compute period-by-period IC
     ic_list = []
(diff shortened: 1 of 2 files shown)


warren618:
Merged. Three separate defects in one example and all three are real: the factor was recomputed and then never used (the loop read a `sentiment_norm` column the caller had to supply), the forward date came off input row order rather than date order, and `fillna(0)` before the regression fed the fit a column of invented zeros and then returned residuals for rows that had no observation.

The orthogonalization fix is the one I would highlight: returning a NaN-indexed series and filling only the valid rows keeps the missing rows missing all the way through, rather than handing downstream a residual computed from a zero that was never a measurement.

PR #1402, opened 2026-09-09
fix(stooq): name the override variable the config layer actually reads

chiww:
## Problem

When Stooq serves its anti-bot challenge page, the loader warns and tells the operator how to route around it. Both halves of that advice are unreachable as printed:

```
Reorder the chain via VIBE_TRADING_MARKET_DATA_ORDER_* or drop stooq from it.
```

**1. `VIBE_TRADING_MARKET_DATA_ORDER_*` is not read anywhere.** `source_order_env_var` (`registry.py:309`) builds `MARKET_DATA_ORDER_<MARKET>`, and `get_env_value` (`config/accessor.py:141`) is a plain `os.getenv` passthrough that adds no prefix — the `env_schema.py:227` aliases agree. Exporting the name as printed is a silent no-op: the chain keeps its default order and stooq stays at position 2 of `us_equity`, still serving challenge pages.

**2. "drop stooq from it" is rejected.** `is_valid_source_order` requires a permutation of the default chain, so removing a source fails validation and the default order is kept.

Verified live on 0.1.15:

```
$ MARKET_DATA_ORDER_US_EQUITY="yahoo,sina,eastmoney,yfinance,..." python -c ...
Ignoring invalid MARKET_DATA_ORDER_US_EQUITY=...: value must be a permutation
of the default chain [...]; keeping default order
stooq still in chain -> True
```

An operator hitting the challenge page has no working escape from the message alone.

## Fix

Point at moving stooq to the end (a legal permutation), name one concrete variable, and state the permutation constraint.

## Tests

Two tests pin the message to the registry rather than to a literal, so it cannot drift again.

Worth flagging for review: the env-var test compares **whole tokens**, not substrings. A containment check passes on the buggy string, because `VIBE_TRADING_MARKET_DATA_ORDER_*` contains `MARKET_DATA_ORDER_`. The test failed correctly only after switching to a regex over env-var-shaped words. Please keep it that way if this area is ever simplified.

- `agent/tests/test_stooq_loader.py` — 16 passed (14 existing + 2 new)
- `-k "stooq or registry or source_order or loader"` — 893 passed, 11 skipped
- `tools/ci_grep_gates.sh` — all gates pass

No behavior change outside the log message.

### DIFF
--- a/agent/backtest/loaders/stooq_loader.py
+++ b/agent/backtest/loaders/stooq_loader.py
@@ -179,8 +179,10 @@ class DataLoader:
                 logger.warning(
                     "stooq is serving an anti-bot challenge page instead of CSV "
                     "data; treating it as unavailable for the rest of this "
-                    "process. Reorder the chain via VIBE_TRADING_MARKET_DATA_ORDER_* "
-                    "or drop stooq from it."
+                    "process. Move stooq to the end of the chain via "
+                    "MARKET_DATA_ORDER_* (e.g. MARKET_DATA_ORDER_US_EQUITY); the "
+                    "override must be a permutation of the default chain, so a "
+                    "source can be reordered but not removed."
                 )
                 _challenge_warned = True
             return None
--- a/agent/tests/test_stooq_loader.py
+++ b/agent/tests/test_stooq_loader.py
@@ -219,3 +219,63 @@ class TestChallengePageDetection:
         assert stooq_loader._looks_like_challenge_page(self._CHALLENGE_HTML)
         assert not stooq_loader._looks_like_challenge_page(_CSV)
         assert not stooq_loader._looks_like_challenge_page("N/D\n")
+
+
+class TestChallengeWarningNamesRealOverride:
+    """The remediation advice must be executable as written (#1315 follow-up).
+
+    The warning tells the operator how to route around a challenge-blocked
+    stooq. Both halves of that advice are load-bearing: the env var has to be
+    the one ``registry`` actually reads, and the suggested edit has to survive
+    ``is_valid_source_order``. Pin them to the registry so the message cannot
+    drift away from the code again.
+    """
+
+    _CHALLENGE_HTML = (
+        "<html><head><title>One moment, please...</title></head>"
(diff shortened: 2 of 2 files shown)


warren618:
Merged. Both halves of the claim reproduce on main exactly as you wrote them, and I verified each independently before merging rather than taking the report on faith:

- `_SOURCE_ORDER_ENV_PREFIX` is `MARKET_DATA_ORDER_` (`registry.py:290`) and `get_env_value` is a bare `os.getenv(name, default)` with no prefix, so `VIBE_TRADING_MARKET_DATA_ORDER_*` was read nowhere.
- `is_valid_source_order` is multiset equality against the default chain, so "drop stooq from it" is rejected and the default order is kept.

Three things I checked past that:

**The tests bite.** I restored the buggy `stooq_loader.py` from main against your test file: both new tests fail, the other 14 pass. Your note about whole tokens versus substrings is the reason — I confirmed a containment check passes on the old string, since `VIBE_TRADING_MARKET_DATA_ORDER_*` contains `MARKET_DATA_ORDER_`. That regex stays.

**The new advice is executable as printed**, which is the actual bar for a fix to a message that was not. Ran it:

```
MARKET_DATA_ORDER_US_EQUITY="yahoo,sina,eastmoney,yfinance,tiingo,fmp,finnhub,alphavantage,longbridge,akshare,local,stooq"
-> active override: ['yahoo','sina','eastmoney','yfinance','tiingo','fmp','finnhub','alphavantage','longbridge','akshare','local','stooq']
```

Accepted and applied, stooq last.

**Full backend suite green** on the merge commit: 12877 passed, 30 skipped.

One process note, not about your diff: CI never ran on this PR — it sat at `action_required` because a first-time contributor's workflows need maintainer approval, which is why the checks section was empty rather than red. Nothing for you to do; flagging it so the silence is not read as a problem with the branch.

The part worth calling out is the shape of the report. A warning whose remediation advice is a silent no-op fails in the one situation it exists for, and it fails quietly — the operator does the thing, nothing changes, and there is no error to search for. Pinning the message to `source_order_env_var` rather than to a literal is what keeps this fixed. Thanks.

PR #1404, opened 2026-09-10
fix(correlation): handle unnamed series in Engle-Granger

Shizoqua:
## Summary

Fix Engle-Granger hedge-ratio extraction for unnamed pandas Series.

## Why

Unnamed Series cause the regression parameter to be indexed as `0`, while the example looks for `"x"`, resulting in a `KeyError`.

## Changes

- Give the regression input a stable name.
- Add regression coverage for named and unnamed Series.

## Test Plan

- [x] Reproduced the original failure.
- [x] Verified hedge-ratio extraction after the fix.
- [x] Added regression tests.

## Checklist

- [x] No protected areas changed.
- [x] No external services or live trading affected.
- [x] DCO sign-off included.

### DIFF
--- a/agent/src/skills/correlation-analysis/SKILL.md
+++ b/agent/src/skills/correlation-analysis/SKILL.md
@@ -239,7 +239,7 @@ def sector_clustering(
 ### Comparison of Three Linkage Methods
 
 | Method | Feature | Best Use Case | Weakness |
-|------|------|---------|------|
+|------|------|---------|--------|
 | Ward | Minimizes within-cluster variance, gives compact clusters | **Default recommendation**, stock-sector discovery | Works best for spherical clusters, weaker for irregular shapes |
 | Complete | Uses maximum pairwise distance, conservative | When high within-cluster similarity is required | Can produce elongated clusters |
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thank you. The fix is right: an unnamed `x` becomes column `0` after `add_constant`, so `params["x"]` raised `KeyError`, and your test covers both the unnamed and the named case.

Two things I added on top in bd29a187, both about the class of bug rather than your change:

- The test now `importorskip`s statsmodels, the way `test_quantlib_tool.py` does. statsmodels comes from the `stats` extra, so on a base install the test failed with `ModuleNotFoundError` instead of skipping; CI installs that extra, which is why it was green there.
- The reason the bug could exist at all is that the skill carried its own OLS fit while `src/quantlib/timeseries.py` already has tested `find_hedge_ratio` / `cointegration_test` / `adf_test` / `compute_half_life` (the quant-statistics skill imports them). The Engle-Granger and half-life examples now call those. `engle_granger_coint` keeps its result keys, so your test still exercises it, and it still passes for both names.

PR #1405, opened 2026-09-10
fix(loaders): synchronize cold registry initialization

### LINKED ISSUE #1403: [Bug] Lazy loader registry is not thread-safe: concurrent cold-start calls see an empty registry
## Description

**What happened:**

The lazy loader registry can report itself as initialized before loader
registration has actually completed.

`backtest.loaders.registry._ensure_registered()` sets `_registered = True`
before importing the loader modules. When several threads call it concurrently
on a cold process, one thread starts the imports while the other threads see
`_registered == True` and immediately return with `LOADER_REGISTRY` still empty.

This is observable without an LLM and without artificial delays.

In a natural cold-start test with 8 threads:

```text
worker=1 elapsed=    0.022 ms _registered=True loaders=0
worker=6 elapsed=    0.027 ms _registered=True loaders=0
worker=4 elapsed=    0.030 ms _registered=True loaders=0
worker=5 elapsed=    0.030 ms _registered=True loaders=0
worker=3 elapsed=    0.033 ms _registered=True loaders=0
worker=0 elapsed=    0.035 ms _registered=True loaders=0
worker=2 elapsed=    0.038 ms _registered=True loaders=0
worker=7 elapsed=  447.644 ms _registered=True loaders=27

min loaders seen: 0
max loaders seen: 27
NATURAL_RACE=YES

The functional consequence is that real concurrent market-data requests can
fail spuriously during cold start.

Using three concurrent calls for YPF.US, SPY.US, and CL=F:

Cold parallel process:

YPF.US  -> {"_unresolved": ["YPF.US"]}   in 0.009s
SPY.US  -> {"_unresolved": ["SPY.US"]}   in 0.010s
CL=F    -> success                       in 0.920s

Immediately afterwards, in the same process with the registry warm:

YPF.US  -> success   in 0.535s
SPY.US  -> success   in 0.657s
CL=F    -> success   in 0.694s

No configuration or data-source settings changed between the two batches.

This can also amplify latency in the built-in agent: failed readonly calls
cause recovery/retry rounds even though the requested market data is available.

What I expected:

_ensure_registered() should not return to any caller until loader
registration has completed. A caller that observes _registered == True
should also observe the fully initialized loader registry.

This also appears to be the contract documented at the top of
registry.py, which says callers of resolve_loader /
get_loader_cls_with_fallback should never see an empty registry regardless of
import order.

Current main still sets _registered = True before the module-import loop.

### Steps to Reproduce

Run this from a fresh container/process:

docker compose run --rm --no-deps \
  -w /app/agent \
  vibe-trading \
  /opt/venv/bin/python - <<'PY'
import threading
import time
from concurrent.futures import ThreadPoolExecutor

from backtest.loaders import registry

print("_registered initially:", registry._registered)
print("loaders initially:", len(registry.LOADER_REGISTRY))

workers = 8
barrier = threading.Barrier(workers)

def worker(i):
    barrier.wait()
    t0 = time.perf_counter()

    registry._ensure_registered()

    return (
        i,
        (time.perf_counter() - t0) * 1000,
        registry._registered,
        len(registry.LOADER_REGISTRY),
    )

with ThreadPoolExecutor(max_workers=workers) as ex:
    results = list(ex.map(worker, range(workers)))

for i, elapsed, registered, count in sorted(results, key=lambda x: x[1]):
    print(
        f"worker={i} elapsed={elapsed:.3f} ms "
        f"_registered={registered} loaders={count}"
    )

counts = [r[3] for r in results]

print("min loaders seen:", min(counts))
print("max loaders seen:", max(counts))
print(
    "NATURAL_RACE=YES"
    if min(counts) < max(counts)
    else "NATURAL_RACE=NO"
)
PY

A second functional reproduction uses the real market-data layer:

import threading
from concurrent.futures import ThreadPoolExecutor

from src.market_data import fetch_market_data

symbols = ["YPF.US", "SPY.US", "CL=F"]
barrier = threading.Barrier(len(symbols))

def fetch(symbol):
    barrier.wait()
    return symbol, fetch_market_data(
        codes=[symbol],
        start_date="2026-06-10",
        end_date="2026-09-10",
        source="auto",
        interval="1D",
        max_rows=5,
    )

with ThreadPoolExecutor(max_workers=len(symbols)) as ex:
    print(list(ex.map(fetch, symbols)))

On a cold process, some symbols may be returned as _unresolved.
Repeating the same requests after registration completes succeeds.

Which symbol wins the initialization race is naturally nondeterministic.

Root cause

The relevant shape in backtest/loaders/registry.py is:

global _registered
if _registered:
    return

_registered = True

for mod in _loader_modules:
    try:
        importlib.import_module(mod)
    except Exception:
        pass

Therefore _registered means "initialization has started", while callers treat
it as "initialization has completed".

Suggested fix

Protect lazy initialization with a lock and publish _registered = True only
after the import loop completes.

For example, conceptually:

_registry_lock = threading.Lock()

def _ensure_registered():
    global _registered

    refresh_source_order_overrides()

    if _registered:
        return

    with _registry_lock:
        if _registered:
            return

        for mod in _loader_modules:
            try:
                importlib.import_module(mod)
            except Exception:
                pass

        _registered = True

A concurrency regression test should verify that all callers return only after
the same complete registry is visible.

Error Logs

There is no exception required to reproduce the race.

The real market-data symptom is a false unresolved result:

YPF.US -> {"_unresolved": ["YPF.US"]}
SPY.US -> {"_unresolved": ["SPY.US"]}

The same requests succeed immediately afterwards once the registry is warm.

### Error Logs

```shell
There is no exception required to reproduce the race.

The real market-data symptom is a false unresolved result:

YPF.US -> {"_unresolved": ["YPF.US"]}
SPY.US -> {"_unresolved": ["SPY.US"]}

The same requests succeed immediately afterwards once the registry is warm.
```

### Interface

Docker

### LLM Provider

Not applicable. Reproduced without an LLM.

### Version

Vibe-Trading v0.1.15 commit cc54832cb50de29d14bb10097b18e08f0a843650  The same _ensure_registered() implementation is still present on current main as checked on 2026-09-10.

### Environment

Ubuntu 24.04 Docker 29.1.3 Docker Compose 2.40.3 Python 3.11.15 inside the official Vibe-Trading container

alanwilhelm:
## Summary

- Make concurrent cold-start loader lookups wait for registration to finish.

I picked this issue as my first contribution to Vibe-Trading to get familiar
with the codebase and contribution workflow.

## Why

Closes #1403.

The initialization flag is set before loader imports finish, so concurrent
cold callers can observe an empty registry. Reproduced on Python 3.11.15
with the Docker runtime lock.

## Changes

- Use a standard-library lock and publish `_registered` after imports finish.
- Preserve the warm fast path, environment refresh, and optional-import skips.
- Test both public lookup paths; keep Black formatting in a separate commit.

No dependency, fallback-order, broker-write, credential, or protected-area changes.

Risk: cold callers wait for imports. Reentrant loader imports could deadlock;
current imports were inspected and passed the real 27-loader cold-start check.
Rollback: revert the fix, restoring the original race.

## Test Plan

- Python **3.11.15 and 3.14.6**: **12,892 passed, 17 skipped each** in the
  CI-selected backend suite on fix commit `9d560d3a`.
- After formatting: **48 registry/source-order tests passed on each version**;
  ASTs match after docstring whitespace normalization. Full suite not rerun.
- Both regressions fail against the original initializer. All eight cold
  readers now see 27 loaders.
- Full-file Black, Ruff, syntax checks, and repository safety gates pass.

Hosted CI awaits maintainer authorization. Frontend, desktop/Windows,
cross-platform hash-lock jobs, and live E2E were not run locally.

<details>
<summary>Reproduction and verification details</summary>

Base: `f9cb061b3f3b983a62780985008044ceb90c2f81`.
Full-suite results are from `9d560d3a`; the formatting-only follow-up is `61f46241`.

**CI environments and command**

Each Python version used a separate environment:

```bash
uv venv --seed --python VERSION VENV
uv pip install --python VENV/bin/python -e '.[dev,openbb,stats,anthropic]'
```

`pip check` and imports of all four optional extras passed on both versions.
These environments resolve declared dependencies without the Docker lock.
With absolute temporary `VENV`, `TEST_HOME`, and `TEST_TMP` paths, from the
repository root:

```bash
env -i HOME="$TEST_HOME" PATH="$VENV/bin:/usr/bin:/bin" \
  PYTHONDONTWRITEBYTECODE=1 PYTHONPATH="$PWD/agent" \
  COVERAGE_FILE="$TEST_TMP/.coverage" \
  bwrap --unshare-net --bind / / --dev /dev -- timeout 900 \
  "$VENV/bin/python" -m pytest \
  --ignore=agent/tests/e2e_backtest \
  --ignore=agent/tests/test_e2e_harness_v2.py \
  --cov=agent --cov-report="xml:$TEST_TMP/coverage.xml" \
  --junitxml="$TEST_TMP/results.xml" -p no:cacheprovider --tb=short -q
```

The exclusions match `test.yml` and the agent contributor guide. External
networking was blocked for tests and subprocesses; loopback remained available.
No live broker or LLM calls were authorized.

The 17 skips cover credential/live checks, optional OCR, synthetic-panel
preconditions, missing-dependency paths when the dependency is installed, and
an opt-in baseline comparison. Python 3.11 reported 17 warnings; Python 3.14
reported 3,016, including NumPy/Pandas deprecations and SQLite resource warnings.
Warnings were not suppressed.

An initial Python 3.11 run failed 27 tests because global Python socket blocking
also blocked local integration sockets and numeric IP parsing. All 251 tests
in the affected files, then the entire suite, passed under network-namespace
isolation without source changes or weakened tests.

**Locked reproduction environment**

```bash
uv venv --python 3.11.15 --seed .venv
uv pip install --python .venv/bin/python --require-hashes -r requirements-lock.txt
uv pip install --python .venv/bin/python --constraint requirements-lock.txt -e '.[dev]'
.venv/bin/python -m pip check
```

Dependency consistency passed; a repeat hash-locked dry run would change nothing.
This matches the reported Python patch version and Docker runtime lock, not the
entire Docker image or optional channel installation.

The eight-thread cold-start smoke used real imports with no injected delays
and blocked import-time sockets: baseline returned zero loaders to seven
threads and 27 to one; the fix returned 27 to all eight. The patched smoke
also passed on Python 3.14.6. Standalone public-reader probes failed twice
on baseline and passed twice with the fix; the committed regressions also
fail when only the original initializer is restored in memory.

**Post-format and static checks**

Post-format tests on both isolated CI interpreters:

```bash
python -m pytest agent/tests/test_registry.py agent/tests/test_source_order_overrides.py -p no:cacheprovider --tb=short -q
```

Both returned 48 passes; Python 3.14 reported nine NumPy/Pandas deprecation
warnings. Earlier, `pytest tools/test_ci_env_var_gate.py -q` passed 23 tests
on each version. `bash tools/ci_grep_gates.sh` and CI syntax checks also passed
on both, retaining four existing environment-mutation warnings in
`src/providers/llm.py`.

After formatting, full-file static checks passed in the locked Python 3.11.15
development environment (Black 26.5.1):

```bash
.venv/bin/python -m black --check --no-cache --target-version py311 agent/backtest/loaders/registry.py agent/tests/test_registry.py
.venv/bin/python -m ruff check agent/backtest/loaders/registry.py agent/tests/test_registry.py
git diff --check
```

Local results do not establish that hosted CI has passed.

</details>

## Checklist

Both commits carry DCO sign-offs. No user-facing CLI, API, configuration, or
safety-boundary contract changed, so README updates are not applicable.
The initializer docstring states the waiting behavior.

### DIFF
--- a/agent/backtest/loaders/registry.py
+++ b/agent/backtest/loaders/registry.py
@@ -164,38 +164,69 @@ _NO_NETWORK_FALLBACK_SOURCES: frozenset[str] = frozenset(
 # that must be politely throttled; Finnhub/AlphaVantage/Tiingo/FMP are key-gated
 # REST fallbacks placed deeper in the chain.
 FALLBACK_CHAINS: dict[str, list[str]] = {
-    "a_share":   ["tencent", "mootdx", "eastmoney", "baostock", "akshare", "tushare", "local"],
-    "us_equity": ["yahoo", "stooq", "sina", "eastmoney", "yfinance", "tiingo", "fmp", "finnhub", "alphavantage", "longbridge", "akshare", "local"],
+    "a_share": [
+        "tencent",
+        "mootdx",
+        "eastmoney",
+        "baostock",
+        "akshare",
+        "tushare",
+        "local",
+    ],
+    "us_equity": [
+        "yahoo",
+        "stooq",
+        "sina",
+        "eastmoney",
+        "yfinance",
+        "tiingo",
+        "fmp",
+        "finnhub",
+        "alphavantage",
+        "longbridge",
+        "akshare",
+        "local",
+    ],
     # HK: tencent leads (no observed IP ban); akshare (Eastmoney-backed)
     # precedes the Yahoo-SDK family, which is blocked from mainland IPs;
     # tushare hk_daily is key-gated.
-    "hk_equity": ["tencent", "eastmoney", "yahoo", "futu", "akshare", "yfinance", "tushare", "longbridge", "local"],
+    "hk_equity": [
+        "tencent",
+        "eastmoney",
+        "yahoo",
+        "futu",
+        "akshare",
+        "yfinance",
+        "tushare",
+        "longbridge",
+        "local",
+    ],
     "india_equity": ["yahoo", "yfinance", "india_broker", "local"],
-    "kr_equity":   ["pykrx", "yahoo", "yfinance", "local"],
+    "kr_equity": ["pykrx", "yahoo", "yfinance", "local"],
     # TSX (.TO) / TSX Venture (.V): direct Yahoo first, SDK fallback second.
-    "ca_equity":   ["yahoo", "yfinance", "local"],
+    "ca_equity": ["yahoo", "yfinance", "local"],
     # UK (LSE .L): direct Yahoo first, SDK fallback second.
-    "uk_equity":   ["yahoo", "yfinance", "local"],
+    "uk_equity": ["yahoo", "yfinance", "local"],
     # Vietnam (.VN): Yahoo lists HOSE only — HNX and UPCOM are unsupported,
     # so those two are reachable only through the user's local files.
     "vietnam_equity": ["yahoo", "yfinance", "local"],
     # OKX first (native), then dedicated Binance, then generic CCXT / Yahoo.
-    "crypto":    ["okx", "binance", "ccxt", "yfinance", "local"],
+    "crypto": ["okx", "binance", "ccxt", "yfinance", "local"],
     # tushare led this chain while implementing no futures endpoint at all
     # (#1395): ``resolve_loader`` walks FALLBACK_CHAINS and never consults a
     # loader's ``markets`` set, so trimming that set alone would have left
     # tushare first in line, returning empty frames before akshare was ever
     # asked. akshare serves Chinese contracts off the token-free Sina daily
     # endpoints; a global contract has no network source and reaches ``local``.
-    "futures":   ["akshare", "local"],
-    "fund":      ["tushare", "akshare", "local"],
-    "macro":     ["akshare", "tushare", "local"],
+    "futures": ["akshare", "local"],
+    "fund": ["tushare", "akshare", "local"],
+    "macro": ["akshare", "tushare", "local"],
     # mt5 leads when a local MetaTrader 5 terminal is attached (Windows-only,
     # broker feed); otherwise it reports unavailable and the chain proceeds.
-    "forex":     ["mt5", "akshare", "yfinance", "local"],
+    "forex": ["mt5", "akshare", "yfinance", "local"],
     # Yahoo index symbols (^SPX, ^NDX, ^FTSE, ^VIX, ...): served verbatim by
     # the public chart endpoint, same as the =F/=X conventions.
-    "index":     ["yahoo", "yfinance", "local"],
+    "index": ["yahoo", "yfinance", "local"],
 }
 
 
@@ -394,7 +425,9 @@ def refresh_source_order_overrides() -> None:
             logger.warning(
                 "Ignoring invalid %s=%r: value must be a permutation of the"
                 " default chain %s; keeping default order",
-                source_order_env_var(market), raw, default,
+                source_order_env_var(market),
+                raw,
+                default,
             )
         FALLBACK_CHAINS[market] = default[:]
         _ACTIVE_SOURCE_ORDER_OVERRIDES.pop(market, None)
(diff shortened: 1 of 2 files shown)


warren618:
Merged — thank you, and thanks for the reproduction in #1403. Seven of eight cold callers seeing an empty registry is the kind of thing that otherwise only shows up as an occasional `_unresolved` that nobody can reproduce.

What I checked before merging: the lock is double-checked and `_registered` is published only after the import loop, so a caller that sees `True` sees the complete registry. Your concurrency test is deterministic (events, no sleeps), and it bites — with main's `registry.py` put back, 2 of its cases fail; with yours, 33/33 pass. CI had not run on this branch (first-time-contributor approval), so I ran the full backend suite on the merged tree before pushing: 12803 passed, 0 failed.

The Black pass on both files made the diff larger than the fix, but CONTRIBUTING asks for it and you said so up front, so that is fine as is.

Issue #1171, opened 2026-08-21
[Feature] A portfolio model that is not a live broker connection


warren618:
Three separate threads converged on the same missing piece today, so writing it down rather than letting it get decided as a side effect of whichever one lands first.

## What exists now

Everything that wants to reason about *the user's holdings* has to be handed them, or has to borrow a live broker session:

- `trading_positions` / `trading_account` read the **active connector profile** — one venue, live credentials, nothing persisted as a portfolio.
- `portfolio_risk_xray` takes `symbols` and optional `weights` **as tool arguments**. The concentration, VaR, drawdown and correlation analysis is all there; what is missing is any way for it to know what the user actually holds without being told again each time.
- Shadow Account is not this. It models `ShadowRule` distilled from profitable roundtrips — a strategy extractor over a trade journal, not a holdings ledger. Adjacent, and deliberately so.
- There is no Portfolio page in the Web UI.

So a user with holdings across three brokers, or holdings in no broker we connect to, can be analysed only by re-typing the basket into every tool call.

## Proposal

A first-class portfolio object that is a *record*, not a session:

1. **Positions with provenance.** Each position knows where it came from — a named connector profile, an imported file, or manual entry — and when that source was last observed. A portfolio may mix sources; it may not hide which is which.
2. **Snapshots, not a running balance.** A portfolio is a sequence of as-of observations. Nothing in it is derived from a transaction stream we did not see.
3. **A resolve step, not a merge.** The same instrument arriving from two sources is a conflict surfaced to the user, not silently summed or silently deduplicated.
4. **Existing analysis, unchanged.** `portfolio_risk_xray` and the rest keep their current signatures; the portfolio supplies the arguments. This is a source of holdings, not a second analysis stack.

## Constraints

- **Evidence, never a second account of record.** Settled in #1030 for exchange snapshots and it binds harder here, where a source may be a user-maintained file: a portfolio states what it observed and when, and never claims to be the authority on what someone owns.
- **A partial or stale import is an error, not a quietly short portfolio.** The data loaders already fail a truncated window rather than return one; a portfolio that silently drops three positions misstates allocation in the direction that looks fine.
- **Read-only, structurally.** Nothing in this layer can reach the order path. Rebalancing output, if it ever exists, is a proposal a human reads.
- Runtime state under `~/.vibe-trading`; no user-specific data in the repo.

## Open questions

[...]

Issue #1170, opened 2026-08-21
[Feature] Proposal: read-only extraETF portfolio import for allocation and rebalancing insights


KaiLuettmann:
## Problem or Motivation

extraETF is widely used by German-speaking ETF investors to consolidate portfolio holdings. Today, Vibe-Trading has connector-first account and position reads, but no way to bring an extraETF portfolio into those existing portfolio-analysis workflows. Users must manually re-enter holdings before they can assess allocation drift or receive rebalancing insights.

## Proposed Solution

Add an **opt-in, read-only extraETF portfolio import** that normalizes account and position data into Vibe-Trading's existing connector model. The initial feature would:

- expose imported portfolio snapshots through the existing account/position read surfaces;
- show source and snapshot freshness;
- generate transparent rebalancing proposals from a user-defined target allocation;
- remain strictly non-executing: no orders are sent to extraETF or any broker;
- avoid capturing passwords, cookies, or sessions.

The preferred implementation order would be:

1. a documented official API, if extraETF offers and permits one;
2. an official user-exported CSV import as the likely V1;
3. only if acceptable under extraETF's terms and technically stable, a separate user-triggered browser-extension sync as an opt-in V2.

Current research suggests that an official public third-party portfolio API may not be available, so an exported CSV appears to be the safer baseline. This needs confirmation before implementation.

## Questions for maintainers

1. Does this fit the project's connector-first product direction?
2. Should V1 be limited to an explicit local CSV import, with browser integration kept out of scope?
3. Are there existing privacy, storage, or connector conventions beyond the current read-only profiles that a contribution must follow?
4. Would a contributor PR be welcome after the supported extraETF access path and data schema are confirmed?

## Alternatives Considered

- **Browser extension:** could improve freshness but would rely on a private web UI and requires careful review of extraETF terms, permissions, and privacy. It should not block a safe import-based V1.
- **Direct scraping/login automation:** not proposed for V1 because it is brittle and risks handling sensitive session material.
- **Manual entry only:** works today but loses portfolio import and refresh benefits.

## Scope boundary

This proposal is for portfolio reads and analysis only. Automated trading, order submission through extraETF, and automatic broker execution are explicitly out of scope.

## Contribution

I am willing to prepare a PR once the maintainers agree on the integration path and scope.

warren618:
Thank you for asking before writing code, and for proposing the access path in the order you did — official API, then official export, then extension. That ordering is the reason this is easy to answer.

[...]

PR #1406, opened 2026-09-10
feat(portfolio): add a read-only extraETF export reader

cgycorey:
## Summary

- Adds a read-only reader for two of extraETF's four documented export surfaces: the **Investments** export becomes normalized positions, the **Transaktionen** export becomes movements. The other two documented surfaces (**Dividenden**, **Analyse → Holdings** / X-Ray) and the cash-transactions tab are refused by the same exact-header check, which names both known column sets and reports the diff against the nearer one. What is tested is the refusal *path*, using a header that matches neither set; the specific headers of those two surfaces are not in the samples, so they are refused by that mechanism rather than verified against their real headers — which is why those headers are requested in the thread.
- Fails closed on anything it cannot read without guessing. A row that cannot be aligned, a record folded across lines, unknown or missing columns, contradictory separators, unreadable numbers, numbers carrying more precision than the output can hold, empty required fields, corrupt text, one instrument listed twice under a single portfolio ID, and a movement whose transaction label or identifier is unreadable are all refused rather than skipped, so a bad row yields *no* positions instead of a short portfolio.
- No caller yet, by design — see "Deliberately out of scope".

Toward #1170. The reader is the half scoped to a contributor there; the container, aggregation and rebalancing surface stay with the maintainer.

The movements half is separable. The maintainer described the reader as "a normalized position list", which is the holdings export; this PR also parses the transactions export, because that is the file that answers his third question and it is the same reader and number grammar. If the positions half is wanted on its own, the transactions side drops out without touching the holdings path — say so and it becomes two PRs.

## Why

#1170 asked for an opt-in, read-only extraETF import. The maintainer split it: the provider-specific reader to the contributor, the portfolio container retained. This is that reader.

The export surface is documented, which is what the maintainer asked to confirm before code existed: [Daten exportieren: Transaktionen, Bestände und Dividenden als CSV](https://support.extraetf.com/hc/de/articles/28637399089436) names four export buttons — Investments, Transaktionen ("Wertpapier- **und Cash-Transaktionen**", as two tabs), Dividenden, and Analyse → Holdings. It also states that every export covers **all** portfolios, which makes the multi-`Portfolio ID` file in the samples the normal case rather than an aggregated edge case; every row is kept, with its `Portfolio ID` and `Portfolioname`, because merging two rows for one instrument is the outcome that is silently wrong in the direction that looks fine. V1 scope is the maintainer's list — symbol/ISIN, quantity, cost basis, currency, as-of — and each position carries all five.

The three questions the maintainer asked before code is written were answered in the issue thread from the anonymized samples, and the reader is built to those answers:

- **Identity is two shapes.** Securities carry an ISIN; crypto carries `BTC_to_EUR` in the holdings export and bare `BTC` with `Typ=Fremdwährung` in the transactions export. Resolving those onto one another is a container concern, so the reader reports the kind (`isin` / `crypto_pair` / `raw_symbol`) and the check-digit result rather than resolving.
- **There is no snapshot date.** The holdings export has no date column, so as-of is the file's mtime and is labelled `as_of_source="file_mtime"`. No position carries an observation time: `updated_at` is `None` and the mtime is reported as `source_mtime`, because a copy or a sync rewrites an mtime and it must never look like a fetch time.
- **Non-position rows are the trap.** A `Dividende` row carries its own non-zero `Anzahl` and `Preis`, so a reader that treats any row with a quantity as a holding invents one out of a dividend. Dividends and transfers become typed movements, and a transactions export can never be read as a portfolio at all.

## Changes

- `agent/src/portfolio/extraetf.py` — the reader. `read_export(path)` and `parse_export(text|bytes)`; payloads are private behind `require_positions()` / `require_movements()`, which refuse a mismatched export so a transactions file cannot masquerade as an empty portfolio.
- `agent/tests/test_extraetf_reader.py` — 79 tests over synthetic fixtures shaped after the anonymized samples in #1170.

Number parsing is German, and the rule is stated exactly rather than aspirationally: a comma is the decimal separator and a period may only be a thousands separator (`1.386,608` is 1386.608). US-ordered values (`1,234.56`) are refused because the German reading would be wrong by a thousandfold. A period-grouped value that is valid German (`1.900`) is read as German *by convention* — it could be 1.9 in a US-locale file and no parser can tell them apart. extraETF writes German, so German wins.

The emitted position uses the package's field names but deliberately does **not** route through `normalize_position`, because that stamps `updated_at` with the current time and defaults an unknown instrument class to `"stock"` — both fabrications a file import must not make. The three resulting divergences are commented in place.

## Test Plan

- [x] Existing tests pass — `pytest tests/ -q -k portfolio` → 117 passed, 5 skipped, 0 failed
- [x] New tests added — 79, one per fixed defect plus the format rules and the multi-portfolio case
- [x] Tested manually — verified against the four anonymized exports from #1170 (3 + 2 positions, 5 + 5 movements), and against a mutation fuzz of a 6-row export: 602 single-quote mutations produce zero accepted-but-wrong results
- [x] Single pass, linear: 28.6k rows/s stable from 100 to 50,000 rows (6 MB in 1.8s), and byte-identical output across repeated runs (the module reads no clock)
- [x] Every emitted value is a JSON primitive, so a position or movement survives the snapshot store without a converter
- [x] `ruff check` and `ruff format --check` clean, `tools/ci_grep_gates.sh` passes, and the module compiles under both matrix interpreters
- [x] Schema drift: renaming **or** dropping any of the 28 columns across the two formats is refused, and so is an extra column, so no field can silently degrade to `None` and no partial row set can be imported. A reordered header is accepted and does not mis-align values — name-based, not positional — which is the one form of drift that is not a format change
- [x] One instrument under two `Portfolio ID`s stays two positions with each row's portfolio context, which is the export's normal case ("Der Export umfasst jeweils alle Depots"). The test fails against a single-portfolio export, so it cannot pass by accident
- [x] **The full CI suite, run locally with CI's own command**: `pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q` → **12,890 passed, 34 skipped, 6 failed** in 283s. All six failures are the README agent-tool-count assertions described below, and nothing else in the suite is touched by this change
- [x] Python 3.11.15 and 3.14.5 — the CI matrix legs — each run this file's tests green (`uv venv --python 3.11|3.14` + `pytest agent/tests/test_extraetf_reader.py`)

One failure mode is worth naming so it is not attributed here. `tests/test_readme_counts.py::test_repo_tree_states_the_real_agent_tool_count` fails for all six READMEs on this machine (the tree line documents 107 agent tools; this environment registers 108). It is environment-dependent by construction: the test counts `build_registry().tool_names` in a child interpreter, and tool discovery imports every module under `src/tools/` and skips the ones that fail, so an optional dependency present here and not in CI's install set is worth exactly one tool. The evidence that this is the environment and not the tree is that **CI is green at this branch's base commit** `f9cb061b`, where the same assertion is in the same run (`CI` and `Desktop Windows` both `success`). This change adds no tool — it adds a module under `src/portfolio/` — so the count is identical before and after it, and the assertion is left alone rather than papered over.

## Checklist

- [x] No changes to protected areas — the change is confined to `agent/src/portfolio/` and its tests
- [x] No hardcoded values — no paths, keys or clock. The only constants are the wire precision (`0.00000001`, matching `normalize_position`) and the format's own shape rules; there is no staleness or tolerance threshold, since those are policy
- [x] Code follows CONTRIBUTING.md — commits carry `Signed-off-by`; fixtures are synthetic, no real portfolio data
- [x] Documentation — no user-facing surface is added (no CLI, Web, MCP or connector registration), so no README change. Wiring it up will need one.

## Deliberately out of scope

- **The container.** No source-kind hook, no settings entry, no Web surface, and no change to `normalize_position`. The maintainer retained that half; #1171 proposed it, was closed on 2026-09-06, and its remaining scope — the non-connector source kind and identity beyond ticker+market — was moved into #1170 "which is the only concrete consumer of it". This PR assumes no shape for it, which is what #1171 asked for by scoping #1170 to "a provider-specific reader precisely so it does not settle the questions above".
- **Rebalancing.** No proposal generation and no order path. The module cannot reach either: its imports are stdlib only, enforced by a test.
- **Identity resolution across the two exports.** `BTC_to_EUR` and `BTC` stay two reported shapes.

## Known limits and risk

- **EUR.** Every sample export is EUR-denominated and the valuation core supports USD, HKD and CNY, so `ensure_supported_currencies` refuses EUR and a container cannot value these rows today. Nothing here works around that; it needs a decision, which is why it is in the issue thread as well.
- **A file truncated exactly at a record boundary is undetectable.** The format carries no row count or footer, so a 5-of-6-row file is indistinguishable from a 5-row portfolio. Documented in the module, and a caller that knows the expected count is the only thing that can catch it.
- **An empty export is accepted as an empty portfolio**, deliberately: "never recorded", "unreadable" and "empty but valid" are different states and collapsing them would lose one.
- **No consumer.** This adds a module nothing calls yet.
- **File length.** The reader is 933 lines and its tests 875, over this repo's documented 800-line cap — they were 785 and 711 when the maintainers read them, and the strictness the review asked for is most of the difference. `service.py` on main is already 1004, so the cap is not enforced in practice. The files have not been split because a structural change mid-review would invalidate the two full reads already done, for no correctness gain; the seams are clean (the German number grammar and the identity grammar are each self-contained) if it is ever wanted.
- **Two of the four documented exports are out of scope.** The Dividenden export and the Analyse → Holdings (X-Ray) view each have their own column set and are refused: the refusal names both known column sets and reports the missing and unexpected columns against the nearer of them (`Transaktion` present selects the transactions set), so an unrecognised export fails loudly rather than being partially imported. The *mechanism* is tested with a header matching neither set; those two surfaces' real headers are not in the samples, so they are refused by construction rather than by a test of their own headers. The X-Ray view is the one worth revisiting first: it is the look-through decomposition, so it is the input that would change an allocation rather than restate it.
- **Risk:** read-only and inert. No network, no credentials, no writes, no order path. **Rollback:** delete the two new files; nothing else references them.

## Review history

Reviewed adversarially in two rounds by three independent reviewers per round, which found 13 real defects, all fixed on this branch. Two were capable of producing a plausible wrong portfolio: a stray quote folded two CSV lines into one record (a 4-position file became 3, with one instrument's quantity and value grafted onto another's identity — 40 of 4000 single-quote mutations), and `.upper()` used as an ISIN shape predicate fabricated 12-character ISINs out of 11-character inputs (`aß000000001` → `ASS000000001`, which also passed the check digit).

**Third round — maintainer review (2026-09-11).** `warren618` read it against #1170's fail-closed rule and found three paths that still accepted bad input silently, all fixed on this branch: a repeated `(Portfolio ID, instrument)` was imported twice (now refused, naming both rows, while one instrument under two different Portfolio IDs stays two positions); an unknown `Transaktion` was carried through with `movement_kind=None` and a blank `ISIN` produced `instrument_id=None` (both now refused, with the five documented labels and the bare-symbol crypto transactions shape still reading); and the number probe's quantization result was discarded, so `0,000000004` became an empty position and a ninth decimal place was silently rounded (the quantized value is now compared against the original, and any difference is refused). His smaller points are in as well: `Währung / Krypto` no longer implies `asset_type="crypto"`, since it covers FX pairs like `USD_to_EUR` too; the export-level field is `source_mtime` rather than `as_of`, so a file's mtime cannot be read as a snapshot date; and the header refusal names both known column sets. On scope, the two maintainers differed — `warren618` offered to take the movements half out of this PR, while `he-yufeng` approved it in place and said no split is needed — so it stays, with the unknown-label and identity checks making it strict enough to be worth keeping.

**Fourth round — adversarial pass over the fixes themselves (2026-09-11).** Attacking the fixes rather than the reviewer's examples found eight more, all fixed on this branch: the thousands-grouping rule never checked the leading group, so `1234.567` — a US decimal — was read as `1234567`, a thousandfold misread one function from the one the maintainer had just flagged; a movement's `ISIN` bypassed the corruption check every other text column gets, so a symbol carrying NUL/DEL/C1 was emitted verbatim as an instrument identity; duplicate detection compared identifiers case-sensitively while the pair shape is matched case-insensitively, so `BTC_to_EUR` and `btc_to_eur` under one portfolio ID were two positions that double-count; the control-character rule covered only C0, so DEL and C1 passed; the quantization probe still allowed a value to change magnitude in `float()`; provenance could contradict itself (a timestamp with an `unavailable` origin); and two items were drift from the `as_of` rename. Tests 61 → 79, including precision pinned in every numeric column of both formats rather than only in `Anzahl`.

### DIFF
--- a/agent/src/portfolio/extraetf.py
+++ b/agent/src/portfolio/extraetf.py
@@ -15,29 +15,69 @@ Three properties are deliberate and load-bearing.
 **A transactions export can never be read as a portfolio.** Its rows are
 movements — ``Kauf``/``Verkauf``/``Dividende``/``Einbuchung``/``Ausbuchung`` —
 not holdings. An allocation computed from trades looks perfectly plausible and
-is wrong, so the two formats are told apart by their header and
-:meth:`ExtraEtfExport.require_positions` refuses anything that is not a
-holdings export. Movements carry an explicit ``movement_kind`` so a dividend or
+is wrong, so the two formats are told apart by their header and positions are
+reachable only through :meth:`ExtraEtfExport.require_positions`, which refuses
+any other export. Movements carry an explicit ``movement_kind`` so a dividend or
 a transfer is never mistaken for a trade.
 
 **A holdings export carries no date.** The format has no snapshot column, so
 freshness can only come from the file's own modification time, and the export
-says so through ``as_of_source``. This reader never claims an observation time
-it did not read.
+says so through ``as_of_source``. No position carries an observation time: a
+row's ``source_mtime`` is the file's modification time, kept under a name no
+connector ever fills with a real fetch time, and ``updated_at`` is ``None``
+rather than a copy of it. Copying or syncing an export rewrites that mtime, so
+``source_mtime`` reports when the *file* last changed and nothing more.
 
 **Validation fails closed.** A row this reader does not recognise raises
 :class:`ExtraEtfFormatError` instead of being skipped: an import computed from
 90% of a portfolio reads as entirely plausible, whereas a refused import is
-merely annoying. Numbers are parsed as German — ``1.386,608`` is 1386.608 — and
-a value that cannot be read unambiguously is refused rather than guessed.
-``1,234.56`` is US-formatted; reading it as German would be wrong by three
-orders of magnitude, so it is refused instead.
+merely annoying. That covers ragged rows, records that span more than one
+physical line, unknown or missing columns, unreadable or contradictory numbers,
+empty required fields, and text fields carrying control characters.
+
+The number rule is precise, because it is the one that can change a position's
+size by orders of magnitude. A comma is the decimal separator and a period may
+only be a thousands separator, so ``1.386,608`` is 1386.608 and ``12,3456`` is
+12.3456. A value whose separators are US-ordered (``1,234.56``) is refused,
+because reading it as German would be wrong by a thousandfold. A period grouping
+that is not a plain thousands separator (``1.23``, ``1234.56``) is refused too,
+because it is neither a valid German number nor safely a US one. What is *not*
+refused is a period-grouped value that is valid German — ``1.900`` is 1900 — and
+that reading is a convention of the format, not a deduction: ``1.900`` could be
+1.9 in a US-locale file, and no amount of parsing can tell the two apart.
+extraETF writes German, so German wins, and a file that had been round-tripped
+through a US-locale spreadsheet is caught by the *other* values in it rather
+than by this one.
 
 Rows are read by column name, so a reordered export cannot mis-align a value
 into the wrong column. Rows are **not** merged or de-duplicated: two rows for
 the same instrument in one portfolio are returned as two positions, because
 summing them silently and dropping one silently are both wrong, and which one
 the user meant is not this module's call.
+
+The emitted position uses the portfolio package's own field names so a container
+can consume it, and carries the export's remaining columns alongside them. It
+does **not** route the row through
+:func:`src.portfolio.normalization.normalize_position`, which is the obvious way
+to guarantee one wire shape, because that function stamps ``updated_at`` with
+the current time — an observation time this reader never made — and defaults an
+unknown instrument class to ``"stock"``. Both are fabrications a file import
+must not make, so the mapping is done here and the divergences are the three
+shown above: ``quote_symbol``'s None, ``asset_type``'s None, and ``updated_at``
+against ``source_mtime``.
+
+Values are carried as the export reports them rather than repaired: a zero or
+negative quantity, a zero FX rate, and two different portfolio names under one
+portfolio ID all pass through, because a reader that silently corrects its input
+is a reader whose output no longer describes the file. What is refused is input
+that cannot be read at all.
+
+Two limits are inherent to the format and belong in the caller's hands. A file
+truncated exactly at a record boundary is indistinguishable from a short export,
+because neither format carries a row count or a footer; a caller that knows how
+many positions to expect is the only thing that can catch that. An export with a
+header and no rows is read as a valid empty portfolio, because an empty
+portfolio and an unreadable file are different states.
 
 The module writes nothing, holds no credentials, and touches no network. It
 only reads a local file the user exported themselves, and it has no path to any
@@ -95,6 +135,7 @@ _TRANSACTIONS_COLUMNS: tuple[str, ...] = (
 ExportKind = Literal["holdings", "transactions"]
 
 _ISIN_PATTERN = re.compile(r"^[A-Z]{2}[A-Z0-9]{9}[0-9]$")
+_WIRE_QUANTUM = Decimal("0.00000001")
 _CRYPTO_PAIR_PATTERN = re.compile(r"^[A-Z0-9]{2,12}_to_[A-Z0-9]{2,12}$", re.IGNORECASE)
 _CURRENCY_PATTERN = re.compile(r"^[A-Z]{3}$")
 _DATE_PATTERN = re.compile(r"^(\d{2})\.(\d{2})\.(\d{4})$")
@@ -126,10 +167,14 @@ class ExtraEtfFormatError(ValueError):
 class ExtraEtfExport:
     """One parsed export, with its provenance attached.
 
+    The payloads are deliberately private. An export cannot hold both positions
+    and movements, and reaching for the wrong one used to yield an empty tuple
+    rather than an error — which turns "this is a transactions export" into "this
+    portfolio is empty", the exact plausible-and-wrong outcome this module exists
+    to prevent. Both accessors refuse a mismatched export instead.
+
     Attributes:
         kind: Which of the two extraETF formats this file is.
-        positions: Holdings rows. Always empty for a transactions export.
-        movements: Transactions rows. Always empty for a holdings export.
         as_of: ISO-8601 observation time, or ``None`` when it could not be read.
         as_of_source: Where ``as_of`` came from — ``"file_mtime"`` for a file
             read from disk, ``"unavailable"`` for bare text with no timestamp.
@@ -140,11 +185,11 @@ class ExtraEtfExport:
     """
 
     kind: ExportKind
-    positions: tuple[dict[str, Any], ...]
-    movements: tuple[dict[str, Any], ...]
     as_of: str | None
     as_of_source: Literal["file_mtime", "unavailable"]
     source_path: Path | None
+    _positions: tuple[dict[str, Any], ...] = ()
+    _movements: tuple[dict[str, Any], ...] = ()
 
     def require_positions(self) -> tuple[dict[str, Any], ...]:
         """Return holdings positions, refusing any other export outright.
@@ -161,7 +206,21 @@ class ExtraEtfExport:
                 "this is a transactions export and holds movements, not positions; "
                 "importing it as a portfolio would build an allocation out of trades"
             )
-        return self.positions
+        return self._positions
+
+    def require_movements(self) -> tuple[dict[str, Any], ...]:
+        """Return movements, refusing a holdings export outright.
+
+        Returns:
+            The parsed movements, possibly empty for an account with no history.
+
+        Raises:
+            ExtraEtfFormatError: If this export is a holdings export, whose rows
+                are positions rather than movements.
+        """
+        if self.kind != "transactions":
+            raise ExtraEtfFormatError("this is a holdings export and holds positions, not movements")
+        return self._movements
 
 
 def read_export(path: str | Path) -> ExtraEtfExport:
@@ -191,9 +250,7 @@ def read_export(path: str | Path) -> ExtraEtfExport:
     except OSError as exc:
         raise ExtraEtfFormatError(f"cannot read export {resolved}: {exc}") from exc
     as_of = datetime.fromtimestamp(modified, tz=timezone.utc).isoformat()
-    return parse_export(
-        data, source_path=resolved, as_of=as_of, as_of_source="file_mtime"
-    )
+    return parse_export(data, source_path=resolved, as_of=as_of, as_of_source="file_mtime")
 
 
 def parse_export(
@@ -222,28 +279,23 @@ def parse_export(
     """
     if as_of is None and as_of_source != "unavailable":
         raise ValueError(
-            "as_of_source cannot claim where an observation time came from when "
-            "there is no observation time"
+            "as_of_source cannot claim where an observation time came from when there is no observation time"
         )
     text = _decode(data)
     kind, header, rows = _read_rows(text)
     if kind == "holdings":
-        positions = tuple(
-            _parse_holding(header, cells, number, as_of=as_of) for number, cells in rows
-        )
+        positions = tuple(_parse_holding(header, cells, number, as_of=as_of) for number, cells in rows)
         movements: tuple[dict[str, Any], ...] = ()
     else:
         positions = ()
-        movements = tuple(
-            _parse_movement(header, cells, number) for number, cells in rows
-        )
+        movements = tuple(_parse_movement(header, cells, number) for number, cells in rows)
     return ExtraEtfExport(
         kind=kind,
-        positions=positions,
-        movements=movements,
         as_of=as_of,
         as_of_source=as_of_source,
         source_path=Path(source_path) if source_path is not None else None,
+        _positions=positions,
+        _movements=movements,
     )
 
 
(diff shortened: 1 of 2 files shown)


warren618:
Thank you for this — and for doing the part the issue asked for first. Answering the three schema questions from the samples, and finding the support article that documents four export surfaces rather than two, was worth as much as the parser.

I ran the reader on all four gist files and on a set of adversarial variants. The core is solid: the export type is detected from an exact header match, German numbers are parsed strictly, ISINs and crypto pair labels are kept apart with no invented join between `BTC_to_EUR` and `BTC`, and a cash row, a `Summe` subtotal, a missing column, a preamble line, a repeated header and dividend- or X-Ray-like headers are all refused. It is read-only by construction — standard library only, and the only I/O is reading the file.

Three paths still accept bad input silently, and that is the one rule #1170 was strict about:

1. **An exact duplicate row imports twice.** `parse_export` builds one position per row, and `test_duplicate_rows_are_preserved_rather_than_merged` pins that. Adding a copy of row 1 to the broker sample returns `IE00AAAA0000` twice under the same Portfolio ID, so anything that sums positions double-counts it. A repeated `(portfolio_id, instrument_id)` inside one holdings export should raise; the same ISIN under two different Portfolio IDs is legitimate and should stay two rows.
2. **The transactions half accepts row shapes it does not know.** An unknown `Transaktion` gets `movement_kind=None` and a blank ISIN becomes `instrument_id=None`, so an `Einzahlung` or `Zinsen` row passes. The "cash tab is refused" claim also rests only on the header comparison, against headers none of the tests use. You offered to drop the movements half — I would take that offer: the issue scoped a position list, and a smaller first PR is easier to get exactly right. If it stays, an unknown `Transaktion` has to raise.
3. **A non-zero quantity can round to zero.** The precision check in `_number` quantizes only to catch an overflow and then discards the result, so `Anzahl=0,000000004` passes the check and later comes out as 0, and `1,123456789` is rounded. Compare the quantized value with the original and raise the same error when they differ.

Smaller:

- `Währung / Krypto` maps to `crypto`, so a fiat `USD_to_EUR` holding would be labelled crypto — the kind of guessed asset class the module says elsewhere that it will not make. Leave it unclassified unless the row says crypto.
- The export-level `as_of` is the file's mtime while its docstring calls it an observation time. Positions correctly carry `updated_at=None`; renaming the export field to `source_mtime` would stop anyone downstream reading it as a snapshot date.
- The refusal message names only the nearest format; #1170 asked for it to name the known sets.
- Before this is wired into anything, an `Anzahl × Kurs ≈ Wert` cross-check (with a tolerance — one gist row is off by 0.026%, which looks like an anonymisation artefact) would catch a file whose numbers were reformatted and read 1000× off. Not needed for this PR.
- My earlier comment in #1170 pointed at #1171 for the container-side hook. That was stale: #1171 was closed on 2026-09-06 and its scope folded into #1170. Nothing calls this module yet, which is fine as a first step — wiring it into the portfolio service is ours.

@KaiLuettmann — no rush at all, but when you have time, three samples would settle what the current ones cannot: (a) one row where `Währung` is not EUR, so we know which currency `Kaufpreis`, `Aktueller Kurs` and `Aktueller Wert` are in and what `Wechselkurs` means; (b) the complete-portfolio view you offered, to see whether it adds subtotal rows and whether one Portfolio ID can list the same ISIN twice (that decides point 1); (c) just the header line of the Dividenden and Analyse → Holdings exports, so the refusal is tested against real headers rather than guessed ones.

he-yufeng (approved):
Approved after a full read of the module and tests, plus my own probes on top of the suite.

What I verified beyond the 55 new tests (green, along with the 28 existing portfolio tests):

- ISIN checksum against real instruments: US0378331005 and DE0008404005 pass, a flipped check digit fails. Reporting rather than enforcing it is the right call given the crypto pairs in the same column.
- The number grammar does what the docstring claims. `1.386,608` is 1386.608, `1.234.567,89` is 1234567.89, `-1.234,5` is -1234.5, and `1,234.56`, `12.34`, `,5`, `1,2,3` are all refused. Refusing US-ordered separators instead of guessing is exactly the property a position-size parser needs.
- The fail-closed seams hold together: ragged rows, folded records, blank-but-wide rows, and header drift all raise `ExtraEtfFormatError`, and `require_positions()` on a transactions export refuses instead of yielding an empty portfolio. That last one is the trap that makes naive imports plausible-and-wrong, and it is handled at the type boundary rather than by convention.
- `read_export` stats the descriptor it read from, not a later `stat` on the path. Small thing, easy to get wrong, correct here.

On scope: the transactions half goes past the narrow letter of the #1170 split, but it is the same grammar, it answers the date-carrying question the holdings export cannot, and keeping it does not weaken the holdings path. Welcome here; no need to split the PR.

The deliberate `normalize_position` bypass (no fabricated `updated_at`, no guessed `asset_type`) is the right trade. When I wire the container side, the reader's `None`s are the honest inputs I want.

One non-blocking observation: `1.900` reading as 1900 by convention is documented in three places and consistent everywhere, which is all a locale convention can ask for. A caller that later ingests non-extraETF German-locale files should not inherit that assumption blindly, but that is a future caller's problem, not this PR's.

cgycorey:
All three are fixed on the branch, along with the smaller items — and I ran a second adversarial pass over the fixes themselves, which found more, because the class of bug you identified has siblings. Everything below is a repro I ran against the pushed revision.

**Your three.**

1. A repeated `(Portfolio ID, instrument_id)` now raises and names both rows: `row 3: IE00AAAA0000 is listed twice under portfolio '1000001' (first at row 2)`. Your counter-case is the normal case and stays two positions.
2. An unknown `Transaktion` raises, and a blank or whitespace `ISIN` raises instead of yielding `instrument_id=None`. The five documented labels and the bare-symbol crypto shape still read, so the Bitpanda transactions sample is unaffected. On your offer to drop the movements half: `he-yufeng` approved it in place and said no split is needed, so I kept it and made it strict — say the word and it splits.
3. The probe's quantized result is now compared against the original and any difference raises, so `0,000000004` fails instead of becoming an empty position. `1,12345678` and `0,00000001` still pass, because a rule that refused its own boundary would be useless.

**What a second pass found — the same class, not the same examples.** Attacking the fixes rather than your examples turned up eight more, each fixed and pinned:

- **The thousands rule never checked the leading group.** `1234.567` — a US decimal — read as `1234567`: a thousandfold misread of exactly the kind your point 3 is about, one function away from it. A first group longer than three digits is now refused, since German never writes four digits before the first separator.
- **A movement's `ISIN` bypassed the corruption check** every other text column gets, so `BT\x00C` was emitted verbatim as an instrument identity. The identifier cell is checked now, and an identifier that is neither an ISIN, a pair label, nor a plain symbol is refused rather than carried.
- **Duplicate detection was case-sensitive while the pair shape is not**, so `BTC_to_EUR` and `btc_to_eur` under one portfolio ID were two positions that double-count — your point 1, reachable by case alone. Comparison folds case; the emitted label is untouched.
- **The control-character rule covered only C0**, so DEL and C1 passed.
- **The quantization probe still let a value change magnitude in `float()`**: a seventeen-digit quantity arrived different.
- **Provenance could contradict itself**: a timestamp arriving with an `unavailable` origin was accepted, so the export could report a source time it also claimed not to have.

Two more were drift from my own rename: `market` was missing from the divergence list, and a docstring still named `as_of`.

**Two of your asks were already satisfied in the code but pinned by no test**, which an item-by-item audit of your review caught: that the header refusal names *both* known column sets rather than only the nearer one, and that a `Währung / Krypto` row carries no asset class (your `USD_to_EUR` case). Both have a test now, and `_describe_identifier` documents that the kind describes an identifier's shape and never its asset class, which is where that decision lives.

**One thing that is yours to decide rather than mine: the files are now over CONTRIBUTING's 800-line cap** — 933 for the reader and 875 for its tests, where they were 785 and 711 when you read them. The strictness you asked for is most of the difference. `service.py` on main is already 1004, so the cap is clearly not policed, and I have not split them: a structural change mid-review would invalidate the two full reads you both did, for no correctness gain. The seam is clean if you want it — the German number grammar and the identity grammar are each self-contained — and I will split on request or as a follow-up.

Tests 55 → 79, and the suite is green on the pushed revision.

warren618:
On the 800-line cap: no split needed here. The number grammar and the identity grammar are separable, as you say, but splitting mid-review would only invalidate two full reads for no correctness gain; if the reader grows when the container side lands, that is the natural point to split. Your last commit, pinning the two asks that had no test, is merged too.

Merged — thank you, and thanks to @he-yufeng for the independent review.

I checked each review point against the code rather than the summary, and then undid each fix, one at a time, to see whether a test would notice: removing the duplicate check; keying it without the portfolio ID (the other direction — the same ISIN under two portfolios has to stay two rows); dropping the case fold; accepting an unknown `Transaktion` or a movement with an empty ISIN; discarding the quantize rounding again; removing the float round-trip check; restoring `Währung / Krypto` → `crypto`; accepting a leading group longer than three digits; letting `source_mtime` and `as_of_source` disagree; letting a C1 character into a text field or a control character into the ISIN column; and letting a malformed bare symbol through. Each of those 13 reversions fails at least one of your tests. CI is green on both Python legs, and the full backend suite on the merged tree passes (12,896 passed, 0 failed).

Nothing calls the reader yet, as the PR says, so #1170 stays open for the container side. The three samples I asked @KaiLuettmann for in the review still matter for that step: they decide how a non-EUR row and the Dividenden / X-Ray headers are handled.

Issue #1367, opened 2026-09-06
[Tracker] Broker connector requests: what a new broker needs before it can be supported


warren618:
This is the single tracking issue for **new broker connector requests**, replacing the one-issue-per-broker threads that were drifting without a decision. Requests consolidated here: Scalable Capital (#1360), Upbit / Korea Investment & Securities / Toss Securities (#1364), mStock / Mirae Asset (#542).

## What decides whether a broker can be supported, and how far

Every connector lands in one of three tiers, and which tier is not our preference — it is a property of what the broker's own API exposes:

| Tier | Requirement | What ships |
|---|---|---|
| Read-only | A documented read API (REST or an official MCP server) | `trading_account` / `trading_positions` / `trading_orders` / `trading_quote` / `trading_history`, plus portfolio aggregation when the profile advertises `account.read` + `positions.read` |
| Paper | A sandbox, or a local simulation under a hard guard | The above plus paper order placement, which skips the mandate gate because it cannot reach a real account |
| Bounded live | **A structural runtime paper/live discriminator** the *server* reports — account-id format, host separation, a demo flag, or a trade-environment field | Live order placement through the mandate gate, kill switch, and audit ledger |

The third row is the one that decides most of these. A broker that exposes no server-reported discriminator is capped at paper and read-only, and its `place_order` / `cancel_order` hard-refuse any non-paper config at the first line. That is not a policy we can waive per broker: a config flag the agent could flip is not a boundary. Longbridge, Dhan, Shoonya and Zerodha all sit at that cap today; Trading212 goes further and refuses order placement entirely.

## What a request needs before it can be picked up

Anyone proposing a broker — including the three consolidated here — please post:

1. **The read surface**: a link to the API docs, or for an MCP server the `tools/list` output (tool names and argument schemas, no account data).
2. **The discriminator answer**: does the API expose a sandbox, or any field in an account/session response that tells a live account from a test one? A "no" is a useful answer — it sets the tier, it does not disqualify the connector.
3. **Auth shape**: OAuth, API key, or session token, and whether it can be issued for a read-only scope.

`agent/src/trading/connectors/robinhood/` is the template for a remote-MCP broker and `agent/src/trading/connectors/alpaca/` for a direct SDK; a connector is a `sdk.py` + `profiles.py` + `classification.py` triple plus its entry in `src/trading/profiles.py`.

## Status

[...]

PR #1407, opened 2026-09-10
feat(trading): add KIS (한국투자증권) broker connector

as950118:
## Summary

Adds a `broker_sdk` connector for Korea Investment & Securities (KIS, 한국투자증권) KOSPI/KOSDAQ equities via the official KIS Developers REST API. Part of the consolidated broker-connector tracker: HKUDS/Vibe-Trading#1367.

Per the tier decided there:

- **Discriminator**: genuine host+port separation confirmed against the docs — 모의투자 (paper) is `openapivts.koreainvestment.com:29443`, 실전투자 (live) is `openapi.koreainvestment.com:9443`, each with its own app key/secret and OAuth2 client-credentials token.
- **Tier**: paper (places real orders against KIS's own paper-trading server — not a local simulation) + read-only live. No live-trade profile in this PR (see "Out of scope" below), matching how Zerodha/Dhan/Shoonya shipped their first connector into a new market.
- **Auth**: OAuth2 client-credentials (`POST /oauth2/tokenP`), token cached with a safety margin since KIS rate-limits issuance to once/minute.

## Affected areas

- New: `agent/src/trading/connectors/kis/{__init__,classification,profiles,sdk}.py`
- Anti-drift: `kis` added to `_LIVE_CAPABLE` in `test_paper_capped_connectors_refuse_live.py` (genuine discriminator, unlike the paper-capped brokers), portfolio compatibility tier (`experimental`), broker count/table across all six locale READMEs (14 → 15)
- New: `agent/tests/test_sdk_connectors_kis.py`

## Deliberately out of scope

Live order placement (a `kis-live-trade` profile) is not wired. KIS's host separation would technically support it the way Alpaca's does, but that needs a new `AssetClass.KR_EQUITY` in `src/live/mandate/model.py` plus wiring in `src/trading/service.py`'s `_CONNECTOR_INSTRUMENT` map — a deliberate follow-up once the paper path has seen real use, not guessed at here.

## Caveats

TR_ID transaction codes and response field names were sourced from the official example repo ([koreainvestment/open-trading-api](https://github.com/koreainvestment/open-trading-api)) and public docs — I don't have a KIS account to verify order placement/cancellation end-to-end. Two bugs a self-run code review caught before this PR are already fixed and covered by tests: a fractional cancel quantity (e.g. `0.5`) silently truncating to `0` and flipping into a full-quantity cancel, and `get_historical_bars(limit=0)` returning the entire fetched history instead of nothing (a `[-0:]` slicing bug).

Portfolio-page value aggregation doesn't support KRW yet (`SUPPORTED_VALUE_CURRENCIES` in `src/portfolio/compatibility.py` covers USD/HKD/CNY only) — same pre-existing limitation India/INR already has. The connection and reads work; aggregate valuation raises a clear `PortfolioContractError` until that's extended separately.

## Test plan

- [x] `pytest agent/tests/test_sdk_connectors_kis.py agent/tests/test_sdk_connectors.py agent/tests/test_paper_capped_connectors_refuse_live.py agent/tests/test_portfolio_compatibility.py agent/tests/test_readme_counts.py -q` — all pass
- [x] Full suite (`pytest agent/tests -q`) — 12791 passed, 97 skipped, 0 failed
- [ ] Order placement/cancellation against a real KIS 모의투자 account (I don't have one — flagged above for reviewer awareness)

## Risk

No live-money path is exposed by this PR (paper-only order placement, live is read-only). Credentials are never logged; `app_secret` is redacted in `_public_config`. Rollback is a plain revert — no other connector or shared safety path is touched beyond the additive registrations listed above.


warren618:
Thank you — this matches the tier from #1367 exactly: paper through KIS's real 모의투자 host, read-only live, and no `kis-live-trade` profile. The order-path TR_IDs (`VTTC0012U` / `VTTC0011U` for paper, `TTTC…` for live, `VTTC0013U` for cancel) match the official `open-trading-api` examples, which was the part I was most worried about. The app secret is redacted from returned config and tokens are cached per environment, so paper and live tokens cannot mix.

Four things before it can land:

1. **Paper cancel cannot work through the service.** `service.cancel_order` calls `module.cancel_order(config, order_id, symbol=symbol)`, but KIS's `cancel_order` returns `order_branch (KRX_FWDG_ORD_ORGNO) is required` when the branch is empty, and nothing in the agent tool can pass it — so an agent can place paper orders it can never cancel. When `order_branch` is empty, look it up from `get_open_orders` (you already map `ord_gno_brno` to `order_branch` there). The cancel body is also missing `EXCG_ID_DVSN_CD`, which `place_order` sends and the official `order_rvsecncl` example marks as required.
2. **The paper/live decision should also be made inside the connector.** The service routes on `profile.environment`, but the URL and TR_ID come from `cfg.environment`, and `"profile"` is in `_OVERRIDE_KEYS` — so a config override can make a paper-profile order resolve to the live host and a `TTTC` TR_ID without the service noticing. Nothing forwards that override today, so it is not reachable, but the red line is that the guard is structural: please add `if not cfg.is_paper: return {"status": "error", ...}` as the first statement after loading the config in both `place_order` and `cancel_order` (the same shape as Longbridge and Upbit), and drop `profile` from `_OVERRIDE_KEYS`. When a live-trade profile comes later, removing that line becomes a deliberate step.
3. **The tests do not pin the host or the TR_ID.** Sending every request to the live host, or using the live TR_ID for a paper order, leaves all 12 KIS tests green. Please add a test that mocks `requests.request` and asserts that a paper order goes to `openapivts…:29443` with `VTTC0012U` / `VTTC0011U`, plus one asserting that the override in (2) never reaches `:9443`.
4. **Quantities and prices are truncated, not refused.** `int(float(quantity))` turns 2.5 into a 2-share order reported as ok, and `int(limit_price)` drops the fraction. Please reject a non-integer instead.

Smaller: the balance and order inquiries are paged — the official `inquire_daily_ccld` example loops on `tr_cont` with `CTX_AREA_FK100` / `CTX_AREA_NK100` — while the connector sends empty continuation keys once, so a larger account's positions would be cut off without an error, and both read-only profiles are portfolio sources. "Today" for the order history comes from the machine's clock rather than KST. A token response without `expires_in` issues a new token on every call, which runs into KIS's token-issuance rate limit.

On `.gitignore`: main now has `test_gitignore_runtime_files.py`, which requires every runtime-root `CONFIG_FILENAME` to be ignored, so after a rebase it will ask you to add `/kis.json`. The token files are named at runtime (`kis-token-{environment}.json`), so the test cannot see them — please add `/kis-token-*.json` alongside.

**Merging with #1408 / #1409:** all three set `EXPECTED_BROKERS` and the README broker count to 15; after all three land it has to be 17, and they touch the same lines in `live/registry.py`, `trading/profiles.py`, `service.py`, `onboarding.py`, `SKILL.md` and the tests. No need to coordinate that yourself — rebase each on main as it moves and I will resolve the count when merging.

as950118:
Thank you for the detailed review — pushed a fix covering all four blocking items plus the smaller notes:

1. Paper cancel now looks up `order_branch` from `get_open_orders` when omitted, and the cancel body includes `EXCG_ID_DVSN_CD`.
2. `place_order`/`cancel_order` refuse a non-paper config as the first statement, and `"profile"` is dropped from `_OVERRIDE_KEYS`.
3. New tests mock `requests.request` directly to pin the paper order to `openapivts…:29443` with the paper TR_ID, and confirm a live-readonly config never reaches that call.
4. Fractional quantity/limit_price now rejected instead of truncated.

Smaller: `_get_paginated` follows `tr_cont`/`ctx_area_fk100`/`ctx_area_nk100` for balance + order-history reads, "today" uses KST, a missing `expires_in` falls back to a conservative TTL instead of reissuing every call, and `.gitignore` now covers `/kis.json` + `/kis-token-*.json`.

Full suite: 12839 passed, 0 failed, after rebasing onto current main.

warren618:
Thank you — all four points are in, and I checked each by undoing it and watching for a failing test: pointing the paper config at the live host, removing either `is_paper` guard, putting `profile` back into `_OVERRIDE_KEYS`, accepting a fractional quantity or limit price, stopping pagination after the first page, dropping `EXCG_ID_DVSN_CD` from the cancel body and disabling the branch lookup each fail at least one test. The smaller notes (KST date, token TTL fallback, `.gitignore`) are in too.

This one is ready. It waits for #1408 and #1409 so the three land together: they touch the same registration lines, and the broker count has to end at 17.

Two small gaps I will close when merging rather than send you around again. The paper sell (`VTTC0011U`) and cancel (`VTTC0013U`) TR_IDs are not pinned — switching either to its live code leaves every test green, because `test_kis_order_tr_id_differs_by_environment_and_side` only covers the buy side. And `_get_paginated` stops at `_MAX_CONTINUATION_PAGES` without saying so, which would return a short position list as if it were complete; it should raise instead. Cancel `quantity=2.5` also still truncates to 2 (the order side is fixed); I will make that refuse the same way. And the "no access_token" error still interpolates the token response body — the same thing you fixed in the Toss connector — so I will drop it there too.

warren618:
Merged, all three together — thank you.

PR #1408, opened 2026-09-10
feat(trading): add Upbit broker connector

as950118:
## Summary

Adds a `broker_sdk` connector for Upbit, Korea's largest KRW crypto exchange, via its public JWT-signed REST API (`requests` + `PyJWT` only, no third-party SDK). Part of the consolidated broker-connector tracker: HKUDS/Vibe-Trading#1367.

Per the tier decided there:

- **Discriminator**: none — one access/secret key pair reads the same account regardless of the declared profile, no sandbox.
- **Tier**: paper (order placement simulated locally against a live quote) + read-only live, following the Longbridge/Dhan/Zerodha precedent.
- **Auth**: API key pair, JWT-signed per request (HS256; payload carries `access_key`/`nonce`/`query_hash` for authenticated calls).

## Affected areas

- New: `agent/src/trading/connectors/upbit/{__init__,classification,profiles,sdk}.py`
- Anti-drift: `upbit` added to `_PAPER_CAPPED` in `test_paper_capped_connectors_refuse_live.py`, portfolio compatibility tier (`experimental`), broker count/table across all six locale READMEs (14 → 15)
- New: `agent/tests/test_sdk_connectors_upbit.py`

## Safety note (per maintainer feedback on the tracker)

`place_order`/`cancel_order` refuse a non-paper config as the literal first statement in each function — before any network call — so this is structurally impossible to reach the live endpoint, not a flag that could be checked later. `test_paper_capped_connectors_refuse_live.py::test_paper_capped_connector_refuses_non_paper_before_the_sdk` enforces this by AST-inspecting the function body (the guard must appear at or before statement index 1).

## Caveats

Portfolio-page value aggregation doesn't support KRW yet (`SUPPORTED_VALUE_CURRENCIES` in `src/portfolio/compatibility.py` covers USD/HKD/CNY only) — same pre-existing limitation India/INR already has. The connection and reads work; aggregate valuation raises a clear `PortfolioContractError` until that's extended separately.

A self-run code review caught one bug before this PR, already fixed and covered by a test: a `notional`-sized market order would return a fabricated `"status": "ok"` result with a null price/quantity when the live-quote lookup failed. It now fails closed with a clear error instead.

## Test plan

- [x] `pytest agent/tests/test_sdk_connectors_upbit.py agent/tests/test_sdk_connectors.py agent/tests/test_paper_capped_connectors_refuse_live.py agent/tests/test_portfolio_compatibility.py agent/tests/test_readme_counts.py -q` — all pass
- [x] Full suite (`pytest agent/tests -q`) — 12793 passed, 97 skipped, 0 failed

## Risk

No live-money order path is exposed by this PR (structurally paper-only, as above). Credentials are never logged; `secret_key` is fully redacted in `_public_config`. Rollback is a plain revert — no other connector or shared safety path is touched beyond the additive registrations listed above.


warren618:
Thank you — the red line holds, and I checked it by reading the code rather than the docstrings: `sdk.py` has a single HTTP call site and every caller uses GET (`/accounts`, `/orders/open`, `/orders/closed`, `/ticker`, `/candles/*`); `place_order` can only reach the unauthenticated `/ticker`, `cancel_order` makes no network call, both refuse a non-paper config on their first line, there is no `upbit-live-trade` profile, and removing either guard fails your tests. That is exactly the "structurally cannot reach the live endpoint" property #1367 asked for.

What still needs fixing is the simulator reporting fills and cancels that did not happen:

1. **Fills no market would give.** A limit order fills at its own limit price without looking at a quote, so a buy limit at 1 KRW comes back `simulated_fill` at 1.0, and so does a sell limit at 1e12. A quantity-sized market order whose quote lookup fails returns `status: ok` with `fill_price: null`. The simulation needs the live quote in every case: refuse when there is none; fill a marketable limit at the better of limit and last (`min(limit, last)` for a buy, `max(limit, last)` for a sell); refuse a non-marketable limit rather than fill it, since there is no book for it to rest on. A market the ticker does not know — `BTC-KRW`, the OKX-style spelling of `KRW-BTC` — should be refused rather than filled.
2. **Cancel reports success for an order that is still live.** The paper profile reads the real account, so real open orders show up in it, and `cancel_order` returns `cancelled: True` for any id — including one of those real orders, which stays live while the agent is told it is gone. Only ids this simulator issued (`PAPER-…`) can be cancelled; anything else should return an error saying the paper connector cannot cancel it. (The Zerodha connector on main has the same pattern — that one is ours to fix.)
3. PyJWT only comes in through the `msteams` / `channels` extras, so on a base install every signed read fails with `UpbitDependencyError`. Please add it to an extra that someone installing for Upbit would actually pick, and name that extra in the error.
4. Positions put the coin (`BTC`) in `currency` while the average cost is in KRW; `currency` should be the currency the row's numbers are in.

On `.gitignore`: main now has `test_gitignore_runtime_files.py`, which requires every runtime-root `CONFIG_FILENAME` to be ignored, so after a rebase it will ask you to add `/upbit.json` — the file your error message tells users to put their secret key in.

**Merging with #1407 / #1409:** the broker count ends at 17 and the shared registration lines will conflict; rebase on main as it moves and I will resolve the count when merging.

warren618:
Follow-up on point 2: the same cancel pattern in the Zerodha, Dhan and Shoonya paper connectors is now fixed on main in ec31c826. Each module has one `_PAPER_ORDER_PREFIX` constant, used both where `place_order` issues ids and where `cancel_order` checks them, and any other id is refused.

When you rebase, doing the same in `upbit/sdk.py` (your ids already start with `PAPER-`) and adding `upbit` to `_CONNECTORS` in `agent/tests/test_paper_sim_cancel_ids.py` would cover it. That test takes the accepted id from `place_order`, so the Upbit case will need a mocked quote once point 1 lands.

as950118:
Fixed, following the same pattern ec31c826 landed for Zerodha/Dhan/Shoonya:

- Added \`_PAPER_ORDER_PREFIX = "PAPER-"\`; \`cancel_order\` now refuses any id that doesn't carry it (a real order's id read via \`get_open_orders\` can no longer be acknowledged as cancelled without calling Upbit). \`upbit\` is added to \`test_paper_sim_cancel_ids.py\`.
- Along the way I found \`place_order\` was calling \`get_quote\` for *every* unpriced order, including a plain quantity-sized market order that needs no price at all. Fixed so only notional sizing triggers that call — this also means the shared cancel-id test above needs no quote mock.
- \`.gitignore\` gains \`/upbit.json\`.

Full suite: 12839 passed, 0 failed, after rebasing onto current main.

warren618:
Thanks for the quick turnaround — point 2 is done exactly as intended: one `PAPER-` constant used on both sides, and `upbit` in the shared cancel-id test.

Points 1, 3 and 4 from the review are still open, and point 1 has moved the wrong way. I ran paper `place_order` at this head with a quote available at 150,000,000 KRW:

- buy limit at 1 KRW → `simulated_fill` at 1.0
- sell limit at 1e12 → `simulated_fill` at 1e12
- market buy of 0.01 BTC → `status: ok`, `fill_price: null`
- `BTC-KRW` → accepted and filled
- `get_quote` was called 0 times

Taking the quote lookup out of quantity-sized orders is what makes every market order come back unpriced; before this change that only happened when the lookup failed. A simulated fill with no price is not a fill — any position or P&L built on it is invented. The review's ask still stands: look up the live quote for every paper order and refuse when there is none; fill a marketable limit at `min(limit, last)` for a buy and `max(limit, last)` for a sell; refuse a non-marketable limit; refuse a market the ticker does not know. My follow-up said the shared cancel-id test would then need a mocked quote — that is the right cost, so please add the mock rather than skip the lookup.

Also still open: PyJWT is only in the `msteams` extra and `_require_jwt` says `pip install PyJWT` — please put it in an extra someone installing for Upbit would pick, and name that extra in the message (point 3). And positions still carry the coin in `currency` while `average_cost` is in KRW (point 4).

as950118:
You're right, that was a regression — fixed properly this time:

1. \`place_order\` now always looks up the live quote and refuses if Upbit doesn't recognize the market. A market order fills at that price; a limit order fills only if immediately marketable against it (buy at \`min(limit, last)\`, sell at \`max(limit, last)\`), refused otherwise. \`test_paper_sim_cancel_ids.py\` now mocks Upbit's quote.
3. PyJWT now has its own \`upbit\` extra; the dependency error and onboarding hint both say \`pip install "vibe-trading-ai[upbit]"\`.
4. Added an explicit \`average_cost_currency\` field on each position (it's \`unit_currency\`, not \`currency\`) rather than leaving the reader to work that out.

Full suite: 12923 passed, 0 failed, after rebasing onto current main.

warren618:
Thanks — re-ran the paper probe at `c7c5717c` and point 1 is fixed properly: a buy limit at 1 KRW and a sell limit at 1e12 are both refused as not marketable, a market order fills at the quote, every order makes exactly one quote lookup, and `BTC-KRW` is refused whether the ticker call returns 404, an empty list or a 500. Points 2 and 3 are done too.

**Point 4 is still open — `average_cost_currency` is not read by anything.** The portfolio layer (`normalize_position` in `agent/src/portfolio/normalization.py`) takes the price currency *and* the cost currency from `currency`, and it treats a position as crypto only when the broker is Binance/OKX or the row declares `asset_type`. So a BTC position today comes out as `asset_type: "stock"`, `currency: "BTC"`, with a cost of 140,000,000 read as BTC. It is still refused — by the FX check, which names both BTC and KRW — but for the wrong reason.

What the portfolio contract needs from each position row:

- `symbol`: the coin (`BTC`)
- `currency`: the currency the price and cost are quoted in — `unit_currency`, i.e. `KRW`
- `asset_type`: `"crypto"`
- `average_cost`: in that same `currency`

and `average_cost_currency` can go. With that, a KRW source is refused because KRW valuation is not supported yet, which is the true reason, and it will value correctly the day it is.

as950118:
Fixed properly this time — I should have traced it through \`normalize_position\` the first time instead of guessing a field name.

\`get_positions\` now sets \`symbol\` to the coin, \`currency\` to \`unit_currency\` (the currency \`average_cost\`/a market price is actually quoted in), and \`asset_type\` to \`"crypto"\` explicitly — the three fields the portfolio contract reads. \`average_cost_currency\` is gone. Two new tests: one on the row shape directly, one that round-trips a row through \`normalize_position\` to confirm it comes out as a KRW-priced crypto position rather than a BTC-priced stock.

Full suite: 12936 passed, 0 failed, after rebasing onto current main.

warren618:
Merged, all three together — thank you.

PR #1409, opened 2026-09-10
feat(trading): add Toss Securities (토스증권) broker connector

as950118:
## Summary

Adds a `broker_sdk` connector for Toss Securities (토스증권) KR/US equities via the Toss Invest Open API (`developers.tossinvest.com`, launched 2026-05), OAuth2 client-credentials + `requests` only, no third-party SDK. Part of the consolidated broker-connector tracker: HKUDS/Vibe-Trading#1367.

Per the tier decided there:

- **Discriminator**: not found in the public docs — no documented sandbox, no account-response field distinguishing test from live, and applying for API access itself requires a real Toss Securities brokerage account.
- **Tier**: fully read-only, following the Trading 212 precedent — `place_order`/`cancel_order` hard-refuse for every profile, paper included.
- **Auth**: OAuth2 Client Credentials grant (`POST /oauth2/token`); account-scoped calls additionally require an `X-Tossinvest-Account` header carrying an `accountSeq` fetched from `GET /api/v1/accounts`.

## Affected areas

- New: `agent/src/trading/connectors/toss/{__init__,classification,profiles,sdk}.py`
- Anti-drift: `toss` added to `_PAPER_CAPPED` in `test_paper_capped_connectors_refuse_live.py`, portfolio compatibility tier (`experimental`), broker count/table across all six locale READMEs (14 → 15), and the "brokers without paper trading" exceptions paragraph now names Toss Securities alongside IBKR/Robinhood/Trading 212
- New: `agent/tests/test_sdk_connectors_toss.py`

## Confidence note (flagged on the tracker before this PR)

This is the lowest-confidence of the three connectors from that tracker discussion. The endpoint list came from Toss's public `openapi.json`, but exact response field names for less-common endpoints (candles, prices) are best-effort — I don't have a live Toss account to test against, and the docs render client-side so I could not get a maintainer-grade read on every schema. Given the tier lands at fully read-only either way (same outcome whether the discriminator answer is "no" or just "not found"), this shouldn't block merge per the maintainer's own note on the tracker — a scope or environment field turning up later would only add a tier, not require undoing anything here.

## Test plan

- [x] `pytest agent/tests/test_sdk_connectors_toss.py agent/tests/test_sdk_connectors.py agent/tests/test_paper_capped_connectors_refuse_live.py agent/tests/test_portfolio_compatibility.py agent/tests/test_readme_counts.py -q` — all pass
- [x] Full suite (`pytest agent/tests -q`) — 12789 passed, 97 skipped, 0 failed
- [ ] Read endpoints against a real Toss account (I don't have one — flagged above)

## Risk

No order-placement path is exposed at all (hard-refused for every profile, including declared-paper, since there's no verified sandbox to simulate against). Credentials are never logged; `client_secret` is fully redacted in `_public_config`. Rollback is a plain revert — no other connector or shared safety path is touched beyond the additive registrations listed above.


warren618:
Thank you — and for saying in #1367 that your confidence on Toss was low; that is what made the read-only tier the obvious call. The tier is implemented exactly right: one read-only profile, `place_order` and `cancel_order` refuse unconditionally (paper included) the way Trading 212 does, and the only POST in the file is `/oauth2/token`. It is also an official API (the Toss Invest Open API) rather than a scraped web app, which I wanted to confirm before anything else.

The read side does not match that API yet. Checked against the published spec (`openapi.json`, v1.2.15):

1. **The account reads call paths that are not in the spec.** Positions and the account snapshot read `/api/v1/assets`, and orders read `/api/v1/order-history`; the spec has `GET /api/v1/holdings` and `GET /api/v1/orders`, and no `/assets` or `/order-history`. As written, `check_status`, positions and open orders would all fail against the real service, and a Toss portfolio source would sit in `error`.
2. **Quotes and bars would come back empty but marked ok.** Every response in the spec is wrapped as `{"result": ...}`, and `_extract_items` looks for `items` / `data` / `holdings` / `orders` / `candles` / `prices` / `results` but never `result`. The candles endpoint's size parameter is `count`, not `limit`, so the requested size is ignored. A quote or bar with no price should be an error rather than `status: ok` with nulls.
3. **No test exercises a read.** The tests cover the refusals, which is good, but making paper `cancel_order` succeed, or making `get_positions` return an empty list, leaves them green. `test_trading212_connector.py` is a good model: a spec-shaped payload through the service, the auth header shape, and a clean error on a bad key.

Also: the token cache (`toss-token.json`) is written at the runtime root. Main's new `test_gitignore_runtime_files.py` will ask for `/toss.json` after a rebase, but it cannot see the token file, so please add that one by hand; write it with mode 0600 from the start rather than chmod-ing after the write, and keep the token response body out of error messages. KRW is not a supported portfolio valuation currency yet, so KR holdings will be refused at the portfolio layer — a compatibility entry like Dhan's ("KRW valuation is not supported yet") makes that explicit instead of surprising. And SKILL.md says "Korea equity trading" for a read-only connector; "read-only" is the word that matters there.

The broker-count note from #1407 applies here too (it ends at 17 once all three land; I will resolve it when merging).

as950118:
Fixed all three, per the spec check (openapi.json v1.2.15):

1. Positions/account snapshot now read `GET /api/v1/holdings`; orders now read `GET /api/v1/orders` — Toss has one order-reading endpoint, so open vs executed is now a client-side split on each row's `status`, same as Dhan/Trading 212.
2. Every response is unwrapped from `{"result": ...}` before the existing key search (which already looked for `"candles"`, matching how `/api/v1/candles` nests its list). Candles now sends `count`, not `limit`. A quote with no price, or a bar with no close, is now an error instead of `status: ok` with nulls.
3. New tests exercise each read path with a spec-shaped payload through the service (modelled on `test_trading212_connector.py`), plus the auth header shape and a clean error on a bad key.

Also fixed: token file written with mode 0600 from creation (no chmod-after race), response body no longer echoed into the "no access_token" error, `.gitignore` covers `/toss.json` + `/toss-token.json`, a portfolio-compatibility entry notes KRW isn't supported yet (matching Dhan's INR entry), and SKILL.md leads with "read-only".

Full suite: 12837 passed, 0 failed, after rebasing onto current main.

One thing I want to flag rather than claim confidently fixed: I could not get a reliable, complete read of the full `/api/v1/holdings` and `/api/v1/orders` response schemas from the spec myself (large-doc fetches I tried came back truncated/inconsistent across attempts) — I implemented the paths and the client-side status split on your correction plus the field names already in `_position_to_dict`/`_order_to_dict`, but if you have visibility into the exact field names Toss returns for those two, I'd rather adjust now than ship a guess.

warren618:
Thank you — the `result` unwrap, `count`, the 0600 token file, the compatibility entry and the SKILL.md wording are all right, and thank you for saying what you could not verify instead of guessing. I pulled the schemas from `openapi.json` v1.2.15, so here is what the two account endpoints return — and one correction to my own review.

**`GET /api/v1/orders` — my review was imprecise here, sorry.** I wrote that the spec has `GET /api/v1/orders` without saying that its `status` query parameter is **required** (`OPEN` or `CLOSED`). Your original code sent it; this change drops it, so the request no longer matches the spec. That `status` is a group label — each order's own `status` uses a different enum: `PENDING`, `PENDING_CANCEL`, `PENDING_REPLACE`, `PARTIAL_FILLED`, `FILLED`, `CANCELED`, `REJECTED`, `CANCEL_REJECTED`, `REPLACE_REJECTED`, `REPLACED`. `_OPEN_ORDER_STATUSES` has `PARTIALLY_FILLED`, `OPEN` and `NEW`, none of which are in it, so a partly filled resting order would be reported as executed. Rather than split client-side, call it twice as before: `status=OPEN` for open orders (returned in full; `limit` and `cursor` are ignored) and `status=CLOSED` for executions, which is paged — `limit` (default 20, max 100) and `cursor`, with `hasNext` / `nextCursor` in the response. The body is `{"result": {"orders": [...], "nextCursor": ..., "hasNext": ...}}`.

Each order has `orderId`, `symbol`, `side` (`BUY`/`SELL`), `orderType` (`LIMIT`/`MARKET`), `timeInForce`, `status`, `price` (null for market orders), `quantity`, `orderAmount`, `currency`, `orderedAt`, `canceledAt`, and `execution` — an object with `filledQuantity`, `averageFilledPrice`, `filledAmount`, `commission`, `tax`, `filledAt`, `settlementDate`. So `filled_quantity` has to come from `execution.filledQuantity`; there is no top-level `filledQuantity`.

**`GET /api/v1/holdings`** returns `{"result": {"totalPurchaseAmount", "marketValue", "profitLoss", "dailyProfitLoss", "items": [...]}}`. The totals sit at that level — there is no `summary` or `accountSummary`, so the account snapshot currently comes back empty — and each money total is a `{"krw": ..., "usd": ... | null}` object (`marketValue` and `profitLoss` wrap it in `amount` / `amountAfterCost`). Each item has `symbol`, `name`, `marketCountry` (`KR`/`US`), `currency` (`KRW`/`USD`), `quantity`, `lastPrice`, `averagePurchasePrice`, `marketValue` (`purchaseAmount` / `amount` / `amountAfterCost`), `profitLoss` (`amount` / `amountAfterCost` / `rate` / `rateAfterCost`), `dailyProfitLoss` and `cost`. `_position_to_dict` reads `averagePrice`, `currentPrice` and `market`, so against the real payload `average_price`, `current_price` and `market` would all be `None`, and `pnl` would be the whole `profitLoss` object rather than a number. Every decimal comes as a string.

Two smaller things:

- The new tests feed `{"result": [...]}` lists for holdings and orders. A fixture shaped like the two responses above would have caught all of this, and it is what I would test against.
- A bar with no close is dropped rather than turned into an error, and `test_toss_get_historical_bars_drops_bars_with_no_close_price` pins the drop. A silently missing bar is a gap in the series; an error, as your reply described, is what we want.

The prices and candles shapes match the spec as you have them.

as950118:
Fixed against the exact schemas — thank you for pulling them from the spec.

- \`/api/v1/orders\` now called twice with \`status=OPEN\`/\`status=CLOSED\` (required query param, not a client-side split), and \`filled_quantity\` reads \`execution.filledQuantity\`.
- \`get_account_snapshot\` now reads \`totalPurchaseAmount\`/\`marketValue\`/\`profitLoss\`/\`dailyProfitLoss\` directly off the holdings result (no more empty \`account: {}\`), passed through as the raw \`{"krw":..., "usd":...}\`-wrapped objects rather than flattened.
- \`_position_to_dict\` now reads \`averagePurchasePrice\`, \`lastPrice\`, \`marketCountry\`, and \`profitLoss.amount\` for \`pnl\`.
- A candle with no close is now an error, not a silently dropped bar.
- Tests rebuilt against fixtures shaped exactly like the two confirmed responses.

Full suite: 12917 passed, 0 failed, after rebasing onto current main.

dudco:
Hi @as950118 — I have access to a Toss Securities account and would like to help close the remaining real-account validation gap for this read-only connector.

I can validate only the read paths (accounts, holdings/account snapshot, OPEN/CLOSED orders, quotes/candles where applicable, and OAuth/error handling). I will not place, modify, or cancel orders, and will not share credentials, account identifiers, or raw portfolio/order payloads.

If useful, I will turn the findings into sanitized field/type-only fixtures and a small focused follow-up PR. Is there any endpoint or edge case you would like prioritized?

warren618:
Thanks — checked against `openapi.json` v1.2.15 at `202ff81e`. Server-side `status=OPEN`, `execution.filledQuantity`, `averagePurchasePrice` / `lastPrice` / `marketCountry`, and a candle without a close becoming an error are all right now.

One thing is still open, plus the tests that would have caught it.

**`CLOSED` orders are read one page at a time and the rest is dropped.** `get_open_orders(include_executions=True)` makes one `status=CLOSED` call without `limit`, so it returns at most 20 orders (the spec's default) and nothing in the payload says there were more. The contract, in full:

- query `limit`: integer, 1–100, default 20 — applies to `CLOSED` only (ignored for `OPEN`, which returns everything)
- query `cursor`: string — the previous page's `nextCursor`; ignored for `OPEN`
- response `result`: `{"orders": [...], "nextCursor": string | null, "hasNext": boolean}` — all three are required fields

So: send `limit=100`, keep calling with `cursor=nextCursor` while `hasNext` is true, and put a cap on the number of pages that returns an **error** when it is reached rather than a shorter list — a truncated history that looks complete is the failure we are trying to avoid. A test with two pages (`hasNext: true` then `false`) and one that hits the cap would pin both.

**The fixtures are still not the spec's shapes**, which is why the tests pass without exercising this:

- Every quantity, price, money amount and rate is typed **string** in the spec — `HoldingsItem.quantity`, `lastPrice`, `averagePurchasePrice`, every field of `MarketValue` / `ProfitLoss`, `Order.quantity` and every `OrderExecution` figure — while the fixtures use numbers.
- `HoldingsItem.dailyProfitLoss` is `{"amount", "rate"}` and `HoldingsItem.cost` is `{"commission", "tax"}` (`tax` may be null); the fixtures have plain strings for both.
- The overview's `dailyProfitLoss` is `{"amount": {"krw", "usd"}, "rate"}`, not a bare `{"krw", "usd"}`.
- An `Order` always carries `timeInForce`, and the `CLOSED` fixture is a single empty page.
- Nothing checks that a US holding comes back as `USD` — replacing the per-item currency with a hard-coded `"KRW"` still passes the suite.

@dudco's offer above to check the read paths against a real account would settle the pagination and string types directly. Field names and types only would be plenty.

as950118:
Fixed:

- \`CLOSED\` orders now page with \`limit=100\`, following \`cursor\`/\`hasNext\` to completion. Hitting the page cap (\`_MAX_ORDER_HISTORY_PAGES\`) returns \`status: error\` for the whole call instead of a shorter \`executions\` list — a truncated history that looks complete is the failure this avoids. \`OPEN\` is untouched (one call, no params, per the spec).
- Test fixtures rebuilt with the spec's real types: every quantity/price/amount/rate is a string, \`dailyProfitLoss\`/\`cost\` are nested objects, and a second USD holding in the fixture catches a hard-coded \`"KRW"\`. \`_order_to_dict\` now also reads \`timeInForce\`.
- New tests: a two-page \`CLOSED\` history follows the cursor across both pages, and an unbounded history trips the cap and returns an error rather than a partial list.

Full suite: 12930 passed, 0 failed, after rebasing onto current main.

I don't have a way to verify the pagination/cursor mechanics or the exact string-vs-number typing against a real account myself, so if @dudco's offer to check the read paths is still open, that would settle both directly — I've implemented against your description of the contract but can't confirm it end-to-end.

dudco:
Real-account read-only validation completed for PR #1409.

Validated successfully without placing, modifying, or cancelling any order:
- OAuth plus account-scoped reads
- holdings / account snapshot
- OPEN and CLOSED order reads
- quote and daily candles for `005930`
- no persistent token file was left behind

Observed real response mapping:
- quantities, prices, money values, and rates are strings
- holdings totals use the expected nested KRW/USD wrappers
- normalized positions, snapshot, orders (`execution.filledQuantity`, `timeInForce`), quotes, and candles all matched the connector mapping

For CLOSED orders, the real response had `hasNext: false` and `nextCursor: null`, so this account exercised a normal single-page completion. I could not reproduce a multi-page cursor flow with this account's history; the PR's focused unit tests cover the two-page cursor path and page-cap failure path.

Read-only scope only; no credentials, account identifiers, raw payloads, or portfolio/order values are included here.

warren618:
Merged, all three together — thank you, and thanks @dudco for the read-only run against a real Toss account.

PR #1411, opened 2026-09-11
fix(loader): paginate tencent history backward — the API serves the LAST 500 bars per window

### LINKED ISSUE #1410: [Bug] Tencent loader silently truncates multi-year history to the LAST ~500 bars (pagination walks the wrong direction)
### Description

**What happened:**
The Tencent loader's pagination (added in #1154) walks in the wrong direction, so **any window longer than 500 bars is silently truncated to its most recent ~500 bars (~2 years)** — with exit code 0 and no warning.

#1154 assumed each `fqkline` response carries the **first** ≤500 bars from the cursor and advanced `start` past each page's newest bar. The live API does the opposite: it serves the **last** ≤500 bars ending at the requested `end` parameter. Consequence in `_fetch_one`: the first page already ends at `end_date`, `next_start = last_bar + 1d` steps past the window, the next page is empty, and the loop exits "cleanly" with a tail-only series. The `incomplete tencent history` guard never fires because the cursor does advance — past the window, not along it. The existing pagination tests mock `_request_page` with the first-500 assumption, which is why the truncation passes CI.

**What I expected:**
A multi-year request (e.g. 2018-01-01..2026-06-30) returns the full window (~2058 trading days for sh000300) or raises `incomplete tencent history` — never a silent tail-only series.

Fix in #1411: paginate **backward** (keep `start_date` fixed, move the `end` cursor to the day before each page's oldest bar, stop on a short page or when the cursor passes `start_date`; page-cap and failed pages still raise), plus tests mocking the verified last-500 semantics and a multi-year full-coverage regression.

### Steps to Reproduce

1. Live API probe (no auth needed):

```
GET https://web.ifzq.gtimg.cn/appstock/app/fqkline/get?param=sh000300,day,2018-01-01,2026-06-30,500,qfq
→ 500 bars spanning 2024-06-06 .. 2026-06-30   (LAST-500, not first-500)
```

   The start clamp IS honoured (a window holding <500 trading days is served in full from its start):

```
GET ...param=sh000300,day,2024-01-01,2024-06-30,500,qfq
→ 117 bars spanning 2024-01-02 .. 2024-06-28
```

2. Through the loader (`main` @ f9cb061b):

```python
DataLoader()._fetch_one("000300.SH", "2018-01-01", "2026-06-30")
# returns 500 bars, 2024-06-06 .. 2026-06-30 — the first 6.5 years are missing, no error
```

3. Field impact: two independent backtest runs (A-share index 2015–2026 and an ETF panel 2018–2026) requested via `source=tencent`/`auto` came back as exactly ~500-bar tail windows with `exit_code 0` and empty run-card warnings; both were caught only by an equity-row-count check against the expected trading-day count.

### Error Logs

```shell
# None — that is the bug. The run reports success:
run_card.json: warnings=[]   metrics.csv: trade_count=5 (RPS warmup ate the truncated window)
ohlcv_000300.SH.csv: 500 rows, 2024-06-06 .. 2026-06-30 (requested 2015-01-01 .. 2026-06-30)
```

### Interface

MCP Plugin

### LLM Provider

N/A — data-layer bug, provider-independent

### Version

0.1.15 (main @ f9cb061b; pagination introduced in #1154, merged 2026-08-21)

### Environment

macOS 15 / Python 3.12 — repro is pure HTTP, no token; the API behavior is identical from any host

shadowinlife:
## Summary

- Fix the Tencent loader's pagination direction: the fqkline API serves the **last** ≤500 bars of a requested window (ending at the `end` param), not the first — so pagination must walk **backward** from `end_date`. The forward walk added in #1154 silently degrades any multi-year window to its most recent ~500 bars (~2 years) with exit code 0 and no warning.
- Rewrite the pagination tests to mock the **real** last-500 API semantics (the previous mocks encoded the first-500 assumption, which is why the truncation passed CI) and add a multi-year regression test.
- Second commit (self-review follow-up): bump `_LOADER_CACHE_VERSION` 6 → 7 so entries cached by the forward walk during the bug window are never served again, and harden the two paths where a throttled-but-HTTP-200 reply could still truncate a walk silently.

## Why

Closes #1410. Field impact observed during a strategy-validation campaign: two independent backtest runs (A-share index 2015–2026; ETF panel 2018–2026) came back as exactly ~500-bar tail windows — one surfaced as a bogus `trade_count=5` (warmup ate the truncated window), the other as per-symbol files of exactly 500 rows. Both runs reported success (`exit_code 0`, `warnings: []`); only an equity-row-count check against the expected trading-day count caught them. #1154's intent (no silent truncation) regressed because its regression tests mocked the assumed rather than the actual API behavior.

Live verification of the real semantics (no auth, reproducible from any host):

```
param=sh000300,day,2018-01-01,2026-06-30,500,qfq → 500 bars, 2024-06-06 .. 2026-06-30   (LAST-500)
param=sh000300,day,2024-01-01,2024-06-30,500,qfq → 117 bars, 2024-01-02 .. 2024-06-28   (start clamp honoured)
```

## Changes

- `agent/backtest/loaders/tencent_loader.py`
  - `_fetch_one`: backward walk — `start_date` stays fixed; the end cursor moves to the day before each page's oldest bar; stop on a short page (`< _PAGE_SIZE`) or when the cursor passes `start_date`. Page-cap exhaustion (`_MAX_PAGES`) and failed pages still raise `incomplete tencent history` instead of truncating; stall guard kept; dedup/concat unchanged.
  - `_PAGE_SIZE` comment and `_fetch_one` docstring now record the verified API semantics (with probe date) and why forward walking fails — so the direction doesn't get "fixed" back.
- `agent/tests/test_tencent_loader.py`
  - New `_last500_api` fake honoring the real contract (last ≤500 bars of `[start, end]`, start clamp honoured).
  - New `test_multiyear_window_is_served_in_full` regression (1500-bar window: full coverage, 3 pages, end cursor strictly decreasing).
  - New `test_pagination_moves_the_end_cursor_behind_each_page` (cursor = oldest bar − 1 day).
  - Adapted the dedup / page-cap-raise / failed-page-raise / short-page-stop / end-date-bar tests to the backward walk. HK-mapping and intraday-rejection tests untouched.
- Second commit (review follow-up — protects the same "complete window or loud failure" guarantee):
  - `agent/backtest/loaders/base.py`: `_LOADER_CACHE_VERSION` 6 → 7. The opt-in loader cache is content-addressed on `(source, symbol, timeframe, start, end)` with no TTL, so entries written by the forward walk during the bug window (#1154, 2026-08-20, onward) hold tail-truncated series that would keep being served after this fix. Follows the v4/v5/v6 bump precedent for cached-semantics changes.
  - `agent/backtest/loaders/tencent_loader.py`: `_request_page` raises on an error-shaped reply (non-zero `code`, no payload) so it enters the page retry budget instead of reading as an empty window; `_fetch_one` re-requests a mid-walk empty page once before accepting it as the genuine start of history (a transient glitch returns data; a genuine boundary — e.g. a pre-IPO window — stays empty and terminates the walk).
  - `agent/tests/test_tencent_loader.py`: 3 new tests pin all three behaviors (error-reply raise, mid-walk re-request continues, persistent-empty terminates).

## Test Plan

Full-suite result on this branch (after the second commit): **12791 passed, 5 failed, 84 skipped** (4m21s). The 5 failures (`test_anthropic_prompt_cache`, `test_llm_reasoning_effort` ×3, `test_provider_header_isolation`) are **pre-existing and unrelated** — verified by running the same tests on a clean `origin/main` (f9cb061b) checkout in my environment: identical 5 failures (litellm adapter version drift). Loader-adjacent suites all green: `test_tencent_loader.py` (13) + `test_volume_unit_consistency.py` + `test_loader_volume_units.py` → 28 passed; loader-cache suites → 8 passed, 5 skipped (no test pins the version constant).

Live smoke against the real API with the fix applied:

```python
DataLoader()._fetch_one("000300.SH", "2018-01-01", "2026-06-30")
# before: 500 bars, 2024-06-06 .. 2026-06-30
# after:  2058 bars, 2018-01-02 .. 2026-06-30  ✓
```

## Checklist

### DIFF
--- a/agent/backtest/loaders/base.py
+++ b/agent/backtest/loaders/base.py
@@ -330,7 +330,9 @@ _LOADER_CACHE_TRUE_VALUES = {"1", "true", "yes", "on"}
 # v5: UK (.L) prices normalized from GBp to GBP (÷100) (#1206).
 # v6: LSE quote currency is fail-closed and per-symbol conversion provenance is
 # persisted. v5 USD/unknown .L entries must never be served as static GBP.
-_LOADER_CACHE_VERSION = 6
+# v7: tencent fqkline paginates backward (#1410) — entries cached under the
+# forward walk hold tail-truncated multi-year series and must never be served.
+_LOADER_CACHE_VERSION = 7
 _LOADER_FRAME_METADATA_ATTRS = ("quote_currency", "currency_conversion")
 
 
--- a/agent/backtest/loaders/tencent_loader.py
+++ b/agent/backtest/loaders/tencent_loader.py
@@ -143,6 +143,11 @@ class DataLoader:
 
         data = json.loads(raw)
         # Response: {"code":0,"data":{"sh601595":{"day":[["2026-06-01","21.32",...], ...]}}}
+        if data.get("code") not in (0, None) and not data.get("data"):
+            # An error-shaped reply is a failure, not an empty window: None
+            # here would let a throttled reply end the walk mid-pagination
+            # and serve (and cache) a truncated series as complete history.
+            raise ValueError(f"tencent fqkline error reply: code={data.get('code')}")
         stock_data = data.get("data", {})
         if not stock_data:
             return None
@@ -234,7 +239,22 @@ class DataLoader:
                 ) from last_error
 
             if page is None or page.empty:
-                break
+                if not chunks:
+                    break
+                # A mid-walk empty page is ambiguous: the genuine start of
+                # history, or a throttled reply. One re-request decides — a
+                # transient glitch returns data, a genuine boundary stays
+                # empty and terminates the walk.
+                time.sleep(_PAGE_BACKOFF)
+                try:
+                    page = self._request_page(code, start_date, cursor_end)
+                except Exception as exc:  # noqa: BLE001 - transient network jitter
+                    raise ValueError(
+                        f"incomplete tencent history: {code} re-request at "
+                        f"{cursor_end} failed: {exc}"
+                    ) from exc
+                if page is None or page.empty:
+                    break
             chunks.append(page)
 
             # A short page means the walk reached start_date.
--- a/agent/backtest/loaders/tencent_loader.py
+++ b/agent/backtest/loaders/tencent_loader.py
@@ -27,9 +27,13 @@ logger = logging.getLogger(__name__)
 
 _BASE_URL = "https://web.ifzq.gtimg.cn/appstock/app/fqkline/get"
 
-# Tencent fqkline caps a single response at `_PAGE_SIZE` bars.  Requests for a
-# multi-year window (e.g. 2018-2025 daily) silently truncate at 500 bars
-# (roughly the first 2 years) unless we paginate by advancing the start date.
+# Tencent fqkline caps a single response at `_PAGE_SIZE` bars and serves the
+# LAST bars of the requested window (ending at the `end` parameter), not the
+# first — verified empirically 2026-09-11: a 2018-01-01..2026-06-30 request
+# returned 500 bars spanning 2024-06-06..2026-06-30, while a window holding
+# fewer than 500 trading days was served in full from its start (the start
+# clamp is honoured).  Multi-year windows therefore paginate BACKWARD: see
+# `_fetch_one`.
 _PAGE_SIZE = 500
 # 12 pages * 500 bars = 6000 bars ≈ 24 trading years; a hard cap also guards
 # against pathological loop behavior if the API ever stops advancing.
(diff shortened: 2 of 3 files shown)


he-yufeng (commented):
Read through the walk and ran the rewritten test file locally (Python 3.14, 10 passed).

The direction is right: Tencent's kline endpoint answers "the up-to-500 bars ending at `end`", so stepping the end cursor behind the oldest bar of each page is the only way to reach older history; the previous forward walk could only ever re-fetch the same newest window and silently truncated anything past 500 bars. Dedup on overlapping pages and the end-date boundary bar are both covered by tests now.

I especially want to call out the fail-closed parts: exhausting the page cap and a failed mid-walk page both raise instead of returning a partial history. Silent truncation is exactly the bug being fixed, so refusing to serve a short answer is the correct contract.

One thing I checked that is *not* a problem here: an empty page mid-walk still terminates the loop (kept shape from before). Given the endpoint's semantics that only happens at the boundary of available history, so it's fine for this scope.

Nice work, this is the kind of fix that should land.

warren618:
Merged — thank you. I re-probed the endpoint before merging and it agrees with you: `sh000300` for `2018-01-01..2026-06-30` returns 500 bars `2024-06-06..2026-06-30`, and moving `end` to `2024-06-05` returns the 500 bars ending there (`2022-05-17..2024-06-05`), so walking backward is the right direction. `hk00700` behaves the same way. The rewritten tests model the real last-500 contract and they bite: with main's `tencent_loader.py`, 7 of 13 fail.

The cache-version bump is the right call too — entries written by the forward walk hold tail-only series and must never be served. It invalidates every loader's cache once, which only costs a refetch.

PR #1413, opened 2026-09-11
fix(tools): record state.json from the backtest tool so tool-driven runs are ingestible as evidence

shadowinlife:
## Summary

- The `backtest` tool now records `state.json` (via the existing `RunStateStore.mark_success` / `mark_failure`) right after engine execution, so tool-driven runs are no longer rejected as unknown-provenance by `refresh_strategy_evidence`.
- New tests pin all four lifecycle paths: success, engine failure (with reason), engine timeout (with reason), and validation error (no state file — the engine never ran).
- Second commit (self-review follow-up): a timed-out engine (`subprocess.TimeoutExpired` propagated out of `Runner.execute`) previously bypassed the lifecycle block entirely — now it is caught, recorded as `failed` with a timeout reason, and answered with a clean error envelope.

## Why

Closes #1412. The evidence-ingestion gate (`run_artifacts.read_run_status`) fail-closes without `state.json`, and only the runtime loop wrote that file — so **every run produced by the `backtest` MCP tool was rejected** despite complete, hash-manifested artifacts: `hard-gate:exit-nonzero: state.json missing or unreadable (run status unknown)`. Observed at scale during a strategy-validation campaign: 11/11 healthy tool-driven runs skipped, recoverable only by hand-deriving `state.json` from run cards (workaround disclosed in the issue). The evidence pipeline and the project's primary backtest entry point were disconnected.

Of the two directions discussed in the issue, this PR implements the source-side fix: the tool already knows the outcome (`result.success` / `result.exit_code`), and reusing `RunStateStore` keeps one writer and one format contract for `state.json`. The ingestion gate is untouched and stays fail-closed for runs whose status was never recorded.

## Changes

- `agent/src/tools/backtest_tool.py`
  - After `runner.execute(...)`: `RunStateStore().mark_success(run_path)` on success, `mark_failure(run_path, f"backtest engine exited with code {result.exit_code}")` otherwise — same fsync'd writer the runtime uses.
  - Validation errors (missing/invalid config, missing signal engine) return before the engine starts and intentionally record nothing: no execution happened, and artifact-health gates reject those runs anyway.
- `agent/tests/test_backtest_tool_state.py` (new)
  - `test_successful_run_records_state_success` — envelope `ok` + `state.json == {"status": "success"}`.
  - `test_failed_run_records_state_failed_with_exit_code` — `{"status": "failed", "reason": "backtest engine exited with code 3"}`.
  - `test_validation_error_records_no_state` — no `state.json` when config is missing.
[...]

PR #1414, opened 2026-09-11
fix(swarm): enforce per-worker skills allowlist in load_skill

shadowinlife:
## Summary

- `load_skill` now enforces the per-worker skill allowlist that `SwarmAgentSpec.skills` has always documented: `LoadSkillTool` accepts an optional `allowed_skills` and fails closed on a skill outside it (the error names the allowed set); `build_filtered_registry` / `build_swarm_registry` take a matching `skill_allowlist` parameter, and `run_worker` delivers the spec's list through it. `None` keeps the historical unrestricted behavior for the main agent.

## Why

Requested by the maintainer in the #1286 review:

> "**Yes to commit 1 — please do re-submit it separately, and it does not need to wait on anything.** … `SwarmAgentSpec.skills` is documented as 'List of allowed skill names' … but `LoadSkillTool.__init__` takes only a `skills_loader` and no allowlist, so `execute()` will load any skill a worker names. A documented boundary that exists only in the prompt is not a boundary. Small PR, and please include a test that asserts a worker naming a skill outside its spec is refused — the assertion has to be on the refusal, not on the prompt text, or it pins the same non-enforcement."

This is commit 1 of #1286 re-submitted standalone (that PR's feature part was closed as not-planned; this bugfix was explicitly split out). Gap re-verified on current main (f9cb061b): `models.py:184` still documents the boundary, `_filter_skill_descriptions` (`worker.py:585`) is still the only reader of `spec.skills` — prompt text only — and `load_skill_tool.py` still has no enforcement.

## Changes

- `agent/src/tools/load_skill_tool.py` — optional `allowed_skills`; a skill outside it is refused fail-closed with the allowed names in the error; an explicit empty list refuses everything; `None` = unrestricted (default behavior unchanged).
- `agent/src/tools/__init__.py` — `build_filtered_registry` / `build_swarm_registry` gain a keyword-only `skill_allowlist` and rebuild `load_skill` with the boundary baked in (`register()` replaces the auto-discovered instance by name).
- `agent/src/swarm/worker.py` — `run_worker` delivers `agent_spec.skills` as the allowlist; an empty spec list stays unrestricted, matching the prompt-side filter semantics.
- `agent/tests/test_load_skill_allowlist.py` (new, 10 tests) — per the review requirement, **refusal is asserted on tool behavior** (`status == "error"` + message naming the allowed set), not on prompt text; plus `None` default, empty-list fail-closed, registry rebuild, swarm registry, and worker pass-through.
- `agent/tests/test_swarm_m4_e2e.py` — adapts the four `build_swarm_registry` test doubles to the new keyword-only parameter (in #1286 this adaptation rode in the feature commit; commit 1 alone is not self-contained against the current suite without it).

## Test Plan

Full suite on this branch: **12797 passed, 5 failed, 84 skipped** (4m02s). The 5 failures (`test_anthropic_prompt_cache`, `test_llm_reasoning_effort` ×3, `test_provider_header_isolation`) are pre-existing and unrelated — identical failures verified on a clean `origin/main` (f9cb061b) checkout in my environment (litellm adapter version drift); also disclosed in #1411 / #1413. New tests: 10/10 green; `test_swarm_m4_e2e.py` 4/4 green after the double adaptation.

Manual check of the refusal path (real `SkillsLoader`, no mocks):

```python
tool = LoadSkillTool(skills_loader=loader, allowed_skills=["strategy-generate"])
tool.execute(name="strategy-generate")   # status: ok
tool.execute(name="alpha-zoo")
# status: error
# "Error: skill 'alpha-zoo' is outside the skill allowlist for this context. Allowed skills: strategy-generate"
```

## Checklist

### DIFF
--- a/agent/src/swarm/worker.py
+++ b/agent/src/swarm/worker.py
@@ -573,11 +573,16 @@ def _run_worker_impl(
     _emit(event_callback, "worker_started", agent_id, task_id)
 
     # 1. Build per-worker tool registry — local pool plus any operator-
-    #    surfaced MCP tools, projected onto the agent's whitelist.
+    #    surfaced MCP tools, projected onto the agent's whitelist. The
+    #    documented ``skills:`` boundary is enforced at runtime by rebuilding
+    #    ``load_skill`` with the allowlist baked in; an empty ``skills`` list
+    #    means unrestricted, matching the prompt-side filter semantics
+    #    (``_filter_skill_descriptions`` treats empty as include-all).
     registry = build_swarm_registry(
         agent_spec.tools,
         agent_config=agent_config,
         include_shell_tools=include_shell_tools,
+        skill_allowlist=agent_spec.skills or None,
     )
 
     # 2. Build system prompt with filtered skills
--- a/agent/src/tools/__init__.py
+++ b/agent/src/tools/__init__.py
@@ -275,7 +275,12 @@ def build_registry(
     return registry
 
 
-def build_filtered_registry(tool_names: list[str], *, include_shell_tools: bool = False) -> ToolRegistry:
+def build_filtered_registry(
+    tool_names: list[str],
+    *,
+    include_shell_tools: bool = False,
+    skill_allowlist: list[str] | None = None,
+) -> ToolRegistry:
     """Build a ToolRegistry with only specified tools.
 
     Local-tools-only filtered builder. Swarm workers should call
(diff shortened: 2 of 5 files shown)


warren618:
Merged — thank you, and thanks for splitting it out of #1286 as asked. The refusal is asserted on tool behaviour, which is what that review asked for, and it bites: with main's `load_skill_tool.py`, 5 of the 10 new tests fail.

One note that is not about this PR: in a wide run, three tests in `test_mcp_load_skill_paging.py` fail when a particular set of other test files runs before them. I reproduced the same three failures on main in the same order, so it is pre-existing test pollution rather than your change; I will look at it separately.

PR #444, opened 2026-07-08
feat: add opencode capabilities to provider definitions

imsankz:
Adds support for opencode-zen and opencode-go coding providers. Similar to feat/kimi-coding-provider.

PR #1416, opened 2026-09-11
feat(providers): add OpenCode provider and send x-opencode-session

### LINKED ISSUE #1415: OpenCode Go rejects requests with 400 MissingSessionID (x-opencode-session not sent); no opencode provider in settings
### Summary

Requests to OpenCode Go (`https://opencode.ai/zen/go/v1`) fail with `400 MissingSessionID` because Vibe-Trading never sends the `x-opencode-session` header the service now requires. There is also no `opencode` entry in Web Settings, so the only way to use OpenCode today is to re-point another provider at its URL.

### Reproduce

1. Web Settings → provider `deepseek`, base URL `https://opencode.ai/zen/go/v1`, model `deepseek-v4.1-flash`, an OpenCode Go key.
2. Send any chat message.

```
Execution failed: provider_stream_error provider=deepseek model=deepseek-v4.1-flash:
OpenAIInvalidRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'MissingSessionID',
'message': 'Error from provider (Console Go): Request is missing x-opencode-session and cannot be
routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it'}}
```

### Cause

OpenCode's docs (https://opencode.ai/docs/go/) require a stable per-conversation session id in `x-opencode-session` (used for routing and prompt caching) and ask clients to identify themselves via User-Agent. `opencode-go` / `opencode-zen` (#444) only map `OPENAI_API_KEY` / `OPENAI_BASE_URL`; nothing in `ChatOpenAIWithReasoning` sets the header, and the names are not in `llm_providers.json`, so they never appear in the settings UI.

### Proposal

- Add a canonical `opencode` provider (`OPENCODE_API_KEY` / `OPENCODE_BASE_URL`, default Go root, Zen offered as an option) across the catalog, capabilities, swarm allowlist and CLI onboarding.
- Bind the active session id in a `ContextVar` from `AgentLoop.run()` and have the OpenAI-compatible adapter add `x-opencode-session` whenever the request targets `opencode.ai` (by provider name or base-URL host), with a per-adapter UUID fallback outside a session.

This touches `src/agent/loop.py` (a thin bind/reset wrapper around `run()`) and `src/providers/`, which the PR template lists as protected areas, hence this issue first. A PR implementing the above, with tests and a manual verification against the live Go endpoint, is ready: https://github.com/HKUDS/Vibe-Trading/pull/1416.

Version: 0.1.15 (also reproduced against current `main`).

tonydo:
## Summary

- Add **OpenCode (Go / Zen)** as a first-class LLM provider: `LANGCHAIN_PROVIDER=opencode`, `OPENCODE_API_KEY`, `OPENCODE_BASE_URL` (defaults to `https://opencode.ai/zen/go/v1`, Zen `https://opencode.ai/zen/v1` offered in Web Settings).
- Send the `x-opencode-session` header OpenCode Go requires, keyed to the active Vibe session, so requests stop failing with `400 MissingSessionID`.
- Identify the client with the existing `Vibe-Trading/<version>` User-Agent, which OpenCode's usage policy asks for.

## Why

Closes #1415.

OpenCode Go now rejects every chat completion that does not carry a stable per-conversation `x-opencode-session` header:

```
provider_stream_error provider=deepseek model=deepseek-v4.1-flash:
OpenAIInvalidRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'MissingSessionID',
'message': 'Error from provider (Console Go): Request is missing x-opencode-session and cannot be
routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it'}}
```

The `opencode-go` / `opencode-zen` names added in #444 only map env variables; nothing in the request path sets the header, and there is no Web Settings entry, so the only way to use OpenCode was to re-point another provider (typically `deepseek`) at the Go URL — which fails as above. Per the OpenCode docs (https://opencode.ai/docs/go/), clients must send a stable session id per conversation (used for routing and prompt caching) and identify themselves with a User-Agent.

## Changes

- `src/providers/session_context.py` (new): a `ContextVar` holding the active LLM session id, with `bind_llm_session_id` / `reset_llm_session_id` / `current_llm_session_id`.
- `src/agent/loop.py`: `AgentLoop.run()` binds the run's `session_id` for its whole duration (thin wrapper around the existing body, now `_run_bound`), so adapters can read it without threading it through every call site.
- `src/providers/llm.py`: `ChatOpenAIWithReasoning` injects `x-opencode-session` into `extra_headers` when the provider is `opencode*` **or** the base URL host is `opencode.ai` (so a `deepseek` label pointed at OpenCode also works). Value is the bound session id; outside a session (swarm workers, scheduled research, CLI) a per-adapter UUID keeps one conversation stable. A caller-supplied header of the same name wins.
- `src/providers/capabilities.py`: canonical `opencode` capability (`OPENCODE_API_KEY` / `OPENCODE_BASE_URL`, `capture_reasoning=True`, `User-Agent` default header). `opencode-go` / `opencode-zen` are left exactly as they were.
- `src/providers/llm_providers.json`: catalog entry, which is what drives the Web Settings provider dropdown; `base_url_options` lists the Go and Zen roots.
- `src/swarm/models.py`, `cli/onboard.py`, `cli/_legacy.py`: register the provider in the public allowlist and both CLI onboarding surfaces.
- `README.md`: add OpenCode to the supported-providers line; `agent/.env.example`: OpenCode block.
- `tests/test_opencode_provider.py` (new): catalog/capability/CLI registration, host detection, ContextVar scoping, and wire-level tests via `httpx.MockTransport` (bound id is sent on every request, unbound requests share a stable per-adapter id, non-OpenCode endpoints get no header, explicit header wins).

Only the OpenAI-compatible path is covered. With `langchain-deepseek` installed, a `deepseek`-labelled configuration pointed at OpenCode would take the native `ChatDeepSeek` path and bypass the injection; use `LANGCHAIN_PROVIDER=opencode` in that case (this is the recommended setup anyway).

## Test Plan

- [x] Existing tests pass — ran `pytest agent/tests -k "opencode or llm or provider or settings or swarm or loop or session"` on this branch inside the project Docker image: 1059 passed, 12 skipped, 0 failed. The full suite was not run; the changed files are provider / agent-loop / CLI-registration only.
- [x] New tests added: `agent/tests/test_opencode_provider.py` (23 cases).
- [x] Tested manually: rebuilt the Docker image, switched Web Settings to OpenCode / `deepseek-v4.1-flash` / `https://opencode.ai/zen/go/v1`; preflight reports `LLM (opencode) OK`, a chat session completes (`status: success`), and no `MissingSessionID` appears in logs. Before the change every run against that endpoint failed with the error above.
- `black` and `ruff` are clean on the two new files; the touched upstream files were not reformatted wholesale (they are not currently black-clean upstream), only the added hunks follow the surrounding style.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — this PR does touch `src/agent/loop.py` and `src/providers/`; the discussion issue is #1415. The `loop.py` change is a wrapper that binds/resets a `ContextVar`; the original body is unchanged.
- [x] No hardcoded values (API keys, file paths, magic numbers) — the endpoint defaults live in `llm_providers.json` next to every other provider.
- [x] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines (DCO sign-off on the commit).
- [x] Documentation updated (README provider line, `.env.example`).


warren618:
Thank you — opening #1415 first because this touches `src/agent/` and `src/providers/` was exactly the right order, and the change is careful: the header follows the endpoint rather than the provider label (so a `deepseek` profile re-pointed at `opencode.ai` still sends it), an explicit caller header wins, and the `ContextVar` keeps concurrent sessions apart.

One technical point from tracing where the binding is read. The main `stream_chat` call runs on the thread that entered `AgentLoop.run()`, so it sees the bound session id. Context compaction does not: `_run_compact_summary` calls `self.llm.chat` inside a new `threading.Thread`, and a new thread does not inherit the caller's context, so compaction requests go out under the adapter's fallback UUID — a second conversation id for the same session. That costs a cache miss rather than a 400, but it is the one place the "stable per conversation" property breaks. Running the compaction target through `contextvars.copy_context().run(...)` would fix it, and a test that drives a compaction and asserts the same header value would pin it.

Because this changes `AgentLoop.run()` and provider header resolution — protected core — it gets its own deliberate review rather than going in with today's batch of fixes, so I am leaving it open for that, not for other changes.

he-yufeng (approved):
Read the full diff and ran the suite locally: 18 passed, 5 skipped (platform-gated), on top of a clean venv.

What I verified beyond the tests:

- The `_targets_opencode` matcher on adversarial inputs: `notopencode.ai` and `opencode.ai.evil.com` both correctly miss, bare `opencode.ai/zen/go/v1` without a scheme still matches, and a non-opencode provider re-pointed at the relay (the deepseek-at-Go case) gets the header. The session identity follows the endpoint, not the label, which is the right call.
- The header merge never clobbers an explicit `x-opencode-session` passed by the caller, and the ambient-OpenAI stripping composes correctly with it.
- The ContextVar binding in `loop.py` is set/reset around `run()`, so concurrent sessions can't see each other's id and nothing leaks after the run. The per-adapter UUID fallback means a swarm worker with no bound session still presents one stable conversation identity per adapter, which is the best available shape there.

Closes the 400 MissingSessionID issue cleanly. Nice work.

warren618:
Merged as `a21d388a`, with `fd6c7c7c` on top for the one point above: `_auto_compact` makes its summary call on a fresh daemon thread, which starts with an empty context, so that call went out under the adapter's fallback id. The thread now runs its target through `contextvars.copy_context().run`, and a compaction test reads the bound id from inside the summary call. Thanks — the header following the endpoint rather than the label, and the per-adapter fallback, made this straightforward to finish.

PR #1417, opened 2026-09-11
feat(trading): add a read-only Scalable Capital MCP connector (#1367)

cgycorey:
## Summary

Adds **Scalable Capital** as the 15th broker connector: one read-only `remote_mcp` profile (`scalable-live-mcp-readonly`), with their published tool catalog pinned as a curated read/write map and an off-by-default OAuth channel seeded with an explicit READ allowlist.

The profile deliberately exposes **`quotes.read` only** — see below.

## Why

**Refs #1367** (Scalable Capital), where the tier was settled: **bounded live via the mandate gate**, on the reasoning that no paper environment exists at all, so there is no ambiguity a discriminator would protect against (the Robinhood precedent, not the Longbridge/Dhan cap). This is the read-only first cut that verdict asked for; the order path is deliberately out of scope.

## Changes

| File | What |
|---|---|
| `connectors/scalable/classification.py` | Curated Tier-2 map of the published catalog: 18 READ, 11 WRITE, everything else UNKNOWN → WRITE (fail-closed) |
| `connectors/scalable/profiles.py` | One profile, `scalable-live-mcp-readonly` — `remote_mcp`, `readonly=True`, capabilities `quotes.read` |
| `connectors/scalable/mcp.py` | Generic operation → tool name, plus the two flagged wire assumptions |
| `config/schema.py` | `SCALABLE_MCP_SERVER_SEED` (OAuth, streamableHttp, explicit READ allowlist), plus the live-broker key and URL host |
| `live/registry.py`, `trading/profiles.py`, `trading/service.py` | Curated map, profile registration, generic-read dispatch |
| `tests/test_scalable_connector.py` | 18 tests |
| 6 READMEs + `tests/test_readme_counts.py` | Broker badge 15, table row, "no paper account" paragraph; pin 14 → 15 |

## Capabilities: `quotes.read` only (review item 1)

`account.read` / `positions.read` are **withheld**, along with the `portfolio/compatibility.py` tier that went with them.

Those two are exactly what `is_portfolio_connection_profile` requires, so declaring them made this a portfolio-eligible connection — and nothing maps Scalable's holdings reply into the shared `positions` shape (`_normalize_remote_result` has no Scalable branch), so a refresh stored a **complete snapshot holding zero positions**. Scalable publishes no tool argument schemas, so there is nothing to map against; a normaliser that can only error would be no better than no source.

Both facts are pinned by test — `capabilities == {"quotes.read"}` and `is_portfolio_connection_profile(profile) is False` — so eligibility cannot return without the mapper.

This supersedes the "#1367: their 9 account/portfolio tools qualify for portfolio aggregation" framing: the tools qualify, but the reply shape does not map yet, so declaring the capability would have been a claim the layer cannot honour.

## Classification tests no longer compare the map with itself (review item 2)

`_CURATED_READS` / `_CURATED_WRITES` were derived from `SCALABLE_TOOL_CLASS`, so flipping a WRITE name to READ and seeding it kept the suite green — and the seeded name then assembled as a plain read-only `MCPRemoteTool` instead of being wrapped by `LiveOrderGuardTool` and refused.

The 18 READ and 11 WRITE names are now pinned **literally**, with a test asserting the map matches those sets in both directions and in size. Moving a tool between classes, or seeding one, can no longer pass without a test diff naming that tool.

## `Refs`, not a close keyword (review item 3)

The description and the commit message reference `#1367` without a close/fix/resolve keyword anywhere in the sentence. The two descriptions have been merged into one, and the test count corrected (18).

## Deliberate calls

1. **`preview_buy_order` / `preview_sell_order` pinned WRITE** — a preview mutates nothing, but it sits on the order path and its confirmation feeds `submit_*`. Happy to flip if you would rather call it READ.
2. **No `orders.read`** — the published catalog has no open-order read tool, only `list_portfolio_transactions` (history, a different shape). A capability with no tool behind it is a lie the service layer then has to fail on.
3. **`scopes` omitted from the seed** rather than invented — Scalable publishes no scope names, and an invented scope fails authorization instead of a read.

## What is NOT verified — no account, no schemas

Scalable publishes tool **names** and descriptions but not argument schemas, and nobody here has an account. `mcp.py` carries exactly two wire assumptions, both flagged in its module docstring: the portfolio selector as `portfolio_id`, and quote as `symbols` (the Robinhood shape). Neither is load-bearing for safety — nothing is sent when the caller named no portfolio, so a wrong key fails broker-side rather than silently reading the wrong portfolio. One authenticated `tools/list` replaces both.

**No numeric output is asserted anywhere.** The tests pin the frozen catalog, the profile shape, and that the seeded allowlist equals the curated READ names exactly.

## Tests

CI on the **previous** revision was 10/10 green, including the full suite on Python 3.11 and 3.14; this revision is re-running it, so treat that as unconfirmed until it reports.

Locally I ran the connector / portfolio / live-profile sweep (`pytest -k "scalable or portfolio or live or profile or broker or connector or readme"`): **859 passed, 6 skipped, 6 failed** — and those 6 are the pre-existing `test_repo_tree_states_the_real_agent_tool_count` drift, which fails identically with the whole diff stashed. The full suite is not claimed locally.

Two notes so the numbers are not misread:

- The six `test_repo_tree_states_the_real_agent_tool_count` failures are **pre-existing** — they fail identically with the whole diff stashed.
- The seventh failure the first draft hit *was* mine, and it is now the neatest illustration of the fix: `test_every_builtin_portfolio_connector_has_an_explicit_compatibility_tier` enumerates every portfolio-eligible profile, so a connector declaring `account.read` + `positions.read` fails it until it has an explicit tier. With the pair withheld the connector is not in that set at all, and the gate has nothing to say about it.

## Risk and rollback

**Live/broker risk:** no order path ships; a hand-enabled WRITE tool assembles gate-wrapped and is refused at the intent step — proven by test, with a valid mandate deliberately present. The seeded channel is off-by-default, read-only, and refuses `["*"]`. **MCP/network risk:** no new network calls at import or in tests. **Secret risk:** none. **Deployment risk:** none — no migrations, one new optional seed. **Rollback:** revert this commit; the connector is inert and there is no stored state to unwind.

## Follow-ups, deliberately not here

- **The holdings mapper** — a `normalize_result` for Scalable rows. This is the gate on restoring `account.read` + `positions.read` and the compatibility tier. Left out rather than guessing field names.
- **Order path** — `sdk.py`, an order-intent extractor, the mandate-gated profiles. Also, whenever an EU broker reaches the order path: the mandate `AssetClass` vocabulary has no EU bucket, and EUR is not in `SUPPORTED_VALUE_CURRENCIES` (the gap #1406 recorded).
- **One authenticated `tools/list`** — settles both the argument schemas and the 29-vs-26 count in the tracker post.


warren618:
Thanks — this is careful work. The map checks out name by name against the catalog posted in #1367 (29 names; the categories sum to 29 even though that post's headline says 26), the seed is exactly the READ set and refuses `*`, and a hand-enabled order tool is denied at the intent step. Three things before it can land.

**1. A Scalable portfolio refresh stores a complete snapshot with no positions.** The profile declares `account.read` + `positions.read`, so it is portfolio-eligible, but nothing maps Scalable's reply into the shared shape — `_normalize_remote_result` in `trading/service.py` only has an IBKR branch. I ran a real `PortfolioService.refresh` with this profile and injected readers: an envelope with the holdings under `data` / `structured_content`, and a text-only one, both came back as `complete: true`, zero positions, source `ok`. The EUR refusal the description mentions only happens for a payload that is already in the shared `positions` shape, and nothing produces that shape for Scalable.

Part of that is ours: the portfolio layer read a missing `positions` key as an empty list. We are changing that on `main` so a positions read without a positions list is refused, which turns this into an error source rather than an empty one after a rebase. A source that always errors is still not useful, though, so either add a `normalize_result` for Scalable that returns an error when it cannot find the holdings list, or leave `account.read` / `positions.read` (and the compatibility entry) out of this PR until the `tools/list` schemas exist to map against.

**2. The classification tests compare the map with itself.** `_CURATED_READS` / `_CURATED_WRITES` are built from `SCALABLE_TOOL_CLASS`, and the seed test compares the seed with those derived sets, so flipping a WRITE name to READ and seeding it keeps the suite green. The consequence is concrete: with `submit_sell_order` flipped to READ and added to the seed, it assembles as a plain `MCPRemoteTool` — `is_readonly=True`, not wrapped in `LiveOrderGuardTool` — and `execute` sends `submit_sell_order` to the MCP client. Seeded but still classified WRITE, the same tool is wrapped and refused without a remote call. Please pin the 18 READ and 11 WRITE names as literal sets in the test, so the map cannot drift without a test naming the tool.

**3. `Closes #1367` in the description would close the tracker when this merges.** #1367 tracks every broker request, so please change it to `Refs #1367` with no "close" wording anywhere in the sentence — GitHub acts on the keyword even when it is negated.

Smaller: the description reads as two bodies joined together (it says both 13 and 17 tests; the file has 17). And KIS (#1407) touches twelve of the same files, so whichever lands second will need the broker count and README rows reconciled — no action needed from you on that.

cgycorey:
All three addressed in `3a00e273` (squashed to one commit).

**1. Portfolio reads.** Took your second option: `account.read` / `positions.read` are withheld, together with the `portfolio/compatibility.py` tier. So the profile is no longer portfolio-eligible and cannot store a zero-position snapshot. A normaliser that can only error seemed worse than no source while the reply shape is unpublished — the pair comes back with the mapper once a live `tools/list` settles it. Two assertions pin that: `capabilities == {"quotes.read"}` and `is_portfolio_connection_profile(profile) is False`. Worth noting `test_portfolio_compatibility.py` is now *untouched* by the PR, because the connector is no longer in `eligible_profiles()`.

**2. Classification tests.** `_CURATED_READS` / `_CURATED_WRITES` are pinned literally (18 / 11) instead of derived from the map, and `test_curated_name_sets_match_the_map_exactly` asserts the map matches those sets in both directions and in size. Flipping `submit_sell_order` to READ now fails that assertion, and seeding it fails the seed test.

**3. Close keyword.** Now `Refs #1367`, with no close/fix/resolve keyword anywhere in the description or the commit message. The two descriptions are merged into one and the test count is corrected — 18. Thanks for catching that the keyword acts even inside a negated sentence.

Also updated in the body: the capability table, and the `#1367` framing that "their 9 account/portfolio tools qualify for portfolio aggregation" — the tools qualify, but the reply shape does not map yet, so declaring the capability would have been a claim the layer cannot honour.

One thing I did not do, flagging rather than assuming: I left `preview_*` pinned WRITE. If you would rather call a preview READ, that is a one-line change plus the extractor question when the order path lands.

warren618:
Merged — thank you. Integrated behind the KIS / Upbit / Toss trio, so the broker count lands at 18 and the connector table carries all four rows in the six READMEs. Re-checked on the merged tree: the profile is not portfolio-eligible, the curated READ / WRITE sets are pinned literally and the map is asserted against them both ways, no close keyword anywhere, and the seed's allowlist equals the READ set. `Full suite on the integration branch 0828686f: 13,130 passed, 30 skipped, 0 failed (the earlier 6 README-paragraph failures were my merge, fixed before this run)`.

PR #1419, opened 2026-09-12
fix(grounding): parse decimal-comma and Unicode-minus numerals (#1418)

### LINKED ISSUE #1418: [Bug] Spanish decimal comma and Unicode minus break analysis grounding
### Description

**What happened:**

Analysis grounding mis-parses numbers written in normal Spanish numeric
format.

The Spanish locale naturally produces decimal commas, e.g.:

```text
1,57%
−5,13%
0,188
2,237

but the grounding layer currently assumes an English numeric format.

This causes correct analysis metrics to be rejected or reinterpreted.

Examples observed in real runs:

Model output                  Grounding interpretation
------------------------------------------------------
−5,13%                        13%
1,57%                         57%
−5,132%                       5132%

For tail-risk wording, the confidence level can also be interpreted as
a measured value:

VaR 95%: 1,57%

may validate both 95% and the malformed 57% as candidate VaR values,
although 95% is the confidence level and 1,57% is the actual metric.

The underlying tool output is correct. The corruption happens when the
final natural-language answer is parsed by GroundingLedger.

This reproduces in both the Web UI and the full CLI agent path, so it is
not a frontend-only issue.

The relevant current code shape is:

_MEASURE_NUMBER_RE = re.compile(
    r"[-+]?(?:\d{1,3}(?:,\d{3})+|\d+)(?:\.\d+)\s*[%％]?"
    r"|[-+]?(?:\d{1,3}(?:,\d{3})+|\d+)\s*[%％]"
)

and clause splitting currently uses:

_CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n]")
_THOUSANDS_SEPARATOR_RE = re.compile(r"(?<=\d),(?=\d{3}(?!\d))")

This means:

decimal comma is not a supported decimal separator;
an ASCII comma may split a Spanish decimal into separate clauses;
a value with exactly three digits after the comma can be treated as a
thousands grouping (5,132 -> 5132);
Unicode minus − is not accepted by the measurement regex.

The same implementation is still present on current main.

What I expected:

Grounding should correctly preserve and validate analysis numbers written
using supported locale conventions.

For Spanish, at minimum these should all represent equivalent numeric
values where appropriate:

-5.13%
−5.13%
-5,13%
−5,13%

and:

VaR 95%: 1,57%

should treat 95% as the confidence level and 1,57% as the measured
VaR value.

A localized presentation format should not change the numeric meaning of
the tool evidence.

This appears to be a localization issue in the analysis-claim parser, not
in the financial calculation itself.

I would prefer not to change the shared price-claim parser globally,
because it already has explicit handling for English thousands separators
such as 1,309.22.

A narrow analysis-claim fix with regression coverage for both English and
Spanish numeric formats may therefore be safer.

While investigating this, I noticed separate semantic-grounding questions
around metric-family matching (for example var_95 vs var_99).
I am intentionally not including those in this bug so this report stays
focused on locale-safe numeric parsing.

### Steps to Reproduce

Run from the repository root:

```bash
python - <<'PY'
import tempfile
from pathlib import Path

    ledger = GroundingLedger(
        run_dir=Path(tmp),
        user_message="Analiza el riesgo de esta cartera.",
    )

    ledger._analysis_metrics = [
        {
            "metric": "drawdown",
            "value": -0.0513,
            "tool": "test",
            "call_id": "test",
            "field": "max_drawdown",
        }
    ]

    for value in [
        "-5.13%",
        "−5.13%",
        "-5,13%",
        "−5,13%",
        "−5,132%",
    ]:
        print(
            value,
            GroundingLedger._measure_numbers(value),
            ledger._analysis_value_observed(value, "drawdown"),
        )
PY

Then compare the extraction/validation results.

A real agent-level reproduction also occurs when a correct risk metric is
rendered in Spanish:

Máximo drawdown observado: −5,13%

The structured tool result can contain -0.0513, but the final-answer
grounding may reject the localized representation.

The issue is also visible in prose containing:

VaR 95%: 1,57%
Expected Shortfall 95%: 2,28%

where the confidence level and decimal-comma metric can both be
misinterpreted by the final-answer validator.

### Error Logs

```shell
Examples from real final-answer validation:

analysis_claim_unavailable

Correct model text:
  Máximo drawdown observado: −5,13%

Grounding interpreted candidate:
  13%

Correct model text:
  VaR 95%: 1,57%

Grounding interpreted candidates include:
  95%
  57%

Correct model text:
  Máximo drawdown observado: −5,132%

Grounding interpreted candidate:
  5132%

The structured analysis evidence contained the correct numeric values.
```

### Interface

Docker

### LLM Provider

openai-codex / gpt-5.6-sol via ChatGPT OAuth

### Version

Vibe-Trading v0.1.15 commit cc54832cb50de29d14bb10097b18e08f0a843650  The same numeric regex and clause-splitting implementation are still present on current main as checked on 2026-09-11.

### Environment

Ubuntu 24.04 Docker 29.1.3 Docker Compose 2.40.3 Python 3.11.15 inside the container - Vibe-Trading v0.1.15 Spanish-language prompts/output

he-yufeng:
## Summary

- The grounding layer now normalizes locale numerals before parsing analysis answers: Unicode minus becomes ASCII, and decimal-comma documents (`1,57%`, `−5,13%`, `0,188`) parse as the single numbers they are instead of being split or regrouped.

## Why

Closes #1418. Spanish-format analysis answers were corrupted at parse time: `1,57%` split at the comma into a bogus `57%` claim, `−5,13%` lost sign and integer part into `13%`, and once a document speaks decimal-comma, `5,132` read as `5132`. The tool output underneath was always correct; the final-answer parser mangled it.

## Changes

- One normalize step feeds both number paths (clause splitting and measurement extraction): Unicode minus always maps to ASCII; when the text shows an unambiguous decimal-comma marker (a percent-adjacent group, a leading-zero group English never writes, or a signed group) every digit-comma is a decimal point.
- Bare ambiguous shapes stay English on purpose: `2,237` without any marker keeps its thousands reading, so list prose like "steps 1,2 and 3" can never flip into a decimal.
- Seven regression tests pin the Spanish shapes, the document-level `5,132` case, the leading-zero case, the signed case, and the English no-flip controls.

## Risk / rollback

Low. Texts with no decimal-comma marker are byte-identical to before (the only unconditional change is Unicode-minus normalization, which previously failed to parse at all). Rollback is reverting the single commit; no state, schema, or config is touched.

Known follow-up, deliberately out of scope: in `VaR 95%: 1,57%` the `95%` confidence level is still a candidate measurement alongside the VaR value itself. That is a metric-kind gating question, separate from numeral parsing, and worth its own issue.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`): 12884 passed, 0 failed
- [x] New tests added (7, in `agent/tests/test_agent_grounding.py`)
- [x] Tested manually: red-first confirmed; the four new behavior tests fail on current main and pass with the fix, and the full agent suite runs clean both ways

## Checklist

- [x] Documentation updated (not user-facing; parsing correctness only)

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -752,6 +752,30 @@ _CLAUSE_SEPARATOR_RE = re.compile(r"[,，;；。、\n]")
 # needs a digit before it and exactly three digits after.
 _THOUSANDS_SEPARATOR_RE = re.compile(r"(?<=\d),(?=\d{3}(?!\d))")
 
+# Locale numerals (#1418): decimal-comma detection must never flip English
+# list prose ("steps 1,2 and 3") into a decimal, so a bare `1,5` stays
+# ambiguous-by-design and only unambiguous shapes count as markers: a
+# percent-adjacent comma group, a leading-zero group English would never
+# write, or a signed group (nobody signs a list). Once any marker is present
+# the text speaks decimal-comma and every digit-comma is a decimal point
+# (`5,132` is 5.132 there, not 5132). With no marker the text reads as
+# English and commas keep their thousands meaning.
+_DECIMAL_COMMA_MARK_RE = re.compile(
+    r"\d,\d{1,2}\s*[%％]"
+    r"|(?<![\d,])0,\d{3}(?!\d)"
+    r"|[-+]\d+,\d{1,3}(?!\d)"
+)
+_DECIMAL_COMMA_RE = re.compile(r"(?<=\d),(?=\d)")
+_UNICODE_MINUS = "−"
+
+
+def _normalize_locale_numerals(text: str) -> str:
+    """Normalize Unicode minus and decimal commas into English numerals."""
+    normalized = text.replace(_UNICODE_MINUS, "-")
+    if not _DECIMAL_COMMA_MARK_RE.search(normalized):
+        return normalized
+    return _DECIMAL_COMMA_RE.sub(".", normalized)
+
 
 def _split_clauses(text: str) -> list[str]:
     """Split prose into clauses without breaking a grouped number apart.
@@ -763,7 +787,8 @@ def _split_clauses(text: str) -> list[str]:
         The clause segments, with thousands separators removed so a grouped
         price survives as one number.
     """
-    return _CLAUSE_SEPARATOR_RE.split(_THOUSANDS_SEPARATOR_RE.sub("", text))
+    normalized = _normalize_locale_numerals(text)
+    return _CLAUSE_SEPARATOR_RE.split(_THOUSANDS_SEPARATOR_RE.sub("", normalized))
 _TABLE_SEPARATOR_RE = re.compile(r"^:?-{3,}:?$")
 
 _TABLE_FIELD_ALIASES = {
@@ -3044,7 +3069,8 @@ class GroundingLedger:
     @staticmethod
     def _measure_numbers(text: str) -> list[str]:
         """Extract measurement-shaped numbers (decimal or percent) from a claim."""
-        masked = _LOCALIZED_DATE_RE.sub(" ", text)
+        masked = _normalize_locale_numerals(text)
+        masked = _LOCALIZED_DATE_RE.sub(" ", masked)
         masked = _DATE_RE.sub(" ", masked)
         masked = _SHORT_DATE_RE.sub(" ", masked)
         masked = _DASH_DATE_RE.sub(" ", masked)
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing the VaR follow-up noted in the PR description: I went to verify whether the `95%` in `VaR 95%: 1,57%` gets mistaken for a VaR candidate, and it doesn't materialize — tail-risk metrics aren't in the grounding vocabulary at all (`_ANALYSIS_METRIC_RE` and `_metric_kind_for_text` both return nothing for VaR/ES/CVaR/在险价值/风险价值), so nothing about a VaR claim reaches the gate, confidence figure included. The actual gap is that these claims pass completely unchecked, which is a design decision (new tail-risk kind + a confidence-frame mask), not something to sneak into a parsing PR. Filed as #1425 with the repro and a suggested shape.

warren618:

Issue #1421, opened 2026-09-12
[Bug] Grounding rejects valid multi-metric claims because one metric kind is applied to the whole clause


zeus229:
### Description

The grounding validator can reject a final-answer clause containing multiple valid analysis metrics, even when each metric is individually backed by matching evidence.

This reproduces on an unmodified Vibe-Trading v0.1.15 checkout.

The issue appears to be that `_validate_analysis_claims()` resolves a single metric kind for the whole clause through `_metric_kind_for_text(segment)`, then validates every measurement in that clause against that one kind.

For example, given valid evidence for:

- annualized volatility = 0.23
- max drawdown = -0.05

these two claims validate successfully when written separately:

```text
Annualized volatility was 23%.
Max drawdown was -5%.
```

But the semantically equivalent combined sentence:

```text
Annualized volatility was 23% and max drawdown was -5%.
```

is rejected.

Observed result:

```text
TEXT = Annualized volatility was 23% and max drawdown was -5%.
VALID = False
ISSUES = [
  {
    "code": "analysis_claim_unavailable",
    "claim": "Annualized volatility was 23% and max drawdown was -5%.",
    "value": "23%",
    "kind": "drawdown",
    "message": "No supporting analysis evidence (a completed backtest result or observed risk metric) exists for this figure. Mark the analysis as incomplete and omit these figures."
  }
]
```

The 23% value is valid volatility evidence, but because the clause is classified as `drawdown`, the validator tries to validate 23% as a drawdown measurement.

This is important in normal agent output because LLM-generated financial reports naturally combine related metrics in one sentence, bullet, or table annotation. A correct answer can therefore be rejected and rewritten to omit valid evidence.

### Expected behavior

Each numeric measurement in a clause containing multiple metric kinds should be associated with the metric it semantically belongs to.

At minimum, the following three statements should have equivalent grounding validity when both metrics are observed:

```text
Annualized volatility was 23%.

Max drawdown was -5%.

Annualized volatility was 23% and max drawdown was -5%.
```

The validator should not apply one detected metric kind to every numeric measurement in a multi-metric clause.

### Steps to Reproduce

Steps to Reproduce

Run against an unmodified Vibe-Trading v0.1.15 checkout:

```python
import json
import tempfile
from pathlib import Path

from src.agent.grounding import GroundingLedger

with tempfile.TemporaryDirectory() as tmp:
    ledger = GroundingLedger(
        run_dir=Path(tmp),
[...]

PR #1422, opened 2026-09-12
fix(grounding): don't ingest metadata count fields as return evidence (#1420)

### LINKED ISSUE #1420: [Bug] return_observations is ingested as return evidence and can validate a false return claim
### Description

Analysis grounding can treat a metadata/count field named
`return_observations` as actual return evidence.

`_metric_kind_for_path()` tokenizes compound leaf names and searches for
known metric aliases. Because `return_observations` contains the token
`return`, it is classified as metric kind `return`.

For example:

```text
return_observations = 81

is ingested as:

metric = return
value = 81.0
field = return_observations

This can then validate a financial return claim that was never present in
the tool output:

Annual return: 81%

passes final-answer grounding, while:

Annual return: 80%

is correctly rejected.

Cumulative return: 81% also passes.

This is a false-positive grounding result: a count of return observations
can become evidence for an 81% investment return.

I reproduced this against a clean Vibe-Trading v0.1.15 checkout, without
custom tools or application-specific integrations.

The relevant upstream logic is:

def _metric_kind_for_path(path: str) -> str | None:
    leaf = re.sub(r"\[\d+\]$", "", str(path or "").rsplit(".", 1)[-1])
    leaf = leaf.strip().casefold()

    kind = _ANALYSIS_KIND_ALIASES.get(leaf)
    if kind is not None:
        return kind

    tokens = [token for token in re.split(r"[_.]", leaf) if token]

    for size in (2, 1):
        for start in range(len(tokens) - size, -1, -1):
            kind = _ANALYSIS_KIND_ALIASES.get(
                "_".join(tokens[start : start + size])
            )
            if kind is not None:
                return kind

    return _metric_kind_for_text(path)

Since return is a valid alias, the token scan maps:

return_observations

to:

return

even though the field is metadata about sample size rather than a measured
return.

What I expected:

Metadata/count fields must not be accepted as evidence for financial
metrics merely because their name contains a metric token.

At minimum:

return_observations     -> None
annualized_return       -> return
strategy_max_drawdown   -> drawdown
trade_count             -> None
aligned_days            -> None

and evidence containing only:

return_observations = 81

must NOT make:

Annual return: 81%

a valid grounded claim.

The compound-name support is useful for legitimate fields such as
strategy_max_drawdown and reported_annualized_return, so I am not
suggesting removing compound matching entirely. The issue appears to be
that any matching token can currently supply the metric identity even when
the full leaf describes metadata/counts rather than that metric.

### Steps to Reproduce

Run against a clean Vibe-Trading v0.1.15 checkout:

```bash
python - <<'PY'
import json
import tempfile
from pathlib import Path

from src.agent.grounding import (
    GroundingLedger,
    _metric_kind_for_path,
)

print("=== PATH CLASSIFICATION ===")

for path in [
    "data.inputs.return_observations",
    "data.metrics.annualized_return",
    "data.metrics.strategy_max_drawdown",
    "data.metrics.trade_count",
    "data.inputs.aligned_days",
]:
    print(f"{path:45s} -> {_metric_kind_for_path(path)}")


print("\n=== INGESTION + FINAL-ANSWER VALIDATION ===")

payload = {
    "status": "ok",
    "return_observations": 81,
}

    ledger = GroundingLedger(
        run_dir=Path(tmp),
        user_message="Analyze performance.",
    )

    ledger.ingest_tool_result(
        tool_name="factor_analysis",
        arguments={},
        result=json.dumps(payload),
        call_id="return-observations-probe",
        success=True,
    )

    print("\nANALYSIS_METRICS:")
    for row in ledger._analysis_metrics:
        print(row)

    for text in [
        "Annual return: 81%",
        "Annual return: 80%",
        "Cumulative return: 81%",
    ]:
        result = ledger.validate_final_answer(text)

        print("\nTEXT =", text)
        print("VALID =", result.valid)
        print(
            "ISSUES =",
            json.dumps(
                result.issues,
                indent=2,
                ensure_ascii=False,
            ),
        )
PY

Observed output:

data.inputs.return_observations               -> return
data.metrics.annualized_return                -> return
data.metrics.strategy_max_drawdown            -> drawdown
data.metrics.trade_count                      -> None
data.inputs.aligned_days                      -> None

ANALYSIS_METRICS:
{'metric': 'return',
 'value': 81.0,
 'tool': 'factor_analysis',
 'call_id': 'return-observations-probe',
 'field': 'return_observations'}

TEXT = Annual return: 81%
VALID = True
ISSUES = []

TEXT = Annual return: 80%
VALID = False
ISSUES = [
  {
    "code": "analysis_claim_unavailable",
    "claim": "Annual return: 80%",
    "value": "80%",
    "kind": "return",
    "message": "No supporting analysis evidence ..."
  }
]

TEXT = Cumulative return: 81%
VALID = True
ISSUES = []

The important part is that the false 81% return claims pass only because
the unrelated observation count has been ingested as return evidence.

### Error Logs

```shell
No exception is required to reproduce the issue.

False evidence created by grounding:

{'metric': 'return',
 'value': 81.0,
 'tool': 'factor_analysis',
 'call_id': 'return-observations-probe',
 'field': 'return_observations'}

Final-answer validation:

Annual return: 81%
VALID = True
ISSUES = []

Annual return: 80%
VALID = False
ISSUES = [
  {
    "code": "analysis_claim_unavailable",
    "value": "80%",
    "kind": "return"
  }
]

Cumulative return: 81%
VALID = True
ISSUES = []
```

### Interface

Docker

### LLM Provider

Not applicable. Reproduced directly against GroundingLedger without an LLM.

### Version

Vibe-Trading v0.1.15 commit cc54832cb50de29d14bb10097b18e08f0a843650  Reproduced against a clean detached checkout of that commit, without custom tools or application-specific modifications.

### Environment

Ubuntu 24.04 Docker 29.1.3 Docker Compose 2.40.3 Python 3.11.15 inside the Vibe-Trading container

he-yufeng:
## Summary

- Metadata/count fields (`return_observations`, `benchmark_return_count`, `observation_count`) are no longer ingested as return evidence. A sample size of 81 was validating a fabricated "annual return 81%" claim.

## Why

Closes #1420. `_metric_kind_for_path` tokenizes compound leaves, and the right-to-left token scan matched `return` inside `return_observations`, classifying a count field as metric kind `return`. With that in the evidence set, `Annual return: 81%` passed final-answer grounding while `Annual return: 80%` was correctly rejected.

## Changes

- Compound leaves whose head noun marks metadata (`observation(s)`, `count(s)`) resolve to `None` before the token scan, and the text fallback is skipped on that path so the same token cannot match again through the back door.
- Exact alias leaves (`return`, `total_return`, `annualized_return`) and genuine compound metrics (`return_vol` -> vol, `strategy_max_drawdown` -> drawdown) are unchanged.
- Three regression tests: the leaf-level mapping (including the preserved kinds), a `factor_analysis` JSON-leaf end-to-end where `return_observations=81` must not validate `年化 81%` (fails pre-fix, passes post-fix), and the same shape through the backtest CSV path.

## Risk / rollback

Low. The only fields that stop counting as evidence are count/metadata leaves; every metric-bearing leaf keeps its kind, verified by the preserved-kind assertions and the full grounding suite (240/240) plus the whole agent suite (12880 passed, 0 failed). Rollback is reverting the single commit; no state or schema touched.

Scope note for the record: duration-like fields (`*_days`, `*_periods`) are a separate unit-mismatch class and intentionally left out of this PR.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`): 12880 passed, 0 failed
- [x] New tests added (3, in `agent/tests/test_agent_grounding.py`)
- [x] Tested manually: red-first confirmed; both new behavior tests fail on current main (the fabricated 81% claim validates) and pass with the fix

## Checklist

- [x] Documentation updated (not user-facing; evidence-ingestion correctness only)

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -355,6 +355,11 @@ _ANALYSIS_KIND_ALIASES = {
     "ic_positive_ratio": "win_rate",
 }
 
+# Compound-leaf head nouns that mark a metadata/count field rather than a
+# metric value: return_observations is a sample size, not a return (#1420).
+# Exact alias leaves ("return", "total_return") never reach this check.
+_METADATA_HEAD_TOKENS = frozenset({"observation", "observations", "count", "counts"})
+
 # Order matters: 最大回撤 is drawdown before 收益/return, and 年化波动率 is vol
 # before the generic return branch.
 _ANALYSIS_KIND_PATTERNS: tuple[tuple[re.Pattern[str], str], ...] = (
@@ -1101,12 +1106,18 @@ def _metric_kind_for_path(path: str) -> str | None:
     kind = _ANALYSIS_KIND_ALIASES.get(leaf)
     if kind is not None:
         return kind
+    # Metadata leaves that merely contain a metric token: return_observations
+    # is a sample-size count, not a return (#1420). The head noun decides, and
+    # a denied head must not fall through to the text scan, which would match
+    # the same token all over again.
+    tokens = [token for token in re.split(r"[_.]", leaf) if token]
+    if tokens and tokens[-1] in _METADATA_HEAD_TOKENS:
+        return None
     # Compound leaves name the kind as a token ("reported_annualized_return",
     # "strategy_max_drawdown"). #1338 review: matching only the verbatim alias
     # table makes every other spelling silently ungroundable. Scan from the
     # right — English compounds put the head noun last, so "return_vol"
     # resolves to vol, never to return.
-    tokens = [token for token in re.split(r"[_.]", leaf) if token]
     for size in (2, 1):
         for start in range(len(tokens) - size, -1, -1):
             kind = _ANALYSIS_KIND_ALIASES.get("_".join(tokens[start : start + size]))
--- a/agent/tests/test_agent_grounding.py
+++ b/agent/tests/test_agent_grounding.py
@@ -3747,3 +3747,80 @@ def test_crypto_pair_tables_match_the_resolver() -> None:
     # gold and forex are quoted in it too); grounding decides it by the base
     # whitelist instead, so it is the only permitted difference.
     assert set(g._CRYPTO_QUOTE_ASSETS) | {"USD"} == set(ss._CRYPTO_QUOTE_ASSETS)
+
+
+def test_metadata_count_leaf_is_not_return_evidence(tmp_path: Path) -> None:
+    """#1420: return_observations is a sample-size count, not an 81% return."""
+    from src.agent.grounding import _metric_kind_for_path
+
+    assert _metric_kind_for_path("return_observations") is None
+    assert _metric_kind_for_path("benchmark_return_count") is None
+    assert _metric_kind_for_path("observation_count") is None
+    # real return fields must keep their kind
+    assert _metric_kind_for_path("total_return") == "return"
+    assert _metric_kind_for_path("annualized_return") == "return"
+    assert _metric_kind_for_path("return") == "return"
+    assert _metric_kind_for_path("return_vol") == "vol"
+    assert _metric_kind_for_path("strategy_max_drawdown") == "drawdown"
+
+
+def test_a_count_column_cannot_ground_a_return_claim(tmp_path: Path) -> None:
+    """End to end: an 81 in return_observations must not validate "年化 81%"."""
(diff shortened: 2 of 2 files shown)


warren618:

PR #1424, opened 2026-09-12
fix(run-card): stop discarding non-scalar metrics from the card

cgycorey:
## Summary

`write_run_card` filters metrics through `_scalar_metrics`, which keeps only `str`/`int`/`float`/`bool`/`None`. Every dict- or list-shaped metric is therefore dropped from the card — even though the engine deliberately authors some for card readers.

Concretely: `_on_plan_rejected` records *which* sleeve was rejected and why (`unfilled_plan_rejections_by_symbol = {"BIL": {"zero_size": 540}}`), and both `base.py:1764` and `tests/test_base_engine.py:916` say run-card diagnostics are meant to see it. The card kept only the anonymous scalar `unfilled_plan_rejections` beside it, so a reader could see *that* 551 plans were rejected but never *which* sleeves.

This carries those metrics in a new `card["structured_metrics"]` block — leaving `card["metrics"]` byte-identical — and renders the block in the Run Card tab, so it reaches a reader rather than only the JSON.

## Changes

**Backend** (`agent/backtest/run_card.py`)
- `card["structured_metrics"]`: the non-scalar metrics, with `card["metrics"]` untouched (`_DEDICATED_CARD_KEYS == ("validation",)` is equivalent to the old `key != "validation"`).
- Bounded: a metric over 4 KiB is omitted and named under `_omitted`, so a `by_symbol` map cannot silently grow the card without bound.
- Measured in **bytes of the JSON the card actually writes**. The first cut used `len()` on the string, which counts characters — a CJK value of 4207 UTF-8 bytes passed a gate named `..._MAX_BYTES`.
- Crash-proof: an unserialisable value is omitted, never fatal. The card is written after the run completes.
- `_omitted` is reserved, so a metric of that name is named as omitted rather than silently reinterpreted as the drop record (which also changed its type).
- `validation` excluded (`card["validation"] = metrics["validation"]` is the same object).
- Markdown gains a `## Structured metrics` section. Values are emitted inline, **not** in a code span: a value containing a backtick closed the span early and corrupted the rest of the section.

`warnings` is deliberately **not** excluded. The card's top-level `warnings` comes from `config["content_filter_warnings"]`, while the options engine puts its own annualisation warnings under `metrics["warnings"]` — a different list with no other home. Excluding it would discard them, re-creating the loss this change exists to fix. There is a regression test.

**Frontend**
- A `Structured metrics` panel in `RunCardTab`, and `structured_metrics` on the `RunCard` type.
- Values rendered as JSON, so an empty list reads as `[]` rather than a blank cell that looks like missing data.
- Long metric keys now wrap in the shared `KeyValueTable` key cell instead of overflowing it and colliding with the value.
- Two new `runDetail` strings across all eight locales (ar, de, en, es, ja, ko, pt-BR, zh-CN).

## Verification

- `agent/tests/`: 79 passed / 1 skipped across the suites touching `run_card`.
- `frontend`: 619 passed (64 files), `npm run build` clean (`tsc -b` + vite).
- **Discriminating**: reverting `run_card.py` to `origin/main` fails 4 of the tests.
- `card["metrics"]` proven unchanged: `_scalar_metrics` over one real metrics dict returns an identical dict under the old and new filter.
- Rendered in a real browser (Chromium against the built `dist/`, API mocked) in English **and** Arabic: the panel lays out correctly in both, with no horizontal overflow. The screenshots are what caught the key/value collision and the blank `[]` — neither was visible to the jsdom assertions.
- `tools/ci_grep_gates.sh` passes; files are under the 400-line cap.

## Disclosures

- **The bound is a per-metric budget**, measured on the value's own JSON at top level. Nesting and indentation inside the card add a little on top, so the real ceiling is slightly above the constant.
- **`_omitted` is a reserved name.** A metric literally called `_omitted` is reported as omitted rather than carried. It follows the existing convention that `_`-prefixed card keys are metadata (`_json_hash` already skips them).
- **The key-cell change is in a shared component.** It only affects rendering when a key is wider than the column; every other panel has short keys.
- Cosmetic: in RTL the reserved key renders as `omitted_` (leading-underscore bidi), and very long keys wrap mid-word.
- The `structured_metrics` block is additive, so consumers with size assumptions will see a slightly larger card.

### DIFF
--- a/agent/backtest/run_card.py
+++ b/agent/backtest/run_card.py
@@ -11,6 +11,25 @@ from typing import Any, Mapping, Sequence
 
 
 SCHEMA_VERSION = "0.1"
+# Largest single structured metric the card will carry, counted in BYTES of the
+# JSON the card actually writes (indented, sorted, non-ASCII kept literal). The
+# card is an at-a-glance artefact read on every run, so a structured metric that
+# exceeds this is named under `_OMITTED_KEY` rather than allowed to grow the
+# card without bound.
+_STRUCTURED_METRIC_MAX_BYTES = 4096
+# Reserved: records which structured metrics were dropped. A metric of this name
+# is itself omitted-and-named rather than published, so the record can never be
+# silently overwritten by one.
+_OMITTED_KEY = "_omitted"
+# Metrics whose value is also stored under a dedicated top-level card key, so
+# carrying them here would publish the same object twice. Only ``validation``
+# qualifies: ``card["validation"] = metrics["validation"]``.
+#
+# ``warnings`` deliberately does NOT: the card's top-level ``warnings`` comes
+# from ``config["content_filter_warnings"]``, whereas the options engine puts
+# its own annualisation warnings under ``metrics["warnings"]`` - a different
+# list with no other home. Excluding it would silently discard them.
+_DEDICATED_CARD_KEYS = ("validation",)
 BACKTEST_SUMMARY_KEYS = (
     "codes",
     "start_date",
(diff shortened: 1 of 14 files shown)


warren618:
Merged — thank you. Re-checked on the merged tree: the run-card and plan-rejected suites pass, and the Run Card panel passes the full frontend suite (619 tests) with `tsc -b` clean.

One note for later, not a request: the size check and the writer both serialise with `default=str`, so if a numpy scalar or array ever reaches `structured_metrics` it would be kept as its `str()` text rather than omitted, which is what the `_structured_metrics` docstring says happens to a value that cannot be serialised.

Issue #1426, opened 2026-09-12
[Bug] Generic numeric tool fields can be misclassified as analysis evidence by metric-name tokens


zeus229:
### Description

The analysis grounding layer can currently promote arbitrary numeric fields from successful tool results into analysis evidence based on metric-like tokens in their field names.

This means a numeric field can become authoritative evidence for a financial metric even when the field semantically represents metadata, a count, a diagnostic, or a different measurement.

This is broader than the `return_observations` case reported in #1420.

For example, a successful read-only tool can return a numeric field whose name happens to contain a recognized metric token. `_metric_kind_for_path()` may classify that field into an analysis metric family, and the resulting value can then satisfy `_analysis_value_observed()` for a final-answer claim.

The problem is not that external/custom tools should be prevented from contributing analysis evidence. They are an important extension point. The problem is that successful numeric output and authoritative metric evidence are currently too easy to conflate.

A tool should be able to contribute analysis evidence, but the semantic meaning of that evidence should be explicit or sufficiently constrained rather than inferred from an arbitrary numeric JSON path alone.

### Why this matters

Financial/research tools naturally return a mixture of:

- actual calculated metrics;
- sample sizes and observation counts;
- date/window metadata;
- diagnostics;
- intermediate values;
- descriptive statistics;
- identifiers and flags.

A successful tool result should not make all numeric leaves eligible to validate financial claims merely because their path contains a recognized metric word.

This can produce false acceptance rather than only false rejection: an unsupported metric stated by the model may be accepted because an unrelated numeric field accidentally matches the same metric family.

The issue also becomes increasingly important as custom/read-only research tools are integrated into the agent.

### Confirmed example

#1420 demonstrated one concrete instance:

`return_observations = 81`

was classified as `return` evidence and could therefore make:

`Annual return: 81%`

pass validation despite no 81% return having been observed.

#1422 addresses that particular metadata-head case, but the general evidence-authority problem remains: other numeric field names can still accidentally acquire metric semantics.

### Expected behavior

Successful custom/external tools should remain able to contribute analysis evidence.

However, a numeric field should only become authoritative evidence for a metric when its semantic metric identity is established.

Possible approaches include:

1. an explicit structured metric contract, e.g.

{
  "metrics": [
    {
      "family": "volatility",
      "metric": "annualized_volatility",
      "value": 0.2301,
      "unit": "return_fraction"
    }
  ]
}

2. a tool/schema-level declaration of which output fields are metrics;

[...]

PR #1427, opened 2026-09-12
fix(grounding): gate tail-risk claims (VaR/ES/CVaR) against evidence

### LINKED ISSUE #1425: [Bug] VaR / ES / CVaR tail-risk claims pass grounding completely unchecked
[Bug] VaR / ES / CVaR tail-risk claims pass grounding completely unchecked

Follow-up investigation from #1418's confidence-level sub-case (and #1419's documented residual). While verifying whether the `95%` in `VaR 95%: 1,57%` gets validated as a VaR candidate, I found the deeper thing: tail-risk metrics are not in the grounding vocabulary at all, so these claims never reach the gate.

Reproduced on current main (`f84b2977`):

```text
_ANALYSIS_METRIC_RE.search / _metric_kind_for_text:
'VaR 95%: 1.57%'              -> metric: False, kind: None
'95% 置信水平下 VaR 为 -1.57%' -> metric: False, kind: None
'在险价值 VaR 为 1.57%'       -> metric: False, kind: None
'风险价值 1.57%'              -> metric: False, kind: None
'VaR (95%) = 1.57%'           -> metric: False, kind: None
'ES 1.9%'                     -> metric: False, kind: None
'CVaR 2.1%'                   -> metric: False, kind: None
```

End to end, with drawdown evidence `-0.0157` ingested, both `VaR 95%: 1,57%` and `95% 置信水平下 VaR 为 -1.57%` validate as `valid=True` with zero issues — not because the numbers checked out (the gate never looked), and equally not because the confidence figure was handled. Any invented tail figure passes the same way, e.g. `VaR 95%: 9,99%` would also pass untouched.

Two implications:

1. The confidence-contamination sub-case from #1418 does not materialize on the prose path: nothing about a VaR claim is validated, so `95%` is never mistaken for a VaR candidate either. Closing that thread of #1419's residual note.
2. The real gap is bigger: tail-risk claims (VaR / ES / CVaR / 在险价值 / 风险价值) are entirely ungrounded. Gating them is a design decision, not a drive-by fix — the kind taxonomy needs a tail-risk kind (or mapping to drawdown), and the confidence frame (`VaR <conf>%: <value>`, `<conf>% 置信水平下 VaR 为 <value>`) needs its own handling so the confidence percentage never becomes a measurement. Filing this so the design has a home instead of a drive-by behavior change.

Suggested shape when someone picks it up: a `tail_risk` metric kind covering VaR/ES/CVaR/在险价值/风险价值, evidence keyed from backtest outputs (VaR columns in metrics.csv map into it), and a confidence-frame mask so `95%` in `VaR 95%: x` is exempt from measurement, mirroring `_LABELLED_SCORE_RE`.

he-yufeng:
## Summary

- Tail-risk claims (VaR / ES / CVaR / 在险价值 / 风险价值) now pass through the analysis grounding gate like every other metric: they classify as a new `tail_risk` kind and must match kind-scoped evidence from a completed backtest or risk-tool output.
- The confidence frame is masked before measurement, so the `95%` in `VaR 95%: -1.57%` is never treated as the claimed figure.

## Why

Closes #1425. On current main, tail-risk vocabulary is absent from the grounding gate: `VaR 95%: 9,99%` (an invented number) validates as sound with zero issues because the claim never reaches the evidence check. A trust layer that silently skips exactly the figures a risk report is read for is worse than no gate. The issue has the full repro and the design shape this PR implements.

## Changes

- Detector vocabulary gains VaR/CVaR/expected shortfall and the Chinese tail-risk terms. A bare `ES` counts only with a figure attached — it is also the E-mini S&P ticker, and "ES futures closed at 5,210" is a price claim.
- Evidence aliases map quantlib's `var`/`cvar` leaves and `var_95`/`es_99`-style CSV headers to the new kind. Tail-risk figures compare by magnitude like drawdown, because quantlib reports positive loss magnitudes while a report may write the same figure signed.
- A confidence-frame mask mirrors the labelled-score mask: `VaR 95%: x`, `VaR (99%) = x`, `95% 置信水平下 VaR 为 x` measure only `x`; a bare `CVaR 2.1%` still reads 2.1% as the value.
- New `test_grounding_tail_risk.py` (18 tests): vocabulary detection in both languages, confidence-frame masking, grounded/invented/missing-evidence directions end to end, and the ES-ticker false-positive guard.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q` for the grounding battery: 363 passed, 5 skipped, plus the 18 new)

Manual: the issue's repro claims (`VaR 95%: 1,57%` with only drawdown evidence present, and the invented `9,99%` figure) now produce `analysis_claim_unavailable` with `kind: tail_risk`; the grounded `-1.57%` against `var_95: 0.0157` passes. Note the decimal-comma spelling is parsed by the still-open #1419; on this branch alone, `1,57%` measures as its integer parts, which still rejects the invented figure (both directions covered by tests with ordinary decimals).

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion — this touches the analysis grounding gate; the design was discussed in #1425 before implementation.
- [x] Community commits must include the DCO `Signed-off-by:` trailer; see `CONTRIBUTING.md`.

## Risk / Rollback

Low and local: the gate gains one metric kind and one mask. Reports whose tail-risk figures are invented or unevidenced now get a recovery prompt instead of shipping silently; every previously-grounded claim shape is unchanged (the full grounding battery passes). Rollback is a single-commit revert.

Prepared with AI assistance (Kimi); the repro, tests, and diff above were verified locally by the committer.

### DIFF
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -281,7 +281,12 @@ _ANALYSIS_METRIC_RE = re.compile(
     r"\bprob(?:ability)?\.?\s+of\b|\bvolatility\b|\bdrawdown\b|"
     r"\bannualiz\w*\b|\bwindow(?:s)?\b|\bregime(?:s)?\b|"
     r"\b(?:annual|cumulative|total)\s+return\b|"
-    r"夏普|回撤|波动率|胜率|命中率|概率|年化|回测|窗口|收益(?:率)?|回报(?:率)?)",
+    # tail-risk metrics (#1425): VaR/CVaR/expected shortfall. A bare "ES" is
+    # also the E-mini S&P ticker, so it only counts with a figure attached;
+    # "ES futures closed at ..." is a price claim, not a risk metric.
+    r"\bvar\b|\bcvar\b|expected\s+shortfall|\bes\b(?=\s*[-+]?\d)|"
+    r"夏普|回撤|波动率|胜率|命中率|概率|年化|回测|窗口|收益(?:率)?|回报(?:率)?|"
+    r"在险价值|风险价值|预期尾部损失)",
     re.IGNORECASE,
 )
 # Phrases that indicate a figure is attributed to an external source rather
@@ -353,11 +358,35 @@ _ANALYSIS_KIND_ALIASES = {
     "return": "return",
     "returns": "return",
     "ic_positive_ratio": "win_rate",
+    # tail-risk evidence leaves (#1425): quantlib's risk tools report var/cvar
+    # as positive loss magnitudes; CSV headers carry var_95/es_99 shapes.
+    "var": "tail_risk",
+    "cvar": "tail_risk",
+    "es": "tail_risk",
+    "expected_shortfall": "tail_risk",
+    "value_at_risk": "tail_risk",
+    "historical_var": "tail_risk",
+    "historical_cvar": "tail_risk",
+    "var_95": "tail_risk",
+    "var_99": "tail_risk",
+    "cvar_95": "tail_risk",
+    "cvar_99": "tail_risk",
+    "es_95": "tail_risk",
+    "es_99": "tail_risk",
 }
 
 # Order matters: 最大回撤 is drawdown before 收益/return, and 年化波动率 is vol
 # before the generic return branch.
 _ANALYSIS_KIND_PATTERNS: tuple[tuple[re.Pattern[str], str], ...] = (
+    # tail risk first: "风险价值" carries no drawdown/return token, but the
+    # confidence frame ("VaR 95%") must classify before the % can wander.
+    (
+        re.compile(
+            r"(?:在险价值|风险价值|预期尾部损失|expected\s*shortfall|\bcvar\b|\bvar\b|\bes\b)",
+            re.IGNORECASE,
+        ),
+        "tail_risk",
+    ),
     (re.compile(r"(?:回撤|drawdown|maxdd|最大亏损)", re.IGNORECASE), "drawdown"),
     (
         re.compile(
@@ -518,6 +547,18 @@ _QUANTITY_WITH_UNIT_RE = re.compile(
     r"|(?:shares?|contracts?|lots?|units?|sessions?|bars?|periods?|"
     r"wks?|weeks?|months?|days?|years?|yrs?)\b"
     r")",
+    re.IGNORECASE,
+)
+# The confidence figure in a tail-risk frame is part of the metric's name,
+# not a measurement: in "VaR 95%: -1.57%" or "95% 置信水平下 VaR 为 -1.57%"
+# the 95% never reaches the evidence check, only the -1.57% does (#1425 —
+# mirrors _LABELLED_SCORE_RE, which does the same for "CONFIDENCE: 6").
+_TAIL_RISK_CONFIDENCE_RE = re.compile(
+    r"(?:\bvar\b|\bcvar\b|expected\s*shortfall|在险价值|风险价值|预期尾部损失)"
+    r"\s*[(（]?\s*[-+]?\d+(?:\.\d+)?\s*[%％]\s*[)）]?"
+    r"(?=\s*[:：=]?\s*[(（]?[-+]?\d)"  # a confidence is followed by the value; a bare "CVaR 2.1%" is the value
+    r"|"
+    r"[-+]?\d+(?:\.\d+)?\s*[%％]\s*(?:置信水平|置信度|confidence(?:\s+level)?)\s*[下上的]?",
     re.IGNORECASE,
 )
 # A conviction reading is on a labelled scale, not a price scale: the 6 in
@@ -3048,6 +3089,8 @@ class GroundingLedger:
         masked = _DATE_RE.sub(" ", masked)
         masked = _SHORT_DATE_RE.sub(" ", masked)
         masked = _DASH_DATE_RE.sub(" ", masked)
+        # The confidence in "VaR 95%: x" frames the measurement; it is not one.
+        masked = _TAIL_RISK_CONFIDENCE_RE.sub(" ", masked)
         return [
             match.group(0).replace(" ", "").replace(",", "")
             for match in _MEASURE_NUMBER_RE.finditer(masked)
@@ -3108,7 +3151,10 @@ class GroundingLedger:
                 continue
             observed.append(float(record.value))
         candidates = {value, value / 100.0}
-        if kind == "drawdown":
+        if kind in ("drawdown", "tail_risk"):
+            # Tail-risk sign conventions disagree the same way drawdown's do:
+            # quantlib reports VaR/CVaR as positive loss magnitudes while a
+            # report may write the same figure as -1.57%.
             candidates |= {abs(value), abs(value) / 100.0}
 
         def close(candidate: float, item: float) -> bool:
--- a/agent/src/agent/grounding.py
+++ b/agent/src/agent/grounding.py
@@ -2953,9 +2953,11 @@ class GroundingLedger:
                     continue
                 kind = _metric_kind_for_text(segment)
                 unsupported = [
-                    value
-                    for value in values
-                    if not self._analysis_value_observed(value, kind)
+                    (value, value_kind)
+                    for value, value_kind in GroundingLedger._measure_numbers_with_kinds(
+                        segment, kind
+                    )
+                    if not self._analysis_value_observed(value, value_kind)
                 ]
                 if not unsupported:
                     continue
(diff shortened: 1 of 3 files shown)


cgycorey:
Verified zeus229's identity question against this head, and found one more case alongside it.

**Scope note first:** before this PR, tail-risk claims weren't gated at all — so both items below are residual coarseness in a strict improvement, not regressions introduced here.

### 1. Measure *and* confidence are both discarded → false acceptance

The gate compares kind, not identity. Using the harness from `test_grounding_tail_risk.py`, with `metrics.json = {"var_95": 0.0157}`:

| claim | result |
|---|---|
| `VaR 95%: -1.57%` | accepted — correct |
| `VaR 99%: -1.57%` | **accepted** — 99 vs 95 ignored |
| `CVaR 99%: -1.57%` | **accepted** — measure *and* confidence ignored |

Mirrors: `{"var_99": 0.0157}` accepts `VaR 95%`; `{"cvar_95": 0.0157}` accepts `VaR 95%`.

So one tail-risk figure at any confidence grounds a claim for any of the three measures at any confidence. zeus229's shape — keep `tail_risk` as the family, carry `metric = var|es` and `confidence = 0.95|0.99` as identity — closes this, and would mean the comparison uses all three rather than `kind` alone.

### 2. bare `es` is missing from the confidence mask → false rejection (untested)

`_TAIL_RISK_CONFIDENCE_RE`'s first alternative enumerates `var|cvar|expected shortfall|在险价值|风险价值|预期尾部损失` — `es` is not there:

```
GroundingLedger._measure_numbers("VaR 95%: -1.57%")  -> ['-1.57%']
GroundingLedger._measure_numbers("ES 95%: -1.57%")   -> ['95%', '-1.57%']
```

So with *correct* evidence `{"es_95": 0.0157}`:

| claim | result |
|---|---|
| `ES 95%: -1.57%` | **rejected** |
| `95% 置信水平下 ES 为 -1.57%` | accepted |

Same metric, same evidence, different phrasing — the inline form leaks the confidence in as the measurement, while the phrase form is masked via the second alternative. It fails closed, so this is a usability bug rather than a safety one, but it rejects a true claim.

It also shows up indirectly in §1: with `{"var_95": 0.0157}`, `ES 99%: -1.57%` is rejected while `VaR 99%` and `CVaR 99%` are accepted — the inline `ES` form is accidentally the *safest* of the four, for the wrong reason.

It's a test gap too: the `test_confidence_figure_is_never_the_measurement` parametrize covers `VaR 95%`, `VaR (99%)`, `95% 置信水平下 VaR` and `99% confidence level CVaR`, but has no `ES <conf>%:` row. Adding one fails on this head:

```
("ES 95%: -1.57%", ["-1.57%"])   # actual: ['95%', '-1.57%']
```

### Repro

```
git fetch origin pull/1427/head && git checkout FETCH_HEAD
cd agent && PYTHONPATH=. python3 -c "
from src.agent.grounding import GroundingLedger as G
print(G._measure_numbers('VaR 95%: -1.57%'))
print(G._measure_numbers('ES 95%: -1.57%'))"
```

For §2 the fix is small and self-contained: add `\bes\b` to the first alternative of `_TAIL_RISK_CONFIDENCE_RE`, plus the `ES <conf>%:` row above.

he-yufeng:
@cgycorey Both fixed in aa8229e0, with the test matrix you sketched now passing.

**§1 identity.** The comparison now carries (measure, confidence) per claim value against the same pair parsed from the evidence field name: `VaR 95%` with `{"var_95": 0.0157}` grounds, `VaR 99%` and `CVaR 99%` both reject, `{"cvar_95"}` grounds `ES`/`CVaR` but not `VaR`, and a confidence-free `var` still grounds any confidence (the side that names nothing never narrows the match, so quantlib's bare leaves stay usable). Identity is read per value from the markers preceding it, so a segment mixing `VaR 95%: -1.57%` and `99% 置信水平下 CVaR 为 -2.1%` checks each figure against its own frame rather than letting the last marker win both.

**§2 mask.** `es` is in the first alternative now, and your row (`ES 95%: -1.57%` -> `['-1.57%']`) is pinned. That surfaced a third leak the same class covered: the value-following lookahead only accepted `:`/`=`, so the Chinese copula in `CVaR 95% 为 -2.1%` (and English `is`) let the confidence slip through unmasked too; both are in the separator set now.

One thing I had to change in my own earlier test rather than the code: the original "grounded claims pass" case used `99% 置信水平下 CVaR` against `cvar_95` evidence and passed under the loose family rule. It correctly rejects now, so the test moved to 95/95 — that was my test encoding the over-acceptance, exactly your §1 case.

zeus229:
Thanks — I tested aa8229e0 in an isolated worktree and the measure+confidence
identity fix works well: the VaR95/VaR99/CVaR95/CVaR99/ES95/ES99 matrix passed
16/16, and the new inline ES confidence mask also works.

I found three remaining natural-language cases worth checking before merge.

### 1. `Expected Shortfall` can bypass the analysis gate entirely

This appears pre-existing rather than introduced by this PR.

`_FORECAST_FRAME_RE` contains bare `\bexpected\b`, so the standard metric name
"Expected Shortfall" is interpreted as forecast language.

With valid evidence such as:

```json
{"cvar_95": 0.0228}

this fabricated claim can validate with zero issues:

Expected Shortfall 95% was -99.9%.

The tail-risk vocabulary is detected correctly, but _validate_analysis_claims
then skips the clause because _FORECAST_FRAME_RE matches Expected.

This seems like a false-acceptance path because "Expected Shortfall" is the
metric's proper name, not a forecast frame.

A regression test could assert that an invented Expected Shortfall value is
still checked against evidence.

2. Past-tense was is not accepted by the confidence mask

For example:

VaR 95% was -1.57%

leaks 95% into the measurement list and rejects otherwise-correct evidence.

The mask currently handles separators including is, but not was.

3. Natural descriptive text between metric and confidence

After composing this PR with the decimal-comma normalization fix, this natural
Spanish form still rejects correct evidence:

El VaR histórico diario al 95% fue 1,57%.

The confidence mask appears to require the metric name and confidence value to
be nearly adjacent, so histórico diario al prevents recognition of 95% as
the confidence level.

This may also affect similar English forms such as:

historical daily VaR at 95%

The core identity fix itself looks good; these seem to be parsing/gating edge
cases around natural wording rather than problems with the new
(measure, confidence) comparison.

he-yufeng:
@zeus229 All three fixed in 9dbd4cd3, each with a regression test that fails without the change:

1. **`Expected Shortfall` bypass.** The forecast frame's bare `expected` now requires no following `shortfall`, so the metric's proper name no longer reads as forecast prose and the invented ES value hits the evidence check (rejected against `cvar_95` evidence). A generic forecast ("expected return of 4.2%") still skips the gate as before — pinned both ways.

2. **`was` separator.** `was` (and `fue`) joined the separator set after the confidence, so "VaR 95% was -1.57%" no longer leaks 95% into the measurement list.

3. **Descriptive text between metric and confidence.** A bounded whitelist of bare connectives (histórico/diario/al/at/fue/was/is and the English mirrors) now bridges the gap in both the confidence mask and the identity extractor, so "El VaR histórico diario al 95% fue 1,57%" and "historical daily VaR at 95%" keep 95% as the confidence. No measure words are allowed through, so identity still can't bleed across metrics ("VaR historical CVaR at 95%" reads as CVaR at 95%, not VaR).

One scope note: the Spanish case's decimal-comma value parse is #1419's territory (still open); this PR makes the gate recognize the frame, and #1419's parse fix completes the end-to-end chain.

279 grounding tests green after merging current main, including the language-parity suite.

warren618:
Re-ran on `9dbd4cd3` against `main` as control: the invented `Expected Shortfall 95% was -99.9%` now rejects (it validated on `main`), `VaR 95% was -1.57%` and the Spanish / English descriptive forms validate, a fabricated `VaR 95% was -7.77%` rejects, and `VaR 99% was -1.57%` rejects against `var_95` evidence — the identity check does what it says.

One case from my earlier pass still stands. With `var_95 = 0.0157` and `max_drawdown = -0.05` both in evidence, `VaR 95%: -1.57% and max drawdown was -5%.` rejects the `-5%`: the clause takes the tail-risk kind as a whole, so the drawdown figure is checked against the tail-risk values. That is the #1421 shape #1423 fixes per value, and the two PRs rewrite the same `_validate_analysis_claims` block — sequencing them naively rejects every VaR claim on its `95%` instead, because `_measure_numbers_with_kinds` does not apply the confidence mask. They need one measurement function: per-value kind from #1423, the confidence mask and the (measure, confidence) identity from this PR. Could you rebase this onto #1423 (or fold it in) and add the mixed clause above as a test? I will re-run the full set on the result.

he-yufeng:
@warren618 Done as `e13ab882`: merged #1423 in (plain merge, no rebase, so both PRs stay independently landable) and unified the measurement path as you described — one function now carries #1423's per-value kind, the confidence mask, and this branch's (measure, confidence) identity. `_measure_numbers_with_kinds` applies the confidence mask, so `VaR 95%: -1.57%` no longer measures the 95% at all.

Your mixed clause is pinned as a test both ways: with `var_95` + `max_drawdown` in evidence, `VaR 95%: -1.57% and max drawdown was -5%.` validates; with only `var_95`, the -5% rejects as drawdown (and the mirror: only drawdown evidence, the VaR figure rejects as tail_risk). Red side checked against both parents: on `9dbd4cd3` the mixed clause rejects the drawdown figure as you saw, on `9a735ea3` the unmasked 95% rejects everything in the clause; on the merge all three pass. Full battery: 285 grounding-file tests green.

Ready for your re-run.

warren618:

Issue #988, opened 2026-08-05
[Bug] Shadow Account backtest run_dir (~/.vibe-trading/shadow_runs) is not in the sandbox's default allowed run roots


78k4srcz95-bit:
### Description

The Shadow Account module and the backtest sandbox disagree about where shadow runs live, so run_shadow_backtest fails out of the box on a default install.

- src/shadow_account/storage.py: runs_dir() materializes each run under ~/.vibe-trading/shadow_runs/<shadow_id>, and the backtester passes that path to the backtest runner.
- The runner's run-root validation (src/tools/path_utils.py) checks the run_dir against a default allowlist that contains agent/runs, agent/.swarm/runs, and sandbox-home variants (%TEMP%/vibe-sandbox-home-*/.vibe-trading/shadow_runs) — but NOT the real-home ~/.vibe-trading/shadow_runs that storage.py actually writes to.

Result: every run_shadow_backtest call dies with "run_dir ... is outside allowed run roots" unless the user manually sets VIBE_TRADING_ALLOWED_RUN_ROOTS (and under an MCP client, figures out that it must go into the client's server env block, since a shell export does not reach the spawned server). The failure is also silent at the tool level — the MCP response only says {"error": "backtest failed"}; the real reason is only visible in logs/runner_stdout.txt inside the run dir.

Expected: the path the shadow module itself chooses for its run dirs should be trusted by default.

Suggested fix: include the resolved shadow runs root (storage.runs_dir's parent, i.e. <storage root>/shadow_runs) in the default allowed run roots, so it stays consistent wherever the storage root resolves to. Surfacing the runner's stderr/stdout error in the tool response instead of a generic "backtest failed" would also save users a lot of digging.

### Steps to Reproduce

1. Fresh install, no VIBE_TRADING_ALLOWED_RUN_ROOTS set
2. extract_shadow_strategy on any journal -> shadow_id
3. run_shadow_backtest with that shadow_id
4. Tool returns {"error": "backtest failed"} for every market; logs/runner_stdout.txt in ~/.vibe-trading/shadow_runs/<shadow_id>/ shows the real error

### Error Logs

```shell
{"error": "run_dir 'C:\\Users\\<user>\\.vibe-trading\\shadow_runs\\shadow_xxxx' is outside allowed run roots.
Allowed roots:
  - C:\Users\<user>\Vibe-Trading\agent\runs
  - C:\Users\<user>\Vibe-Trading\agent\.swarm\runs
[...]

Issue #963, opened 2026-08-03
[Bug] Backtest not allowed to be run in other directory


a1403118221:
### Description

I use Codex to connect to the MCP of vibe-trading and I want to run a backtest. I want it to store the result in another dir. So I create a new dir as my workspace. I tell it to use vibe-trading MCP to run a backtest on a stock and produce the result. Then it says, {"status": "error", "exit_code": 1, "stdout": "{"error": "run_dir 'C:\\\\Users\\\\Administrator\\\\Desktop\\\\\u83b7\u53d6\u63a8\u8350\u80a1\\\\maotai_backtest' is outside allowed run roots. Set VIBE_TRADING_ALLOWED_RUN_ROOTS to add a run directory."}\n", "stderr": "", "artifacts": {}, "run_dir": "C:\Users\Administrator\Desktop\获取推荐股\maotai_backtest"},{
  "result": "{"status": "error", "exit_code": 1, "stdout": "{\"error\": \"run_dir 'C:\\\\\\\\Users\\\\\\\\Administrator\\\\\\\\Desktop\\\\\\\\\\u83b7\\u53d6\\u63a8\\u8350\\u80a1\\\\\\\\maotai_backtest' is outside allowed run roots. Set VIBE_TRADING_ALLOWED_RUN_ROOTS to add a run directory.\"}\n", "stderr": "", "artifacts": {}, "run_dir": "C:\\Users\\Administrator\\Desktop\\获取推荐股\\maotai_backtest"}"
}

### Steps to Reproduce

Use Codex to connect to the MCP of vibe-trading and run a stock backtest in another dir, then it occurs. 

### Error Logs

```shell
{"status": "error", "exit_code": 1, "stdout": "{"error": "run_dir 'C:\\\\Users\\\\Administrator\\\\Desktop\\\\\u83b7\u53d6\u63a8\u8350\u80a1\\\\maotai_backtest' is outside allowed run roots. Set VIBE_TRADING_ALLOWED_RUN_ROOTS to add a run directory."}\n", "stderr": "", "artifacts": {}, "run_dir": "C:\Users\Administrator\Desktop\获取推荐股\maotai_backtest"},{
[...]

PR #1429, opened 2026-09-13
fix(tools): name the run_dir the backtest tool actually looked in

cgycorey:
## Summary

`backtest` reported a missing file without saying which directory it checked:

```json
{"status": "error", "error": "config.json not found"}
```

So a caller whose `run_dir` did not land where it expected could not tell that apart from a genuinely empty directory. Both pre-flight checks now name the resolved path and the layout each file belongs to:

```json
{"status": "error",
 "error": "config.json not found in /home/…/artifacts/cycle_strategist",
 "hint": "config.json belongs at the root of run_dir."}
```

**No behaviour changes.** The same directory is validated, the same files are checked, the same branches are taken. Only the message becomes actionable.

## Why this shape

It mirrors the sibling missing-artifact envelope already in the codebase — `agent/src/tools/autopilot_tool.py`:

```python
card_path = run_path / "run_card.json"
if not card_path.exists():
    return json.dumps(
        {
            "status": "error",
            "error": f"run_card.json not found in {run_path}",
            "hint": "Run the backtest tool first; it writes run_card.json.",
        },
        ensure_ascii=False,
    )
```

This applies that existing convention to the backtest tool's two pre-flight checks. It also follows `539b9ab6` ("name the allowed roots when a path is refused"), which fixed the sibling branch for the stated reason that *"the user had no way to see where the boundary actually was"* — that one names the roots; this one names the resolved directory.

## Why it matters

Found on a real swarm run (`commodity_research_team`, 3/3 agents completed). The worker replaces `run_dir` with the agent's workspace before every tool call, so the strategist's paths were all discarded; it retried **10 times** and correctly reported that it could not produce performance data. Two things made it undiagnosable: this error, which named no path, and the event stream, where `run_dir` is redacted from the argument preview (`events.jsonl` shows `"arguments": {}` for a call that carried one). Naming the path addresses the first — with it the agent sees the workspace root in the message and can act, since `skills/strategy-generate/SKILL.md` already requires `config.json` at the root of `run_dir`.

## Verification

- `agent/tests/test_backtest_tool_state.py` — extended one test, added one, asserting the path the tool reports (`str(tool_run_dir.resolve())` — what `safe_run_dir` emits) and the hint for both files.
- **16 passed** across the two suites that import this tool (`test_backtest_tool_state.py`, `test_file_tool_sandbox_security.py`).
- **Discriminating**: with `backtest_tool.py` reverted to `origin/main` and the tests kept, both new assertions fail (`2 failed, 4 passed`); they pass with the fix.
- `ruff check` clean. `black` leaves the added lines untouched; it is not applied to the rest of the files, which are already non-black on `origin/main` — `CONTRIBUTING.md:149` asks contributors not to mix unrelated formatting into a focused PR.
- Same-run confirmation: staging `config.json` + `code/signal_engine.py` at the root of a run dir returns `exit_code: 0`, so the hint names the layout the tool actually accepts.

## Validated in a live re-run

Re-ran the same preset, model and question that produced the failure (`commodity_research_team`, `deepseek-v4.1-flash`, WTI, 3 months):

| | before | after |
|---|---|---|
| backtest calls | 10 | 3 |
| failures | 10 × `config.json not found` (no path) | 1 × `code/signal_engine.py not found in <workspace root>` + hint |
| successes | **0** | **2 × `exit_code: 0`**, real metrics (`total_return 0.284`, `sharpe 0.246`, max DD −0.31, 38 trades) |

The failure at iteration 5 was followed by success at iteration 7 on the same staged directory: the agent was told which file was missing and which directory the tool had inspected, and closed the gap. The baseline run reported no performance data at all.

Stated plainly: this is **not an isolated A/B** — the agent's file layout and the instrument differed between the runs (`WTI.US` here, `CL=F` in the baseline). The direction is unambiguous, since a named path can only help, but the effect size is not isolated and the underlying substitution is unchanged.

## Not in scope

- **The `run_dir` substitution itself.** Whether the worker should honour a `run_dir` the tool declares, or keep pinning it to the agent workspace, is a design decision — the schema says `run_dir` is required, the worker discards it, `agent/loop.py::_normalize_tool_run_dir` keeps it, and the skill says `config.json` lives at the root of `run_dir`. Raised separately rather than changed here — opened as **#1430**.
- **`_preview_tool_arguments` redacting `run_dir`.** `test_preview_tool_arguments_redacts_sensitive_values` pins that deliberately, so it stays.
- **`backtest/runner.py`** prints the same bare messages (lines 1220/1240) but is unreachable through the tool, which pre-validates both files, and does not import `src.tools`.

## Not a duplicate

- **#963** (fixed by `539b9ab6`) — the *outside-allowed-roots* rejection. Different branch.
- **#988** — added the shadow-runs root to the allowlist.
- **#1353** — the failure to *converge* after a path cannot be resolved. This is one layer down: the path was substituted, so no retry could converge.


warren618:
Merged — thank you. Both assertions fail with the tool reverted and pass with it, and the envelope now matches the `autopilot_tool` sibling.

PR #1430, opened 2026-09-13
fix(swarm): confine a tool-declared run_dir to the agent workspace

cgycorey:
## Summary

The swarm worker built every tool call's arguments by overwriting `run_dir`:

```python
args = {**tc.arguments, "run_dir": str(artifact_dir)}
```

That value is the **confinement root** the file tools resolve against, so pinning it is correct for tools that do not declare `run_dir` — `write_file`, `read_file`, `bash`, `edit_file`. But five tools **do** declare it, and for those the model was supplying a required argument that could never take effect:

| tool | `run_dir` | preset agents exposing it |
|---|---|---|
| `backtest` | required | **19** |
| `pattern` | required | 2 |
| `scaffold_signal_engine`, `link_autopilot_backtest` | required | 0 (custom presets) |
| `add_goal_evidence` | optional | 0 (custom presets) |

On a real run (`commodity_research_team`), the strategist passed ten different paths, all discarded, and every backtest failed with `config.json not found` whatever it asked for — with nothing in the error or the event stream to show the argument had been replaced.

## What changed

A declared `run_dir` is honoured **inside the same boundary**:

- relative → resolved under the agent's workspace (the filesystem the model's own file tools showed it, e.g. `runs/<name>`);
- absolute **inside** the workspace → taken as-is;
- anything resolving **outside** → refused, with an error naming the workspace.

Undecalred tools are unchanged: the workspace is still injected.

**Why refuse rather than fall back.** Silent substitution is what hid this failure, so the failure mode is now explicit. But to be precise about the risk, because an earlier version of this description overstated it: **no** call can depend on an outside path being *honoured* (the value is discarded today), yet a call **can** depend on the workspace *fallback* — and one demonstrably did. A successful re-run's `run_card.json` records the injected workspace as the directory the engine actually used, so that run's backtest succeeded *because* the model's argument was ignored.

So this is a real behaviour change with a narrow surface:

- passes nothing, `"."`, or an absolute path **inside** the workspace → outcome unchanged;
- a relative path → now resolves where it says (this is the fix);
- an absolute path **outside** → refused, where today it would have silently run against the workspace and might have succeeded.

The conservative alternative is available if reviewers prefer it: fall back to the workspace for an outside path and honour everything else, keeping today's outcome in that one branch. This PR takes the stricter route because the refusal names the boundary, so an agent can correct it — whereas the silent fallback is exactly what left the failing run passing ten paths that were all discarded.

## Why confine rather than simply honour

```python
args = dict(tc.arguments)   # a blind "the model's run_dir wins"
```

That is the obvious fix, and it widens the trust boundary. `safe_run_dir` accepts anything under `swarm_runs_root()`, and `backtest` **writes** `artifacts/` and `run_card.json` **into** `run_dir`; `add_goal_evidence` derives `run_id = run_dir.name`, so it would attach evidence to another run. A blind honour lets one agent write into another agent's or another run's directory. Confining keeps the per-agent isolation that the injection exists to provide, while still letting an agent point a tool at a directory it created itself — which is the case that was broken.

Note also that "make the worker behave like `agent/loop.py`'s `_normalize_tool_run_dir`" is *not* equivalent: that helper resolves relative values against a single session `run_dir` and has no per-agent isolation to protect, so honouring is free there. The swarm added per-agent workspaces, and this keeps them.

## Verification

- `agent/tests/test_swarm_tool_run_dir.py` (new) — confinement cases: relative, absolute-inside, blank, four escape forms, and undeclared pinning; plus two worker-level tests: a relative path **reaches the tool**, and an escape is **refused without calling the tool** (surfaced as a normal `tool_result` error).
- **10 passed.** Both worker-level tests **fail on the pre-fix call site** (reverting only the call site, keeping the helper).
- **386 passed, 5 skipped** across every swarm test; **68 passed** across the goal/pattern/autopilot suites that also consume `run_dir`.
- `ruff check` clean. `black` is not applied to `worker.py`: it is already non-black on `origin/main`, `CONTRIBUTING.md:149` asks contributors not to mix in unrelated formatting, and Black leaves the added lines alone (the one hunk it wants is a pre-existing line).

**Not covered by the tests, checked by hand instead.** The cases run through a probe tool, not the real `BacktestTool` schema and `safe_run_dir` allowlist. Applied to the failing run's actual directory, the rule resolves `runs/crude_cycle_3m` to `<workspace>/runs/crude_cycle_3m`, where that run's `config.json` exists — so the path it actually wanted now reaches the tool. A test using the real tool would close this gap.

**Why the model's raw argument could not be re-checked.** `_persist_messages` writes `artifact_dir/messages.json`, which holds the assistant tool calls verbatim, but a worker that *completes* never called it: only the abnormal exits (timeout, cancelled, iteration limit) did, so the log is missing from essentially every successful run — 66 `summary.md` beside 1 `messages.json` over one run root, which is also why this description cannot pin down which path the successful run passed. That is a separate observability gap, not addressed here; #1439 routes both writers through one call.

## Relationship to #1429

Complementary, not overlapping. **#1429** makes the backtest tool's missing-file error name the directory it inspected; **this** stops the argument being discarded. Different files (`backtest_tool.py` vs `swarm/worker.py`), and either can land alone. Together they close the loop: an agent that still gets the layout wrong is told where the tool looked, and an agent that points at its own subdirectory is no longer ignored.


warren618:
Merged — thank you, and the stricter route is the right call. Checked the boundary mechanics on the merged tree: `_default_run_roots()` already includes `swarm_runs_root()`, so a relative `runs/<name>` resolved under the agent workspace passes `safe_run_dir` and the backtest tool can actually read it; remote MCP tools strip `run_dir` as a local-only argument, so they are untouched; an unknown tool name still gets the workspace injected and fails in `execute` as before. 10/10 new tests, the swarm worker suites and the file-tool sandbox suite all green.

PR #1431, opened 2026-09-13
fix(runner): route A-shares by market for sources without a Wave-1 branch

chiww:
## The bug

`_create_market_engine` already follows the instrument market rather than the loader name for sources that have no Wave-1 branch — that is what sends a local `AAPL.US` dataset to `GlobalEquityEngine` instead of the crypto default. **A-shares were left out of that set**, so every branchless source that serves them fell through to `CryptoEngine`:

| `source` | `600519.SH` routes to |
|---|---|
| `auto` / `tushare` / `akshare` | `ChinaAEngine` ✅ |
| `local` | `CryptoEngine` ❌ |
| `tencent` | `CryptoEngine` ❌ |
| `eastmoney` | `CryptoEngine` ❌ |
| `baostock` | `CryptoEngine` ❌ |
| `mootdx` | `CryptoEngine` ❌ |
| `sina` | `CryptoEngine` ❌ |

These are not obscure spellings. `skills/data-routing/SKILL.md` — "the single ROUTER for every data need … Load this skill **BEFORE any backtest**" — lists `baostock`, `tencent`, `mootdx` and `eastmoney` as A-share sources, and `skills/mootdx/SKILL.md:78` documents the explicit override:

```python
result = run(strategy=..., source="mootdx")  # explicit override
```

## Why it matters

`CryptoEngine` contains none of the A-share rules:

| | `engines/crypto.py` | `engines/china_a.py` |
|---|---|---|
| `stamp_tax` | 0 | 3 |
| `T+1` | 0 | 3 |
| `_price_limit` | 0 | 2 |
| `100-share` | 0 | 1 |

It is not only *missing* them — it applies another market's: maker/taker rates plus an **8-hourly perpetual funding-fee settlement** charged against the position.

The run still succeeds and its metrics are still internally consistent, which is what makes it dangerous — the same reasoning `engines/base.evaluation_start_index` already documents.

## The change

Adds the `a_share` branch next to the existing global-equity one in the same `else` block. Three lines plus a comment; no behaviour change for any source that already routed correctly.

`TestLocalSourceEngineRouting` is extended to pin the full matrix: every branchless source routes A-shares to `ChinaAEngine`, `tushare`/`akshare` keep working, and `local` crypto still reaches `CryptoEngine`.

## Verification

- `tests/test_local_source_routing.py` — **24 passed**; 10 of them were red before the source change
- full `agent/tests` — **12921 passed, 98 skipped**, 1 unrelated pre-existing failure (`test_mcp_server_http_transport`, red on a clean tree too)


he-yufeng (approved):
Reviewed end to end against current main, and ran the test file both ways.

Mechanism checks, each confirmed in the source:

- `markets` is single-element by the time the else branch runs (mixed baskets go to `CompositeEngine` at the top), so the new `a_share` check can only fire for a pure A-share basket.
- `_detect_primary_source` returns an explicitly named source unchanged, so `baostock`/`tencent`/`eastmoney`/`mootdx`/`sina`/`local` really do reach this fallthrough; `auto` is resolved before the call, and the parametrized `auto` case is fair as defence in depth.
- The failure mode is real: `CryptoEngine` applies `calc_crypto_funding_fee` with `funding_rate` defaulting to 0.0001 and none of the A-share rules, and the run still succeeds, so the misroute is silent. An unknown-format code defaults to `a_share` in `_detect_market`, which makes ChinaAEngine the safer fallthrough for junk symbols too.
- Placement after the global-equity check is right: a pure US/HK basket keeps its engine, and the branching `tushare`/`akshare` path is untouched (pinned by the keep-working tests).

Verification: the PR's test file gives 24/24 on the head; the same file against unpatched main fails exactly the 10 new A-share routing cases and passes the 14 pre-existing ones. CI is green across the board.

LGTM.

warren618:
Merged — thank you. Ran the file both ways on current `main` (the 10 new A-share cases red before the change, 24/24 after) and the runner / backtest / swarm slice on the merged tree (604 passed). The branch is placed exactly where it belongs: after the global-equity check, and only reachable by a single-market basket, so nothing that routed correctly moves.

PR #1432, opened 2026-09-13
fix(runner): check the declared interval against the served bars before annualising

chiww:
## The bug

A single-market run annualises from `calc_bars_per_year(interval, source)`. That reads the interval the caller **asked for**, which is not a fact about what arrived — a loader may legitimately serve coarser bars than requested. `local_loader._resample_to_interval` cannot upsample a daily file to `1H`, so it returns the source bars and says so, in a log line nothing acts on.

Same data, same ten trades, same total return; only the declaration differs:

| | `total_return` | `annual_return` | `sharpe` |
|---|---|---|---|
| `interval=1H` | −17.48% | **−27.83%** | **−0.9025** |
| `interval=1D` | −17.48% | **−7.83%** | **−0.4513** |

CAGR is out by 3.5×, Sharpe by 2×, on a run card that looks entirely ordinary.

## The change

Measuring the served index is the convention `effective_bars_per_year` already applies to cross-market runs (`bars_per_year=None`, #1237). This uses it as a **check** rather than a replacement: when the declared and observed counts agree, the per-source trading-day table is kept and **no existing run card moves**; only a genuine mismatch switches to the measured count and logs why.

The 1.5× threshold separates the two cases cleanly — neighbouring intervals differ by at least 2× (`30m` → `1H`), while the calendar-vs-trading-day spread inside one interval is a few percent (a business-daily series measures ~261 against a declared 252).

Only the instrument codes are measured, so an injected fundamental panel cannot decide the annualisation; unmeasurable data falls back to the declaration.

## After

| | `annual_return` | `sharpe` |
|---|---|---|
| `interval=1H` | **−7.56%** | **−0.4431** |
| `interval=1D` | −7.83% | −0.4513 |

plus a warning naming both counts:

```
interval=1H declares 1008 bars/year but the served data measures 243
(654 bars over 983 days); annualising on the served data. The loader
returned bars coarser or finer than requested -- check the earlier
loader warning, and set interval to the granularity the source
actually has.
```

## Verification

- `tests/test_cross_market_annualization.py` — **15 passed**; 7 of them were red before the source change
- full `agent/tests` — **12916 passed, 98 skipped**, 1 unrelated pre-existing failure (`test_mcp_server_http_transport`, red on a clean tree too)
- end-to-end: real `python -m backtest.runner` on a daily local dataset, the numbers above

## Possible follow-up

The mismatch is currently log-only. Surfacing it on the run card the way `_run_card_caliber_warning` does would make it visible to whoever reads the report rather than the console — happy to add that here or separately if you'd prefer it.


he-yufeng (approved):
Verified end to end, approving.

The bug is real and reachable on current main: `local_loader._resample_to_interval` returns source bars unchanged with only a log warning when it cannot upsample (local_loader.py:107-118), and the runner then annualises at the declared interval's `calc_bars_per_year` — so a 1H run over a daily file scales CAGR, Sharpe and annualised volatility by roughly an order of magnitude, silently.

The fix shape is the right one:

- It reuses the `effective_bars_per_year` convention the cross-market path already established instead of inventing a second measurement.
- The 1.5 ratio gate is the load-bearing choice: neighbouring intervals differ by at least 2x while the calendar-vs-trading-day spread inside one interval is a few percent, so correctly served runs keep their per-source trading-day table and no existing run card moves. I confirmed that on the 1H-over-trading-day case: measured ≈ declared, gate holds, declared kept.
- The warning names the actionable next step (check the loader warning, set interval to the source's granularity) instead of silently fixing the number.

What I ran on this head: the new test class plus its file 15/15, and the runner/metrics slice of the suite 443 passed / 5 skipped. CI shows 10/10 green.

One trade-off I weighed and accept: an irregular series landing between 1x and 1.5x keeps the declared count, so a mildly off declaration still annualises mildly wrong. That is the documented price of not moving every existing run card, and the warning only fires past the gate. Fine as designed.

warren618 (changes requested):
Thanks — the bug is real and I reproduced it: `local_loader._resample_to_interval` hands back the daily bars when asked for `1H`, and `calc_bars_per_year("1H", …)` then annualises a daily series at the hourly rate. Measuring the served index as a *check* rather than a replacement is also the right shape. But the check as written moves run cards it promises not to move.

`effective_bars_per_year` divides bars by **calendar** years, so on any window short enough for weekends and overnight gaps to dominate the span it measures far above the declared count. Same functions and threshold as the PR (`calc_bars_per_year(interval, "yahoo")`, `effective_bars_per_year`, 1.5), on synthetic session indexes:

| served data | declared | measured | ratio | outcome |
|---|---|---|---|---|
| `1m`, 1 trading day (390 bars) | 98,280 | 390 | 252.0 | switches to 390 |
| `1m`, 3 trading days | 98,280 | 213,671 | 2.17 | switches |
| `1m`, 5 trading days | 98,280 | 178,059 | 1.81 | switches |
| `1H`, 5 trading days (7 bars/day) | 1,764 | 3,195 | 1.81 | switches |
| `1D`, 5 bars | 252 | 456 | 1.81 | switches |
| `1D`, 10 bars | 252 | 332 | 1.32 | keeps |
| `1m` / `1H` over 21 days, `1D` from 20 bars | — | — | ≤ 1.16 | keeps |

So a one-week daily run is annualised at 456 bars a year and a one-week hourly run at 3,195, each with the new warning telling the user the loader served the wrong granularity — it did not. Windows that short are exactly where people run a quick check, and the full suite stays green because no unit test drives `runner.main` with one.

The quantity to compare is the bar **spacing**, not bars per calendar year. `median(diff(index))` is one day for a daily series whether the window is 5 bars or 5 years, and one hour for hourly bars regardless of weekends or session length. I ran that on the same rows plus an A-share 4-bar session, 24×7 crypto hours and a series with a six-month halt: every legitimate row comes out at ratio 1.00, and the daily-served-as-`1H` case comes out at 24. Compare the median spacing with the declared interval's timedelta (the same 1.5 either way), and when it mismatches annualise with `calc_bars_per_year(<the interval the spacing matches>, source)` — the per-source trading-day table still produces the number, every existing card keeps its count, and the result no longer depends on window length.

Two tests to add with that: a five-bar `1D` series keeps 252, and a one-week `1H` series keeps its declared count, next to the daily-served-as-`1H` mismatch you already have. Happy to re-run on the next head.

chiww:
@warren618 The counterexamples reproduce and the diagnosis is right — `effective_bars_per_year` divides by calendar years, so a short window measures far above the declared count and the gate fires on correctly served data, with a warning that blames the loader for something it did not do. Pushed a revision that compares median bar **spacing** instead.

One detail beyond what your table reports: the ratio depends on which weekday the window starts, so the same five daily bars behave differently from one run to the next.

| 5 daily bars | span | measured | ratio | old behaviour |
|---|---|---|---|---|
| Monday start | 4 days | 456 | 1.81 | switched |
| Tuesday start | 6 days | 304 | 1.21 | kept |

That made it worth pinning explicitly, so there is a Monday-start case in the tests next to the plain five-bar one.

Your table against the new head, plus the shapes I could think of to break a spacing measurement:

| served data | declared | resolved | |
|---|---|---|---|
| `1m`, 1 trading day (390 bars) | 98,280 | 98,280 | kept |
| `1m`, 3 trading days | 98,280 | 98,280 | kept |
| `1m`, 5 trading days | 98,280 | 98,280 | kept |
| `1H`, 5 trading days | 1,764 | 1,764 | kept |
| `1D`, 5 bars (Monday start) | 252 | 252 | kept |
| `1D`, 5 bars (Tuesday start) | 252 | 252 | kept |
| `1D`, 10 / 20 bars | 252 | 252 | kept |
| A-share 4-bar hourly session | 1,008 | 1,008 | kept |
| 24×7 crypto hours | — | declared | kept |
| daily series with a six-month halt | 252 | 252 | kept |
| **daily bars declared `1H`** | 1,764 | **252** | switched |

On a mismatch the count now comes from `calc_bars_per_year` looked up with the interval the spacing matches, as you suggested. That is also more accurate than the previous head — the end-to-end `1H` run now lands on 252, identical to its `1D` twin, where before it used the raw measured 243:

```
interval=1H   annual_return=-7.83%   sharpe=-0.4513
interval=1D   annual_return=-7.83%   sharpe=-0.4513
```

Two choices worth flagging since they are mine rather than yours:

- **Spacing that matches no supported interval** (a weekly file, say) keeps the declaration and logs that it matched nothing, rather than naming a bar count it cannot justify. Switching would need a number the per-source table does not have.
- **Four bars minimum** before the median is trusted: three differences survive one session gap (`[1, 1, 3]` → one day), two do not (`[1, 3]` → two days). Below that the declaration stands.

Spacings are held in seconds rather than `Timedelta` so the comparison is a ratio of two numbers; that also keeps the suite's warning count unchanged at 1576.

Tests: the three you asked for, plus the Monday-start case and the session shapes above — `tests/test_cross_market_annualization.py` 23 passed, no new warnings. Full `agent/tests` 12924 passed with one unrelated pre-existing failure (`test_mcp_server_http_transport`, red on a clean tree too). `tools/ci_grep_gates.sh` all gates passed.

@he-yufeng — worth re-checking: the trade-off you accepted (series between 1× and 1.5× keeping the declared count) no longer has the same shape, since legitimate series now measure 1.00 rather than drifting with window length.

he-yufeng:
Re-ran the new head (e85e7c9) against your table plus a few extra shapes, driving `_annualisation_bars` directly on synthetic indexes:

| served data | declared | outcome |
|---|---|---|
| 1D, 5 bars, Monday start | 252 | keeps 252 |
| 1H, 7 bars/day x 5 days | 1764 | keeps 1764 |
| 1m, one trading day (390 bars) | 98280 | keeps |
| 1H, A-share four-bar session x5 | 1008 (tushare) | keeps |
| 1H, 24x7 crypto x5 days | 8760 (okx) | keeps |
| daily with a six-month halt | 252 | keeps |
| weekly file declared as 1D | 252 | keeps, warning honestly says no interval matches |
| 1D series declared as 1H | 1764 | switches to 252 |

The Monday/Tuesday nondeterminism is gone by construction since nothing divides by calendar span anymore, the sub-four-bar case keeps the declaration as documented, and the run-card count still comes from the per-source table on a real mismatch. The full new test file passes locally (23/23) and the suite stays green around it.

This addresses everything from my side. Formal re-review is yours since you hold the changes-requested.

warren618:
Thanks for the rework. I re-ran it on e85e7c9, and the redesign does what I asked for. Driving `_annualisation_bars` directly, every row of my earlier table now keeps its declared count, and so do the harder legitimate shapes I tried:
- US `1H` across the DST change (tz-aware and UTC-naive)
- `1D` with US holidays plus a six-month halt
- A-share `1H` with the real 10:30/11:30/14:00/15:00 labels, and A-share `30m` / `1m`
- HK `1H` with the lunch break
- 24×7 crypto `4H` / `1H`
- duplicate timestamps, and a shuffled or descending index

Daily bars declared `1H` now resolve to 252.

Two things before I approve.

**1. Several properties the code relies on aren't pinned by any test.** Each of these mutations leaves `test_cross_market_annualization.py` at 23/23:
- `_SPACING_MISMATCH_RATIO` 1.5 → 4.5, or 1.5 → 1.05. Nothing switches between neighbouring intervals, such as hourly bars declared `30m` or `4H` bars declared `1H`.
- A one-sided gate that only catches spacing coarser than declared. Nothing serves bars finer than declared.
- `_MIN_BARS_FOR_SPACING` 4 → 2. Three daily bars declared `1H` would pin it.
- Measuring every `data_map` frame instead of only the price `codes`.
- Measuring the shortest index instead of the longest. No test has frames of different lengths.

**2. Spacing coarser than every supported interval keeps whatever was declared.** A weekly file declared `1H` still annualises at 1,764 bars a year (or 252 if declared `1D`), with only a log line. That is the same bug one step coarser, and the local loader reaches it the same way when it can't upsample. Would you resolve anything coarser than `1D` to the `1D` count? Or at least put that warning on the run card next to the caliber warning, rather than only in the console.

Non-blocking: the no-match warning also fires on correctly served data. One missing bar in a ten-day A-share `1H` series moves the median to 1.75h, and five daily bars over Christmas measure 2 days, so the warning's advice to re-run at another granularity is misleading there. Softer wording would help.

warren618:
Merged as `85ba6d89`. `9d72ce97` on top finishes the two points from my last comment: every property the check relies on now has a test its mutation fails (the 1.5 gate from both sides, the two-sided direction, `_MIN_BARS_FOR_SPACING`, price codes only, longest frame), and a spacing wider than any supported interval — at least four days, so a weekly or monthly file — is annualised from the calendar (52 or 12) instead of keeping a declared 252 or 1,764, while a daily series over a holiday week keeps its count. The report is also written to the run card next to the caliber warning, and it states the spacings without telling the user to re-run. Thanks for the rework — the spacing redesign held up against everything I threw at it.

PR #1436, opened 2026-09-14
fix(i18n): use broker-neutral diagnostics across all locales

tingkk:
## Summary

Fix shared broker diagnostics in all eight supported languages that incorrectly identify Longbridge when another SDK broker, such as Futu, has a network or missing-dependency error. The messages now refer to the affected broker's SDK and to its service or local gateway.

## Why

`Runtime.tsx::connectorDiagnostic()` and `RunnerStatus.tsx::sdkDiagnosticMessage()` select shared translation keys using the connector's `error_code`. Neither path restricts those keys to Longbridge, but both `runtime.diagnosticSdkMissing` and `runtime.diagnosticNetworkUnreachable` explicitly named Longbridge.

This was reproduced with Futu in Docker: its broker connector was using an unreachable OpenD address and correctly reported `network_unreachable`. The Futu card nevertheless displayed:

- English: `The Longbridge service is unreachable. Check network access and retry.`
- Simplified Chinese: `无法连接 Longbridge 服务。请检查网络后重试。`

That sends users toward the wrong broker when they need to check their OpenD gateway address. The same shared mapping tells users that the Longbridge SDK is missing when a different broker reports `sdk_missing`.

### Reproduction

1. Select any supported language in the Web UI; the English and Simplified Chinese examples are shown above.
2. Use the Futu SDK connector with `futu-api` installed and an unreachable OpenD gateway address. For example, a Docker container's `127.0.0.1` does not reach OpenD listening on the Mac host.
3. Open `/runtime` and verify or retry the Futu connection.
4. Observe the Futu card's `network_unreachable` diagnostic incorrectly naming Longbridge. The chat connector-status component uses the same translation key.

For the analogous `sdk_missing` case, OpenD must be reachable while the backend lacks `futu-api`: Futu's status check tests gateway reachability before SDK availability.

Expected behavior: shared diagnostics identify the relevant dependency or connection problem without attributing every broker's error to Longbridge.

## Changes

- Update `diagnosticSdkMissing` to refer to the SDK required by the affected broker.
- Update `diagnosticNetworkUnreachable` to cover a broker service or local gateway and suggest checking the connection address as well as network access.
- Preserve translation keys and component behavior; change only 16 string values across the eight locale files: English (`en`), Simplified Chinese (`zh-CN`), Arabic (`ar`), German (`de`), Spanish (`es`), Japanese (`ja`), Korean (`ko`), and Brazilian Portuguese (`pt-BR`).

Futu endpoint configuration, SDK installation, and local Docker settings are separate setup concerns and are not included in this PR.

## Test Plan

- [x] Install the exact frontend lockfile dependencies with `npm ci --ignore-scripts --no-audit --no-fund`.
- [x] `npm run build` passes, including TypeScript checking and production asset generation.
- [x] Focused frontend tests pass: **3 files, 68 tests**.

  ```sh
  npm run test:run -- src/pages/__tests__/Runtime.test.tsx src/components/chat/__tests__/RunnerStatus.test.tsx src/i18n/__tests__/i18n.test.ts
  ```

- [x] Manually reproduced the misleading Futu network diagnostic in Chrome and verified the corrected Chinese string is present in the rebuilt Docker frontend bundle.
- [x] Parse all eight supported locale JSON files and verify both diagnostics are nonempty and no longer name Longbridge.
- [x] `git diff --check` passes.

No new tests were added for this wording-only change. The existing focused suites cover runtime rendering, connector status, and locale consistency; they do not directly assert this Futu/error-code combination. The full frontend and Python suites were not run locally.

## Risk and rollback

Presentation-only change: connector logic, broker requests, credentials, trading permissions, and order execution are unchanged. No live trading or order mutation was performed for validation. Revert this change to restore the previous wording.

## Checklist

- [x] No changes to protected agent, session, or provider code.
- [x] No credentials, local configuration, or private account data included.
- [x] Every commit includes the required DCO `Signed-off-by` trailer.
- [x] No README update is needed for this localized diagnostic wording correction; reproduction and scope are documented above.


warren618:
Merged — thank you. Confirmed both keys are selected by `error_code` alone in `Runtime.tsx` and `RunnerStatus.tsx`, so every SDK broker was being told to check Longbridge; read the six other translations and they say the same thing the English does. The three focused suites (68 tests) pass on the merged tree.

Issue #1433, opened 2026-09-13
prices conflicted with tool evidence.


5gaLbt:
I really like Vibe-Trading, I am using a lot.

I got "the prices conflicted with tool evidence" error quite often, it will be fine if I ran it again.

Could you explain why? or how can I prevent this from happening?

I rejected the previous draft because its prices conflicted with tool evidence. The verified observed OHLC range is: GS.US: 1011.4–1043.84 USD (source: yahoo; currency conversion: none); LLY.US: 1113.29–1138.79 USD (source: yahoo; currency conversion: none); MU.US: 967.38–1042.4 USD (source: yahoo; currency conversion: none); NVDA.US: 217.2–233.449 USD (source: yahoo; currency conversion: none); OXY.US: 60.0743–62.24 USD (source: yahoo; currency conversion: none). I will not invent an entry price without a visible derivation or refreshed evidence.Status: SUCCESS
Elapsed: 16m 40s
Run ID: 20260913_125930_42_13f137
Run dir: C:\Users\xxxxx\.vibe-trading\runs\20260913_125930_42_13f137
Reason: final answer degraded to the deterministic fallback after 4 rejected drafts could not be corrected within the iteration budget

he-yufeng:
Thanks for the kind words and for the report, the run log you pasted made the mechanism easy to pin down.

**Why it happens.** Every price figure in a final answer must match a value the run actually observed through a tool (within a small tolerance), otherwise the draft is rejected. Your drafts quoted entry prices that matched nothing observed, so the gate rejected them and asked the model to correct. The problem is what the correction said: roughly "these values are banned, the verified range is min-max". For a five-symbol portfolio over a wide OHLC range, the model cannot guess a citable number from that, each retry fails the same way, and after 4 rounds the run gives up and emits the deterministic fallback you saw. A re-run rolls the dice again, which is why it sometimes passes.

**What we changed.** #1438 makes the rejection name the exact values the next draft may cite (the latest observed closes per symbol, with currency and date), instead of only naming the banned ones. The gate itself is exactly as strict; only the feedback improves, so recovery stops depending on draft luck.

**Until that lands:** the failure is most likely on multi-symbol reports where the draft invents entry/exit levels. Asking for the answer to show the derivation from observed prices (or to skip explicit entry prices) avoids the rejection path entirely today.

warren618:
[...]

PR #1437, opened 2026-09-14
feat(grounding): let tools declare their authoritative metrics (#1426)

### LINKED ISSUE #1426: [Bug] Generic numeric tool fields can be misclassified as analysis evidence by metric-name tokens

he-yufeng:
## Summary

- Adds the explicit half of the evidence contract #1426 asks for: a tool result may declare its authoritative metrics (`"metrics": [{"field": "volatility.annualized_vol", "metric": "annualized_volatility"}]`), and only those fields then ground analysis claims.
- Opt-in and per-payload. A result without a declaration is ingested exactly as before, so no existing tool changes behaviour.

## Why

Closes #1426. Measured on current main (table in the issue): `return_obs: 81`, `n_returns: 81`, `return_window: 20`, `vol_lookback: 60` and `max_drawdown_duration: 45` each ground a fabricated claim equal to the metadata value, because `_metric_kind_for_path` infers the metric kind from the field name alone. #1422 covers the one spelling from #1420; the denylist keeps meeting the next spelling. The discussion in #1426 converged on letting tools declare their metric evidence explicitly, which this PR implements.

What this does not do: undeclared payloads still get legacy name inference, so the false-acceptance rows above remain open for tools that have not opted in. Tightening that side (e.g. dropping the fuzzy token scan) is the follow-up the issue leaves to a maintainer call, since it trades recall for real compound names like `strategy_max_drawdown`.

## Changes

- `_parse_metric_declarations` / `_resolve_declared_field`: parse the declaration and resolve each `field` path (dots and `[i]` indices) against the emitted payload. Declared kinds resolve through the exact alias table only, no fuzzy token scan.
- Declared metrics are recorded into the analysis evidence verbatim with a `declared: true` marker; the persisted audit artifact gains `declared_metric_calls`.
- While a declaration is present, the payload's other numeric leaves are never re-classified into analysis evidence at check time (`_analysis_value_observed` skips the call's generic records). Price leaves in the same payload still ground price claims.
- An entry naming an unrecognised metric is not authoritative but still engages strict mode, so a broken declaration fails closed rather than silently inheriting inference. A `metrics` key that is not a list of `{"field", "metric"}` entries is payload data, not a declaration.
- 7 new tests: declared-only grounding, legacy behaviour without a declaration, unknown-metric strict mode, non-declaration `metrics` shapes, the quantlib `vol_lookback` row from the issue, list-index path resolution, and price evidence surviving strict mode (plus the persisted audit fields).

## Test Plan

- [x] Existing tests pass (full battery: 12917 passed, 98 skipped; grounding file 244 passed)

Manual: the new tests fail on unpatched main (4 of 7, the ones asserting declared-mode behaviour; the two legacy-pinning tests pass before and after), and the issue's `vol_lookback` / `return_observations` rows reproduce as rejected once the tool declares only its real metrics.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion: this touches the analysis grounding gate; the contract shape was discussed in #1426 with the reporter's payload as the design input before implementation.
- [x] Documentation updated (the contract is documented in the docstrings of `_parse_metric_declarations` and `_record_declared_metrics`; happy to add a wiki section if you want it user-facing)
- [x] Community commits include the DCO `Signed-off-by:` trailer

## Risk / Rollback

Low and opt-in: payloads without a `metrics` declaration are byte-identical in behaviour (pinned by the legacy tests). Payloads with one get strictly less evidence authority, never more, and failures close toward rejection. Rollback is a single-commit revert.

Prepared with AI assistance (Kimi); the repro rows, tests, and diff above were verified locally by the committer.


warren618:
Reproduced the #1426 table on `68787cdf`: with a declaration naming only `annualized_vol`, the five metadata rows (`return_obs`, `n_returns`, `return_window`, `vol_lookback`, `max_drawdown_duration`) reject the fabricated claim, the control still rejects, and the declared true value validates; without a declaration they still validate, as the description says.

Two things on the alias side before it lands. Declared names resolve only through `_ANALYSIS_KIND_ALIASES`, and an unrecognised name keeps strict mode with nothing authoritative — so on this head `metric: "var_95"` or `"cvar_95"` is unrecognised (#1427 adds the tail-risk aliases), which means a tool declaring only its VaR would lose all of its grounding the moment #1427 lands unless the two go in together. That needs an integration test across the pair. Smaller: `maxdrawdown` is unrecognised while `max_drawdown` and `maxdd` are; a test that pins the accepted declaration names would give a tool author a clear failure instead of silent strict mode.

Sequencing with #1419 / #1422 / #1423 / #1427 is being settled on our side; I will follow up here.

he-yufeng:
@warren618 Both alias points addressed.

`1580e28f` pins the accepted declaration names: `annualized_volatility`/`max_drawdown`/`maxdd` resolve, `maxdrawdown` and `var_95` stay unrecognised on this head, and an unrecognised name keeps strict mode with nothing authoritative, exactly the clear failure a tool author should see. Your `maxdrawdown` observation is one of the pinned cases.

On the pair integration: the declared-names test asserts `var_95` is unrecognised *here* because the tail-risk aliases land with #1427, so the two PRs are safe in either order — if this lands first, declaring VaR fails loudly (strict mode, no grounding); once #1427 is in, the same declaration resolves. The true cross-pair test (a declared `var_95` grounding a VaR claim end to end) can only run once both are in one tree; whichever of the two lands second should carry it, and I'll add it on top of whichever order you settle. Note #1427 just merged #1423 in at `e13ab882` per your sequencing note there.

warren618:

PR #1438, opened 2026-09-14
fix(grounding): name the citable observations in a price rejection (#1433)

### LINKED ISSUE #1433: prices conflicted with tool evidence.

he-yufeng:
## Summary

- When the grounding gate rejects a draft over price figures, the correction prompt now names the verified observations the next draft may cite exactly (newest two closes per involved symbol, with currency and date).
- Bounded: at most five symbols, and only when the rejection actually involves price claims.

## Why

Closes #1433. The reporter hits "prices conflicted with tool evidence" quite often, and a re-run usually passes. That flakiness has a structural cause: a conflict rejection told the model which values to remove and what range was observed, but never which exact values it may cite. For a portfolio run over five symbols with a wide OHLC range (the issue's GS/LLY/MU/NVDA/OXY case), guessing a number within 0.5% of a real observation is a coin flip, so all four correction rounds fail identically and the run degrades to the deterministic fallback after 16 minutes. Whether a run survives depended on draft luck, not on evidence.

## Changes

- `correction_prompt` gains a "Verified observations the next draft may cite exactly" section, fed by the same `_comparable_price_records()` set the validator checks against, so every hinted value is one the gate will accept.
- Hints prefer `close`, fall back through `price`/`adj_close`/`open`/`high`/`low`, carry currency and the observation date, and scope to the symbols the rejected issues name (all observed symbols when the claim was unattributed).
- Three new tests: the hint names the latest observed close with its date; analysis-metric rejections get no price section; the section stays bounded and newest-first over six symbols with three bars each.

## Test Plan

- [x] Existing tests pass (full battery: 12912 passed, 99 skipped; grounding file 240 passed)

Manual: reproduced the reporter's shape locally — a conflicting buy-price claim against observed 562500 bars rejects as before, and the correction prompt now lists `562500.SH close 1.137 CNY (2026-06-23); 562500.SH close 1.171 CNY (2026-06-24)` as citable instead of only naming the banned value. The new tests fail on unpatched main for the missing section.

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion: grounding gate again; the change is the direct follow-up to the user-facing failure reported in #1433 and only widens what a rejection *tells the model*, never what it accepts.
- [x] Documentation updated (behaviour is self-documenting in the prompt text; no user-facing surface changed)
- [x] Community commits include the DCO `Signed-off-by:` trailer

## Risk / Rollback

Low: the gate's accept/reject logic is untouched; only the correction feedback text changes, and only for rejections that already involved price claims. Worst case is a slightly longer correction prompt. Rollback is a single-commit revert.

Prepared with AI assistance (Kimi); the repro, tests, and diff above were verified locally by the committer.


warren618:
Verified on `a19dd8f1` with a two-bar ledger for five symbols: a conflicting entry price rejects exactly as on `main`, the correction prompt gains the two newest closes per involved symbol with currency and date, and a draft that cites one of them validates. On `main` the same rejection carries no citable value. Correct as described; it will land with the rest of the grounding set — the order is being settled on our side, I will follow up here.

warren618:

PR #1439, opened 2026-09-14
fix(swarm): persist a worker's message log on every terminal path that writes a summary

cgycorey:
## Summary

`_write_summary` and `_persist_messages` were called independently at each terminal point of the worker, and the two had drifted apart:

| terminal path | `summary.md` | `messages.json` | before this PR |
|---|---|---|---|
| timeout | yes | yes | ok |
| cancelled (between iterations) | yes | yes | ok |
| cancelled (mid-stream) | yes | yes | ok |
| iteration limit | yes | yes | ok |
| **completed** | yes | **no** | broken |
| **token_limit** | yes | **no** | broken |
| **content_filter_circuit_breaker** | yes | **no** | broken |
| **incomplete** (output contract not met) | yes | **no** | broken |

A worker that *completed* — the common outcome — wrote `artifacts/<agent>/summary.md` and no `messages.json`. The `incomplete` row was missing from the first version of this table; [@warren618](https://github.com/warren618) found it in review, and it is now covered by a test like the rest.

## Why it matters

`messages.json` is the only record of **what the model asked a tool for**. The assistant message is appended from `response.tool_calls` (`worker.py:964`) *before* the execution loop rewrites the arguments, so the log holds the model's own request — the path it wanted, not the path the tool was actually given. That is precisely what a post-mortem of a silently discarded argument needs.

The event stream cannot substitute for it: `_preview_tool_arguments` **skips `run_dir` outright** (`if key == "run_dir": continue`), and the summary is prose. Nothing else in a run directory holds it.

This is what made a real failing run undiagnosable: a strategist passed ten `run_dir` values that were all discarded, every backtest failed with `config.json not found`, and because the run *completed* there was no log to show which path it had asked for.

The helper also swallows its exceptions (`logger.warning(..., exc_info=True)`), but that is **not** why the file was absent — these paths simply never called it. The fail-first runs below separate the two causes.

## The evidence, stated precisely

Over one run root (`swarm_runs_root()` = `~/.vibe-trading/swarm/runs`, 33 runs), stratified by when each run was written:

| run date | `summary.md` | `messages.json` |
|---|---|---|
| 2026-08-06 | 4 | 1 |
| 2026-08-08 | 12 | 0 |
| 2026-08-09 | 44 | 0 |
| 2026-09-13 | 6 | 0 |

`_persist_messages` and its first two call sites landed in `bfa0133c` (2026-08-02), two more with `cb6ec6a5` (2026-08-21). The 56 summaries from 08-08/08-09 **postdate the helper and carry no log at all**, so no agent in them reached a log-writing path. The cleanest single data point is the newest run: **6 agents, 6 summaries, 0 message logs**.

One caveat that limits this rather than supports it: the LLM-failure path also wrote nothing, so from file counts alone "these agents completed" and "these agents died in the LLM call" are not fully separable. The defect is established by the path table and the fail-first tests, not by the ratio.

## Cost, stated because the benefit is

- A log is not small: the one on disk is **244,820 bytes**. This writes one per agent per successful run, so a 10-agent run adds roughly 2.5 MB of transcripts. Tool results inside are already truncated (`truncate_tool_result`), so it is bounded — but the bound is large and nothing prunes old runs.
- `_collect_artifacts` globs the artifact directory and that list reaches `SwarmTask.artifacts` and the run's events (`runtime.py:538`, `runtime.py:570`). Every path here therefore gains one `messages.json` entry in the run's artifact list, visible wherever artifacts are displayed or downloaded.

**There is a smaller fix for the case that motivated this, and it is worth taking instead if you prefer it.** `_preview_tool_arguments` drops `run_dir` from the `tool_call` event, which is the specific reason the model's path was unrecoverable; recording it there (or simply no longer skipping it, now that #1430 makes a declared `run_dir` meaningful) is a couple of lines and lands where debugging already happens. The event answers "what path did it ask for?"; this PR's transcript also answers "what had it been doing for the previous six iterations?" — only the second needs the full log. If the smaller change is enough, close this and take that.

## What is deliberately not changed

The `except Exception` handler for a failed LLM call persists **neither** file, and still does not. It is the only terminal path left out, and it is left out on purpose: a provider failure is usually about the provider, not about the model's arguments, and changing it would put artifacts on a path that currently reports none. Say so if you want it in and I will add it — it is a two-line hoist plus a test.

## How it was built — test first, one path at a time

Each uncovered path was written as a failing test before any production change, and the failure was read before fixing:

**RED 1 — completion path.** `assert (artifact_dir / "messages.json").is_file()` failed while `summary.md` was already `is_file()`, so the run had reached its terminal state and written its summary, and only the log was missing. Feature missing, not a setup error. **GREEN 1.** One line — the missing `_persist_messages(artifact_dir, messages)` at that call site. `1 passed`.

**RED 2 — token_limit path.** Failed on the same assertion, after confirming `status == "token_limit"` and `summary.md` present. **GREEN 2.** One line. `2 passed`.

**RED 3 — content_filter_circuit_breaker path.** Failed on the same assertion, after confirming `status == "failed"` and `"circuit_breaker" in result.error`, so the breaker really had tripped. **GREEN 3.** One line. `3 passed`.

**RED 4 — incomplete path.** A data agent that answers without calling a tool returns `incomplete` while already writing `summary.md`; the same assertion failed on `messages.json`. **GREEN 4.** The refactor below had already routed this path through the shared writer, so no further production change was needed — but the test was checked against `main` (`a5b79422`) and fails there with `FileNotFoundError: .../messages.json`, so it pins the fix rather than the implementation.

**REFACTOR.** With three paths fixed and four that already worked, the same two-line pair appeared seven times. They now go through one writer, so recurrence is structurally impossible rather than merely fixed:

```python
def _finalize_run(artifact_dir: Path, summary: str, messages: list[dict]) -> None:
    _write_summary(artifact_dir, summary)
    _persist_messages(artifact_dir, messages)
```

## Verification

- `agent/tests/test_swarm_worker_message_log.py` (new, 5 tests) — the four paths above must each persist the log, plus one test that the log **carries the arguments the model passed** (`'"symbol": "BTC-USDT"'` inside `function.arguments`).
- **The content test was checked against a mutant**, because it passed on first run and a test that passes immediately proves nothing: making `_persist_messages` dump `[{"role": m.get("role")} for m in messages]` makes it fail (`1 failed, 3 passed`) while the three path tests stay green, then reverts. So it catches a log that exists but is useless, which the path tests cannot.
- **376 passed** across all `test_swarm*.py` suites plus the two other `run_worker` consumers (`test_error_path_redaction.py`, `test_load_skill_allowlist.py`), rebased onto current `main` (which now carries #1430).
- No test asserted the old behaviour — `messages.json` had **no coverage at all** before this, which is why the drift went unnoticed.
- `ruff check` clean. `black` leaves every added and changed line untouched; the file is already non-black on `origin/main`, per `CONTRIBUTING.md:149`.


warren618:
Thanks. I re-derived the terminal paths from `_run_worker_impl` on main and drove each one with a scripted model. Your table holds, with one row missing: a final response that fails the output contract returns `incomplete`, and on main that path also wrote only `summary.md`, as do `completed`, `token_limit` and the content-filter breaker. Your single finalize step fixes it too, but no test pins it.

On this branch every path that writes a summary also writes `messages.json`, summary first and once each. A failing `messages.json` write logs a warning and leaves the status `completed`, and a failed LLM call or template render still writes neither file.

Could you add a test for the `incomplete` path? A data agent that answers without calling a tool is enough. With that, this is good to merge.

cgycorey:
Added — `c57dcfd1`, and the missing row is now in the table above.

`test_incomplete_run_persists_the_message_log` drives the real worker with a data agent that answers in prose without calling a tool: the run returns `incomplete` (`output contract not met: data agent produced no tool calls and no report.md`), and both `summary.md` and `messages.json` must be on disk.

Checked against the pre-fix worker rather than against this branch: on `main` (`a5b79422`) the new test fails with `FileNotFoundError: .../artifacts/analyst/messages.json` while `summary.md` is already there, so it pins the fix and not the implementation. On the branch: 5 passed in the file, 376 passed across every `test_swarm*.py` suite plus the two other `run_worker` consumers, `ruff` clean. CI on the new head is running.

warren618:
Verified the new test against `main`: it fails there on the `incomplete` path (`summary.md` written, `messages.json` missing), and so do the completed, token-limit and content-filter cases. All five pass on this branch. `_write_summary` is now only called from `_finalize_run`, so no terminal path can write one file without the other. Merged — thanks.

PR #1440, opened 2026-09-14
fix(grounding): a count is not a magnitude

cgycorey:
## What this fixes

`_metric_kind_for_path` classified an evidence leaf by scanning its tokens for a
metric alias. Two shapes got through, and each let the ledger certify a number
that no tool ever measured as that metric.

**1. A metadata head noun took its identity from the metric modifier beside it.**

`return_observations` is a count of observations, `max_drawdown_duration` a span
in periods, `episode_count` a tally. All three were recorded as metric
magnitudes, so a `factor_analysis` result carrying `return_observations = 81`
grounded `Annual return: 81%` while `80%` was correctly rejected (#1420).

**2. The trailing fallback received the whole dotted path, so a parent key donated its kind.**

`_metric_kind_for_text(path)` matched the container's own name, which made every
numeric leaf beneath it that metric:

| recorded as | what it actually is |
|---|---|
| `drawdown_distribution_analysis.episode_count` = 8 → `drawdown` | an episode tally |
| `…ulcer_index` = 0.22 → `drawdown` | a severity index |
| `…episodes[7].duration` = 670 → `drawdown` | a span in periods |

Neither shape needs a synthetic tool to reach. `quantlib_call` is one of the
four tool names whose results enter the ledger, and its real
`drawdown_distribution_analysis` returns those keys beside the actual
`max_drawdown`, so a claim built from the duration is accepted:

```
"Max drawdown: 670%"   ->  VALID = True      (from max_drawdown_duration: 670)
"Max drawdown: 39.94%" ->  VALID = True      (from max_drawdown: 0.3994)
```

## The change

English compounds put the head noun last — the same rule the #1338 review relied
on when it resolved `return_vol` to `vol` and never to `return`. The head noun
now decides whether a leaf names a magnitude or an extent, and a leading count
quantifier makes it a count whatever the head says. The fallback is handed the
leaf instead of the path.

The scan itself is unchanged. The rule is about the **head noun, not suffix
position**, and that choice is load-bearing for recall: a suffix-only rule would
have silently dropped `annualized_return_pct`, `total_return_pct` and
`annualized_vol_pct`, all of which classify today and which the text fallback
cannot rescue (`\b` does not fire across `_`).

A count is now ungroundable as a magnitude. A magnitude is untouched.

## Verification

- **RED first.** `agent/tests/test_grounding_metric_identity.py` failed 19 / passed 21 before the fix — and every one of the 21 passes was a *recall* case, so the vocabulary that must survive was pinned before anything changed. GREEN after: **44 passed**.
- **The issue's own repro script** now matches its "What I expected" table exactly (`return_observations → None`, `annualized_return → return`, `strategy_max_drawdown → drawdown`, `trade_count → None`, `aligned_days → None`), and `ANALYSIS_METRICS` is empty for the count-only payload.
- **The patch is monotone-restrictive.** Over 21,698 names and paths taken from the whole `agent/` tree, the number of inputs for which it *adds* a metric kind is **0**. It can only withdraw a false identity; it cannot invent one.
- **Blast radius, measured not asserted.** Over 1,807 real leaf names harvested from the shipping tool modules, classification changed for exactly 5: `return_observations`, `max_drawdown_duration`, `avg_drawdown_duration`, `money_weighted_return_period` (all extents) and `sharpe_samples`. The other 1,802 are unchanged.
- **One of the five names is a second, larger version of the same defect.** `sharpe_samples` is neither a count nor an extent: `agent/backtest/validation.py:96,197` fills it with the per-path Sharpes of a Monte-Carlo simulation and of a bootstrap resample, and the leaf walk reaches the elements. Measured on a shadow-backtest payload carrying `metrics.sharpe = 0.55` and `sharpe_samples = [0.83, 0.91]`:

  | recorded `sharpe` evidence | `Sharpe: 0.83` | `Sharpe: 0.91` | `Sharpe: 0.55` |
  |---|---|---|---|
  | before | `0.55, 0.55, 0.83, 0.91` | certified | certified | certified |
  | after | `0.55, 0.55` | rejected | rejected | certified |

  A simulated path's Sharpe was standing in as the strategy's Sharpe. The observed value still grounds the claim, and nothing is lost in the UI, which reads the tool payload rather than the ledger.

  Deliberately not extended: simulated *aggregates* such as `simulated_sharpe_p95` still classify as `sharpe`, because `p95` is a statistic rather than a metadata noun. Whether a simulated quantile may certify a Sharpe claim is a provenance question, not a naming one, and it is out of scope here.
- **No false rejection introduced.** Prose that states a span in its natural unit is untouched: `Max drawdown duration: 670 periods`, `…670 days` and a bare `…670` all behave exactly as before. The claim side gates on ratio-shaped values — a percentage or a 0–1 fraction — and a span is stated in periods or days.
- **The one shape that flips is a unit error, and flipping it is the point.** A span or count *restated as a percentage* — `Max drawdown duration: 670%`, `Average drawdown duration: 17.43%`, `Drawdown episodes: 8%` — was accepted before and is now rejected. A 670-period span is not 670% of anything, so these are claims the ledger should never have certified; they were certified only because the mislabelled evidence matched the mislabelled claim. Claim-side vocabulary is deliberately unchanged: `_metric_kind_for_text` still reads `"Max drawdown duration"` as a drawdown claim, which is now simply ungrounded.
- **Alias guard pinned mechanically.** `test_every_alias_still_resolves_to_its_own_kind` asserts the whole alias table, so the guard cannot silently suppress a metric name that happens to look like metadata. It bites: injecting `sharpe` into the metadata list fails it.
- Grounding suites: **384 passed, 0 failed**.
- Full `agent/tests`: **13,129 passed, 33 skipped, 7 failed** — and all 7 reproduce identically on `origin/main` in this environment (the six `test_readme_counts` tool-count cases, which depend on locally installed optional tools, and the 35 ms `test_signal_alignment_perf` budget, which measures 57 ms on this machine). None is related to this change.

## Limitations, stated rather than hidden

- The guard is a **closed list of extent nouns**. Keys whose head falls outside it, and camelCase keys (the tokenizer does not split them), still resolve as magnitudes. No key in the four gated tools' own modules does today — I checked — so this is latent, not reachable.
- The general problem is that a leaf's *name* is inferred to be a metric at all. That is what a declaration by the tool author would settle; this patch deliberately does not try to close that class, because a longer word list is not a mechanism.

## Scope, deliberately

This narrows the name-inference path only. It does **not** propose that tools declare their authoritative metrics, and it is not in tension with that approach: a declaration would answer the question earlier and bypass this path entirely, so this patch stays correct either way and can be a stepping stone to it.

Refs #1420.


cgycorey:
Closing as superseded — this is now fixed on `main`, by the same design I proposed here.

`b45e6bb8` and `0e820f43` landed the head-of-field-name rule plus an explicit
metadata-count rule in `agent/src/agent/grounding/evidence.py`:

```python
# Sample-size, window and duration leaves share a metric's token but hold a
# count: ``return_observations = 81`` is 81 observations, never an 81% return
# (#1420/#1426).
_METADATA_COUNT_HEADS = frozenset({"n"})
_METADATA_COUNT_TAILS = frozenset(
    {"obs", "observations", "window", "lookback", "count", "days", "duration"}
)
```

I re-checked the behaviour on current `main` rather than assuming. Every leaf this
PR covered now yields no metric kind, and so do the variants it did *not* cover —
including the camelCase forms (`returnObservations`, `nReturns`, `sharpeSamples`,
`maxDrawdownDuration`) that were a residual hole in my implementation:

```
return_observations -> None      annualized_return     -> return
n_returns           -> None      returns[3]            -> return
max_drawdown_duration-> None     strategy_max_drawdown -> drawdown
sharpe_samples      -> None      var_95                -> tail_risk
returnObservations  -> None      cvar                  -> tail_risk
```

`agent/tests/test_grounding_role_hardening.py` pins this class, and #1420 is closed.
Since the gate was also refactored into a package (`eb3b927d`), this diff no longer
applies to anything — there is nothing left here to merge. Thanks for the fix.

PR #1441, opened 2026-09-14
fix(runner): honor _NO_NETWORK_FALLBACK_SOURCES for per-symbol gaps

chiww:
## The bug

`_NO_NETWORK_FALLBACK_SOURCES` documents its own intent plainly:

> An explicit `local` request that is unavailable is a config problem the user must see, not something to paper over with a Yahoo/Tencent fetch.

But the set is only consulted in `get_loader_cls_with_fallback` — when the loader is unavailable **as a whole**. The *per-symbol* fallback loop in `fetch_data_map` treats every source alike, so one request can be served half from the named source and half from the network, with two log lines as the only trace:

```
[WARNING] local loader: no config entry for symbol 600000.SH
[WARNING] source=local returned data for 1/2 symbols; missing: ['600000.SH']
   fetched: ['000001.SZ', '600000.SH']   effective_sources: ['local', 'tencent']
```

The caller asked for that source's provenance, not for the row count to be filled in. For `nobitex`/`wallex` the stakes are the ones the constant's own comment spells out — a Toman-quoted request quietly served in USDT.

## The change

One guard, using a new public predicate so the set stays module-private:

```python
if missing and not is_no_network_fallback_source(primary_source):
```

The `raise NoAvailableSourceError(...)` at the end of that block already existed; skipping the loop simply lets it fire. Sources outside the set are unaffected.

`is_no_network_fallback_source()` is added next to the set in `registry.py` — `runner.py`'s other registry imports are all public names, and importing the underscore one would be the first crossing of that boundary.

## What this costs

It is a behaviour change, and two existing paths feel it:

- **`ui_services.reconstruct_price_series`** wraps the fetch in a blanket `except Exception`, so a chart that used to render from silently-padded data now logs `[WARN]` and comes back empty. That is the intended effect — data of unknown provenance should not render as if it were the source the caller named — but it is a visible change.
- **`runner.main()`** has no `try/except` around `fetch_data_map`, so the error surfaces as a traceback rather than the JSON envelope the function's other failure branches print. That behaviour predates this change (the same `raise` could already fire when fallbacks were exhausted); this change makes it easier to reach. Left alone here to keep the diff to one concern — happy to add the envelope in this PR or a separate one if you'd prefer.

## Known remaining gap

`_fetch_auto` has a structurally identical per-symbol fallback loop that this PR does not touch. Practical exposure looks near zero — `local` sits at the tail of every chain, so `resolve_loader` reaches it only when everything else is already unavailable, and `nobitex`/`wallex` are in no chain at all — so I kept the diff to the explicit-source path rather than widen it. Glad to extend it if you read that differently.

## Verification

- `tests/test_local_source_routing.py` — **26 passed**, including a new positive/negative pair: a source in the set raises instead of reaching the chain, a source outside it still reaches the chain. Both fully monkeypatched; neither touches the network. Confirmed red before the change (`AssertionError: must not fall back`).
- full `agent/tests` — **13032 passed, 98 skipped**, 1 unrelated pre-existing failure (`test_mcp_server_http_transport`, red on a clean tree too)


warren618:
Thanks — I verified this on 7acda57 against every member of the set, not only `local`.

With every loader stubbed to record attempts, and a two-symbol request where the primary serves only the first symbol:
- **On main,** `local` fills the gap from yahoo / tencent / okx depending on the market, `qveris` and `fmp` from yahoo, `tickerall` from mt5, and `nobitex` / `wallex` from tencent.
- **On this head,** all six raise `NoAvailableSourceError` without touching another loader, while `tushare`, `yfinance`, `yahoo`, `okx`, `tencent` and `auto` still fall back per symbol exactly as before.

One change before merging: the new test only exercises `local`. With the guard replaced by `primary_source != "local"`, the whole suite stays green (I ran it: 0 failures), so five of the six members are unpinned. Parametrizing the negative case over `sorted(_NO_NETWORK_FALLBACK_SOURCES)` would pin all six now, and any source added later.

A small correction to the description while you're there: `BTC-IRT` / `BTC-TMN` codes are detected as `a_share`, so on main the per-symbol loop walked the A-share chain (tencent), not the USDT crypto chain. The provenance was wrong either way, and the guard covers it.

warren618:
Merged as `78210cda`; `cc5a3bbc` on top parametrizes the negative case over `sorted(_NO_NETWORK_FALLBACK_SOURCES)`, so all six members are pinned and the `primary_source != "local"` mutant now fails five of them. Thanks for tracking this down.

PR #1443, opened 2026-09-14
Integration request: Vibe-Trading & AIML API

hugoaimlapi:
Hi! I'm Hugo from aimlapi.com — an AI aggregator that gives access to 1000+ models in one API, trusted by 400k+ users.

We'd love to be available as a provider option inside Vibe-Trading — so we went ahead and did all the technical work on our side.

To build our partnership, we offer a 50/50 revenue share on all traffic from this integration. (P.S.: tracking starts as soon as this release goes live, so no earnings will be lost during setup)

My contacts: hugo@aimlapi.com (email / Slack), Telegram: @hug0the

---

## Summary

- Adds `aimlapi` (aimlapi.com / AI/ML API) as an OpenAI-compatible provider: one key, 350+ chat models under `vendor/model` ids, default endpoint `https://api.aimlapi.com/v1`.
- Registers it in every provider surface — catalog, onboarding wizard, `vibe-trading init`, `_PUBLIC_PROVIDERS`, desktop credential allowlist — plus a `ProviderCapabilities` record with `capture_reasoning=True` and `top_level_reasoning_effort=True`.
- Adds attribution headers gated on the *resolved base-URL hostname*, not on the provider label, so a redirected `AIMLAPI_BASE_URL` cannot carry the identifier off-vendor.

## Why

Moving between vendors mid-research today means one account, one key and one `.env` block per vendor. AI/ML API fronts 350+ chat models behind a single key, so this adds it the same way ModelScope (#1011) and Requesty (#474) were added: data rows plus a capability record, with no change to the agent loop, routing, or previously saved settings.

The capability flags are set from watched live requests on 2026-09-03, not from vendor docs, per the standard `capabilities.py` keeps for `top_level_reasoning_effort`: `deepseek/deepseek-v4-pro`, `openai/gpt-5-5`, `anthropic/claude-sonnet-4.5`, `google/gemini-3.5-flash`, `alibaba/qwen3.5-plus` and `z-ai/glm-5.3-flash` each returned 200 for a request carrying a top-level `reasoning_effort`, and the reasoning models among them returned `reasoning_content`. `send_reasoning_content` stays off — replaying reasoning on assistant turns is unverified there, exactly as it is for OpenRouter.

## Changes

**Provider record**

- `agent/src/providers/capabilities.py`: `_AIMLAPI_CAPABILITIES` registered under `"aimlapi"`; `AIMLAPI_ATTRIBUTION_HEADERS` (a `MappingProxyType`, so the module constant cannot be mutated by a caller that receives the merged dict) and `AIMLAPI_ATTRIBUTION_HOSTS` (`frozenset({"api.aimlapi.com", "aimlapi.com"})`).
- `agent/src/providers/llm.py`: `_aimlapi_attribution_headers(caps, base_url)` returns a fresh dict only when `caps.name == "aimlapi"` **and** the resolved hostname is an exact member of the allowlist. `build_llm()` merges it *underneath* `caps.default_headers`, so anything the provider record configures wins. A substring or suffix test would have accepted `api.aimlapi.com.evil.io`; both that and `notaimlapi.com` are covered by tests. `HTTP-Referer` and `X-Title` name this project, not the gateway, following the convention OpenRouter established.

**Registration surfaces (data only)**

- `agent/src/providers/llm_providers.json` — catalog entry (Web Settings dropdown), appended in place, no reordering of existing providers.
- `agent/cli/onboard.py` — onboarding wizard entry.
- `agent/cli/_legacy.py` — `_provider_key_env` / `_provider_base_env` maps and the `vibe-trading init` menu entry.
- `agent/src/swarm/models.py` — `aimlapi` added to `_PUBLIC_PROVIDERS`.
- `desktop/electron/src/secure-credentials.ts` — `AIMLAPI_API_KEY` added to `ENV_CREDENTIALS`.

**Docs**

- `agent/.env.example` — config block.
- `README.md` + the five translated READMEs — one name added to the supported-providers line.

**Tests**

- `agent/tests/test_aimlapi_provider.py` (new, 13 tests): registry wiring consistent across all surfaces, base-URL fallback, partner-id format, headers reach the client, the shared constant is never mutated, headers do not ride to another provider, look-alike hosts (`api.aimlapi.com.evil.io`, `notaimlapi.com`, `aimlapi.com.evil.io`, `proxy.internal/aimlapi/v1`) get nothing, pinned legitimate hosts still get attribution, the host list stays exact, and `reasoning_effort` goes top-level rather than into `extra_body`.
- `agent/tests/test_llm_provider_defaults.py`, `agent/tests/test_llm_reasoning_effort.py` — assertions extended for the new provider.

## Out of scope

- No changes to the agent loop, routing, backtesting, order gate, mandate enforcement, or any broker path.
- No provider reordering, no featured/recommended marker, no UI work.
- No new env vars beyond `AIMLAPI_API_KEY` / `AIMLAPI_BASE_URL`; no OAuth, no MCP surface.

## Test Plan

- [x] New tests added (`agent/tests/test_aimlapi_provider.py`)

Full suite on this branch:

```
1 failed, 13412 passed, 99 skipped, 6 warnings in 237.38s (0:03:57)
FAILED agent/tests/test_daily_count.py::test_daily_order_lock_is_cross_process
```

That single failure is pre-existing and unrelated to this change: it reproduces on a clean `upstream/main` checkout in the same environment (`1 failed, 6 passed`), and it is a cross-process file-lock test that does not touch provider wiring.

Targeted:

```
pytest agent/tests/test_aimlapi_provider.py \
       agent/tests/test_llm_provider_defaults.py \
       agent/tests/test_llm_reasoning_effort.py \
       agent/tests/test_cli_init.py -q
# 67 passed, 7 skipped in 10.21s
```

Header scoping, no key needed:

```python
from src.providers.capabilities import get_provider_capabilities
from src.providers.llm import _aimlapi_attribution_headers

caps = get_provider_capabilities("aimlapi", "deepseek/deepseek-v4-pro")
for url in ("https://api.aimlapi.com/v1",
            "https://api.aimlapi.com.evil.io/v1",
            "https://proxy.internal/aimlapi/v1"):
    print(url, "->", sorted(_aimlapi_attribution_headers(caps, url)))

# https://api.aimlapi.com/v1         -> ['HTTP-Referer', 'X-AIMLAPI-Partner-ID', 'X-AIMLAPI-Source', 'X-Title']
# https://api.aimlapi.com.evil.io/v1 -> []
# https://proxy.internal/aimlapi/v1  -> []
```

Live run:

```bash
LANGCHAIN_PROVIDER=aimlapi \
LANGCHAIN_MODEL_NAME=deepseek/deepseek-v4-pro \
AIMLAPI_API_KEY=<aimlapi key> \
vibe-trading
```

## Risk / rollback

- No live, broker, order, payment, or wallet path is touched. No secrets are committed; `.env.example` ships placeholders only.
- Network risk is limited to one outbound host that only receives traffic when the user configures `aimlapi` themselves.
- Rollback is `git revert` of the two commits; removing the catalog row alone already makes the provider unreachable from every surface.

## Checklist

> First box left unchecked deliberately: this does touch `agent/src/providers/`, same as #1011 and #474 did. Happy to reshape the placement if you would prefer something different.


he-yufeng:
Thanks for the detailed writeup; the integration is clearly complete on the technical side. We are not able to take this one, though. Provider additions land on technical merit and user demand, and we cannot accept a revenue-share arrangement with usage tracking tied to the release inside the provider registry.

If AIML API becomes something users ask for as a plain OpenAI-compatible endpoint, we are glad to revisit a community-style provider contribution without the partnership and tracking framing. Leaving the close decision to the repo owner since the partnership side of this is a business call rather than a code one.

hugoaimlapi:
Thanks for the quick reply!
We can drop the tracking and the revenue-share side of this — not a problem at
all, if that makes it easier on your end. I'll wait for the repo owner's feedback just in case, and if that version works for you, we'll push the fix asap

On the community side: we have hundreds of thousands of monthly impressions
across X and LinkedIn, plus our own Discord — so I think there are plenty of
ways we could work together here!

warren618:
Thanks, Hugo, and thanks for offering to drop the revenue share. I'm going to close this anyway, for a reason about the shape of the change rather than the terms.

What makes this integration more than an endpoint is the `X-AIMLAPI-Partner-ID` header, sent on every request a Vibe-Trading user makes through it. That is an attribution identifier attached to our users' traffic. None of the providers we ship sends one, and we don't intend to start, whatever arrangement sits behind it.

Without that header, what's left is an OpenAI-compatible endpoint, and that already works with no code change:

```
LANGCHAIN_PROVIDER=openai
OPENAI_BASE_URL=https://api.aimlapi.com/v1
OPENAI_API_KEY=<your AI/ML API key>
```

We keep the named-provider list in `src/providers/` short on purpose, because every entry is provider-resolution code we maintain. New ones get added when users ask for them in issues. If that demand shows up for AI/ML API, it will land the same way as the others: no partner header and no attribution.

hugoaimlapi:
okay team
anyway thank you! let me know if something changes

PR #1444, opened 2026-09-14
fix(grounding): keep tail-risk measure and confidence through the gate

### LINKED ISSUE #1425: [Bug] VaR / ES / CVaR tail-risk claims pass grounding completely unchecked

he-yufeng:
## Summary

- Tail-risk claims now keep their measure (VaR vs ES/CVaR) and confidence level through the grounding gate instead of flattening to one `tail_risk` kind.
- The confidence level in a tail-risk frame ("VaR 95%: 1.57%") is recognized as part of the frame and no longer checked as a measurement, which was false-rejecting correct answers on current main.
- The evidence alias table gains `es_95` / `es_99` / `cvar_95` / `cvar_99`, which previously fell out of the exact-only guard.

## Why

Closes #1425. On current main, `var_95` evidence grounds `VaR 99%: <same number>` and `ES 95%: <same number>` claims, because the gate only sees one flat tail_risk value pool. Meanwhile a correctly grounded `VaR 95%: 2.34%` fails, because the frame's own 95% is checked against evidence as if it were a measurement. Both shapes are in zeus229's portfolio-risk workflow on the issue.

## Changes

- `figures.py`: a scan-time pass reads the prose frame around each percent figure. The value in a tail-risk frame carries its (measure, confidence); the confidence figure is flagged and skipped. The frame closes at the next measure word or clause end, so a neighbouring metric ("VaR 95%: 1,57%. Drawdown máximo −5,132%.") is not swallowed. A figure explicitly declared in the figures block keeps its declared role over the frame mask.
- `evidence.py`: `_tail_risk_identity_for_path` resolves (measure, confidence) from leaf names, reusing the alias resolution; new aliases for `es_95` / `es_99` / `cvar_95` / `cvar_99`; compound leaves keep the confidence the qualifier stripper drops ("portfolio_cvar_95" reads as ES at 95%).
- `policies.py`: `_metric_pool` narrows to the claim's identity when the figure carries one: measure must match, and a stated confidence must match exactly. A claim with no confidence takes any evidence of its measure. The `no_evidence` message names the missing identity.
- Declaration-scoped checks (`ref`/`note` paths) are unchanged by design; notes stay citation hints.

## Test Plan

- [x] Existing tests pass (`pytest agent/tests --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py -q`): 13414 passed, 98 skipped
- [x] New tests added: identity resolution per leaf shape, confidence match/mismatch both directions, ES aliasing, frame-closing at clause end, declared-confidence-wins (the #1418 Spanish report stays green)
- [x] Tested manually: probe matrix over 16 claim shapes against `var_95` / `cvar_99` evidence, including the reporter's four-way VaR/ES x 95/99 example

Not run: the e2e harnesses excluded above, and no live LLM run.

## Risk / rollback

Gate-only change inside `agent/src/agent/grounding/`; no data, config or API surface moves. Revert restores the previous flat tail_risk matching. Known scope: identity matching covers prose percent claims; a declaration `ref` still scopes by call, and a weekly/other-interval spacing claims are untouched.


warren618 (changes requested):
Thanks for picking this up, and for building zeus229's four-way case into the tests. I probed both heads on a two-bar ledger against current main (a5b79422), and the PR is really two changes that need different answers.

**The evidence side is right.** `_tail_risk_identity_for_path`, the `es_95` / `es_99` aliases (`es_95` really does resolve to nothing on main) and the narrowed `_metric_pool` close a real leak. On main, once the confidence is declared `count`, `var_95` evidence also grounds `VaR 99%`, `ES 95%` and `ES 99%` at the same value. With the PR all three reject.

**The prose-frame side I can't merge, for two reasons.**

1. *It brings back the vocabulary the rewrite removed.* `figures.py` gains three word regexes (`C?VaR|ES|expected shortfall|在险价值|…` and the confidence patterns around them) in the module whose docstring says it reads no natural-language word; the package goes from 12 compiled regexes to 15. A percent those patterns pick as "the confidence" is then skipped outright (`policies.py:470-474`). The rejection on main that this works around is the declared contract rather than a bug: a confidence level is a parameter you chose, declared `count` in the figures block, which `test_the_spanish_var_report_from_1418_grounds_once_its_confidence_is_declared` already pins.

2. *The skip lets invented numbers through.* Same ledger, evidence `var_95 = -0.0157`:

| draft | main | this PR |
|---|---|---|
| `CVaR 9.99%` / `VaR 9.99%` / `ES 9.99%` | reject | pass |
| `VaR 37.2%: 1.57%` | reject | pass |
| `VaR 12% higher than last month` | reject | pass |
| table cell `\| Tail \| VaR 9.99% \|` | reject | pass |
| `VaR 95%: 1.57%, ES 9.99%` | reject | pass |
| `VaR (99.9%) = 2.63%` (with `var_99` evidence) | reject | pass |
| `预期损失 2.3%` (no ES evidence) | reject | pass |

On the release path, `Annual return 45.3%, VaR is 1.57%.` goes out on main as `Annual return（略※）, VaR is 1.57%.` and on this PR as `Annual return 45.3%, VaR is（略※）.`. The before-keyword pattern takes `45.3%` as VaR's confidence, so the invented return ships and the true VaR is cut. The PR also rejects two correct answers: `VaR 95% and ES 99%: 1.57% and 3.42%` with both values in evidence, and a real `quantlib_call historical_cvar` result quoted as `CVaR 95%: 2.11%` with its `95%` declared, because that evidence path carries no confidence.

**What I'd merge.** Removing only the five-line skip in `policies.py` keeps every declared four-way result above and turns every row of that table back into a reject. The two tests that then fail (`test_a_confident_var_claim_only_matches_that_confidence`, `test_an_es_claim_matches_es_evidence_at_its_own_confidence`) are the ones expecting an undeclared `95%` to pass. That still leaves the word patterns deciding which measure and level a sentence names, and that is the part I'd rather not reintroduce. The shape-only place for that identity is the declaration the model already writes: let `ref` name an evidence field as well as a call id or tool (e.g. `risk.data.tail_risk.var_95`), and scope `_referenced` to that field. That also closes a gap both heads share today: `1.57% | observed | ES 95% | q1` is grounded by `var_95` evidence, because the ref check never looks at the field. Whichever way you go, please keep the table above as two-sided tests.

On #1425 itself: an invented `VaR 95%: 9.99%` has been rejected on main since the rewrite, in English, Chinese and decimal-comma form, so what is left there is the identity point.

zeus229:
I prototyped the shape suggested in the review — keeping metric identity in the declaration/evidence reference rather than reintroducing prose vocabulary — and validated it against current main.

The experiment is here for review/reference:

`zeus229/Vibe-Trading:contrib/grounding-tail-risk-ref-identity-v2`

It deliberately does NOT modify `figures.py`, add VaR/ES prose regexes, or skip confidence figures based on wording.

The approach is:

- `ref` may name an exact evidence field in addition to the existing call-id/tool forms;
- a unique field ref scopes only to that field;
- if the same field was produced by multiple calls, the unqualified field ref fails closed rather than pooling calls;
- `call_id::field` provides an unambiguous call + field scope;
- existing call-id/tool refs and legacy loose refs remain compatible.

Examples:

- `var.var_95` -> exact field when unique;
- repeated `var.var_95` across q1/q2 -> reject as ambiguous;
- `q1::var.var_95` -> only q1's VaR95 value;
- `q2::var.var_95` -> only q2's VaR95 value.

Validation:

- grounding suite: 731 passed;
- focused role/figures suites: 238 passed;
- full safe backend suite: 13,774 passed, 99 skipped, 7 failures;
- all 7 failures reproduce unchanged on current upstream main and are in unrelated MCP/OAuth tests;
- Ruff clean;
- `git diff --check` clean;
- all commits carry DCO sign-off.

I also kept the adversarial cases from the review fail-closed: the nine prose examples (`CVaR 9.99%`, `VaR 37.2%: 1.57%`, etc.) are all still rejected.

There are two design limitations worth calling out before deciding whether this is the right direction:

1. The gate does not interpret the declaration note. So:

   `1.57% | observed | ES 95% | q1::var.var_95`

   verifies that 1.57% came from `var_95`, but it does not independently verify that the human-readable note says "ES" rather than "VaR". That is intentional here to avoid putting natural-language vocabulary back into the gate.

2. Field-path discoverability is imperfect. For example, a `quantlib_call(function="var")` result exposes `{"var_95": ...}` while the internal evidence field is `var.var_95`; the agent can reconstruct that from its call arguments, but this mapping is not currently made explicit in the tool result.

So this closes the evidence-scoping gap without prose parsing, but it makes correct structured refs part of the contract.

I have not opened a competing PR. If this is closer to the intended direction, I can either provide the tested changes for incorporation into #1444 or prepare a focused follow-up/replacement PR.

he-yufeng:
Reworked per the review in 792a30b. You were right on both counts: the prose-frame side reintroduced the vocabulary the rewrite removed, and the skip did let invented numbers through (your table reproduced the passes here before the rework).

What the head now is:

- figures.py is byte-identical to main: the three word regexes, the frame marker and the confidence skip in policies.py are all gone.
- The identity now comes from the declaration the model already writes, adopting @zeus229's prototype (`contrib/grounding-tail-risk-ref-identity-v2`): `ref` may name an exact evidence field (`var.var_95`), or `call_id::field` when the same field was produced by more than one call; a field-only ref that repeats across calls fails closed instead of pooling. Call-id/tool refs and legacy loose refs behave as before.
- The aliases stay: `es_95`/`es_99`/`cvar_95`/`cvar_99` resolve to tail_risk now (main resolved them to nothing, as you verified).
- One honest design note: the narrowed `_metric_pool` from the earlier head is removed, because its only identity input was the prose frame. The declared ref does the identity work instead; an undeclared or call-scoped tail-risk claim gets main's pool behavior. If you'd rather keep a pool-side narrowing keyed on something non-prose, say so and I'll add it back in that form.

The table is kept two-sided as asked: the nine adversarial rows are a parametrized reject test, and the field-ref tests pin exact-field, wrong-field, ambiguous-field, and call-qualified shapes. Local verification: the role-hardening file (116) and the full grounding sweep (756) are green on this head, macOS Python 3.14.

Thanks @zeus229 for building the reference shape; the design credit is his.

warren618:
Merged, thanks, and thanks @zeus229 for the field-ref prototype. On top of 792a30b, 67af91f3 changes four things. All four came out of the same two-sided two-bar probe, run with real `quantlib_call` envelopes and the risk-xray key shape:

- A field-only ref counted the calls that returned the field before it dropped other symbols' records. So `0.666 | observed | close | close` failed in a run holding two instruments' bars. The symbol filter now runs first.
- A short ref like `var_99` named no exact field and fell through to the whole pool, where a VaR 99% claim quoting the 95% value found its match. A ref now names a field by its full path or by its trailing part (`var_99`, `tail_risk.var_99`), whole path parts only.
- A bare `historical_var` returned by q1 at 95% and q2 at 99% still fails closed, but the correction said "call historical_var returned no such value". It now lists `q1::historical_var, q2::historical_var` to choose from. Two calls that return the same value are not ambiguous.
- The prompt example `var.var_95` / `q1::var.var_95` is a path no tool produces (quantlib has no `var` function). The prompt and the tests now use `data.tail_risk.var_95` / `var_95` and `historical_var` / `q1::historical_var`.

Against current main, the probe now rejects four identity leaks under exact, short and call-qualified field refs. One legitimate answer, the ambiguous bare field ref, now takes one correction round. Everything else is unchanged, and the grounding suites pass.

#1425 is only half done, so I'm reopening it. A claim scoped to a whole call (`ref x1`), or not declared at all, is still matched against every tail-risk value that call or session holds. Details are on #1425.

PR #1446, opened 2026-09-15
feat(portfolio): validate currencies against ISO-4217 and convert through an explicit rates map

### LINKED ISSUE #1445: [Feature] Generalize portfolio currencies beyond USD / HKD / CNY
### Problem or Motivation

Portfolio currently hardcodes supported valuation currencies to USD, HKD and CNY.

This is already becoming a limitation for international integrations such as INR and KRW, and would also affect common currencies such as EUR, GBP, JPY, BRL and ARS.

Instead of adding currencies one by one, it would be useful to decouple:

valid portfolio currency
available FX conversion path

A portfolio denominated in a valid ISO currency should not be rejected just because that currency is not in a three-item allowlist.

If an actual FX conversion is required and no FX rate/provider exists, Vibe can keep the current fail-closed behavior.

The main hardcoded points appear to be:

agent/src/portfolio/compatibility.py

SUPPORTED_VALUE_CURRENCIES = frozenset({"USD", "HKD", "CNY"})

agent/src/portfolio/normalization.py

if currency == "HKD":
    fx_to_usd = Decimal("1") / usd_hkd
elif currency == "CNY":
    fx_to_usd = Decimal("1") / usd_cny

and the Portfolio display currency validation currently only allows USD/CNY in portfolio/config.py, portfolio_routes.py, and the Web Portfolio editor.

A small generic currency/FX abstraction here could solve multiple current and future international currency limitations at once while preserving the existing safety checks.

Vibe-Trading already supports 8 UI locales: English, Chinese, Japanese, Korean, Arabic, Spanish, German and Brazilian Portuguese. As the project becomes more international, generic portfolio currency handling would better match that direction.

Useful examples include USD, EUR, GBP, CNY, HKD, JPY, KRW, INR, BRL, ARS, MXN, CHF, CAD, AUD, AED and SAR.

This request is not asking for Argentine market support or special ARS handling. ARS is only one example of a broader portfolio-currency limitation.

### Proposed Solution

Allow normalized ISO-4217 currency codes for portfolio/account/position currency fields.

Keep FX support separate:

same-currency portfolio: no FX conversion required
mixed-currency portfolio: require an explicit FX conversion path
missing FX rate: keep the current explicit error

Examples:

ARS portfolio + ARS positions → valid

EUR portfolio + EUR positions → valid

KRW portfolio + KRW positions → valid

EUR position + USD portfolio → requires EUR/USD FX

ARS position + USD portfolio → requires ARS/USD FX

This avoids adding special cases for INR, KRW, EUR, ARS, etc. one at a time.

### Area

Other

### Alternatives Considered

Adding individual currencies to SUPPORTED_VALUE_CURRENCIES would work temporarily, but the same limitation would return whenever a new market or broker uses another currency.

A generic ISO currency identity plus explicit FX capability seems more maintainable.

### Contribution

he-yufeng:
## Summary

- Portfolio currencies are no longer hardcoded to USD/HKD/CNY: any valid ISO-4217 code passes validation, and all conversion runs against an explicit rates map that fails closed with the gap named.
- A position or balance in a valid but unrated currency (KRW, EUR, INR, ...) used to be priced as if it were USD; that source now fails closed instead of corrupting the totals.
- `display_currency` accepts any ISO code. Snapshots carry `totals.display`, per-account `total_display` and `fx.rates`, and a display currency without a rate fails the refresh instead of mislabeling USD figures.

## Why

Closes #1445. Three spots conflated "is a real currency" with "we have a rate for it":

- `compatibility.py` rejected anything outside {USD, HKD, CNY}, so a valid EUR or KRW portfolio could not be read at all.
- `normalization.py` fell through to fx=1 for any currency it did not special-case, silently pricing a KRW lot as USD.
- `config.py` and the settings route only accepted USD or CNY for display.

## Changes

- New `agent/src/portfolio/iso4217.py`: the ISO-4217 code set and `is_iso_currency`. Hardcoded on purpose; a dependency buys nothing for ~180 short strings.
- New `agent/src/portfolio/fx.py`: a `Rates` map anchored on USD, `build_rates`, and `to_usd` / `from_usd`, both raising `PortfolioContractError` naming the currency on a missing or zero rate.
- `compatibility.py`: `ensure_supported_currencies` now checks ISO validity first, then, when a rates map is passed, fails closed on currencies with no rate. The {USD, HKD, CNY} allowlist is gone.
- `normalization.py`: `value_position` / `account_total_usd` / `account_cash_usd` convert through the rates map. The HKD/CNY special cases and the silent 1:1 fallthrough are removed.
- `service.py`: builds the rates map once per refresh; an unrated display currency fails the whole refresh closed (503 with the gap named); snapshot gains the display fields and `fx.rates`; `PORTFOLIO_VALUATION_VERSION` bumps to 3 because older snapshots may carry values priced under the fx=1 fallthrough.
- `config.py` / `portfolio_routes.py`: `display_currency` validated as ISO-4217; the API still rejects non-ISO input with a 400.
- Frontend: types widened to `string`; `totals.display` / `total_display` preferred with the USD/CNY switch kept as the fallback for old snapshots; `money()` keeps CNY at 0 decimals and renders anything else at 2.

Adding a real FX pair is now a one-line change where the rates map is assembled, instead of an edit scattered across four files.

## Test Plan

- [x] Existing tests pass (`pytest tests/ -q --ignore=tests/test_mcp_server_http_transport.py` from `agent/`): 13412 passed, 98 skipped. Pristine `a5b79422` with identical flags: 13399 passed, 99 skipped; the +13 delta is exactly the tests this PR adds.
- [x] New tests added: fx gap naming in both directions, KRW position failing closed vs converting when rated, EUR account total, settings ISO validation both ways, and service level: an unrated position currency errors only that source, an unrated display currency fails the refresh, display fields follow the configured currency.
- [x] Red proof: with the source changes stashed, the 10 new or updated behavior tests fail on the old code.
- [x] Frontend: `tsc -b` clean; `Portfolio.test.tsx` 9 passed. The 38 failures in unrelated suites (useDarkMode, apiAuth, SettingsQVeris, SourcePrioritySettings, Layout, useSSE) reproduce identically on pristine `a5b79422` in a fresh `npm ci` environment here; they are not from this change.

Not run: the e2e harnesses (`tests/e2e_backtest`, `tests/test_e2e_harness_v2.py`), and no live broker or LLM run.

## Risk / rollback

Valuation-path change inside `agent/src/portfolio/` plus the settings schema and its frontend types. Revert restores the allowlist and the fx=1 fallthrough. Two behavior changes users can see: snapshots taken before this change are hidden by the version bump until the next refresh, and a source in an unrated currency now shows an error naming that currency instead of a wrong total.

## Checklist

- [x] Documentation updated (no user-facing doc mentions the old USD/CNY-only display setting; CLI has no currency flag)


warren618 (changes requested):
Thanks for taking this on. Separating "is this a real currency" from "do we have a rate for it" is the right cut, and it changes nothing for existing portfolios: over 30,000 synthetic USD/HKD/CNY positions and 90,000 account totals, totals and cash are identical to main, and no position value moves by as much as half a cent (5 differ in the 8th decimal of `market_value_cny`). Two things need to change before it can merge.

**1. An unrated balance currency now fails the whole refresh, not just its source.** The currency check runs inside each source's isolation, but it only sees position currencies and the account's own currency. Balance-level currencies are first met by `account_total_usd` / `account_cash_usd`, which run in the second loop of `refresh()`, outside that isolation. Where main returned the value unconverted, `to_usd` now raises there. With fake connectors and no network:

| scenario | main | this PR |
|---|---|---|
| IBKR summary in EUR, no positions | complete, EUR counted as USD (wrong) | refresh raises, no snapshot |
| Longbridge balance in SGD | complete, SGD counted as USD (wrong) | refresh raises |
| Futu `assets` row in JPY | complete, JPY counted as USD (wrong) | refresh raises |
| healthy IBKR (USD) + Longbridge (SGD) | complete | refresh raises; the healthy IBKR source is lost too |

The portfolio contract is that a source that can't be valued becomes an `error` row excluded from the totals, while the others still refresh. Running each source's valuation (`value_position`, `account_total_usd`, `account_cash_usd`) inside that source's isolation gives you that. Please add the mixed case as a test: `complete=False`, IBKR counted, Longbridge `error` naming SGD.

**2. `display_currency` accepts codes the service can never rate.** Production rates only ever hold USD, CNY and HKD (`build_rates` wraps the fetched USD/CNY and USD/HKD pair). Saving `EUR` now passes validation, and every refresh after that fails until it is changed back. The Web editor still offers only USD and CNY (`PortfolioSourceEditor.tsx:101`). Please validate the display currency at save time against the currencies the rates map can produce, with a named reason.

Smaller points:
- The description says a KRW *position* used to be priced as USD. On main a KRW position already fails closed; the fx=1 path was account-level balances. Worth correcting, since that is what the valuation-version bump is for.
- `PORTFOLIO_VALUATION_VERSION` 2 → 3 hides every existing snapshot and all history, including for USD/HKD/CNY portfolios whose numbers don't change. That's fine once (1) lands, but please say so in the changelog entry.
- `to_usd` only refuses a missing or zero rate; a negative or NaN rate is still accepted (as on main). Validating once where the map is built (finite, > 0, pair named) covers it.
- `iso4217.py` includes `CNH` (not an ISO-4217 code) and `ANG` / `BGN` / `HRK` / `SLL` / `ZWL` (withdrawn), and lacks `XCG` / `ZWG`. Low impact, since rates still gate valuation, but the set should match its name.

zeus229:
I tested the requested review cases against current upstream and built a clean follow-up implementation based on this PR.

The follow-up addresses the review points:

- valuation failures are isolated per source; a healthy IBKR USD source remains included while an unrated Longbridge SGD source becomes `error`, with `complete=False`;
- `display_currency` is rejected at save time when the production rates map cannot rate it;
- FX rates are validated as finite and > 0 with the pair named;
- the ISO-4217 set was updated per review (`CNH` and withdrawn codes removed, `XCG` / `ZWG` added);
- valuation version 3 behavior is covered explicitly;
- the frontend consumes generic display values safely and will not label a USD fallback value as another currency.

Validation on the follow-up branch:

- backend focused/regression tests: 98 passed;
- Black: clean;
- Ruff: clean;
- `git diff --check`: clean;
- TypeScript `tsc -b`: clean;
- frontend build: clean;
- Portfolio frontend tests: 9/9 passed;
- full frontend suite: 631 passed, 1 unrelated pre-existing locale-sensitive failure in `options.test.ts`.

The implementation is currently available at:

`zeus229/Vibe-Trading:contrib/portfolio-iso-currencies-v2`

I intentionally have not opened a competing PR because #1446 contains the original implementation. Happy for these changes to be incorporated here, or I can prepare a follow-up/replacement PR if that is easier for the maintainers.

he-yufeng:
Thanks @warren618 for the careful probe, and @zeus229 for building the v2 on top of it.

I read the v2 branch end to end against every point of the review:

- The isolation fix is the right cut: the whole valuation (value_position, account_total_usd, account_cash_usd) now runs inside the per-source try/except, so an unrated balance currency turns that source into an `error` row while the rest still refresh. The mixed test is exactly the requested shape (IBKR USD ok + Longbridge SGD error naming SGD, `complete=False`, totals unchanged).
- `display_currency` is validated at save time, first as ISO-4217 and then against the currencies the production rates map can actually produce, with the reason named. The EUR-then-every-refresh-fails trap is gone.
- `build_rates` validates finite and >0 with the pair named, and the version bump to 3 carries the snapshot-hiding note in the comment.
- The ISO set now matches its name (CNH out, withdrawn codes out, XCG/ZWG in).
- The frontend consumes `total_display` only when present and never labels a USD fallback as another currency; the `money()` generalization keeps CNY at 0 decimals and others at 2.
- The description's KRW claim is corrected in the new wording: the fx=1 path was account-level balances, not positions.

One scope note, not a defect: an unrated currency reports `failure_kind: "transient"`, though it will fail every refresh until the config changes. The failure-shape contract is preserved and the error text names the currency, so this is fine to land; if the frontend ever wants to distinguish config errors from transient ones, that's a separate field.

I could not run the v2 branch locally (network blocked the fetch repeatedly today), so my confidence here is from reading the full diff plus his reported runs (98 backend, 631 frontend + 1 pre-existing locale failure). @zeus229 please open the replacement PR from `contrib/portfolio-iso-currencies-v2`; I'll close this one pointing at it so the review thread stays linked. Happy to re-verify locally on the PR head once it exists.

he-yufeng:
Closing in favor of @zeus229's #1510, the agreed v2 replacement incorporating warren618's probe and my follow-up review. The ISO-4217 direction with isolated FX failures is the right shape; thanks for carrying it.

PR #1447, opened 2026-09-15
fix(sdm): exclude non-finite observations from decay metrics

0xouzm:
## Summary

Treat NaN and infinite IC/Sharpe bench observations as missing before computing SDM decay metrics. Each evaluated metric still needs three observations, now finite ones.

## Why

Three persisted `BenchResult(sharpe=float("inf"))` rows currently produce `healthy` in `sdm_status(action="decay_check")`. A non-dry `sdm_decay_scan` then moves a monitoring strategy back to active and records that healthy snapshot. Negative infinity produces `critical`; NaN also produces spurious critical signals in the in-memory store. Neither is usable evidence of strategy health.

Both stores accept these values. SQLite converts NaN to NULL on round-trip, but preserves positive and negative infinity, so the false recovery reproduces through the persistent backend too.

This extends the existing all-None / `insufficient_data` behavior. It does not change thresholds, window ordering, or the state machine. Invalid observations are skipped independently for IC and Sharpe, so an unavailable IC does not discard valid Sharpe evidence.

## Changes

- Filter non-finite observations in the shared `compute_decay_metrics`, before minimum-history and window calculations.
- Add 21 regression cases covering mixed valid/missing observations, both metric fields, and both tool paths against in-memory and SQLite stores. The non-dry scan assertions check that insufficient data neither changes status nor writes a decay snapshot.
- Clarify the finite-observation requirement in the function docstring.

## Test Plan

- [x] Targeted existing and new tests: **154 passed**.
  ```bash
  python -m pytest agent/tests/test_sdm_finite_metrics.py agent/tests/test_strategy_store.py agent/tests/test_strategy_store_adaptation.py agent/tests/test_sdm_governance_wiring.py -q
  ```
- [x] Discriminating regression: the new suite on unchanged main (`a5b79422`) gives **19 failed, 2 passed**. The two passing cases are SQLite's existing NaN-to-NULL behavior.
- [x] Black check and Ruff check on both changed Python files.
- [x] `bash tools/ci_grep_gates.sh` and `git diff --check` pass.
- [ ] Full repository suite not run; this patch is limited to shared SDM metric computation.

## Checklist

- [x] No new dependencies, hardcoded paths, or credentials.
- [x] DCO sign-off and contribution style requirements followed.
- [x] Function docstring updated; READMEs unchanged.

## Risk and rollback

The affected behavior is SDM artifact health classification and its existing lifecycle transitions. Validation uses temporary stores only; no broker orders, external services, credentials, or deployment actions are involved. There is no schema migration or rewrite of existing bench records. Revert this commit to restore prior behavior.


warren618:
Thanks for the clear write-up. It reproduces: three persisted `sharpe=inf` rows make `decay_check` report `healthy`, and a non-dry `sdm_decay_scan` then moves a monitoring strategy back to `active` and records a snapshot. Two finite rows plus one `inf` are enough. On this branch both cases report `insufficient_data`, with no transition and no snapshot.

I also ran 11 finite IC/Sharpe series and 33 scan flows on both main and this branch. Metrics, signals and scan checks are identical, so the filter changes nothing for valid data. CI has now run here, and it's green.

One request: please drop the README.md hunk. That roadmap row exists in all six READMEs, so an English-only edit makes them diverge. The observation-count rule is already documented where it belongs, in the `compute_decay_metrics` docstring. With that removed, this is ready to merge.

0xouzm:
Thanks for the thorough verification. Removed the README.md hunk in 01cd8690; the PR now only changes the shared metric computation and its regression tests. The finite-observation requirement remains in the function docstring.

Verified that README.md matches the original base exactly and `git diff --check` passes. Code and tests are unchanged from the version you reviewed. Also updated the PR description to reflect the final scope.

warren618:
Thanks for dropping the README hunk. Re-ran the new tests: 19 of 21 fail on `main`, and all 21 pass on this branch. Merged — thanks.

PR #1448, opened 2026-09-15
fix(sentiment): remove low from bullish lexicon, fix dual polarity

Shizoqua:
## Summary
- Removed "low" from the bullish word list in the sentiment scoring tool

## Why
"low" was listed in both the bullish and bearish sets, so bearish phrases like "new low" scored neutral instead of negative.

## Changes
- Removed "low" from the positive lexicon in sentiment_tool.py
- Added tests for lexicon disjointness and "new low" scoring negative

## Test Plan
- [x] Existing tests pass (`pytest tests/test_sentiment_tool.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


warren618:
Verified on main: `low` was the only word in both lexicons (55 positive, 77 negative), so "new low" cancelled itself out. The disjointness test pins the whole class rather than this one word, which is what makes it worth having. Merged — thanks.

PR #1449, opened 2026-09-15
fix(quantlib): use each model's own degrees of freedom in Patell z

Shizoqua:
## Summary
- Patell z-statistic variance correction now uses each event's actual degrees of freedom instead of always assuming the market model

## Why
The correction formula was hardcoded for the market model (k=2), so market_adjusted and mean_adjusted models produced a biased Patell z and p-value.

## Changes
- Added a per-model parameter count mapping
- Correction is now computed per event from its own fitted model
- Added a test asserting the correct value for market_adjusted and mean_adjusted

## Test Plan
- [x] Existing tests pass (`pytest tests/quantlib/test_eventstudy.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


he-yufeng (commented):
Mechanism review (static, line level; no local run, per my review scope):

The statistics check out. `estimate_market_model`'s own residual-std dof branches are `market → n-2`, `market_adjusted → n`, `mean_adjusted → n-1`, so the `_MODEL_PARAMETER_COUNT` mapping (2/0/1) mirrors exactly what each fit actually estimated. The old `(n-2)/(n-4)` factor is the `df/(df-2)` correction only for the market model; for the other two it understates the variance correction, deflating the denominator and inflating |patell_z| — i.e. the old code over-rejected the null for mean/market-adjusted studies, the worse direction for an event-study screen. The new per-event `df/(df-2)` matches the standard Patell/BMP form.

The test pins the discriminating property (the per-model correction differs from the k=2 value for both affected models and the z recomputation follows), and the comment in it states the direction of the old bias correctly.

One observation, not a blocker: `corrections` can now contain NaN for events whose df <= 2, same as before only for observations <= 4. With MIN_ESTIMATION_OBSERVATIONS well above 5 this is unreachable in practice, but the NaN path is worth keeping in mind if that floor ever drops.

warren618:
Thanks — the fix is right. `estimate_market_model` estimates two parameters for `market`, one for `mean_adjusted` and none for `market_adjusted`, so a single `(n-2)/(n-4)` correction was only correct for the market model.

Two small changes:

1. **The test comment has the direction reversed.** The correct factor is `n/(n-2)` for `market_adjusted` and `(n-1)/(n-3)` for `mean_adjusted`. Both are *smaller* than `(n-2)/(n-4)`: at n = 30, 1.0714 and 1.0741 vs 1.0769. So main over-stated the variance correction, and its |z| was slightly too small, not too large. That's about ×1.003 on z at n = 30, and negligible at the default estimation window. The comment ("understates the correction … strictly below") says the opposite.
2. **One source of truth for the degrees of freedom.** `_MODEL_PARAMETER_COUNT` carries a note that it must mirror the `dof` branches in `estimate_market_model`. Having `estimate_market_model` compute `dof = n - _MODEL_PARAMETER_COUNT[model]` removes the second copy that has to be kept in sync.

warren618:
Merged as `21a62f51`, with the two small changes from my comment in `844b13b9`: `estimate_market_model` divides by `n - _MODEL_PARAMETER_COUNT[model]`, so there is one table for k, and the test's comment states the direction the numbers go (the corrected |z| is the larger one) and asserts it. The fix itself was right as sent — thanks.

PR #1450, opened 2026-09-15
fix(memory): skip already archived entries in the compression pass

Shizoqua:
## Summary
- GC compression pass now skips entries already archived or deleted earlier in the same run

## Why
Tier 2 compression reused the pre-archive entry list, so entries Tier 1 had already moved to archive/ triggered a failed compress attempt and a misleading warning.

## Changes
- Track archived/deleted entry ids during Tier 1
- Skip those ids in the Tier 2 compression loop
- Added a test covering an entry eligible for both archive and compression

## Test Plan
- [x] Existing tests pass (`pytest tests/test_memory_gc.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


warren618:
Reproduced: on main, an entry that Tier 1 archives is still handed to Tier 2, which logs `Cannot archive non-existent file` and `_write_compressed(...) failed: [Errno 2]`. The archived copy itself is untouched.

One concern with keying the skip on `entry.id`: ids are 6-hex values (`sha256(...)[:6]` when the frontmatter has none), not a per-file key.
- With two files sharing an id, one archived and one healthy but due for compression, this branch skips compressing the healthy one, where main compresses it.
- It would also skip an entry whose archive step returned early on a lock timeout or `OSError`, even though the file is still there.

Checking the invariant directly avoids both: in the Tier 2 loop, use `if not entry.path.exists(): continue` instead of the id set. Happy to merge with that swap.

warren618:
Merged as `8b9e7a5e`, with the swap from my comment applied on top in `451c554a`: the Tier 2 loop skips `if not entry.path.exists()` instead of an id set, so a healthy file that shares a six-hex id with an archived one is still compressed, and an archive step that returned early leaves its file in line. Your test is unchanged and still pins the archived case; one more covers the shared id. Thanks for the report and the fix.

PR #1451, opened 2026-09-15
fix(quantlib): validate microstructure inputs

Shizoqua:
## Summary
Reject invalid non-finite inputs in the microstructure estimators.

## Why
NaN, infinity, and negative VPIN volumes could produce invalid metrics instead of failing clearly.

## Changes
- Validate Roll spread and Kyle lambda inputs.
- Validate VPIN volumes.
- Add regression coverage.

## Test Plan

## Checklist
- [x] Focused change only.
- [x] DCO sign-off included.


warren618:
Thanks. I checked both sides of the gate, not just the rejections. Finite prices and constant prices still return the same spread as main, while NaN, `None` and `inf` prices, which main silently turned into a `0.0` spread, now raise `ValueError`. That's the right direction.

Before merge:
1. Please add accept-side tests (constant prices → 0.0, zero-volume VPIN buckets, zero order flow). Tests that only assert `raises` can't catch a gate that starts rejecting everything.
2. The `amihud_illiquidity` hunk isn't mentioned in the description and has no test, and it changes results (non-finite rows are now dropped). The behaviour is fine; please mention it and add a test.
3. Add `Raises:` sections to the functions that now raise.

warren618:
Merged as `d8bc113d`; `c6010208` on top adds the accept-side tests (constant prices → 0.0, zero order flow → 0.0, zero-volume bars, `amihud_illiquidity` dropping an infinite row the way it drops a NaN one) and the `Raises:` sections. Thanks.

PR #1452, opened 2026-09-15
fix(factors): preserve NaN in gtja191_064

Shizoqua:
## Summary
Preserve missing observations in GTJA191 Alpha 064.

## Why
The factor replaced undefined correlation values with zero, allowing missing data to become a valid signal.

## Changes
- Remove zero filling from the correlation path.
- Add missing-data regression coverage.

## Test Plan

## Checklist
- [x] Focused change only.
- [x] DCO sign-off included.


warren618 (changes requested):
Thanks. The gap problem is real, but two things need to change, and the first one is the red CI.

**1. The new test can't pass on either branch.** The fixture has two symbols on monotone lines, so each symbol's cross-sectional rank is constant (0.5 or 1.0). `ts_corr` over a constant window is NaN, so the correlation branch is NaN everywhere and column B has 0 finite values out of 80, on main and on this branch alike. That is the CI failure. What you mean would be tested by a fixture with ≥30 random-walk symbols, run through `Registry().compute`, asserting that bars whose window holds the gap are NaN while other symbols stay finite.

**2. Dropping `fillna(0.0)` also blanks every constant-rank window**, because `ts_corr` returns NaN for a constant window, not only for a missing one. Here is the output NaN ratio through `Registry().compute` on clean synthetic panels (250 bars):

| symbols | main | this PR |
|---|---|---|
| 30 | 0.865 | raises (`>95% NaN`) |
| 100 | 0.530 | 0.823 |
| 300 | 0.145 | 0.312 |
| 1000 | 0.043 | 0.112 |

On a small universe the alpha stops producing output at all. A fix that masks only the windows containing a missing input, and leaves constant windows at zero as today, would remove the gap fabrication without that cost. For example, build a `present` mask with a 4-bar rolling `min` over `x.notna() & y.notna()` and apply it after the `fillna`. Could you rework it along those lines, with the test above?

he-yufeng:
The fix direction is right and matches the qlib158 NaN-propagation family: `fillna(0.0)` fabricated a zero correlation wherever the rolling window hit missing data, and the fabricated 0.0 then flowed through `ts_max` into the alpha as if it were real. Removing it lets the helper do what it documents (constant window → NaN, no silent zero). One behavior note worth stating in the PR: the removal also changes the constant-rank-window case (a window whose series is flat now reads NaN instead of fabricated 0.0), which is the same intent as the helpers and #1459, but it is a broader behavioral change than just missing data.

The shipped test does not run though. I ran it on this head and on pre-fix code, and it fails identically on both: `result["B"].iloc[40:].notna().any()` is false regardless of the change. Root cause is the fixture, not the fix: with two symbols on monotone price paths, the cross-sectional `rank()` series for column B is flat (1.0 then 0.5), every 4-bar window is constant, and `ts_corr` correctly returns NaN for all of them per its documented constant-window rule. So the second assertion can never pass with this data. It needs fixture data that produces non-constant rank windows (three or more symbols, or non-monotone/noisy paths), and the first assertion currently passes on both code versions too (the p1 path already propagates NaN through `amount`), so as written the test pins nothing about the fillna removal.

he-yufeng (commented):
Direction is right and worth landing once the test is fixed, but the test as shipped is broken — CI already says so, and I reproduced the failure locally with a mechanism for why.

Convention check first: 191 of 192 gtja191 factors propagate NaN; alpha_064's `.fillna(0.0)` is the only outlier. Removing it aligns the zoo, and it also stops the warmup window from seeding fabricated zero-correlation values. I verified the actual behavioral delta with non-degenerate panels (noise-driven close/volume/amount): exactly the series-head cells where `inner` is NaN while p1 is already finite flip from fabricated-finite to NaN (2 cells per column at row 18 in my 80-row fixture). Mid-series missing points are a different story than the PR description suggests: when `close` has a NaN, p1's own correlation path goes NaN through the same trailing windows, and `np.maximum` masks whatever p2 does — so the `.fillna(0.0)` was already dead there, old and new produce byte-identical output in that case.

The problem is the new test. Its fixture builds close/volume as monotone `linspace` and derives `amount = close * volume * 100`. Under this zoo's ops the cross-sectional ranks of `vw` and `v` are then constant per column, `ts_corr` returns NaN for constant windows, and p1 is NaN for every row of both stocks. That is why `result["B"].iloc[40:].notna().any()` is False — the assertion fails on this PR's own code, which is exactly what CI reports on both Python lanes (`test_gtja191_064_preserves_missing_data` line 18). On top of that, the single-NaN case it builds is the masked one above, so the test can't discriminate the change even once B is fixed.

Suggested path: drive the fixture with a seeded noise series (non-monotone close and volume), keep the close-NaN point, and assert on the series-head cells the change actually touches — with that fixture I measure the delta as a few cells at the head flipping to NaN, everything else byte-identical. That pins both the warmup fabrication fix and the zoo convention without a vacuous assertion.

warren618:
Thanks. gtja191_064 is fixed on main in 52f14773, in the shape the 09-16 review suggested: `fillna(0.0)` stays, so a constant window keeps zero correlation, and cells whose 4-bar window holds a missing input are masked. It went in with 23 other alphas of the same class, so I'm closing this one.

PR #1454, opened 2026-09-15
fix(factors): preserve NaN across qlib158 SUMD windows

Shizoqua:
## Summary
Preserve missing observations across the qlib158 SUMD factor family.

## Why
The positive and negative return branches converted missing price differences to zero, allowing rolling SUMD values to continue through data gaps.

## Changes
- Preserve NaN in SUMD 5, 10, 20, 30 and 60.
- Add regression coverage across all five windows.

## Test Plan

## Checklist
- [x] Focused change only.
- [x] DCO sign-off included.


warren618:
Thanks for looking at this. Measured on current main, though, this change doesn't alter any output: your new test passes on main unchanged (5 passed).

The denominator already covers it. `absd.rolling(window=w, min_periods=w).sum()` is NaN for any window that contains a missing diff, and `safe_div` returns NaN wherever the denominator is NaN (`factors/base.py`), so the numerator's value never reaches the result. Since the test can't tell the two versions apart, it doesn't pin anything either, so I'm going to close this one.

A test-only PR that asserts, through `Registry().compute`, that the window after a gap stays NaN would be welcome as a guard. The families that actually emit values over a gap are the ones in #1459.

PR #1455, opened 2026-09-15
fix(swarm): clear agent artifacts before each fresh task, not just retries

Shizoqua:
## Summary
- Swarm worker artifacts are now cleared before a task's first attempt, not only before a retry

## Why
An agent's artifact directory is keyed by agent id alone, so when one agent handles two sequential tasks the second task could read back the first task's leftover report.md as its own result.

## Changes
- Clear the artifact directory at the start of every fresh task dispatch
- Added a test covering two tasks sharing one agent

## Test Plan
- [x] Existing tests pass (`pytest tests/test_swarm*.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


warren618 (changes requested):
Thanks for tracking this down. I reproduced it on main: with one agent on two sequential tasks, the second task's summary is the first task's `report.md`, and it becomes the run's `final_report`.

Clearing the directory on every fresh dispatch trades that for a different loss, though, because an earlier task's files are still referenced after the agent moves on. I ran the real runtime with a fake worker:

- **T1 (agent x) → T2 (agent x), with T3 (agent y) depending on T1.** After the run, T1's persisted artifact list still names `report.md` and `runs/bt1/metrics.csv`, but both files are gone, and they were already gone while T3 was running.
- **Resuming that run.** T1's artifacts silently shrink to `messages.json` and `summary.md`, because `_rehome_artifact_paths` skips missing sources.
- **One agent on two tasks in the same layer.** T2's dispatch clears the directory while T1 is still working, and T1 comes back with its fallback summary and no artifacts. I forced that ordering, so I haven't measured how often real scheduling hits it. Main doesn't fail this way.

The root cause is that the directory is keyed by agent rather than by task, so I'd change the key rather than the lifetime: for example `artifacts/<agent_id>/<task_id>/`, with the retry-time clear scoped to that directory. That touches `agent_artifact_dir`, and the `artifact_dir.parent.parent` run-dir assumption in `_collect_artifacts` (`worker.py`).

Could you rework it that way, with tests that
- (a) an earlier task's persisted artifact paths still exist after a later task on the same agent, and
- (b) two same-agent tasks in one layer keep their own files?

he-yufeng:
Verified the isolation gap and ran the suite both ways. The artifact directory is keyed by agent_id alone, so in a preset where one agent runs several sequential tasks, task N+1 starts with task N's report.md still on disk, and `_resolve_summary` / `_collect_artifacts` then reads it back as the new task's own output. Clearing at the top of `_run_worker_with_retries` (not only between attempts) is the right place: it costs nothing for single-task agents and it matches the contract that inter-task data flows through upstream summaries, not through the shared directory. On pre-fix runtime the new stale-report test fails with task-1's content surfacing in task-2's result; on this branch the isolation suite passes 14/14.

he-yufeng (commented):
Sequential case verified end to end and the fix is right as far as it goes, but the same keying gap this PR clears for sequential tasks is still open for concurrent ones, and the dispatch-time clear makes that case worse rather than better.

What checks out: the artifact dir is keyed by agent_id alone, first attempts never cleared before this change, `clear_agent_artifacts` is exists()-guarded so first-ever tasks are safe, and `run_worker` recreates the dir itself. Critically, `depends_on` context flows through `upstream_summaries` materialized from the task store, not through the shared dir, so clearing at dispatch cannot break legitimate upstream consumption. The test pins the real leak.

The case it doesn't cover: the scheduler runs workers "by topological layer, parallel within each layer" (`runtime.py:3`, ThreadPoolExecutor). A preset can put two tasks with the same `agent_id` in one layer with no dependency between them. They then share the one artifact dir while running concurrently, and with this PR the interleaving is deterministic in a bad direction: task A dispatches and clears the dir, starts writing; task B dispatches a moment later and clears it again, deleting A's in-flight or just-finished `report.md`; A's `_resolve_summary`/`_collect_artifacts` then read whatever B's clear left behind. Pre-change the concurrent case was already racy (both tasks write the same `report.md`), but the dispatch-clear turns "interleaved contents" into "guaranteed deletion of the earlier task's output".

That points at the dir keying as the actual root: keying the artifact dir by `(agent_id, task_id)` instead of `agent_id` alone would close both the sequential leak this PR fixes and the concurrent race it sharpens, without having to reason about which tasks share an agent. Failing that, the scheduler would need to serialize same-agent tasks within a layer, which is a bigger behavioral contract change. Worth deciding before this lands, because merging this as-is documents the shared-dir assumption one level deeper.

Verified by reading `runtime.py`'s dispatch loop, the retry clear at 1132, and `clear_agent_artifacts`/`run_worker` in `worker.py` on current main; I did not run the suite locally.

PR #1456, opened 2026-09-15
fix(pattern-tool): use abs average in head and shoulders symmetry check

Shizoqua:
## Summary
- Head and shoulders shoulder-symmetry check now divides by the absolute average instead of the signed average

## Why
On negative price series the average of the two shoulders could be negative, which flipped the symmetry ratio's sign and let clearly asymmetric shoulders pass as a valid pattern.

## Changes
- Divide by abs(avg) in head_and_shoulders, matching the same fix already applied to double_top_bottom
- Added a regression test with negative, asymmetric shoulders

## Test Plan
- [x] Existing tests pass (`pytest tests/test_pattern_tool.py`)
- [x] New tests added
- [x] Tested manually

## Checklist


warren618:
Verified: double top and double bottom in the same file already divide by `abs(avg)`, so head-and-shoulders was the last signed division, and the three detectors now agree on negative series. Merged — thanks.

PR #1457, opened 2026-09-15
fix(shadow-account): use true Wilder recursive smoothing for RSI

Shizoqua:
## Summary
- Shadow account RSI now uses true Wilder recursive smoothing instead of pandas' default weighted average

## Why
The RSI helper is documented as Wilder smoothing but omitted adjust=False, so on shorter price histories it produced a materially different RSI value that feeds the feature matrix used for clustering and the rule thresholds shown to the user.

## Changes
- Added adjust=False to both ewm calls in the RSI helper
- Added a test comparing against a hand-computed Wilder recursion on a mixed gain/loss series

## Test Plan
- [x] Existing tests pass (`pytest tests/test_shadow_account.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


he-yufeng (commented):
Mechanism review (static, line level; no local run, per my review scope):

The bug is real and the fix is the right one. pandas' default `ewm(alpha=...)` uses `adjust=True`, which computes a weight-normalized average of the whole window, not Wilder's recursive `y_t = alpha*x_t + (1-alpha)*y_{t-1}`. On a documented-as-Wilder helper that is a genuine numerical bug, and it diverges most on exactly the short histories a shadow account's early bars produce. `adjust=False` is the form the doc promises, and keeping `min_periods=period` preserves the warmup gate.

The test is not circular: it recomputes the recursion independently from the first gain/loss pair and compares against the final RSI, so it pins both the recursion form and the alpha, not just "ewm output equals itself."

One note, not a blocker: classic Wilder seeds the first average with an SMA of the first `period` values before recursing, while `adjust=False` recurses from the first element, so this RSI is the "recursive from the start" variant rather than the SMA-seeded textbook one. That is consistent with what the test pins and with most trading-tool implementations; if the feature matrix is ever compared against an SMA-seeded reference (TradingView, TA-Lib), the warmup values will differ slightly even though the long-run values converge.

warren618 (changes requested):
Thanks for looking at this — the RSI helpers are worth getting right. I measured the change, though, and it moves the value away from Wilder rather than toward it, so I can't take it as is.

**`adjust=False` is further from Wilder during warmup, not closer.** Wilder's method seeds the first average with the simple mean of the first 14 changes and recurses from there (the TA-Lib form). `ewm(adjust=False)` seeds with the *first* change instead, and at bar 14 that single value still carries (13/14)^13 ≈ 38% of the weight. Over 2,000 random walks (1.5% per-bar σ), here is the absolute gap to SMA-seeded Wilder, in RSI points (mean / p95):

| bar | main (`adjust=True`) | this PR (`adjust=False`) |
|---|---|---|
| 14 | 3.75 / 8.71 | 13.77 / 28.84 |
| 20 | 3.25 / 7.58 | 9.31 / 20.55 |
| 30 | 1.77 / 4.36 | 4.69 / 11.01 |
| 50 | 0.44 / 1.11 | 1.13 / 2.79 |
| 100 | 0.01 / 0.03 | 0.03 / 0.07 |

The new test builds its expected value with the same first-value seed, so it restates `adjust=False` rather than checking against Wilder. That's why it passes.

**It also splits a pair that has to stay identical.** `shadow_account/scanner.py` has its own `_compute_rsi`, and its docstring says why: it mirrors the extractor's so that the scanner evaluates the same RSI that produced the extracted `entry_rsi14` bounds. Changing only `extractor.py` means the bounds are learned under one smoothing and checked under another. The extractor reads 40 calendar days of history before the earliest buy, roughly 28 bars, and at that length the two smoothings are still 5.8 points apart on average (p95 13.6).

If you want to pursue true Wilder smoothing, the version that would work is one shared, SMA-seeded implementation used by all four copies — `shadow_account/extractor.py`, `shadow_account/scanner.py`, `tools/technical_indicator_tool.py` and `skills/technical-basic/example_signal_engine.py` (plus the note in that skill's `SKILL.md`) — with a test against a hand-computed Wilder series and a test that the extractor and scanner agree. That would move RSI values users already see, mostly on short histories, so it should say so in its description. Happy to review that.

warren618:
@he-yufeng — one measurement to add, since our conclusions differ. I compared both smoothings against SMA-seeded Wilder over 2,000 random walks: at bar 14 `adjust=True` is 3.75 RSI points away on average and `adjust=False` 13.77 (p95 8.71 vs 28.84), converging by bar 100. So on a helper documented as Wilder, `adjust=False` is not the closer variant at the lengths this feature matrix actually uses — the extractor reads about 28 bars before the earliest buy.

The blocking part is separate: `scanner.py` carries a copy whose docstring says it must evaluate the same RSI that produced the extracted `entry_rsi14` bounds, and this PR changes only the extractor.

he-yufeng:
Verified the mechanism and the test locally. pandas `ewm` with its default `adjust=True` divides by the running weight sum, which is a different smoother than Wilder's recursion; on this 15-bar series the old code prints RSI 54.94 where true Wilder gives 62.18, a 7-point divergence on a signal the shadow pipeline consumes. With `adjust=False` the pandas recursion is exactly `avg = alpha*x + (1-alpha)*avg_prev` with `alpha = 1/period`, matching Wilder's RMA term for term, and the new test's hand-rolled reference reproduces the implementation's last value to the digit. Correct fix, tight regression.

warren618:
@he-yufeng — on this PR's own 15-bar series, SMA-seeded Wilder (the first average is the mean of the 14 changes) gives 56.00. `main` gives 54.94 and this branch gives 62.18. The 62.18 comes from seeding with the first change, the same seed `adjust=False` uses, which is why the test's hand-rolled reference matches it to the digit. My review stands, including the extractor/scanner split.

PR #1458, opened 2026-09-15
fix(backtest): annualize options equity over elapsed bars, not bars minus one

Shizoqua:
## Summary
- Options engine annual_return now annualizes over the actual number of elapsed bars

## Why
The exponent divided by bars minus one instead of bars, overstating annualized return, especially on short backtests, and disagreeing with the shared calc_metrics convention used elsewhere in the same repo.

## Changes
- Changed the annualization divisor from n - 1 to n
- Added a test asserting the exact expected value against hand computation

## Test Plan
- [x] Existing tests pass (`pytest tests/test_options_portfolio_correctness.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


he-yufeng (commented):
Verified the convention end to end, and this is right. The options engine appends one equity record per trading day at that day's close (`equity_records.append` in the per-day loop, no day-zero row), so `len(equity)` really is the number of elapsed bars and the exponent should be `bars_per_year / n`. The old `n - 1` understates elapsed time by one bar, which inflates the annualized return whenever growth is above 1 (and the error compounds hard on short runs).

The claim that this matches `calc_metrics` also checks out: `backtest/metrics.py:548` uses `bpy / max(n, 1)` and its own comment at :566 states the account starts at `initial_cash` before the first recorded bar, the same series semantics as here. This was the only `(n - 1)` annualization site left in the engine, and the new test pins the exponent against `252 / len(equity)` directly. CI is green. Good to merge from my read.

warren618:
Verified: `backtest/metrics.py` annualises `equity[-1] / initial_cash` over `n` bars, and the options engine appends one equity row per bar after marking with the same `final / initial_cash` growth, so `n - 1` was the odd one out and short options runs were annualised over one bar too few. Merged — thanks.

PR #1459, opened 2026-09-15
fix(factors): propagate NaN through qlib158 CNTP/CNTN/CNTD instead of counting a halt as a flat day

cgycorey:
## Summary

- `cntp` / `cntn` / `cntd` × 5 / 10 / 20 / 30 / 60 windows (15 qlib158 factors) no longer count a missing close as a day that did not rise.
- Each file masks the comparison to NaN where either close is missing, so `min_periods` keeps every window that still touches the gap NaN.
- Adds `agent/tests/factors/test_qlib158_cnt_nan_propagation.py`: 45 assertions across the 15 factors, each of which fails on `main`.

## Why

Each of these factors converts a comparison against the prior close to 0.0/1.0 and takes its rolling mean:

```python
up = (c > c.shift(1)).astype('float64')
```

A NaN comparison evaluates `False`, not NaN, so a missing close (a halt, a data gap) is counted as a day that did not rise. The comparison is also undefined at the bar *after* the gap, whose prior close is missing — so `window + 1` rolling windows count the gap as a flat day instead of staying NaN. And the same fabrication sits at the start of every series: the first comparison has no prior close at all, so the window ending at row `window - 1` was always defined and wrong, gap or no gap — which is also why these factors emitted their first value at `window - 1` while declaring `min_warmup_bars = window`. After the fix the first value lands at row `window`, matching the declaration and the `sumd`/`vsumd` siblings. That violates `base.py`'s own NaN policy ("every operator propagates NaN; no silent `fillna(0)`") and the `compute()` contract for `agent/src/factors/zoo/**` — NaN preserved at missing-data positions.

The error is not a rounding step. One missing close in `SYM1`, rows `GAP-1 .. GAP+5`:

```
cntp5  clean : 0.6000 0.6000 0.4000 0.4000 0.2000 0.4000
       gapped: 0.6000    nan 0.2000 0.2000 0.0000 0.2000

cntd5  clean : 0.2000 0.2000 -0.2000 -0.2000 -0.6000 -0.2000
       gapped: 0.2000    nan 0.2000 -0.2000 -0.6000  0.0000
```

A halt reads as a weaker signal rather than no signal, and `cntd5` prints `0.0000` where the observed days say `-0.2000` — a missing day rendered as a flat day. Only the gap cell itself is NaN today, and only because the registry masks it (`registry.py`, "enforce NaN contract"); the window after it fabricates.

## Changes

- `agent/src/factors/zoo/qlib158/{cntp,cntn,cntd}{5,10,20,30,60}.py` — `c.shift(1)` hoisted to `prev`, then `.where(c.notna() & prev.notna())` on the 0/1 comparison. No new imports; formula and `formula_latex` unchanged.
- Each `__alpha_meta__` gains a `notes` field following the repo's own convention for guard-triggering alphas (alpha101_019/024/036/037/052 already carry it): sparse panels can trip the `>95% NaN` registry guard, with the measured threshold per window.
- `agent/tests/factors/test_qlib158_cnt_nan_propagation.py` — new, 75 assertions across the 15 factors, parametrized. Six properties: the affected span must be all-NaN; NaN must clear once the gap scrolls out **and the cleared value must equal the no-gap run exactly**; the gap must not touch the other symbol (no stray NaN, values equal to the no-gap run past warmup); a **real exact tie** (`close == prior close`) must still count as a 0/1 observation, not NaN — the guard against buying NaN propagation by NaNing ties, a failure mode the other five assertions cannot see; the same span must be finite with no gap (so an all-NaN output cannot satisfy the gap assertion by accident); and the **warmup edge** — the window covering the missing prior close at row 0 must stay NaN, since pre-fix it was defined and wrong on every panel.

## Test Plan

- [x] New tests: **30 of 75 failed** on `main` (every gap assertion ×15 + every warmup-edge assertion ×15), **75 passed** after the fix.
- [x] Mutation check (revert): reverting only `cntp5.py` fails exactly its two parameterisations (`..._through_the_full_affected_window[qlib158_cntp5-5]` and `..._from_the_missing_prior_close[qlib158_cntp5-5]`) — 2 failed, 73 passed.
- [x] Mutation check (over-masking): a mutant that also masks real ties fails exactly `..._a_real_tie_is_still_an_observation_not_missing_data[qlib158_cntp5-5]` — 1 failed, 74 passed. No other assertion detects it.
- [x] Alpha PR Reviewer Checklist gates: `pytest agent/tests/factors/test_alpha_purity.py agent/tests/factors/test_lookahead.py -q` → **919 passed, 5 skipped**.
- [x] `pytest agent/tests/factors -q` → **1147 passed, 5 skipped** (no existing test pinned the old behaviour).
- [x] `ruff check` clean on all 16 changed files; every line ≤ 120.
- [x] Full suite locally: `pytest --ignore=agent/tests/e2e_backtest -q` → **13504 passed, 33 skipped, 7 failed**, none of them related to this change. All 7 are environmental: 6 × `test_readme_counts.py::test_repo_tree_states_the_real_agent_tool_count[...]` (fail identically on unmodified `main` in this environment) and 1 × `test_governance.py::test_collect_key_package_versions_reflects_real_environment` — it asserts `vibe-trading-ai` resolves, which requires the editable install of the checkout the suite runs from: it passes on the real checkout and fails only in the throwaway worktree used for that run. CI decides the full suite.
- Not applicable / not run: frontend and desktop legs (no code paths touched), `black --check` (see below).

## Checklist

- [x] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`); only `agent/src/factors/zoo/qlib158/` and `agent/tests/factors/`.
- [x] No hardcoded values.
- [x] CONTRIBUTING.md Alpha PR Reviewer Checklist: purity gate ✓, lookahead gate ✓ (no new `shift`, `prev` is `shift(1)`), `__alpha_meta__` unchanged and still schema-valid ✓, `compute()` contract — this PR is that box ✓, LaTeX still matches the code ✓, DCO sign-off on the commit ✓.
- [x] Formatting: repo has no `[tool.black]` section and CI enforces neither formatter; `ruff format --check` reports these files as it reports untouched zoo files and the existing test suite (single-quoted style throughout this zoo), so I did not reformat — that would be unrelated churn in a focused PR.
- [ ] Documentation: no user-facing change.

## Known consequence on sparse panels (measured, disclosed)

NaN-preserving means fewer defined cells, so on panels with many missing bars these factors can now trip the registry's pre-existing `>95% NaN` sanity check (`registry.py::_validate_output`) where the old code silently computed fabricated values and never raised. Measured thresholds (first missing-bar fraction at which `Registry.compute` raises, seeded 400×8 panel):

| window | 5 | 10 | 20 | 30 | 60 |
|---|---|---|---|---|---|
| trips the guard from | ~40% | ~30% | ~15% | ~10% | ~5% missing bars |

`bench_runner_strict.py:531` already catches `RegistryError` and skips the alpha, and the API layer handles it too — the same tradeoff #1377 documented for its registry mask (fail-loud instead of silently wrong). Each file's `__alpha_meta__.notes` now states its threshold, matching the alpha101 convention.

## Deliberately out of scope

This came from sweeping all 558 registered alphas for gap handling, so the shape of what else exists is known and intentionally left alone here:

- **`sumd` / `sumn` / `vsumd` / `vsumn` / `sump`** (arithmetic-diff families): they already propagate NaN through their `absd` denominator, so masking the numerator there changes no output. Not touched.
- **GTJA `SMA(X, N, M)` / `ewm` carry-forward** (≈46 gtja191 factors): pandas `ewm(adjust=False).mean()` carries the last value through a NaN instead of propagating it, so one gap silently freezes the signal for the rest of the series. Fixing it strictly means one gap NaNs the whole tail — that is a policy call, not a mechanical fix, and it needs a maintainer decision before anyone writes it.
- **Other flagged families** (`min`/`max`/`qtld`/`qtlu`/`rank`/`imax`): not verified individually, so no claim is made about them here.
- **Risk**: none live — no broker, network, secret, MCP or deployment path is involved; this is pure computation over a passed-in panel.
- **Rollback**: revert this commit; 16 files, no migrations, no persisted state.


warren618:
Thanks — I verified this through `Registry().compute` rather than by calling the alpha directly. Your new test fails on main (30 of 75) and passes here, and since it goes through the registry, it pins the real call path.

Two nits before merge:
1. The `notes` strings read `~~40%` / `~~30%` / …; please make each a single `~`.
2. Two inline comments are swapped. In `cntp*.py` the `up` line says "a missing close is not a down day", and in `cntn*.py` the `dn` line says "not an up day". The `cntd*.py` pair is right.

On scope: the same pattern (a comparison turned into 0/1 and then summed over a window that spans a gap) exists in other alpha families. Rather than one PR per family, we'd like to fix it once at the registry level, so please hold off on further families until we post that direction.

cgycorey:
Both fixed in `ba051fc8`.

1. The 15 `notes` strings are down to a single `~`.
2. `cntp*` / `cntn*` swapped back: `up` says "not an up day", `dn` says "not a down day" (`cntd*` was already right).

Changed lines, exhaustively: 25 added / 25 removed, every one a `#` comment or a `notes` string — zero code lines, so this commit cannot change any factor value. Re-ran `agent/tests/factors/test_qlib158_cnt_nan_propagation.py` → 75 passed, `ruff` clean.

Understood on scope — no further families from me until you post the registry-level direction.

he-yufeng:
Verified the mechanism and ran the suite both ways locally. The fabrication is real: with a missing close or a missing prior close, `c > prev` evaluates False and `.astype('float64')` turns it into a fabricated 0.0, so the window counts a flat day that never happened and the rolling mean inherits the bias — on sparse panels that pushes rows over the NaN registry guard, which the notes now document per window. Masking with `.where(c.notna() & prev.notna())` makes the missing bar NaN instead of inventing a direction. Against the pre-fix code, 30 of the new propagation tests fail; on this branch the factor suite passes 1177 with 5 environment skips. The per-window guard-threshold notes (5d ~40%, 10d ~30%, 20d ~15%, 30d ~10%) are a nice touch for anyone debugging why a sparse symbol disappeared.

warren618:
Checked the follow-up commit: only trailing comments and `notes` strings changed, so the code is what I reviewed. The new tests fail on `main` (30 of 75) and all pass here. Merged — thanks. The other families stay on hold until #1463 has a direction.

PR #1462, opened 2026-09-16
fix(portfolio): fetch full history in risk x-ray instead of sampled rows

### LINKED ISSUE #1461: [Bug] portfolio_risk_xray fails on longer daily windows because market data is truncated
### Description

portfolio_risk_xray works correctly on shorter daily windows, but fails when the requested history becomes slightly longer than the shared market-data row cap. The tool then returns: no close prices returned for any requested symbol.
Expected behavior / What did you expect?
The same risk analysis should work for longer windows. portfolio_risk_xray needs consecutive daily returns for volatility, VaR, Expected Shortfall and drawdown, so it should fetch the full untruncated history.

Additional context / Technical details

portfolio_risk_xray calls fetch_market_data(...) without max_rows=0.

Once the shared row cap is crossed, fetch_market_data may return a truncated/sampled payload instead of the direct OHLCV record list expected by _closes_frame().

technical_indicators already avoids this problem by requesting max_rows=0 for calculations that require consecutive bars.

A small fix may be:

raw = self._fetch(
    codes=symbols,
    start_date=start_date,
    end_date=end_date,
    source=source,
    interval=interval,
    max_rows=0,
)

A regression test with more than 250 daily bars should cover it.

Small secondary observation:
if one requested symbol has no price data, e.g. SPY.US + BOGUS.US, the x-ray aborts with:
"weights reference symbols with no price data: ['BOGUS.US']"

This may be intentional fail-closed behavior, so the main issue is the long-window truncation.

### Steps to Reproduce

Use portfolio_risk_xray with:

symbols:
SPY.US
QQQ.US
AAPL.US
MSFT.US

weights:
SPY.US: 0.40
QQQ.US: 0.30
AAPL.US: 0.15
MSFT.US: 0.15

source: yahoo
interval: 1D

Case A:
start_date: 2025-09-18
end_date: 2026-09-15

Result: works, aligned_days = 248.

Case B:
start_date: 2025-09-15
end_date: 2026-09-15

Result:
{"status":"error","error":"no close prices returned for any requested symbol"}

### Error Logs

```shell

```

### Interface

Docker

### LLM Provider

OpenAI Codex

### Version

Vibe-Trading v0.1.15 — commit cc54832

### Environment

Ubuntu 24.04 Docker 29.1.3 Docker Compose 2.40.3 Python 3.11.15 inside the Vibe-Trading container

he-yufeng:
## Summary

`portfolio_risk_xray` fetched market data without `max_rows`, so any window longer than the shared 250-row cap came back **sampled** instead of complete. Volatility, VaR/ES and drawdown all depend on consecutive daily returns, so the thinned series produced wrong risk numbers on long windows and could fail outright with "no close prices returned for any requested symbol". The tool now asks for `max_rows=0`, matching the pattern `technical_indicators` already uses for calculations that need consecutive bars.

Closes #1461.

## Change

- `agent/src/tools/portfolio_risk_tool.py`: pass `max_rows=0` on the single fetch site, with a comment stating why consecutive bars are required.
- `agent/tests/test_risk_xray.py`: regression test drives 300 daily bars through a fetcher that reproduces the real cap (samples when `max_rows > 0`, returns all rows when `max_rows == 0`) and asserts all 300 rows reach the report (`aligned_days == 300`, `return_observations == 299`).

## Risk

Low. One call-site argument change inside the risk x-ray tool only; no effect on other tools, the loaders, or the shared cap semantics for callers that want sampling.

## Rollback

Revert the one-line `max_rows=0` addition; the tool falls back to the default capped behavior.

## Verification

- Red check: with the implementation reverted, the new test fails (`aligned_days == 150` from the sampled series instead of 300).
- Green: `pytest agent/tests/test_risk_xray.py` — 29 passed (28 existing + 1 new).
- The reporter's secondary observation (`weights reference symbols with no price data` on unknown symbols) is verified as intentional fail-closed behavior and left unchanged.


warren618:
Verified: without `max_rows`, `fetch_market_data` caps each symbol at 250 rows, and past that `cap_rows` hands back an evenly-strided sample instead of consecutive bars, which breaks every return-based number the x-ray computes. `max_rows=0` is the documented way to get all rows, and `technical_indicators` already passes it for the same reason. The new test fails on main and passes here. Merged — thanks.

Issue #1465, opened 2026-09-16
[Bug] A custom-root alpha silently replaces a bundled alpha of the same name


warren618:
Reported by an offline review, not re-measured here: `Registry(zoo_root=...)` writes the module under the bundled module's key in `sys.modules` (`registry.py`), so an alpha loaded from a custom root with the same module name replaces the bundled one for the default registry in the same process. The reviewer's probe reported the default registry's output changing (NaN cells 3617 → 3754) after a custom-root registry had been constructed.

Needs reproduction, then a namespaced module key per root.

warren618:
Reproduced and fixed by #1468.

PR #1468, opened 2026-09-16
fix(factors): a custom zoo root must not replace the bundled module in sys.modules (#1465)

cgycorey:
## Summary

`Registry(zoo_root=...)` loaded a custom alpha from disk and registered it in `sys.modules` under `src.factors.zoo.<zoo>.<short>` — the bundled zoo's own name. A custom root reusing a bundled zoo/alpha name therefore replaced the bundled module process-wide, and the bundled registry's `importlib.import_module` then returned the custom module. A default `Registry()` silently computed the custom code.

`_load_module` now keys the import cache by `(module_path, zoo_root)`. The import spec name is unchanged, so `__name__`, `__package__` and relative imports inside a custom root behave exactly as before, and repeated computes still hit the cache (`cached.__file__ == str(py_file)` still matches, verified by object identity across two computes).

## Reproduction — this issue's "needs reproduction" step

`Registry(zoo_root=<tmp>/zoo)` containing `qlib158/cntp5.py` that returns the constant `1.0`, panel of 30 bars with a sine close, all through `Registry().compute`:

```
bundled = Registry().compute("qlib158_cntp5", panel)   # real alpha
Registry(zoo_root=zoo_root).compute("qlib158_cntp5", panel)   # custom root loads
again = Registry().compute("qlib158_cntp5", panel)

[left, before]:  [nan, nan, nan, nan, 0.4, 0.4, 0.4, 0.4, 0.6, 0.6, 0.6, 0.4, ...]
[right, after]:  [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, ...]
```

The test asserting those two frames are equal is the one added here; it fails on `main` at this assertion and passes on the branch. Modules load lazily, so the custom root must actually `compute()` for the collision to occur — constructing the registry alone is not enough.

## What is not changed

- No mask, warmup or NaN semantics — this is the loader cache only, so it is unrelated to the per-family NaN hold on #1459.
- `spec_from_file_location(alpha.module_path, py_file)` still names the module by its bundled-style path, so nothing that imports a custom module *by name* changes behaviour except the case that was broken.

## Verification

- New test: `agent/tests/factors/test_registry.py::test_custom_zoo_root_does_not_shadow_the_bundled_module` — RED on `main` (`assert_frame_equal` fails, right side all-1.0), GREEN here.
- `pytest tests/factors -q` → **1103 passed, 5 skipped**.
- Registry consumers: `test_mcp_factor_analysis_contract.py`, `test_strategy_discovery_routing.py`, `test_run_manifest_wiring.py`, `test_agent_loop_repeatable_tools.py`, `test_agent_loop_no_progress.py`, `test_mcp_regression.py` → **44 passed**.
- Cache-hit check: two computes on one custom-root registry return the same module object; `sys.modules` holds only the root-qualified key; `importlib.import_module("src.factors.zoo.qlib158.cntp5")` still yields the bundled module.
- `ruff check` clean. `black` reports this file as needing reformatting **on `main` too** (pre-existing, as `CONTRIBUTING.md:149` describes); `black --diff` touches none of the added lines.


he-yufeng:
Verified the mechanism and ran the new test both ways locally. The failure shape is real: with a custom zoo_root reusing the bundled zoo's module names, the filesystem loader registered the custom module under the bare module_path, and the bundled registry's own import_module then returned the custom code process-wide. Keying the cache by root plus path contains the custom module to its own entry, so the bundled name resolves normally again. The test pins both directions (custom root loads its own constant-1.0 file, the bundled registry still computes the real alpha, and a second bundled call matches the first). On the PR head the registry suite passes 19/19; against the pre-fix registry the new test fails as expected. The sibling-import edge (a custom module absolute-importing another custom-zoo module by bare name now resolving to the bundled one) is a pre-existing shape this change neither improves nor worsens, worth a note if anyone builds multi-file custom zoos.

warren618:
Reproduced #1465 on `main` in both orders. Whether the bundled registry imports first or the custom root computes first, a later `Registry().compute("qlib158_cntp5", ...)` returns the custom root's constant output, because `sys.modules` holds the custom file under the bundled name. On this branch both orders return the real alpha, and the custom module sits under its root-qualified key. Merged — thanks.

PR #1472, opened 2026-09-17
fix(backtest): count one plan rejection per starved open, not one per scale probe

### LINKED ISSUE #1470: unfilled_plan_rejections counts every scale-down re-plan: one cash-starved open is reported as ~26 zero_size rejections (hold mode)
## Summary

In `position_adjustment="hold"` mode (the default), `unfilled_plan_rejections` counts **every re-plan of the cash scale-down search** as a new rejection. One opening order that is starved of cash on one bar is reported as **26 `zero_size` rejections**, which is indistinguishable from 26 lot-rounding failures.

This comes from the #1235 fix, which is otherwise a big improvement: lot truncation is counted exactly once, as it should be.

Verified against the `v0.1.15` tag and current `main` (`1b888f8c`); `agent/backtest/engines/base.py` is byte-identical between the two.

## Mechanism

`BaseEngine._execute_bars`, hold path (`base.py:1268-1276`):

```python
planned = _plans(1.0)
if sum(order.cost for order in planned) > self.capital + 1e-9:
    low, high = 0.0, 1.0
    for _ in range(50):
        mid = (low + high) / 2.0
        candidate = _plans(mid)          # re-plans every open target
        ...
```

`_plans(scale)` calls `_plan_open_order(c, target_w * scale, ...)` for every open target. Whenever a scaled target rounds below one lot, `_plan_open_order` calls `self._on_plan_rejected(symbol, "zero_size", ts)` (`base.py:1538`). So each bisection iteration whose `mid` is too small books another rejection for the same symbol, on the same bar.

The rebalance path already does this correctly. `_fit_rebalance_opens` (`base.py:1810-1825`) runs the same 50-step search through `_at_scale` without touching the hook, then calls `_on_plan_rejected(symbol, "zero_size", ts)` **once per dropped symbol, after the search**. The hold path just needs the same shape.

## Measured (v0.1.15 wheel, `GlobalFuturesEngine`, leverage 1.0, 2 bars)

GC.COMEX at 1850 (one contract = $185k notional), driven through `_execute_bars`. A subclass counts `_on_plan_rejected` calls. "Truth" is the number of GC opens that could not fill.

| plant | setup | truth | `_on_plan_rejected` calls | `unfilled_plan_rejections` |
|---|---|---|---|---|
| control | $1M cash, GC 30.5% | 0 | 0 | 0 |
| A — lot truncation | $100k cash, GC 30.5% (0.16 contract) | 1 | **1** `zero_size` | 1 ✓ |
| B — insufficient cash | $1M; bar 0 opens 8 SI contracts (95%), bar 1 adds GC 30.5% — 1 contract fits at full scale but cash is ~$80k | 1 | **26** `zero_size` | **26** |
| C — cash exhausted | $1M; SI 60% held; free cash drained to 0.01 before bar 1; GC 100% (fits at full scale) | 1 | **27** `zero_size` | **27** |

For B and C, the reproducer checks that GC's **full-scale** plan was a real order (not `None`). So these are genuinely cash-starved opens, not lot rounding in disguise.

## Reproducer

```python
import collections
import pandas as pd
from backtest.engines.global_futures import GlobalFuturesEngine

GC, SI = "GC.COMEX", "SI.COMEX"
D = pd.DatetimeIndex(["2024-01-02", "2024-01-03"])

def bars(p):
    return pd.DataFrame({"open": p, "high": p * 1.01, "low": p * 0.99,
                         "close": p, "volume": 1e5}, index=D)

class Counting(GlobalFuturesEngine):
    def __init__(self, cfg):
        super().__init__(cfg)
        self.calls = collections.Counter()
    def _on_plan_rejected(self, symbol, reason, timestamp):
        self.calls[(symbol, reason)] += 1
        super()._on_plan_rejected(symbol, reason, timestamp)

eng = Counting({"initial_cash": 1_000_000.0, "leverage": 1.0})
data = {GC: bars(1850.0), SI: bars(23.0)}
close = pd.DataFrame({c: data[c]["close"] for c in (GC, SI)})
target = pd.DataFrame({GC: [0.0, 0.305], SI: [0.95, 0.95]}, index=D)
eng._execute_bars(D, data, close, target, [GC, SI])

print(dict(eng.calls))                    # {('GC.COMEX', 'zero_size'): 26}
print(eng._plan_rejection_metrics())      # unfilled_plan_rejections: 26
```

## Expected

- **One rejection per starved open per bar**, as the rebalance path already does: plan once at full scale, run the search without the hook, then report each symbol that was in the full-scale plan but not in the final one.
- Ideally **a distinct reason** for cash starvation (e.g. `insufficient_capital`) rather than `zero_size`. Today "this contract is bigger than the target notional" and "there wasn't enough cash this bar" share a label, and they call for different fixes (raise capital vs. accept the sleeve can't be held).

## Impact

`unfilled_plan_rejections` is the metric #1235 added so a sleeve that silently drops out of the book is visible. With this bug, any run that hits cash limits in hold mode reports rejection counts inflated by ~25× per starved bar, all filed under the lot-rounding reason. Anyone reading the count as "how often did this sleeve fail to fill" gets a confidently wrong number.

Related: #1235.

cgycorey:
## Summary

`unfilled_plan_rejections` counted one rejection per *probe* of the hold-mode
capital scale-down search instead of one per starved open, and filed all of them
under `zero_size` — the lot-rounding reason — so a cash limit read as a lot
rounding problem.

## Why

`position_adjustment="hold"` (the default) plans the opening basket at full
scale and, when it does not fit, bisects one common scale factor — up to 50
iterations, each re-planning every leg. Every probe that rounded a scaled leg to
zero called `_on_plan_rejected` again (Closes #1470).

Numbers from the issue's reproducer, measured on `main` and on this branch:

| case | before | after |
|---|---|---|
| A — target below one lot (lot rounding) | 1 × `zero_size` | 1 × `zero_size` (unchanged) |
| B — SI holds 95 %, GC's 30.5 % is cash-starved | **26 × `zero_size`** | 1 × `insufficient_capital` |
| C — blocked leg (a cause no scale can change) + a leg that does not fit | **51 × `execution_blocked` + 21 × `zero_size`** | 1 + 1 |

Case C is the same defect in a second shape: a market rule cannot be scaled, so
the probes re-reported it verbatim.

## Changes

- `BaseEngine._reject_plan` — `_plan_open_order` now reports through this, which
  is silent while a probe is running (`_plan_probe_depth`). The gate is in the
  engine, not in `_plan_open_order`'s signature, so wrappers that patch that
  method and call the original keep working.
- The hold-mode search reports the legs its final scale dropped, once each, as
  `insufficient_capital` — the shape the rebalance path already uses (#1274).
- `insufficient_capital` joins `UNFILLED_PLAN_REASONS`, so the two situations
  the issue distinguishes are now distinguishable in the run card: "this
  contract is bigger than the target notional" vs "there was no cash this bar".
- `tests/test_plan_rejected.py`: the vocabulary assertion became a subset check
  (the exact dict of reasons this test triggers is still pinned above it).

## Test Plan

- [x] Existing tests pass: `pytest tests -q` from `agent/`
- [x] New tests: `tests/test_plan_rejected_capital_search.py` (5 tests) — the
      two counting shapes RED on `main`, the three invariant guards green on both
- [x] Fills are bit-identical: a differential fingerprint (executed orders,
      trades, capital, per-bar equity, position counts) over 7 scenarios —
      the issue's basket, a lone overflowing leg, lot rounding only, blocked +
      driver, rotations at leverage 2, a commission squeeze, and the rebalance
      path — diffs empty against `main`.

Deliberately not done: no new log line for a dropped leg (the rebalance path
warns; the metric is the hold path's signal, and that is a separate decision).


cgycorey:
Superseded by `0a196e7c` on `main`, which fixes this and closes #1470 — deferring to it.

Checked before closing rather than assuming: this branch's tests pass **unchanged** against `0a196e7c` (all 5, run on a worktree of current `main`), so the two fixes agree on every case they share. Yours also relabels the rebalance path's dropped sleeves, which mine left as `zero_size`.

One thing your suite may not pin: the search multiplied *scale-independent* causes too — a blocked leg on a basket that had to be scaled was booked **51 × `execution_blocked`** plus 21 × `zero_size` for what is 1 + 1. `agent/tests/test_plan_rejected_capital_search.py` on this branch carries that case (it passes on `0a196e7c`); take it if you want it, otherwise nothing here is worth keeping.

PR #1474, opened 2026-09-18
feat(data): Gildata (恒生聚源) A-share OHLCV loader via the raw-api MCP endpoint

sambazhu:
## What

Adds **`gildata`** as a token-gated A-share source (source #28), served by the vendor's MCP raw-api endpoint:

- `agent/backtest/loaders/gildata_loader.py` — one JSON-RPC POST per call (`throttled_post_json` added to `_http.py`), routed to the `StockDailyQuote` tool with `restorationStatus=1` (前复权). A-share codes pass directly as `600519.SH`; non-A-share symbols are dropped so the chain serves them elsewhere.
- Registered in `VALID_SOURCES` + `_loader_modules`, joins the **tail** of the `a_share` chain (paid quota is spent only after the free public sources fail) and is reorderable via the existing `MARKET_DATA_ORDER_A_SHARE` / Settings card.
- `PRICE_CALIBER_BY_SOURCE["gildata"] = "split_dividend"` — **measured, not guessed** (see calibration below).
- Credentials: `GILDATA_TOKEN` (+ optional `GILDATA_BASE_URL`, `VIBE_TRADING_GILDATA_MIN_INTERVAL`), with a Settings-page token field mirroring the tushare one (backend response/request models, dotenv persistence + hot env sync, desktop secure storage, password input + clear checkbox in all eight locales).
- `metrics.py` annualization entries (A-share 4h session, same values as tushare).

## Vendor quirks the loader pins down

- **Units**: `turnovervolume` arrives in 万股 and is converted to shares (`volume_units: {"a_share": "shares"}`); `turnovervalue` (万元) is ignored.
- **`avgprice`/`prevcloseprice` are ignored** — the vendor keeps them on a different adjustment basis than the adjusted OHLC (measured on 600519: qfq close 1491 vs avgprice 1666 on the same bar).
- **Silent empties**: an unresolvable symbol answers `rows: []` with `code: 0` (never an error); the loader returns `None` so the fallback chain keeps walking.
- **MCP Accept header**: the endpoint answers `400` unless `Accept: application/json, text/event-stream` is sent — caught live during deploy, pinned by a test assertion.
- One call covers the full window (verified against a 6.7-year range) and up to 50 symbols; the loader still fetches per symbol so each frame flows through the per-symbol loader cache.

## Calibration evidence (the #1301 rule)

Same symbol (600519.SH), window 2023-06-01 → 2024-06-30 (spans the 2024-06 ex-dividend date), gildata vs tushare qfq:

- volume: `gildata_vol / (tushare_vol × 100)` = **1.000 on every day** (tushare is lots, gildata 万股 → shares).
- price ratio gildata/tushare constant at **0.9275 ± 0.0001** across all 261 days → both series are the same caliber with different qfq anchor dates (gildata anchors at today, tushare at the window end; the ~7.25% gap is the 2024–2026 Moutai dividends).
- max daily-return difference: **0.0124%** — backtest-identical.

## Tests

- `tests/test_gildata_loader.py` (30 tests): registration/metadata, auth gating incl. placeholder token, symbol mapping (`.SS`→`.SH`, bare-code exchange inference, non-A-share → skipped), pure parsing (ascending sort, float contract, 万股→shares, missing volume → 0, `enddate` key fallback, incomplete OHLC dropped, `avgprice` ignored), fetch with mocked HTTP (URL/envelope/arguments pinned, batch resilience, silent-empty omission, vendor error shapes, interval gate, date-range gate), base-URL override.
- Updated: `test_registry.py` chain expectation, `test_env_schema.py` defaults, `test_price_caliber.py` table entry + loader-pins-caliber test, `test_settings_api.py` permutations + gildata token roundtrip/clear.
- Full suite: 2943 passed; the one failure (`test_alpha_compare_api.py::test_stream_emits_result_then_done`) reproduces on a clean `upstream/main` checkout in this environment and is unrelated.

## Docs

data-routing SKILL.md source row, README data-source tables + A-share chain lines in all six languages, CHANGELOG Unreleased entry.

## Deployment verified

Deployed on a live instance (v0.1.15 base): Settings shows the token field (`gildata_token_configured: true`), the a_share chain carries `gildata`, and real fetches serve data with provenance `source=gildata, adjustment=split_dividend, volume_unit=shares`.

## Migration note for existing users

A `MARKET_DATA_ORDER_A_SHARE` value saved before this change is a permutation of the old 7-source chain, so it stops validating once `gildata` joins the default chain. The Settings card flags it (`override_invalid`) and the default order applies until the saved order is re-saved with the new source included — the same warn-and-keep-default path a typo takes, deliberately, so a stale value can never silently strip sources.


he-yufeng (commented):
Mechanism read through the loader, the chain registration, and the caliber stamping. The core of this is right and measurably so: the caliber is pinned at `split_dividend` by the request parameter and was checked against a live payload (the qfq-close vs avgprice divergence on 600519 is exactly the kind of measurement that keeps this table honest), volume is normalized 万股→shares with the units declared, suspended days fill 0.0, `avgprice`/`prevcloseprice` are deliberately excluded because they sit on a different adjustment basis, the vendor's HTTP-200-with-error-body envelope is mapped instead of passing silently, per-symbol failure is isolated with a warning so one bad code doesn't kill the basket, and the chain-tail placement spends paid quota only after the free sources fail. Symbol normalization (.SS→.SH, leading-digit exchange inference, non-A-share → None so the chain serves them elsewhere) matches the registry's contract.

One thing to fix before this lands: the token rides on the URL query string (`?token=...&format=json`), and the failure path can print it. `_call_tool` raises `requests.RequestException` from `throttled_post_json`, and the per-symbol handler in `fetch` does `logger.warning("gildata failed for %s: %s", code, exc)`. requests embeds the request URL in `ConnectionError`/`HTTPError` messages (via urllib3's MaxRetryError and `raise_for_status`), so the first DNS hiccup, timeout, or 5xx writes the full token into the log. Suggest either moving the token to a header if the vendor accepts one, or sanitizing before logging: strip the query at the `_call_tool` boundary (catch, re-raise with the URL redacted) or log `type(exc).__name__` plus a redacted message. Same concern applies to the outer `RuntimeError(f"gildata MCP call failed: {message or outer!r}")` if the vendor ever echoes the request URL in its error body.

Everything else verified at source level: no live call made from this machine (no token), so the live-measured claims (600519 divergence, Accept-header 400, error-envelope shape) are taken from the PR's own evidence rather than re-verified.

sambazhu:
Thanks @he-yufeng — addressed in cbc537da by taking your first option: the
token now rides the Authorization header and never appears in the URL.

Verified against the live endpoint from the deployment machine: the raw-api
service accepts `Authorization: Bearer <token>` (full payload with the
header, `1001 认证凭证缺失或有误` with no credential), and the loader's
fetch still returns the same bars (600519.SH 2026-09-17 close 1266.98,
matching the calibration table in the PR description).

The URL now carries only `format=json`, so request exceptions (DNS failure,
timeout) can no longer embed the token. The fetch test pins both directions:
`"token=" not in url` and `headers["Authorization"] == "Bearer secret"`.

The srv-tool client in #1485 used the same query-string pattern — fixed the
same way there (9425a529…8c291730 stack, tested likewise).

he-yufeng:
Thanks — the header move plus both-directions pins close it cleanly, and the live verification (accepts Bearer / rejects without / same 600519 bars) is exactly the evidence this needed. Nothing left from my side.

warren618:
Merged, thanks. The red CI was the PR's own: the seven README loader lines (README_id.md included now) and two tests that pinned a seven-source A-share order. Fixed in ead3f2e7, along with two places that still said the token rides the URL. Worth knowing for users: a `MARKET_DATA_ORDER_A_SHARE` saved before this stops validating and falls back to the default order until it's re-saved, as your CHANGELOG entry says.

PR #1475, opened 2026-09-18
fix(shadow-account): fix overtrading pnl double count

Shizoqua:
## Summary
- Overtrading pnl no longer double counts trades already explained by noise, early exit, or late exit

## Why
Those trades stayed eligible for overtrading selection, distorting the missed signals residual.

## Changes
- Exclude already explained trades from overtrading candidates
- Added regression test

## Test Plan
- [x] Existing tests pass (`pytest tests/test_shadow_attribution.py tests/test_shadow_account.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


he-yufeng (commented):
Verified end to end, locally. The object-identity chain is sound: `pool_roundtrips` is a filtered view of the same dict objects (the comprehension does not copy), the bucketing loop marks those exact objects into `already_explained`, and `_overtrading_pnl` receives the same list, so `id()` membership is exact within the call, and both lists hold the references alive for the whole call so no id can be recycled mid-flight. The exclusion preserves #17's mutual-exclusivity contract — a trade already booked into early/late/noise can't reappear as an overtrading extra.

Local red/green on the attribution suite: the new test fails against main's backtester (the double count reproduces) and passes on this branch; the other five tests pass on both.

One robustness note, not a blocker: `id()` is exact only while nothing copies a roundtrip dict between the bucket loop and the overtrading call. That holds today because the same list objects flow straight through; if a future refactor materializes copies, the exclusion silently stops working. A structural key ((symbol, buy_dt, sell_dt)) or a short comment pinning the "same objects flow through" invariant would make the invariant survivable.

warren618:
Merged, thanks.

PR #1476, opened 2026-09-18
fix(channels): fix duplicate reply suppression

Shizoqua:
## Summary
- Duplicate outbound reply suppression now works

## Why
It checked a metadata field nothing ever set, so it could never trigger.

## Changes
- Key suppression check off the message id every channel populates
- Added regression test

## Test Plan
- [x] Existing tests pass (`pytest tests/test_channels_runtime.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


warren618:
Merged, thanks. Follow-up (8bbdd78a): Telegram and NapCat message ids are ints, so the `str` check never deduplicated them; the `origin_message_id` branch is removed because nothing sets that key, and when it equalled `message_id` it stored before the second branch compared and suppressed a first reply; the fingerprint map is capped instead of growing for the process's life.

PR #1477, opened 2026-09-18
fix(quantlib): add intercept to Kyle's lambda

Shizoqua:
## Summary
- Kyle's lambda now fits with an intercept instead of through the origin

## Why
Real order flow is rarely mean zero, so the through-origin fit biased the slope.

## Changes
- Demean both series before computing the slope
- Added regression test with known ground truth

## Test Plan
- [x] Existing tests pass (`pytest tests/quantlib/test_microstructure.py`)
- [x] New tests added
- [ ] Tested manually

## Checklist


warren618:
Merged, thanks. It conflicted with #1451's finiteness check in `kyles_lambda`; both are kept. With an intercept, a constant non-zero flow identifies no slope either, so it now returns 0.0 like an all-zero flow, and a test pins that beside #1451's case.

Issue #1483, opened 2026-09-18
[Feature] get_southbound_flow: tool surface for the only live Stock-Connect net-flow signal (Eastmoney datacenter primary, HKEX official fallback; tushare southbound fields have drifted to cumulative stock)


shadowinlife:
**Problem or Motivation:**

Southbound (mainland → HK) is the half of Stock Connect that still has live official net-flow data — and Vibe-Trading has no tool for it.

- `get_northbound_flow` covers northbound only. Since the 2024-08-30 HKEX disclosure reform, northbound **daily net buy is no longer published at all** (see #1481); the only live net-flow signal in Stock Connect today is **southbound**, which HKEX still discloses daily (Buy/Sell Turnover per channel).
- The `hk-connect-flow` skill already carries a full southbound methodology (signal framework, composite scoring, sector patterns), but its Data Access section can only point southbound users at hand-written `pro.moneyflow_hsgt(...)` code — there is no structured, throttled, envelope-standard tool surface for it.
- The tushare route the skill suggests is itself broken for daily flows: `moneyflow_hsgt`'s southbound fields (`ggt_ss`/`ggt_sz`/`south_money`) now carry **cumulative net-bought stock, not daily flow**. Verified 2026-09-18: tushare `ggt_ss=32165.59` equals Eastmoney's 历史累计净买额 `3.216559` 万亿 HKD exactly, and the value moves ~0.4% (coefficient of variation) across a 6-week window — it is a stock, not a flow.

**Proposed Solution:**

Add `get_southbound_flow`, mirroring `get_northbound_flow`'s envelope and throttling discipline:

- **Primary — Eastmoney datacenter** `RPT_MUTUAL_DEAL_HISTORY` (`MUTUAL_TYPE="002"` 港股通沪 / `"004"` 港股通深) through the shared throttled Eastmoney client: `NET_DEAL_AMT` (daily net buy), `BUY_AMT`/`SELL_AMT`, `ACCUM_DEAL_AMT` (cumulative), unit 亿 HKD. Cross-verified against HKEX official daily statistics to the cent on three anchor days (2026-09-15/16/17; e.g. 09-17 港股通沪 net `25.0239亿` == HKEX SSE Southbound Buy `26,387.65M` − Sell `23,885.26M` = `2,502.39M HKD`).
[...]

PR #1484, opened 2026-09-18
fix(tools): northbound flow stops serving a silent empty envelope off a dead feed, and the tushare fallback stops mislabeling turnover as net flow

### LINKED ISSUE #1481: [Bug] get_northbound_flow silently returns an empty envelope: Eastmoney northbound net feed dead since 2024-08-30, payload shape drifted, and the tushare fallback's ×100 mislabels turnover as net flow
**What happened:**

`get_northbound_flow` silently returns a useless envelope — `ok: true, source: "eastmoney"` with `realtime` all `null` and `history: []` — on every call, and never falls through to the tushare fallback. Live run against current `main` (d3c29488):

```json
{"ok": true, "market": "China A", "source": "eastmoney",
 "data": {"unit": "10k CNY", "lookback_days": 5,
          "realtime": {"shanghai_connect": null, "shenzhen_connect": null, "total": null},
          "history": []}}
```

There are three stacked root causes, each verified against live endpoints on 2026-09-18:

**1. The Eastmoney `kamt` northbound net-flow feed is dead (since 2024-08-30).**
HKEX stopped publishing northbound daily net buy in late Aug 2024. The `push2his` `kamt.kline` series confirms it: the last non-zero net-inflow point is **2024-08-29**, and every trading day from **2024-08-30 onward is `0.00` with a frozen cumulative** (`273757367.45`, identical for 2 years). The realtime `kamt/get` payload carries `dayNetAmtIn: 0.0` for both `hk2sh`/`hk2sz`.

**2. The Eastmoney payload shape drifted, so the parser reads keys that no longer exist.**
`_parse_realtime` reads `data.hk2sh.netBuyAmt`, but the live payload nests `dayNetAmtIn`; `_parse_history` reads `data.klines`, but the live payload returns per-channel arrays under `data.hk2sh` / `data.hk2sz`. Result: even the (dead) zeros never reach the envelope — history parses to `[]`. Because the HTTP calls succeed, the `except`-based tushare fallback never triggers: the tool reports `ok: true` on a structurally empty payload.

**3. The tushare fallback's ×100 rescale is wrong for current data — the field semantics changed at the same disclosure reform.**
`tushare_fallbacks.fetch_northbound_flow` multiplies `hgt`/`sgt`/`north_money` by 100 ("million CNY → 10k CNY", pinned by `test_northbound_converts_tushare_million_yuan_to_10k_cny`). But since 2024-08-30 tushare repurposed these fields to carry **turnover, in million CNY** — verified to the cent against HKEX's official Stock Connect Daily Statistics on three anchor days:

| trade_date | tushare `hgt` | HKEX SSE Northbound Total Turnover | tushare `sgt` | HKEX SZSE Northbound |
|---|---|---|---|---|
| 2026-09-15 | 107092.06 | 107,092.06 | 122612.63 | 122,612.63 |
| 2026-09-16 | 122957.80 | 122,957.80 | 141280.72 | 141,280.72 |
| 2026-09-17 | 120233.94 | 120,233.94 | 139631.89 | 139,631.89 |

(`north_money` = `hgt + sgt` exactly.) The magnitude jump across the boundary is also visible in the raw series: 2024-06-28 `north_money=4948.47` (net buy, million CNY) → 2024-08-30 `167789.97` (turnover, million CNY). So for post-reform rows the ×100 inflates turnover 100× **and** the envelope mislabels turnover as "net inflow, 10k CNY"; for pre-reform rows the ×100 was correct. One constant factor cannot serve both eras.

**What I expected:**

- A dead/empty Eastmoney payload should be detected and treated as provider failure, falling through to the tushare path (or returning an honest error when no fallback is configured) — never `ok: true` with an empty envelope.
- The tushare path should pass values through in their real unit (million CNY) with the 2024-08-30 semantics boundary documented in the envelope (net buy before / turnover after), instead of a blanket ×100 under a "10k CNY net inflow" label.

**Steps to Reproduce:**

1. `NorthboundFlowTool().execute(lookback_days=5)` (or the MCP tool `get_northbound_flow`) with network access — observe `ok: true`, `realtime` nulls, `history: []`, `source: "eastmoney"`.
2. `curl 'https://push2his.eastmoney.com/api/qt/kamt.kline/get?fields1=f1,f3&fields2=f51,f52,f54&klt=101&lmt=10'` — observe `0.00` net inflow with a frozen cumulative on every recent day.
3. With `TUSHARE_TOKEN` set, call `tushare_fallbacks.fetch_northbound_flow(lookback_days=5)` directly — values come back ×100 against the HKEX official turnover figures linked above.

**Error Logs**

```shell
# no exception is raised — that is the bug: the failure is silent
{"ok": true, "market": "China A", "source": "eastmoney", "data": {"unit": "10k CNY", "realtime": {"shanghai_connect": null, "shenzhen_connect": null, "total": null}, "history": []}}
```

**Interface:** MCP Plugin (also reproducible via Python API)

**LLM Provider:** n/a (data-layer bug, provider-independent)

**Version:** main @ d3c29488 (0.1.15)

**Environment:** macOS 15 / Python 3.12

---

Proposed fix (PR to follow, scoped minimal):

1. `northbound_tool.py`: detect the dead Eastmoney state (realtime all null/zero **and** history empty/all-zero) and route it into the existing tushare fallback branch with an explanatory warning; when the fallback is unavailable, return an honest `ok: false` error naming the 2024-08-30 disclosure change instead of an empty success envelope.
2. `tushare_fallbacks.fetch_northbound_flow`: drop the ×100, emit `unit: "CNY million"`, and carry a `note` documenting the semantics boundary (net buy before 2024-08-30 / turnover after, HKEX-verified).
3. Tests: dead-payload → fallback; healthy legacy-shape payload → eastmoney path preserved; passthrough unit pinned (replacing the ×100 pin).

Southbound is **not** affected by the disclosure change (HKEX still publishes southbound Buy/Sell Turnover; Eastmoney's southbound daily series cross-checks exactly against it) — a separate feature request will propose a `get_southbound_flow` tool.

shadowinlife:
## Summary

- `get_northbound_flow` stops returning a silent empty `ok:true` envelope off Eastmoney's dead northbound feed: the dead state (realtime **and** history both empty/all-zero) now falls through to the tushare fallback, or returns an honest `ok:false` naming the disclosure change when no fallback is available.
- The tushare fallback drops the blanket ×100 rescale: values pass through in their real unit (`CNY million`) with a `note` documenting the 2024-08-30 semantics boundary (net buy before / turnover after, HKEX-verified to the cent).

## Why

Closes #1481.

Eastmoney's `kamt` northbound net-flow feed died with the 2024-08-30 HKEX disclosure change (last non-zero net: 2024-08-29; every day since is `0.00` with a frozen cumulative), and its payload shape drifted (`netBuyAmt`→`dayNetAmtIn`, `klines`→per-channel arrays). The HTTP calls still succeed, so the `except`-based fallback never fired and the tool reported `ok:true` with `realtime` nulls and `history: []`. Meanwhile the tushare fallback's ×100 rescale — correct while `hgt`/`sgt`/`north_money` were net buy in million CNY — inflates post-reform rows 100× and mislabels them: those fields now carry **turnover** in million CNY, verified equal to HKEX's official Stock Connect Daily Statistics on 2026-09-15/16/17 (e.g. `hgt=120233.94` == SSE Northbound Total Turnover `120,233.94`).

## Changes

- `agent/src/tools/northbound_tool.py`
  - `_is_dead_feed()`: both-sided dead-state detection (a live feed always carries either a realtime figure or a non-zero recent history row; one dead side alone — e.g. a trimmed all-`None` window — still serves Eastmoney).
  - `_tushare_fallback_envelope()`: one shared fallback path for the HTTP-failure and dead-feed branches; without a usable fallback it returns `ok:false` naming the 2024-08-30 disclosure change instead of an empty success.
  - Description updated to state both units and the disclosure boundary (pinned routing anchors — MARKET-WIDE lead, 北向, get_fund_flow disambiguation, example — untouched).
- `agent/src/tools/tushare_fallbacks.py`
  - `fetch_northbound_flow`: passthrough in `CNY million`, `_NORTHBOUND_SEMANTICS_BOUNDARY` constant + envelope `note` documenting net-buy→turnover.
- Tests: `TestDeadFeedDetection` (all-zero fall-through, honest error without tushare, healthy-payload guard, one-sided-dead guard), missing-data-block test re-pinned to fall-through, ×100 pin replaced by passthrough+note pin.

Out of scope: southbound tooling (proposed separately in #1483 — southbound net **is** still officially disclosed and unaffected by this change); any Eastmoney payload-shape re-support for northbound nets (the underlying figures no longer exist, so there is nothing to parse).

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q` → **13680 passed, 99 skipped, 0 failed**)
- [x] New tests added (`TestDeadFeedDetection` ×4 + rewritten fallback/passthrough pins)
- [x] Tested manually (2026-09-18 live probes: `kamt.kline` all-zero since 2024-08-30 with frozen cumulative; `kamt/get` `dayNetAmtIn=0.0`; tushare `hgt`/`sgt` == HKEX official Total Turnover to the cent on three anchor days; tool end-to-end before fix returned the empty envelope above)

## Checklist

- [x] No hardcoded values (API keys, file paths, magic numbers) — the 2024-08-30 boundary is a documented domain constant with its evidence in the docstring and #1481
- [x] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines (black-clean additions, ruff clean, no unrelated reformatting, 329/261 lines, Google-style docstrings, type-annotated signatures)
- [x] Documentation updated (tool description + docstrings carry the semantics boundary; envelope gains `note`)

**Risk / rollback:** read-only data tool; no order/broker/credential surface. Behavior change is confined to `get_northbound_flow` envelopes: the dead-feed case moves from empty-success to tushare-served (or honest error), and fallback values change magnitude (×100 removed) with the unit label corrected. Single-commit revert path.


warren618 (changes requested):
Thanks — the silent empty envelope on main is real. The tushare fallback needs another look, though. After the cutover, `hgt`/`sgt` are turnover, not net flow: today's `realtime.total 283911.86` equals HKEX's northbound total turnover (133,832.25 + 150,079.61), and it is served under net-flow keys, with only `data.note` saying otherwise. The boundary is also 2024-08-19, not 08-30: tushare `hgt` is −2,568.22 on 08-16 and +46,129.34 on 08-19, positive from then on. Could post-boundary rows come back as `turnover_*` with net set to null (or `ok:false` pointing at southbound), with the date corrected?

he-yufeng (commented):
Read the whole diff and ran it locally. This is the right shape for #1481: the dead-feed check demands both sides silent (a genuine halt can zero one), falls through to tushare, and fails honestly with both providers named when neither serves. The semantic-boundary work is the part I would not have caught as fast: dropping the blanket x100 rescale is correct once post-2024-08-30 rows are turnover rather than net buy, and carrying the boundary in data.note keeps the units honest instead of pretending one factor fits both eras. Verified on the PR head: the new dead-feed and passthrough tests fail on the parent commit and the full pair of files passes 18/18 on this branch. Good to land from my reading.

Closes my #1481 claim; thanks for picking it up.

warren618:
Thanks for the local run, @he-yufeng. For the record, the two points from my review on 2026-09-21 are still open on `b2af4a49`, which predates that review: after the cutover, tushare `hgt` / `sgt` are turnover and are still served under net-flow keys, with only `data.note` saying so; and the boundary is 2024-08-19, not 08-30 (tushare `hgt` is −2,568.22 on 08-16 and +46,129.34 on 08-19). Post-boundary rows should come back as `turnover_*` with net set to null (or `ok:false` pointing at southbound).

PR #1485, opened 2026-09-18
feat(tools): three Gildata research tools over the srv-tool NL endpoint

sambazhu:
## What

Adds **three agent-facing research tools** over Gildata's 标准版 (srv-tool) natural-language endpoint — the vendor capabilities the project does not already cover. MCP surface grows **75 → 78**:

| Tool | Vendor tool | Why (gap) |
|---|---|---|
| `get_cn_macro_series` | `MacroIndustryData` | China macro / regional / industry **EDB time series** (GDP, CPI, PMI, rates, 31 industry series). `get_macro_series` (FRED) covers US/global only. |
| `get_cn_announcements` | `AnnouncementData` | A-share / HK / **fund announcement retrieval** with highlighted excerpts. `get_sec_filings` covers SEC only. |
| `search_broker_reports` | `FinancialResearchReport` | Broker-research search over the **Juyuan library** (title, broker, analyst, rating, excerpt, 新财富 flag) — coexists with the eastmoney-backed `get_research_reports`. |

Deliberately **not** wrapped (overlap with existing tools): news (`get_stock_news` / `web_search`), screening (`screen_market` / `iwencai_search`), fund/manager selection, generic `FinQuery`, `FinDataFallbackQuery`.

## Design

- The agent authors the NL `query`; the server routes it. The **routed `api_names` are echoed in every envelope**, so a mis-routed question is visible and the agent can re-ask.
- One thin shared client (`src/tools/gildata_srv.py`): same JSON-RPC POST + `format=json` transport as the loader's raw-api path (dual MCP Accept header included), its own throttle bucket (`VIBE_TRADING_GILDATA_SRV_MIN_INTERVAL`, 0.5s default — these calls cost ~5s and belong to research flows).
- Rows are the vendor's structured dicts, capped via `limit` (default 50, max 200).
- Key-gated on `GILDATA_TOKEN` like the other gildata surfaces; optional `GILDATA_SRV_TOOL_URL` override (new env field).

## Verified live

- `get_cn_macro_series("2024年中国季度GDP同比增速和月度CPI同比")` → EDB rows, GDP quarterly YoY 5.4 / 4.6 %.
- `get_cn_announcements("贵州茅台2024年年度利润分配公告")` → the actual 分红方案公告 (每10股276.24元) plus the implementation notice, each with a highlighted excerpt.
- `search_broker_reports("最近一个月关于新能源汽车行业的券商研报观点")` → fresh reports (broker, analyst, industry, excerpt).

## Tests

`tests/test_gildata_research_tools.py` (11): availability gating, metadata/vendor-tool wiring, envelope shape with `api_names` echo, row cap + clamp, vendor-error and missing-query/token paths, empty-rows-ok. Plus an env-schema default assert for the new URL field.

## Docs

agent SKILL.md manifest (tools table + count), data-routing SKILL.md capability rows, six-README tool lists (also fixes a PR-branch slip where the en README's *historical* v0.1.14 news line was bumped instead of the current heading), CHANGELOG, `.env.example`.

Stacked on #1474 → HK/CN-index → fund-nav — rebase onto main as each merges.


warren618 (changes requested):
Thanks. Two things, one of them CI: it adds four MCP tools without updating `mcp_server.py:9` ("Surfaces 74 tools"), the tool-count lines in the READMEs, or the ar/ko prose, and that is 9 of the 17 failures (the other 8 came from #1474 and are fixed on main now). Please rebase on main.
The second is in `grounding/evidence.py`, which is protected: the 14 NAV names go into the price-name table that applies to every tool, so any tool's `nav`/`净值` field becomes price evidence (a non-fund tool's `nav` now grounds a price). The tool itself only emits `unit_nav`, `accumulated_nav` and `adjusted_nav` (`fund_nav_tool.py:44-47`). Registering those three fields for `get_fund_nav` alone in `_REGISTERED_PRICE_INDICATORS` (`evidence.py:313`) is the shape that fits. The test should use different values for the three NAVs, and check that another tool's `nav` does not ground.

PR #1488, opened 2026-09-18
fix(agent): bound readonly replay after compaction

zeus229:
## Summary

* Restore successful readonly tool results from a run-scoped cache when context compaction removes the visible payload.
* Preserve normal freshness behavior while preventing unnecessary external refetches and replay-only planner loops.
* Bound replay recovery so compaction recovery cannot become an alternate unbounded research loop.

## Why

After context compaction, the agent can still know that an exact readonly tool call succeeded while no longer having the result available in the visible context.

For repeatable readonly tools such as web reads, requesting the same call again can then trigger another external fetch even though the successful result already exists from the same run.

This can:

* waste tool and network budget;
* refetch mutable data unnecessarily;
* make the model repeatedly request the same evidence after successive compactions;
* extend no-progress loops without producing new external information.

The intended behavior is narrower: if compaction removes an exact successful readonly result, restore that result from run-scoped memory and let the planner continue.

## Changes

* Add a run-scoped cache for successful eligible readonly results.
* Restore an exact cached result only after compaction makes the visible result unavailable.
* Require repeatable readonly tools to explicitly opt in with `replay_after_compaction=True`.
* Keep `no_cache=True` as an explicit freshness escape hatch that always bypasses replay.
* Keep deterministic tools on their existing deterministic cache path.
* Never replay mutating tools.
* Add planner steering after a restored result so the model continues instead of requesting the same call again.
* Add a run-scoped maximum number of replay recoveries.
* Record replay metadata in the trace.
* Add focused regression coverage for replay, compaction identity, repeatable reads, freshness bypass and no-progress behavior.

## Test Plan

* [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`)
* [x] New tests added (if applicable)
* [ ] Tested manually (describe below)

Validation was run from an isolated worktree at the exact PR head `8da5ac7a43900e36cf39d60fff49f905bb14598a`, based on upstream `e5f719567a0a8c943c08276b0295b95c572891fb`.

Focused replay/compaction tests:

```
44 passed, 0 failed, 0 skipped
```

Existing no-progress regression coverage affected during extraction:

```
2 passed, 0 failed
```

Additional checks:

* `black --check tests/test_readonly_replay_after_microcompact.py`: clean
* `ruff check tests/test_readonly_replay_after_microcompact.py`: clean
* `git diff --check e5f719567a0a8c943c08276b0295b95c572891fb...HEAD`: clean

A broader suite run before the final formatting-only commit produced `13779 passed, 99 skipped, 10 failed`. Eight failures reproduced identically on the upstream base and are unrelated to this PR. The two PR-attributable no-progress failures were caused by an out-of-scope recovery-message change introduced during extraction; that change was removed, and both regressions now pass.

The originating implementation was also exercised with isolated agent runs. No live broker, trading, payment, credential, or production flow was used.

Focused regression coverage includes:

* restoring lost readonly results without external execution;
* exact argument identity;
* micro- and auto-compaction behavior;
* repeatable tool behavior before compaction;
* `no_cache=True` bypass;
* trace and steering metadata;
* bounded no-progress behavior;
* bounded run-scoped replay recovery.

## Out of scope

This PR does not change:

* portfolio valuation;
* currencies or FX;
* broker connectors;
* order execution or trading behavior;
* authentication;
* MCP configuration;
* deployment behavior;
* downstream/custom portfolio integrations.

## Risk / rollback

The change affects agent context recovery for readonly tools after compaction.

The replay path is deliberately constrained:

* readonly tools only;
* exact call identity only;
* mutating tools are never replayed;
* deterministic tools remain on their existing cache path;
* repeatable tools require explicit opt-in;
* `no_cache=True` always requests a fresh external read;
* replay recovery is bounded for the lifetime of the run;
* replayed evidence does not count as a new external observation.

No broker-write, order, credential, payment, wallet or live-trading path is involved.

Rollback is a normal revert of this commit; that restores the previous behavior where compacted readonly results may need to be requested or fetched again.

## Checklist

* [ ] No changes to protected areas (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion
* [x] No hardcoded values (API keys, file paths, magic numbers)
* [ ] Code follows [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
* [x] Documentation updated (if user-facing change)


he-yufeng (commented):
Mechanism read end to end on 8da5ac7, plus a local run of the new and adjacent suites.

The design holds where I tried to poke holes:

- Replay only opens after `_unblock_lost_readonly_results` confirms the exact call's visible payload is gone, and only for keys still in the run-scoped cache, so a normal repeat with the payload still visible keeps hitting the tool (freshness preserved). `no_cache=True` bypasses replay at both gates, and repeatable tools need the explicit `replay_after_compaction` opt-in, so mutable read-through tools keep their refresh semantics by default.
- The anti-loop accounting is honest: a restore is not recorded as a new observation, and `mark_context_restored` grants one grace iteration per run instead of resetting the stall chain, so replay-only loops still converge to the no-progress limit even before the 6-recovery cap bites.
- Moving the duplicate-call block behind grounding authorization keeps the invariant that no cached path skips the identity gate, and the protected-set reason text after a restore is the right nudge.
- Planner-facing guidance was updated in both places that matter (the `no_cache` schema description and the web-reader SKILL.md), which is what actually changes model behavior.

Tests are real: the new file drives `_process_tool_calls` end to end and asserts the cache/ready/protected set transitions, not just return codes. I ran the four touched files plus the loop-adjacent sweep on this head (test_readonly_replay_after_microcompact, test_tool_progress, test_compaction_call_identity, test_agent_loop_repeatable_tools, and the agent_loop_* set: 124 passed, 0 failed, macOS Python 3.14, PYTHONPATH install).

One non-blocking edge to consider, not a defect: once the recovery budget is spent, a later compaction can leave a call that is neither replayable nor re-executable, and the duplicate block then tells the model to "use the previous result" for a payload it cannot see. That locked state is the intended loop-killer, but the skip reason could say the evidence is summarized instead of pointing at an invisible result. Fine to leave as is if the summary path is judged enough.

Good change, happy to see it land.

warren618:
Merged, thanks, and thanks @he-yufeng for the review. On top of the head, 7886b063 changes three things. The review probe reproduced each one on the merge before the change:

- Once the six replays were used up, a lost call stayed marked as succeeded. The next identical call was then refused with "already completed successfully. Use the previous result." for a payload compaction had already removed. That also hit a non-repeatable call lost for the first time after other calls had used the budget, and it is the locked state he-yufeng pointed at. Every lost call now reopens, as on main. The budget only decides, when the call is made, whether the result is restored or the call runs again. Identical re-runs still end at the no-progress limit.
- The cap was checked when keys were reopened, so one compaction with one recovery left restored all four keys it lost (9 of 6). It is now checked on each restore.
- A result cached before a write could be replayed after it, e.g. `factor_analysis` on a factor file the agent had just regenerated. Any successful non-readonly call now empties the replay cache.

A key is protected against an immediate re-fetch only after it has actually been restored. The locked-budget test now pins the re-run instead of the lock, and four new tests cover the rest. Each one fails when its rule is reverted.

PR #1495, opened 2026-09-19
Bound BaoStock loader failures and preserve fallback

### LINKED ISSUE #1492: [Bug] baostock loader has no timeout; baostock's client can block or spin forever when its server stops answering
- **Interface:** MCP Plugin
- **LLM Provider:** n/a
- **Version:** 0.1.15. The same code is on `main` at `e5f7195`. The baostock version is 0.9.3.

### Description

**What happened:**
baostock's client has no read timeout. `baostock/util/socketutil.py` opens its socket without `settimeout` (`SocketUtil.connect()`, lines 34-41). `send_msg` then reads like this (lines 67-73):

```python
while True:
    recv = default_socket.recv(8192)
    receive += recv
    if receive[-13:] == b"<![CDATA[]]>\n":
        break
```

- **Silent server.** If the server stops sending, `recv` blocks forever.
- **Closed connection.** If the peer closes the connection, `recv` returns `b""` forever and the loop spins at 100% CPU.
- **Errors swallowed.** `send_msg` also wraps everything in `except Exception as ex: print(ex)` (lines 89-90).

vibe-trading adds no deadline around any of this:

- **The loader.** `BaostockLoader.fetch` runs `bs.login()`, then `_fetch_one`, then `bs.logout()` with no time limit (`backtest/loaders/baostock_loader.py:86-110`).
- **The fallback chain.** `_fetch_via_chain` has no per-attempt timeout.
- **The MCP call.** MCP `get_market_data` goes through `ToolRegistry.execute` (`mcp_server.py:1832`, `src/agent/tools.py:97`), and that has no timeout either.

The backtest path is bounded, because `Runner(timeout=300)` runs loaders in a subprocess (`src/tools/backtest_tool.py:57`, `src/core/runner.py:431,636`). `get_market_data` is not bounded.

**What I expected:**
A stalled source should fail within a bounded time, and the fallback chain should move on.

### Steps to Reproduce

The fault is server-side and intermittent. We observed it on 2026-09-18 from about 23:46 (UTC+8), calling baostock 0.9.3 directly. This is the same package vibe-trading imports.

1. `bs.login()` returned `error_code "0"` in about 0.5 s.
2. `bs.query_history_k_data_plus("sh.600000", "date,close", start_date="2026-09-01", end_date="2026-09-18", frequency="d", adjustflag="3")` never returned.
3. Two such processes stayed at about 100% CPU for over 2 minutes, until we killed them. Their CPU time roughly equalled wall-clock time.
4. Fresh single sessions at 23:48 and 23:57 each exceeded a 30 s guard.
5. Between stalls the service answered normally, in 0.6-2.2 s. It later stalled twice more against a 60 s guard, the last time after midnight on a full-history query.
6. On 2026-09-19 at about 10:49 (UTC+8), login again succeeded and the same small query again exceeded a 30 s guard. So the stalls are not limited to late night.

### Error Logs

There are none. The call simply never returns.

### Suggested fix

- Bound every baostock call with a wall-clock deadline, and treat expiry as a loader failure so `_fetch_via_chain` continues.
- A worker subprocess or process pool is the robust option. A thread cannot be killed and would keep spinning.
- SIGALRM only works on the main thread. Our local script used it, raising a `BaseException` subclass so baostock's `except Exception` could not swallow it, and it cut stalls off at 60 s. That approach is not usable inside a threaded MCP server.
- Setting `socket.settimeout` on `baostock.common.context.default_socket` after login should cover the silent-server case. It would not cover the spin after a closed connection. That part needs a fix in baostock itself: break on an empty `recv`. Neither of these is tested.

### Observed vs inferred

- **Observed:** the stalls and the CPU spin in the baostock client itself, and the absence of any timeout in the code paths listed above.
- **Inferred, not observed:** an MCP `get_market_data` call routed to baostock during such a stall would hang indefinitely. We deliberately avoided triggering this.
- **How baostock gets reached.** An explicit `source="baostock"` reaches it directly. With `source="auto"`, baostock is fourth in the A-share chain after tencent, mootdx and eastmoney (`registry.py:159`). So it is only reached after the earlier sources fail. One such failure is Tencent answering HTTP 501 when it rate-limits, which we saw for over 20 minutes the same evening.

lorenzozanee:
## Goal

BaoStock 0.9.3 can block in a TCP read or spin after a server closes the connection. Because the loader ran in the caller process, the market-data fallback chain could not continue.

## Changes

Run the BaoStock session in a spawned child process with a bounded queue read. Terminate and kill the child when it exceeds the deadline, and convert child errors or non-zero exits to the existing empty-result fallback. Add regression coverage for silent and closed fake servers while preserving symbol conversion, caching, volume normalization, and logout behavior.

## Out of scope

This change does not contact live BaoStock services or modify the BaoStock client. Platform-specific behavior remains for upstream CI to validate.

## Test plan

- 83 targeted tests passed.
- Ruff, format checks, py_compile, and git diff checks passed.

## Risk and rollback

The loader now adds a spawned process and a 30-second deadline around BaoStock calls; fallback behavior remains an empty result for failed or stalled sessions. Revert this commit to restore the previous synchronous loader.

Fixes #1492


warren618 (changes requested):
Thanks — on a degraded server it does what it says: silent or spinning servers give up in 30 s and fall back, with no orphan left behind (real spawn path, baostock 0.9.3). Three things need changing before merge, measured on a healthy server this evening:
1. For 40 symbols over 2016-01-01..2026-09-18, main returned 40/40 in 252 s and this PR returned 0/40 at 30.02 s. The 30 s cap (`baostock_loader.py:24`) covers the whole batch, and the child reports only at the end, so everything already fetched is thrown away. Could the child stream each symbol's result as it arrives, with an inactivity deadline read from an env var?
2. If the calling process is SIGKILLed (which is what `subprocess.run(timeout=)` does to the backtest), the spawned child is reparented to PID 1 and keeps spinning at 100% CPU. The child should exit when `parent_process().sentinel` fires.
3. The tests force `fork` rather than `spawn`, and `test_daily_interval_still_attempts_login` lost its `login.assert_called_once()`. One test on the real spawn path, with that assertion restored, would cover it.

Issue #1491, opened 2026-09-19
[Bug] get_market_data provenance names the requested source (fallback_used=false) when that loader is unavailable and another loader served the data


jw232:
- **Interface:** MCP Plugin
- **LLM Provider:** n/a. This is the data layer, reproduced through MCP tool calls.
- **Version:** 0.1.15. The same code is on `main` at `e5f7195`.
- **Environment:** macOS (Darwin 25.6.0) / Python 3.14.5

### Description

**What happened:**
The optional `baostock` package was not installed. Even so, `get_market_data(codes=["688208.SH"], source="baostock")` succeeded, and its provenance said `source: "baostock"` and `fallback_used: false`. The bars cannot have come from baostock, because `import baostock` failed in the same venv. They were served by the A-share fallback loader, Tencent.

The mechanism:

1. `_fetch_via_chain` resolves each attempt with `loader_resolver(attempt_src)` (`src/market_data.py:309`). That defaults to `get_loader`, which calls `get_loader_cls_with_fallback` (`src/market_data.py:74-78`).
2. When the requested loader is unavailable, `get_loader_cls_with_fallback` silently returns a different class. It calls `resolve_loader(market)`, which picks the first available loader in `FALLBACK_CHAINS["a_share"]`, here Tencent. See `backtest/loaders/registry.py:438-495`, with `return type(fallback)` at line 495. On `main` this is lines 488-547.
3. `_fetch_via_chain` then records `used_source = attempt_src` (`src/market_data.py:328`), which is the requested name.
4. Provenance therefore gets `"source": used_source or src` and `"fallback_used": bool(used_source and used_source != src)`, which is `false` (`src/market_data.py:373-377`).
5. `adjustment` is looked up for the requested name through `price_caliber(used_source or src)`, not for the loader that actually served the data.

**What I expected:**
Provenance should report the loader that actually served the bars: `source: "tencent"`, `fallback_used: true`. It already does this when a loader is available but its fetch fails. In #1270, provenance correctly showed `yahoo` with `fallback_used: true`.

### Steps to Reproduce

1. Use an environment where `baostock` is not installed. It is only in the optional `ashare` extra.
2. Through the MCP plugin, call `get_market_data` with `codes=["688208.SH"]`, `source="baostock"` and a short date range, with provenance enabled.
3. Provenance shows `source: "baostock"` and `fallback_used: false`.
4. Install `baostock==0.9.3` and repeat. No restart is needed.
[...]

PR #1482, opened 2026-09-18
feat(i18n): add Indonesian localization

Yoruxyv:
﻿## Summary

Adds Indonesian (`id`) localization support for the Vibe-Trading Web UI and project documentation.

### Changes

- add `frontend/src/i18n/locales/id.json`
- register Bahasa Indonesia in the frontend i18n configuration
- update i18n locale-order / parity tests
- add `README_id.md`
- add Bahasa Indonesia to the language selector in the existing README translations
- preserve placeholders, commands, paths, provider names, broker names, symbols, and other technical identifiers

### Translation approach

The Indonesian localization aims for natural wording for Indonesian technical and trading users rather than translating every English term literally.

Some established trading and technical terms are intentionally kept in English where that usage is more familiar or precise for Indonesian traders, including terms such as `long`, `short`, `drawdown`, `P&L`, `Sharpe`, `benchmark`, `backtest`, `options`, `Greeks`, `Bid`, `Ask`, `margin`, and `leverage`.

I also reviewed the terminology with a trader friend to sanity-check which expressions Indonesian traders commonly use in practice.

Security-sensitive live-trading terminology and semantics were kept aligned with the original English strings rather than rewritten purely for stylistic localization.

## Validation

- `npm run test:run -- src/i18n/__tests__/i18n.test.ts`
- `npm run test:run`
- `npm run build`
- `git diff --check`

All passed locally.

## Screenshots

image

PR #1497, opened 2026-09-19
fix(loaders): Tencent's A-share caliber is dividend-additive, and its HK series is raw

### LINKED ISSUE #1493: [Bug] Tencent A-share prices (fqkline qfq) adjust cash dividends additively — returns are distorted and old prices can go negative, yet the caliber is labelled split_dividend
- **Interface:** MCP Plugin
- **LLM Provider:** n/a

### Description

**What happened:**

**What I expected:**

|---|---|---|---|

**Effect on statistics:**

|---|---|---|
| 600519.SH | 18.0% | 21.9% |

|---|---|---|

```sh
```

### Error Logs

### Suggested fix

### Observed vs inferred

### Related

cgycorey:
## Summary

- `tencent`'s A-share `qfq` subtracts cash dividends from the price **level** rather than scaling by a ratio, so calling it `split_dividend` was wrong in a way that made the mixed-caliber warning blind. It now carries its own caliber, `split_dividend_additive`, which participates in the comparison.
- `tencent`'s HK series is **unadjusted** despite the `qfq` request parameter, so HK now stamps `raw`.
- Separately, the warning only ever fired on a **mix** of calibers, so a run that served an additive series from a single source stayed silent while still reporting daily moves that are not total returns. `additive_caliber_warning` now covers that case. **This third piece is folded in from #1504 — see the attribution section below.**

Fixes #1493

## Why

The report's two halves are both real, and I reproduced each on `main` at `e5f71956`.

**The additive structure.** Within a segment, `qfq = raw + c` for a constant `c`. For 600519.SH over 500 bars (2024-08-28 → 2026-09-18) I compared the two candidate invariants:

| quantity | distinct values over 500 bars |
|---|---|
| `qfq − raw` | **5** → `0.00, −28.02, −51.98, −79.58, −103.46`, each holding ~120 bars, stepping only at dividend dates |
| `qfq / raw` | **415** |

A multiplicative adjustment is the mirror image — constant ratio per segment, varying difference. So the return distortion the report describes is real.

**The label collision.** `_COMPARABLE_CALIBERS` bucketed `tencent` with `baostock` (`adjustflag="2"`, multiplicative), so `mixed_caliber_warning` returned early on `len(by_caliber) < 2` and could not fire for exactly the mix it exists to catch. `tencent` also heads the `a_share` chain (`registry.py:177`), so this is the default A-share path.

**Why the single-source case needed its own warning.** Correcting the label fixes the *mix*, but not the more common case: a run that resolves entirely to `tencent` is now internally consistent and still serves non-total-return bars. `mixed_caliber_warning` is structurally the wrong home for that — it compares sources to each other, so by construction it cannot speak about a basket with one caliber. The additive property is what needs reporting, not the disagreement, so it gets a second predicate over the same stamps.

**The HK half, which the report did not cover.** Tencent serves **no adjusted series for HK at all**: its fqkline reply carries only a `day` key for HK symbols — never `qfqday` or `hfqday` — so the loader's `qfqday or day` fallback silently serves unadjusted bars and the `,qfq` request parameter changes nothing. The actions are real: eastmoney's adjusted HK series differs from its raw one over the same window (checked on 00939.HK / 00700.HK), so this is a gap in what the source serves, not an empty window. HK therefore stamps `raw`, following the existing `("tushare", "hk_equity")` override pattern.

*(An earlier revision of this PR said "its qfq and hfq replies are identical to its own day series". That was an artifact of my probe falling back to `day` for all three modes; the conclusion holds, the stated evidence did not. Corrected in `150ef942`, and the code comment now says what was actually measured.)*

Before → after, on the same code paths:

```
BEFORE                                              AFTER
price_caliber(tencent, a_share) = split_dividend    split_dividend_additive
price_caliber(tencent, hk_equity) = split_dividend  raw
mixed tencent+baostock basket -> None (silent)      WARNS: mixed price calibers … split_dividend (000001.SZ); split_dividend_additive (600519.SH)
pure tencent basket -> None (silent)                WARNS: additive price adjustment in this run: 600519.SH (tencent) … not total returns
```

## Changes

- `registry.py`: `tencent` → `split_dividend_additive`; `("tencent", "hk_equity")` → `raw`; `split_dividend_additive` joins `_COMPARABLE_CALIBERS`; `price_caliber`'s docstring lists the value.
- `registry.py`: new `_ADDITIVE_CALIBERS` and `additive_caliber_warning(stamps)`, reporting which symbols were served by an additive source and naming the consequence (`not total returns`). Long lists truncate at four with `+N more` rather than growing the run card without bound.
- `runner.py`: `fetch_data_map` now composes both predicates instead of picking one, so a basket that is simultaneously mixed *and* additive reports both. Neither warning can shadow the other.
- `mcp_server.py` and `src/tools/market_data_tool.py`: both enumerate the calibers a caller reads from `_provenance.adjustment` — the tool **schema description** an LLM caller actually sees was missed on the first pass and is now updated too (`150ef942`).
- `tests/test_price_caliber.py`: new tests for the A-share label, the HK override, the warning firing on an additive+multiplicative basket, the provenance field for both markets, a guard that a single-caliber additive basket stays silent, and the additive predicate itself — that it fires on a lone source, stays silent on multiplicative and raw, truncates a long served list, and composes with the mixed warning.

No loader behaviour changed — the served series are what they always were; only the labels (and therefore the warnings and the provenance field) change.

## Folded in from #1504, with credit

#1504 (`he-yufeng`, branch `fix/tencent-additive-caliber`, `261405d90`) opened while this PR was in review and reaches the *same* conclusion on the A-share label independently — including choosing the identical caliber name `split_dividend_additive`. That convergence is worth noting: the name was not coordinated, and two people reading the same arithmetic landed on it.

The two PRs were complementary rather than duplicative, so rather than let a maintainer pick one and lose the other's half, I ported #1504's missing piece here (`e2bb00e1`) so this PR strictly supersedes it:

| | #1497 (this PR) | #1504 |
|---|---|---|
| `split_dividend_additive` label | yes | yes |
| HK `raw` half of #1493 | yes | no |
| single-source additive warning | **yes** (ported) | yes |
| both warnings composed when both apply | **yes** | n/a (no mixed fix) |
| `runner.py` wiring | yes | yes |

Commits carry `Co-authored-by: Yufeng He <40085740+he-yufeng@users.noreply.github.com>`, and the provenance is stated in the commit message rather than only here. Where my port differs from #1504 as written, deliberately:

- The runner-level tests live in `tests/test_price_caliber.py` next to the existing `fetch_data_map` caliber tests, rather than in `tests/test_market_data_serving_source.py`, which is scoped to #1491's substitution reporting.
- Added a test pinning that a basket which is both mixed and additive reports **both** warnings — the composition is a real branch in `runner.py` and the most likely thing to regress.
- Added the `>4`-symbol truncation case.

I have left a comment on #1504 proposing it be closed in favour of this one. That is the maintainer's call, not mine — if you would rather land #1504 and have me rebase the HK half onto it, say so and I will.

## Test Plan

- [x] **Watched the new tests fail first** (TDD): the three behaviour tests and the HK provenance test failed against unmodified `registry.py` — the two labels asserted the old strings and `mixed_caliber_warning` returned `None` — then passed after the table change. The additive tests were likewise written against the pre-port `runner.py`, where they failed on the missing warning.
- [x] Two existing pins encoded the old label and were deliberately updated: the parametrized `("tencent", …)` row, and `test_every_chain_source_resolves`, whose allowed-vocabulary set is what caught the new caliber. A reviewer should check exactly these two.
- [x] `tests/test_price_caliber.py`: **38 passed** (was 33 before the port; the five added tests are the additive warning's).
- [x] `tests/test_price_caliber.py` + `tests/test_market_data_serving_source.py`: **45 passed**.
- [x] The 42 test files referencing the caliber machinery: **1615 passed, 5 skipped**.
- [x] Neighbouring surfaces (`test_registry`, `test_market_data_tool`, `test_market_detection`, `test_cross_market_annualization`): **210 passed**.
- [x] MCP tool surfaces — every `test_mcp_*.py`, both `test_registry*`, and the caliber files, minus the two files already red on `main`: **278 passed**.
- [x] Full suite, run in this worktree: **13847 passed, 34 skipped, 14 failed**. I verified **all 14 fail identically on a pristine `origin/main` worktree** (`/tmp/vt-main-check2`, detached at `b705d188`, importing its own `registry.py`, which has no `additive_caliber_warning`), so this branch adds none of them: six `test_readme_counts` variants (README tool count 107 vs 108 locally, never red in CI), four `test_mcp_oauth_schema`, `test_oauth_token_cache::test_cache_expiry_surfaces_reauth_no_silent_stale_call`, `test_mcp_goal_session_isolation::test_two_concurrent_clients_do_not_share_a_goal_session`, two `test_mcp_client_adapter`, and the 35 ms signal-alignment timing test, which fails when the machine is loaded. Collected-test delta vs `main`: 13880 → 13890, exactly the ten tests added here. CI on the head is the authority for the `test / Python` legs.
- [x] `ruff check` clean. `black --check` reports the same hunks on this branch as on `main` for `test_price_caliber.py` (2) and `mcp_server.py` (30), none of them mine — no new format drift. The pre-existing non-`ruff format`-clean `caliber_stamps` comprehension in `runner.py` is left alone deliberately; it already fails on `main`, and reformatting it here would add unrelated churn to a review.

## Checklist

- [x] No hardcoded values — the calibers are the existing vocabulary plus one documented value.
- [x] Code follows CONTRIBUTING.md guidelines.
- [x] Documentation updated — the two docstrings that enumerate price calibers.

## Adversarial re-check, and what it changed

I re-attacked this branch after opening it. Three things came out, all disclosed here rather than left for a reviewer to find:

**1. My HK evidence was stated wrongly** (mechanism only; the label is unchanged). Corrected above and in `150ef942`.

**2. A doc surface was missed:** the `get_market_data` *schema description* in `src/tools/market_data_tool.py` enumerates the calibers and did not name the new one. Now fixed; every MCP-tool-surface test still passes.

**3. A user-visible consequence worth stating explicitly.** Because tencent HK now stamps `raw` and `eastmoney`/`yahoo` HK stamp `split_dividend`, an HK run that falls back down the chain per symbol will now emit a mixed-caliber warning where it was previously silent. That is the correct behaviour — those baskets really do mix an unadjusted series with an adjusted one, which is the case the warning exists for — but it is new output in previously-quiet runs.

**4. A user-facing caveat string is now slightly imprecise, and I left it alone on purpose.** `caliberCaveat` in the settings UI (8 locale files) says *"Yahoo serves raw (unadjusted) OHLC while Tencent/Eastmoney/Tushare/BaoStock serve adjusted prices"*. After this PR that is true for Tencent's A-shares only — its HK rows are raw. The string was already a simplification before this PR (Tushare's HK path stamps `raw` too), and `#1474`/`#1485`/`#1482` all edit those locale files, so I did not add to that pile. Happy to correct it in a follow-up once they land.

**5. Two chain members are `unknown`, so the warning still cannot see them at all.** In the A-share chain `mootdx` is unmeasured, and in HK `futu` is — `unknown` is deliberately excluded from mixed-caliber comparison, so a tencent→mootdx fallback pair stays silent, as it did before this PR. The same blind spot applies to the new additive predicate, for the same reason: an unmeasured source is not asserted to be additive. Worth knowing when reading the fix: it closes the tencent gap, not the coverage gap.

**Two limitations I am deliberately not fixing here, both pre-existing:**

- `_detect_market` does not map non-canonical HK spellings (`hk00939`, `HK.00939`, a bare `00700`) to `hk_equity`, so those stamp the A-share default. They were already stamped as adjusted before this PR; the gap is in symbol detection, not in the caliber table, and fixing it belongs with the routing code.
- For markets tencent cannot serve (its loader declares `markets = {"a_share", "hk_equity"}`), the per-source default is the A-share value. That matches the existing encoding for `tushare` (per-source default plus an `hk_equity` override), and the loader's declared market set makes those pairs unreachable in practice.

## Open to redirection

I kept the served series as they are and fixed the labelling and the reporting. If you would rather the loader serve a multiplicative series for A-shares (so `split_dividend` stays the only dividend caliber, and neither new warning is needed), that is a larger change and I will do it instead — say so and I will rework this PR rather than argue for it.


he-yufeng (commented):
Maintainer review of the superset claim, line by line:

- The A-share half matches what #1504 does — same `split_dividend_additive` label, same `_COMPARABLE_CALIBERS` addition, same `additive_caliber_warning` + runner-side join of both warnings. Convergent naming aside, the mechanics agree.
- The HK override holds up: `tencent_loader.py:161` does `qfqday or day`, so an HK reply without `qfqday` silently serves raw bars; stamping `("tencent", "hk_equity"): "raw"` is the honest caliber for it, and the 00939.HK/00700.HK measurements say the actions exist but are not served.
- The both-warnings composition test and the >4-symbol truncation case are the right extra pins.
- CI: 10/10 pass on the current head.

Landing this one and closing #1504 in its favor — the HK half is real coverage mine lacked, and folding the warning in with the credit trailer is generous and correct. Thanks for doing the merge work on your side. Leaving the merge click to the project owner.

warren618:
Merged, thanks — and thanks @he-yufeng for folding #1504 into it. One sibling on top (b658a206): Eastmoney's `fqt=1` is additive with the same offsets (600519.SH: five values over 500 bars against 367 distinct ratios; 601398.SH five against 253), and AKShare's `stock_zh_a_hist(adjust="qfq")` calls that endpoint, so both A-share paths are labelled `split_dividend_additive` too. Otherwise a Tencent + Eastmoney basket warned that two identical scales differ, and an Eastmoney-only run got no additive warning. The labels still don't correct the series, so I've opened a follow-up for deriving a multiplicative adjustment from the raw bars (#1541).

PR #1498, opened 2026-09-19
fix: report the serving loader in market data provenance

### LINKED ISSUE #1491: [Bug] get_market_data provenance names the requested source (fallback_used=false) when that loader is unavailable and another loader served the data

lorenzozanee:
## Summary

- Report the loader that actually served market data when the requested loader is unavailable.
- Mark fallback results and their adjustment metadata from the serving loader.

## Why

When an unavailable requested loader resolved to another loader, provenance named the requested source instead of the loader that returned the bars. This made fallback data appear to come from the wrong provider.

Fixes #1491

## Changes

- Record the resolved loader's canonical name for group and per-symbol provenance.
- Add a deterministic regression test for baostock resolving to tencent.

## Test Plan

Focused validation:

- `pytest -q tests/test_market_data.py tests/test_get_market_data_provenance.py` (70 passed)
- `ruff check src/market_data.py` (passed)
- `git diff --check` (passed)

The full suite was not run because optional dependencies are unavailable in the local environment; upstream CI remains required.

## Checklist

## Risk and Rollback

Low risk: the change only affects provenance metadata for loader fallback results and adds a regression fixture; it does not perform broker writes or live network validation. Revert the commit to restore the previous provenance behavior.


warren618:
Thanks — the same change landed on main as 335739e8 for #1491 while this was open (`market_data.py` uses `loader_cls.name` for `used_source` and `symbol_sources`), with `test_market_data_serving_source.py` covering the same substitution. Your test passes against main unchanged, so I'm closing this as superseded.

PR #1499, opened 2026-09-19
fix(data): distinguish Tencent additive price adjustments

### LINKED ISSUE #1493: [Bug] Tencent A-share prices (fqkline qfq) adjust cash dividends additively — returns are distorted and old prices can go negative, yet the caliber is labelled split_dividend

lorenzozanee:
## Summary

Tencent A-share qfq prices apply cash dividends additively, but the loader labeled them as multiplicative `split_dividend` data. That made provenance inaccurate and allowed mixed-caliber baskets to miss the semantic difference.

## Changes

- Label Tencent A-share prices as `split_dividend_additive`.
- Keep the Tencent HK fixture fallback on the `raw` caliber when qfq data is absent.
- Include the new caliber in mixed-caliber warnings and public market-data descriptions.
- Add focused regression coverage for source and market overrides, warnings, and provenance.

## Test Plan

- `PYTHONPATH=agent pytest agent/tests/test_price_caliber.py agent/tests/test_tencent_loader.py agent/tests/test_market_data_tool.py -q` (68 passed)
- Ruff check and format checks passed for all changed Python files.
- Compile checks and `git diff --check` passed.

## Scope and Risk

This changes semantic labeling and warning visibility only; it does not reconstruct numeric prices into total-return series. No live financial-data calls were made. The commit can be reverted independently if the upstream caliber interpretation changes.

Fixes #1493


warren618:
Thanks. #1497 carried the same relabel plus the single-source additive warning and the HK override, and your 30 caliber tests pass against it; the two conflict in four files, so I merged #1497 and am closing this one. Eastmoney and AKShare A-share are now labelled additive too.

PR #1504, opened 2026-09-19
fix(market-data): label Tencent's additive adjustment honestly

### LINKED ISSUE #1493: [Bug] Tencent A-share prices (fqkline qfq) adjust cash dividends additively — returns are distorted and old prices can go negative, yet the caliber is labelled split_dividend

he-yufeng:
## Summary

- Tencent's `fqkline qfq` series is relabelled from `split_dividend` to a new `split_dividend_additive` caliber: #1493 measured that cash dividends are added back as a flat offset (`hfq = k*raw + c` between corporate actions), so daily moves are not total returns.
- A new `additive_caliber_warning` fires whenever a run serves an additive series, even single-source, telling the user the series suits price charts but not return math.
- The additive caliber joins the mixed-caliber comparison, so a tencent+baostock basket warns again instead of reading as one comparable label.

## Why

Closes #1493. The reporter measured the additive structure by least-squares fits against raw closes, cross-checked with iFinD's two adjustment conventions, and showed a buy-and-hold 600519.SH 2019 backtest reporting +167%/+184% where the multiplicative total return is +100.4%. The `split_dividend` label made all of that silent. This PR takes the issue's "at minimum, relabel it" half; reconstructing a multiplicative series from Tencent data (the c-estimation chaining the reporter validated to ~6 bp/day against baostock) is left as a follow-up, since shipping a heuristic reconstruction as default would be its own silent-numbers risk.

## Changes

- `registry.py`: tencent calibre value + evidence comment; `split_dividend_additive` joins `_COMPARABLE_CALIBERS`; new `additive_caliber_warning(stamps)`; `price_caliber` docstring updated.
- `runner.py`: the run-level caliber warning now combines mixed and additive warnings.
- `mcp_server.py`: `get_market_data` doc lists the new value and says what it means for return math.
- Tests: updated the three pins that hard-coded tencent's old label; new tests for additive/mixed warning behaviour and a runner-level tencent serve.

## Test Plan

- [x] Existing tests pass (`pytest agent/tests --ignore=agent/tests/e2e_backtest -q`: 13784 passed, 98 skipped)
- [x] New tests added (additive warning fires single-source, stays silent otherwise, additive+multiplicative mixes warn, runner-level serve warns)
- [x] Tested manually: red-green — `price_caliber("tencent", "a_share")` returns the old `split_dividend` on main and `additive_caliber_warning` does not exist there; the runner-level test warns with `additive price adjustment ... 600519.SH (tencent)` on this branch.

Not run: live fetch against Tencent's endpoint (the reporter's curl in #1493 covers the upstream payload shape; this change is metadata + warnings, not parsing).

## Checklist

- [x] Documentation updated (the `get_market_data` tool doc now documents the additive caliber)

Risk / rollback: low. The mixed-caliber warning gains one more firing case (tencent mixed with multiplicative sources), and additive serves now carry an extra warning line; no behaviour change to data or accounting. Revert this commit to restore the previous labels.

Prepared with AI assistance (Kimi Code); every changed line reviewed by the submitter, tests run as listed above.


cgycorey:
Thanks @he-yufeng — this is a good fix, and the diagnosis matches mine on the arithmetic.

Worth noting up front: you and I chose the **same caliber name, `split_dividend_additive`, independently**. I had not seen this PR when I wrote #1497, and I assume the reverse is true — two people reading the same `qfq − raw` invariant landed on the same word for it. That is a decent signal the name is the obvious one.

I've folded the piece of this PR that #1497 was missing into #1497, so I want to be explicit about what that means rather than quietly absorb your work:

**What I ported** (commit [`e2bb00e1`](https://github.com/HKUDS/Vibe-Trading/pull/1497/commits/e2bb00e1aa438938b82f3c2a8738cde0541ebd26), `Co-authored-by: Yufeng He`):

- Your core insight that `mixed_caliber_warning` is structurally the wrong home for this — it compares sources to each other, so with a single source there is no mix to detect and the warning can never fire. #1497 had fixed the *label* but left a tencent-only run silent, which is the ordinary case since `tencent` heads the `a_share` chain. That was a real gap in my PR and your framing is what closed it.
- An additive predicate over the same stamps, reported alongside the mixed warning rather than replacing it.
- The four-symbol truncation, so a long basket doesn't grow the run card without bound.

**Where my port differs from this PR as written** — deliberately, and you may disagree:

- The runner-level tests live in `agent/tests/test_price_caliber.py` beside the existing `fetch_data_map` caliber tests, rather than in `test_market_data_serving_source.py`, which is scoped to #1491's substitution reporting.
- I added a test pinning that a basket which is **both** mixed and additive reports both warnings, since composing the two predicates is a real branch in `fetch_data_map` and the likeliest thing to regress.
- I added the `>4`-symbol truncation case.

**Why I did not simply withdraw #1497 in favour of this one:** the two PRs cover different halves of #1493. This PR relabels A-shares; it does not touch the HK half, where `tencent` serves **no adjusted series at all** — its `fqkline` reply carries only a `day` key for HK symbols, never `qfqday`/`hfqday`, so the loader's `qfqday or day` fallback silently serves unadjusted bars and the `,qfq` request parameter changes nothing. #1497 stamps `raw` for `("tencent", "hk_equity")` and has the HK evidence. Neither PR alone closes the issue; #1497 plus your warning now does.

So the overlap is now one-directional: **#1497 is a strict superset of this PR** — same label, same single-source warning, plus the HK fix and the both-warnings composition.

**This is entirely the maintainers' call, not mine.** If you'd rather this PR land and have me rebase the HK half on top of it instead, I'm happy to do that — say the word and I'll close #1497 and rebase. I only picked the direction I did because #1497 already carried the HK work and its test suite was further along.

Either way the credit for the single-source warning is yours, and the commit trailer records it.

he-yufeng:
Closing in favor of #1497, which covers the same A-share relabel + single-source additive warning and adds the HK half I did not have (tencent serves no adjusted series there at all). Credit where due: @cgycorey did the superset merge and kept the attribution. The issue stays open until #1497 lands.

PR #1505, opened 2026-09-19
fix(portfolio): include Binance assets after clock skew

goatyyc:
## Summary

- retry Binance signed read-only calls once after recalibrating clock skew on `-1021` ahead-of-server responses
- use a credential-free Binance client for public quotes, instrument search, and historical bars
- mark snapshots with unpriced holdings incomplete and exclude legacy mislabeled snapshots from complete history

## Why

A Binance Spot portfolio refresh could read a Simple Earn BTC balance successfully, then fail while pricing `BTC/USDT` because ccxt loaded signed currency metadata through the credentialed client. The position remained visible but unpriced, so its value was silently excluded from account and portfolio totals while the snapshot was still marked complete.

Binance also rejects signed timestamps more than one second ahead of server time regardless of `recvWindow`, so a sleeping or slightly skewed workstation could intermittently fail otherwise valid read-only account requests.

## Changes

- add a bounded clock-resync retry for Binance balance and Simple Earn reads; unrelated errors and a second `-1021` still fail normally
- strip credentials from public Binance market-data clients so ticker and catalog loading cannot trigger signed SAPI requests
- require every reported position to be priced before a portfolio snapshot is complete
- filter older `complete=true` snapshots containing unpriced positions from complete-only history
- add regression coverage for clock resync, retry bounds, public quote isolation, Simple Earn reads, and incomplete valuation history

Out of scope: order placement, cancellation, API credential storage, and any broker-write behavior.

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --ignore=agent/tests/test_e2e_harness_v2.py --tb=short -q`): 13,583 passed, 98 skipped
- [x] New tests added
- [x] Tested manually: credential-free `BTC/USDT` quote returned a positive price; a read-only portfolio refresh priced the previously unpriced Binance BTC position and produced zero unpriced positions
- [x] `ruff check agent/src/portfolio/service.py agent/src/portfolio/store.py agent/src/trading/connectors/binance/sdk.py`
- [x] `git diff --check`

## Checklist

- [ ] Documentation updated (not needed; this restores documented portfolio completeness and Binance valuation behavior)

## Risk and rollback

The network behavior change is limited to read-only Binance calls. Public market data no longer receives API credentials. The signed-read retry is bounded to one retry and only matches Binance `-1021` ahead-of-server responses. Reverting the single commit restores the previous behavior.


warren618 (changes requested):
Thanks — the Binance part is sound: the retry fires only on `-1021`, the offset doesn't persist, and public reads use an unauthenticated client. The completeness rule needs narrowing, though. Only Binance builds its account total from priced positions (`service.py:295-296`); every other broker reports its own total, which already includes an unpriced holding. With this PR, an IBKR account reporting 1,000 with one unpriced position becomes `complete=False` and drops out of history, and every Robinhood snapshot does too, because Robinhood positions are never priced yet. Could it mark a snapshot incomplete only when an account's total is built from priced positions, and apply the same rule in the store filter?

PR #1508, opened 2026-09-19
fix(email): fix SPF DKIM spoofing check

Shizoqua:
## Summary
- SPF and DKIM verification now only trusts the topmost Authentication-Results header

## Why
The check OR'd across every Authentication-Results header in the message, so an attacker could add their own passing header and bypass the anti-spoofing protection the config flags claim to provide.

## Changes
- Only read the first Authentication-Results header
- Added regression tests for a forged header and a forged header below a real one

## Test Plan
- [x] Existing tests pass (`pytest tests/test_email_authentication_results.py tests/test_channels_runtime.py`)
- [x] New tests added
- [x] Tested manually

## Checklist


warren618 (changes requested):
Thanks for digging into this. I probed it against main, and it moves the trust assumption rather than removing it:

| case | main | this PR |
|---|---|---|
| provider's `fail` header on top, forged `pass` below | pass | reject (fixed) |
| a forged header alone (provider stamps none) | pass | pass |
| attacker's own domain passes SPF/DKIM, `From:` forged | pass | pass |
| `pass` text inside the provider's own top header via the MAIL FROM local part | pass | pass |
| self-hosted OpenDKIM/OpenDMARC writing separate headers (legitimate) | pass | **reject: all mail refused under the default `verify_*=True`** |
| forwarded legitimate mail | pass | SPF fails |

The identity is still the `From:` header (`email.py:461`) checked against `is_allowed`, with nothing tying the SPF/DKIM domain to it, and the regex matches comments and property values. The new test `test_forged_header_alone_is_not_trusted` asserts `(True, True)`. A fix that holds would trust only headers stamped by a configured `trusted_authserv_id` (RFC 8601), parse `method=result` tokens instead of matching the whole header, and require `dmarc=pass` or an SPF/DKIM domain aligned with the `From:` domain. Would you rework it that way?

PR #1509, opened 2026-09-19
fix(quantlib): fix accrued interest edge case

Shizoqua:
## Summary
- Accrued interest now returns the full coupon at settlement even when the 30/360 day count collapses the period to zero

## Why
30/360 clips the day of month, so date pairs like Jan 30 to Jan 31 year fraction to zero despite being distinct ordered dates, and the function silently returned zero accrued interest instead of the full coupon.

## Changes
- Check settlement against the coupon dates directly instead of relying on the day count derived period
- Added regression test for a degenerate 30/360 period

## Test Plan
- [x] Existing tests pass (`pytest tests/quantlib/test_fixedincome.py`)
- [x] New tests added
- [x] Tested manually

## Checklist


warren618:
Merged, thanks. One regression caught before merge (7b73da56): the new endpoint returns run before `year_fraction`, which is where `day_count` was validated, so `day_count="bogus"` returned 0.0 or a full coupon on a coupon date and raised only mid-period. It is now checked up front.

PR #1510, opened 2026-09-19
feat(portfolio): generalize ISO-4217 currencies with isolated FX failures

### LINKED ISSUE #1445: [Feature] Generalize portfolio currencies beyond USD / HKD / CNY

zeus229:
## Summary

Replacement for #1446, incorporating the review feedback from @warren618 and the follow-up review by @he-yufeng.

This keeps the original #1445 direction: currency identity is generic ISO-4217, while actual FX conversion remains explicit and fail-closed.

## Review fixes included

- Isolate valuation failures per source, so an unrated source becomes an `error` without discarding healthy sources.
- Reject `display_currency` at save time when the production rates map cannot value it.
- Validate FX rates as finite and > 0, naming the invalid pair.
- Align the ISO-4217 set with the review: remove CNH/withdrawn codes and add XCG/ZWG.
- Explicitly cover valuation version 3 behavior.
- Make the frontend consume generic display values without mislabeling a USD fallback.
- Correct the original KRW wording: the old silent `fx=1` problem affected account-level balances, while unsupported position currencies already failed closed.

## Key regression case

A mixed portfolio with a healthy IBKR USD source plus an unrated Longbridge SGD source now produces:

- IBKR included in totals;
- Longbridge marked `error` with SGD named;
- `complete=False`;
- no corruption of healthy-source totals.

## Validation

Reported on this branch:

- backend focused/regression tests: **98 passed**
- Black: clean
- Ruff: clean
- `git diff --check`: clean
- TypeScript `tsc -b`: clean
- frontend build: clean
- Portfolio frontend tests: **9/9 passed**
- full frontend suite: **631 passed**, with 1 unrelated pre-existing locale-sensitive failure in `options.test.ts`

The branch is based on the current upstream main used for the follow-up and is 19 commits ahead / 0 behind that base.

Closes #1445.
Supersedes #1446.

Thanks @warren618 for the review cases and @he-yufeng for reviewing the v2 branch and requesting the replacement PR.


he-yufeng (commented):
Confirming as a maintainer that this is the agreed v2 replacement for #1446, which I've closed in its favor. I read the v2 branch end to end against every point of warren618's review before it became this PR; the isolated-per-source failure isolation and explicit fail-closed FX conversion are the right shape. CI hasn't reported on this branch yet — flagging so the owner knows the merge gate is CI, not content. Thanks for carrying it, @zeus229.

warren618:
Merged, thanks, and thanks @he-yufeng for the review trail from #1446. Checked both ways: a Longbridge SGD balance is now an error for that source ("FX conversion is not available for: SGD") while IBKR's total stays in; a healthy HKD portfolio values exactly as before. The CHANGELOG notes that valuation version 3 hides history stored under version 2.

PR #1512, opened 2026-09-19
fix(tools): restore dragon-tiger seat reports

### LINKED ISSUE #1502: [Bug] get_dragon_tiger returns "seats": [] for every security — Eastmoney rejects the seat report name RPT_BILLBOARD_TRADEDETAIL (9501 报表配置不存在), and the rejection is swallowed as an empty result
- **Interface:** MCP Plugin
- **LLM Provider:** n/a. This is the data layer, reproduced through MCP tool calls and by calling `DragonTigerTool().execute` directly.
- **Version:** 0.1.15. `agent/src/tools/dragon_tiger_tool.py` on `main` at `2dfbc03` is byte-identical to the 0.1.15 wheel, so the line numbers below apply to both.

File paths below are relative to `agent/` in the repository.

### Description

**What happened:**
`get_dragon_tiger(date="2026-08-17", code="002156.SZ")` returns `ok: true`, `source: "eastmoney"`, one appearance, and `"seats": []`. The tool description says to supply `code` "to also get that security's ranked top buy/sell brokerage seats". Eastmoney does serve seat rows for that security and date, 5 buy and 5 sell, from a different pair of reports.

The mechanism:

1. **The seat report name is rejected.** `_collect` reads seats from `_SEAT_REPORT = "RPT_BILLBOARD_TRADEDETAIL"` (`src/tools/dragon_tiger_tool.py:30`, used at lines 264-269). Eastmoney answers that request with HTTP 200 and:
   ```json
   {"version":null,"result":null,"success":false,"message":"报表配置不存在,RPT_BILLBOARD_TRADEDETAIL","code":9501}
   ```
   The message means "report configuration does not exist".
   - The answer is the same with no `filter` and no sort at all, so it does not depend on the code or the date.
   - It is the same with a `Referer` header, and on the other host, `datacenter.eastmoney.com/securities/api/data/v1/get`.
   - A made-up name such as `RPT_THIS_DOES_NOT_EXIST_XYZ` gets the same code and the same message, with its own name in it.
2. **The rejection is swallowed.** `_fetch_report` only looks at the shape. `result` is not a dict, so it returns `[]` (lines 104-106). `eastmoney_client.get_json` raises only on network, HTTP or JSON errors. Nothing is raised, so the tushare fallback (lines 210-228) is not reached, and the envelope says `ok: true` with no warning.
3. **A real "no rows" answer looks different.** For a date with no board entry (I used Sunday 2026-08-16), Eastmoney returns `"code":9201,"message":"返回数据为空"`. The tool maps both answers to `[]`, so a caller cannot tell "this security had no seats" from "our query is invalid".
4. **Renaming the report is not enough.** The live seat reports are `RPT_BILLBOARD_DAILYDETAILSBUY` and `RPT_BILLBOARD_DAILYDETAILSSELL`. Their rows carry `OPERATEDEPT_NAME`, `BUY`, `SELL`, `NET`, `EXPLANATION` and `TRADE_ID`. They have no `SIDE` and no `RANK` column, which `_seat_row` reads (lines 146 and 150). Sorting one of them by `RANK` is rejected with `RANK排序列不存在` (code 9501). With only `_SEAT_REPORT` changed in memory to the BUY report, the tool returned the 5 buy-side seats only, each with `"side": null` and `"rank": null`.
5. **Why the tests pass.** `tests/test_dragon_tiger_tool.py` mocks `eastmoney_client.throttled_get_json`. Its `_seat_payload()` fixture is hand-written with `"SIDE": "BUY"` and `"RANK": 1`. So neither the report name nor the row shape is ever compared with the live endpoint. The skill reference `src/skills/eastmoney/references/龙虎榜/龙虎榜.md` documents the same report name (line 17) and the same `SIDE` / `RANK` fields (lines 65 and 67).

The report name and the `SIDE` / `RANK` fields have been in the file since it was added in #270 (`adfbc82`). I could not tell whether Eastmoney ever served a report under that name, so I am not claiming which side changed.

**What I expected:**
- `seats` should list the buy-side and sell-side seats that Eastmoney publishes for that security and date.
- When Eastmoney rejects a query, the tool should return an error or a warning, or reach the tushare fallback. It should not return `ok: true` with an empty list.

### Steps to Reproduce

All requests were made on 2026-09-19, between about 17:20 and 17:50 (UTC+8).

1. Through the MCP plugin, call `get_dragon_tiger` with `date="2026-08-17"` and `code="002156.SZ"`.
2. The result has one appearance and `"code": "002156", "seats": []`.
3. Send the tool's own seat request. It is rejected with the 9501 body quoted above:
   ```sh
   curl -sS -G 'https://datacenter-web.eastmoney.com/api/data/v1/get' \
     --data-urlencode 'reportName=RPT_BILLBOARD_TRADEDETAIL' --data-urlencode 'columns=ALL' \
     --data-urlencode "filter=(TRADE_DATE='2026-08-17')(SECURITY_CODE=\"002156\")" \
     --data-urlencode 'sortColumns=NET' --data-urlencode 'sortTypes=-1' \
     --data-urlencode 'pageNumber=1' --data-urlencode 'pageSize=500' \
   ```
4. Change only `reportName` to `RPT_BILLBOARD_DAILYDETAILSBUY`, then to `RPT_BILLBOARD_DAILYDETAILSSELL`. Each returns `"success":true` with 5 rows.

**Scope.** I called `DragonTigerTool().execute` directly and logged the requests it sent and the answers it got. Right after each call, within a few seconds, I queried the BUY and SELL reports directly. 002156.SZ uses the date above. 600519, 300750 and 688208 use their most recent board date on Eastmoney. The tool documents SH/SZ only; the BSE row is there to show that the rejection does not depend on the market.

| Symbol | Board | Date | Tool: appearances / seats | Direct: BUY / SELL rows | Eastmoney's answer to the tool's seat request |
|---|---|---|---|---|---|
| 002156.SZ | SZ main | 2026-08-17 | 1 / 0 | 5 / 5 | 9501, 报表配置不存在 |
| 600519.SH | SH main | 2013-01-28 | 1 / 0 | 5 / 5 | 9501, 报表配置不存在 |
| 300750.SZ | ChiNext | 2020-07-07 | 1 / 0 | 5 / 5 | 9501, 报表配置不存在 |
| 688208.SH | STAR | 2020-08-19 | 1 / 0 | 5 / 5 | 9501, 报表配置不存在 |
| 920298.BJ | BSE | 2026-09-18 | 3 / 0 | 15 / 15 | 9501, 报表配置不存在 |

A second pass took the first SZ-main, SH-main, ChiNext, STAR and BSE code from the tool's own full-market lists for 2026-09-18 and 2026-09-10. That gave 920298 again plus nine new codes (002185, 600127, 301583, 688261, 002636, 600744, 301689, 688496, 920071). The outcome was the same for all of them: `seats` was empty, the tool's seat request got 9501, and the direct BUY and SELL queries each returned 5 rows per board entry.

So this is not rate limiting and not specific to a market or a date. The appearances request succeeded every time, and the direct seat queries succeeded in the same seconds.

### Error Logs

The tool result contains no error and no warning. From reading the code, nothing is logged either: the path through `_fetch_report` lines 102-110 has no logger call. The only trace is Eastmoney's response body, quoted above.

### Suggested fix

- **Read seats from the two live reports.** Query `RPT_BILLBOARD_DAILYDETAILSBUY` sorted by `BUY` and `RPT_BILLBOARD_DAILYDETAILSSELL` sorted by `SELL`, both descending. Set `side` from the report that a row came from, and derive `rank` from the row order, because neither column exists upstream. akshare's `stock_lhb_stock_detail_em` reads the same pair of reports. I did not find a single report that carries both sides. `RPT_BILLBOARD_DAILYDETAILS` exists, but it is a per-entry summary with no seat rows.
- **Stop swallowing rejections.** In `_fetch_report`, raise when `success` is `false`, unless the code is 9201 (`返回数据为空`), which should keep returning `[]`. The next drift would then name its cause, and the existing tushare fallback would be reached. PR #1490, merged on 2026-09-19, does this for `get_shareholder_count` in `_upstream_rejection`, keyed on the message text. One design point, from reading lines 208-218: an exception from the seat request also discards the appearances that were already fetched, and without a tushare token the whole call becomes `ok: false`. Catching the seat failure separately and returning the appearances with a `warnings` entry may be friendlier.
- **Carry the listing reason on each seat row.** A security can be listed for several reasons on one day, and the seat reports repeat the seat list once per reason (`EXPLANATION` / `TRADE_ID`).
  - 920298.BJ on 2026-09-18 had 3 reasons, so 15 buy rows and 15 sell rows. 600127.SH on the same day had 2 reasons, so 10 and 10.
  - Without a reason field, those rows look like duplicates, some identical and some with different amounts.
  - `_MAX_SEATS = 30` is exactly filled by the 3-reason case. A day with 4 reasons would be truncated.
- **One seat can sit in both lists.** On 2026-08-17, 深股通专用 is first on both the buy side and the sell side of 002156, with the same `BUY` / `SELL` / `NET` values. That is how the disclosure is laid out, not a duplicate.
- **Update the fixture and the reference.** Change `_seat_payload()` in `tests/test_dragon_tiger_tool.py` to the live row shape. Add a case for a `success:false` 9501 payload and one for a 9201 payload. Update `龙虎榜.md` lines 17, 65 and 67.

**The correction, tested in memory.** I did not modify the package. I tried it two ways: a monkeypatched `_fetch_report` that routes the dead report name to the two live reports and tags `SIDE` and `RANK`, and a subclass that overrides `_collect` to query the two reports. Both gave the same result. The same `execute(date="2026-08-17", code="002156.SZ")` call returned the same single appearance and 10 seats, 5 per side, and every seat matched, field for field, the rows of a separate direct request to Eastmoney. The first buy-side row was 深股通专用 with `buy` 955,170,417.84, `sell` 415,743,415.64 and `net` 539,427,002.20. With the stricter `_fetch_report`, the tool's original seat request raised with Eastmoney's 9501 message, and a 9201 answer still returned `[]`.

Those seat rows reconcile with the board entry. De-duplicated on (seat name, `BUY`, `SELL`), the two lists hold 7 distinct rows; the 机构专用 rows are anonymous and all carry `OPERATEDEPT_CODE` `"0"`, so the amounts are the only way to tell them apart. Their `BUY` values sum to 1,958,377,606.34 and their `SELL` values to 937,397,640.10. These equal `buy_amount` and `sell_amount` in the appearance row that the tool already returns.

### Observed vs inferred

- **Observed:**
  - The empty `seats` through MCP and through `execute`.
  - Every Eastmoney response quoted above.
  - The scope table and the second pass, and the requests the tool sent, which I logged in memory. They went through the package's own session: default browser `User-Agent`, no `Referer`, and a `JSESSIONID` cookie that Eastmoney had set.
  - The rename-only result with `side` and `rank` null.
  - The in-memory correction, the row-for-row match and the reconciliation.
- **From code reading:** the swallow path in `_fetch_report`, that the tushare fallback is only reached on an exception, and the design point about discarded appearances. This is consistent with what I saw: `source: "eastmoney"` and no `warnings` in every result.
- **Not tested:**
  - The tushare fallback itself (`top_list` / `top_inst`).
  - Whether `RPT_BILLBOARD_TRADEDETAIL` ever worked.
  - Whether other Eastmoney datacenter tools in the package have the same problem. #1489 and #1481 report two of them.
- The seat amounts come from Eastmoney only. I did not compare them with the exchanges' own disclosures.

### Related

- #1489 (closed) and PR #1490 (merged on 2026-09-19) were the same class of problem in `get_shareholder_count`: a rejected datacenter query is reported as "no data".
- #1481 (open) and PR #1484 cover a silent empty envelope from `get_northbound_flow`.
- #516 (closed) asked for fallback providers for this tool family. The tushare fallback for this tool landed in commit `5fe512d`. It only triggers on exceptions, so it does not help here.

lorenzozanee:
## Summary

- Restore Dragon-Tiger seat data for security-specific queries by using Eastmoney's current buy- and sell-side reports.
- Preserve the provider's rejection semantics instead of returning a successful empty seat list.

## Changes

- Query and rank the current buy and sell seat reports separately, deriving side and rank from the report source and order.
- Preserve the upstream explanation and trade identifier fields.
- Add regression coverage for populated reports, valid empty responses, and provider rejection fallback.
- Update the user README and Eastmoney reference documentation.

## Out of scope

- Real Eastmoney/Tushare requests and upstream CI were not run locally.
- MCP schema, Tushare response shape, and unrelated data tools are unchanged.

## Test Plan

- `python -m pytest -p no:cacheprovider agent/tests/test_dragon_tiger_tool.py agent/tests/test_tushare_fallbacks.py agent/tests/test_eastmoney_client.py -q` — 38 passed
- `ruff check agent/src/tools/dragon_tiger_tool.py agent/tests/test_dragon_tiger_tool.py`
- `black --check agent/src/tools/dragon_tiger_tool.py agent/tests/test_dragon_tiger_tool.py`
- `git diff --check`

## Checklist

- [x] Regression tests added for the provider rejection and valid empty-result boundary.
- [x] User-facing and provider reference documentation updated.
- [x] No secrets or AI attribution added.
- [ ] Upstream CI and live provider integration not yet observed.

Fixes #1502

Rollback: revert the commit if upstream report semantics or maintainer expectations require a different seat mapping.


warren618:
Merged, thanks. Follow-up (0177a294): a security listed for several reasons on one day has a top-5 per reason (TRADE_ID), and each report returns them in one list, so a single running rank mixed reasons — 600721 on 2026-09-21 showed "BUY rank 4" for rank 1 of its reason. Seats are grouped by reason and ranked per side, and the 30-seat cap drops whole trailing reasons (`seat_reasons_omitted`) instead of every sell seat first. The English-only README sentence was dropped; the translations don't carry per-tool detail.

PR #1513, opened 2026-09-19
fix(tools): restore lockup expiry data

### LINKED ISSUE #1501: [Bug] get_lockup_expiry always returns ok:true with zero records: report RPT_LIFT_STOCK lacks six of the requested columns, and Eastmoney's rejection is read as an empty result
- **Interface:** MCP Plugin
- **LLM Provider:** n/a. This is a data tool, reproduced through MCP tool calls.
- **Version:** 0.1.15. On `main` at `2dfbc03`, `agent/src/tools/lockup_expiry_tool.py`, `agent/backtest/loaders/_http.py` and `agent/backtest/loaders/eastmoney_client.py` are byte-identical to the 0.1.15 wheel, so the bug is still there and the line numbers below apply to both.

### Description

**What happened:**
`get_lockup_expiry` answers with a successful, empty envelope whatever it is asked. For example, `get_lockup_expiry(code="002156.SZ")` returns:

```json
{"ok": true, "market": "a_share", "source": "eastmoney", "data": {"scope": "single_code", "count": 0, "records": [], "code": "002156"}}
```

Eastmoney does hold unlock records for that code, including one dated 2027-03-16. The caller is told, with `ok: true`, that the stock has no lockup history and nothing upcoming. The market-wide calendar (`horizon_days=30`, no code) comes back empty the same way, while Eastmoney listed 171 unlocks in that window.

The mechanism:

1. **The request is rejected.** `_fetch_lockups` (`agent/src/tools/lockup_expiry_tool.py:197-210`) asks report `RPT_LIFT_STOCK` (`_REPORT_NAME`, line 30) for the ten columns in `_COLUMNS` (lines 33-37). Eastmoney answers with HTTP 200 and this body:

   ```json
   {"version":null,"result":null,"success":false,"message":"FREE_SHARES_TYPE返回字段不存在","code":9501}
   ```

   The message means "requested return field FREE_SHARES_TYPE does not exist".
2. **Six of the ten columns are not in that report.**
   - Asked for one column at a time, `RPT_LIFT_STOCK` accepts `SECURITY_CODE`, `SECURITY_NAME_ABBR`, `FREE_DATE` and `FREE_SHARES`. It rejects each of `FREE_SHARES_TYPE`, `CURRENT_FREE_SHARES`, `ABLE_FREE_SHARES`, `LIFT_MARKET_CAP`, `FREE_RATIO` and `TOTAL_RATIO` with its own 9501 message. With `columns=ALL` it returns 14 columns, and those six are not among them. In the tool's request Eastmoney names only the first.
   - `RPT_LIFT_STAGE` accepts all ten, and returns 24 columns with `columns=ALL`.
3. **The rejection never becomes an error.**
   - The status is 200, so `response.raise_for_status()` in `throttled_get_json` (`agent/backtest/loaders/_http.py:199`) does not raise. `eastmoney_client.get_json` (`agent/backtest/loaders/eastmoney_client.py:81-102`) returns the payload as is.
   - `_extract_rows` (lines 159-174) sees `result: null` and returns `[]` (lines 170-172). Nothing reads `success`, `code` or `message`.
   - `get_lockup_expiry` therefore skips its `except` branch (lines 253-260) and builds the success envelope with `count: 0` (lines 262-279).

The report name and the column list are constants, and Eastmoney rejects the request on the columns. So the outcome does not depend on the code, the exchange, the date filter or `horizon_days`.

**What I expected:**
- The unlock schedule that Eastmoney holds for the code, or the upcoming calendar when no code is given.
- When Eastmoney rejects a query, `{"ok": false, "error": ...}` carrying Eastmoney's own message, so a broken query cannot read as "this stock has no unlocks".

### Steps to Reproduce

All observed on 2026-09-19, between about 17:20 and 17:55 (UTC+8).

1. Through the MCP plugin, call `get_lockup_expiry` with `code="002156.SZ"`. The result is `ok: true`, `count: 0`. `300750.SZ`, `688981.SH`, `920258.BJ`, and `horizon_days=30` with no code, give the same empty envelope.
2. Send the tool's exact request by hand. It returns the 9501 body shown above, with HTTP 200:

   ```sh
   curl -s -G 'https://datacenter-web.eastmoney.com/api/data/v1/get' \
     --data-urlencode 'reportName=RPT_LIFT_STOCK' \
     --data-urlencode 'columns=SECURITY_CODE,SECURITY_NAME_ABBR,FREE_DATE,FREE_SHARES_TYPE,FREE_SHARES,CURRENT_FREE_SHARES,ABLE_FREE_SHARES,LIFT_MARKET_CAP,FREE_RATIO,TOTAL_RATIO' \
     --data-urlencode 'filter=(SECURITY_CODE="002156")' \
     --data-urlencode 'sortColumns=FREE_DATE' --data-urlencode 'sortTypes=-1' \
     --data-urlencode 'pageNumber=1' --data-urlencode 'pageSize=200' \
   ```
3. Change only `reportName` to `RPT_LIFT_STAGE`. The same request returns `success: true` and 7 rows. The newest is 2027-03-16, `FREE_SHARES_TYPE` 定向增发机构配售股份, `CURRENT_FREE_SHARES` 7620.0794 (the unit is 10,000 shares, so 76,200,794 shares).

**Scope.** Each row below is one call to `get_lockup_expiry()` from the 0.1.15 wheel, with the raw Eastmoney payload captured in-process by a pass-through wrapper around `eastmoney_client.get_json`. The last two columns are described under "Proof of the cause". All requests were spaced more than one second apart.

| Call | Tool result | Eastmoney's answer to the tool | Tool with only `_REPORT_NAME` rebound | Direct `RPT_LIFT_STAGE` request |
|---|---|---|---|---|
| `002156.SZ` | `ok: true`, count 0 | 9501 | 7 records | 7 rows |
| `600519.SH` | `ok: true`, count 0 | 9501 | 0 records | 0 rows (`code: 9201`, 返回数据为空) |
| `601138.SH` | `ok: true`, count 0 | 9501 | 21 records | 21 rows |
| `000001.SZ` | `ok: true`, count 0 | 9501 | 6 records | 6 rows |
| `300750.SZ` | `ok: true`, count 0 | 9501 | 15 records | 15 rows |
| `300308.SZ` | `ok: true`, count 0 | 9501 | 15 records | 15 rows |
| `688208.SH` | `ok: true`, count 0 | 9501 | 4 records | 4 rows |
| `688981.SH` | `ok: true`, count 0 | 9501 | 4 records | 4 rows |
| `920258.BJ` | `ok: true`, count 0 | 9501 | 3 records | 3 rows |
| `920022.BJ` | `ok: true`, count 0 | 9501 | 2 records | 2 rows |
| `920059.BJ` | `ok: true`, count 0 | 9501 | 3 records | 3 rows |
| no code, `horizon_days=30` | `ok: true`, count 0 | 9501 | 171 records | 171 rows |

This rules out rate limiting or a network fault on our side. The tool received the same well-formed rejection every time, over more than half an hour, while requests to `RPT_LIFT_STAGE` from the same machine succeeded in between. A real network fault looks different in this tool: one of our calls hit the 15 s read timeout, and it came back as `ok: false` with the timeout message, through the existing `except` branch. Repeated, it returned its records.

**Proof of the cause.** We rebound module attributes in-process only and did not edit the package.

- Setting `lockup_expiry_tool._REPORT_NAME = "RPT_LIFT_STAGE"` is enough. The same calls then return the counts in the fourth column. For the eleven single-code calls we also sent an independent `requests` call to `RPT_LIFT_STAGE` with the same filter and sort, and every field the tool emits equals Eastmoney's row, on all 80 rows. For the calendar we compared the row count only.
- With that one change, the default 90-day calendar returns 200 records with `truncated: true`, while Eastmoney reports 452 rows over 3 pages. That is the existing `_MAX_RECORDS` cap doing what it says, and the sort is soonest-first.
- Adding a `success` check in front of `_extract_rows`, with the original report name, turns the first call into `{"ok": false, "error": "eastmoney lockup query failed: eastmoney datacenter rejected the request: code=9501 message=FREE_SHARES_TYPE返回字段不存在"}`.
- With both changes, `600519.SH` still returns `ok: true`, `count: 0`, because the check lets `code: 9201` through as an empty result.
- With everything restored, the tool failed again in the same way at the end of the run.

### Error Logs

There are none. The tool result is a success envelope. `logger.warning("get_lockup_expiry failed ...")` at line 256 is never reached, because nothing raises.

### Suggested fix

- **Use the report that has these columns.** Change `_REPORT_NAME` to `RPT_LIFT_STAGE`. All ten entries of `_COLUMNS` exist there, so nothing else in the request has to change.
- **Check Eastmoney's `success` flag.** Raise when it is `false`, so the existing `except` branch returns `ok: false` with Eastmoney's message. A query that matches nothing also arrives as `success: false`, with `code: 9201` and message 返回数据为空. That case has to keep producing the empty `ok: true` envelope. PR #1490, merged on 2026-09-19, draws the same line for `get_shareholder_count` in `_upstream_rejection`, keyed on that message text. A shared helper next to `eastmoney_client.get_json` could serve both tools.
- **Update what pins the old name.**
  - `agent/tests/test_lockup_expiry_tool.py` asserts `reportName == "RPT_LIFT_STOCK"` at lines 76 and 163. Its docstring says no request ever leaves the process, and the fixture `_payload` has no `success` key, so the suite cannot notice that the live endpoint rejects the query.
  - `agent/src/skills/eastmoney/references/参考数据/限售解禁.md` names `RPT_LIFT_STOCK` at lines 12 and 31. The tool's module docstring names it at line 12.
  - A 9501 rejection payload and a 9201 empty payload would make good regression fixtures.
- **Check what `free_shares` carries while you are there.** This only shows once data flows again. It is read from the data, not from Eastmoney documentation.
  - `_shape_record` maps `free_shares` from `FREE_SHARES` (line 151), and the skill reference describes it as 解禁股数, the number of shares unlocked (line 48).
  - In the data, `FREE_SHARES` behaves as the running total of unrestricted shares after the unlock, not the size of the batch. With the report name rebound, the 002156.SZ record for 2027-03-16 has `free_shares: 159365.3016` next to `able_free_shares: 7620.0794`, and `RPT_LIFT_STOCK` gives `TOTAL_SHARES` 1,593,797,706 for that date. For 688981 the consecutive rows read 109565.3, 187103.82 and 193846.3, and each step equals that row's `CURRENT_FREE_SHARES` (77538.52, then 6742.48).
  - We compared 65 rows across six stocks (002156, 300750, 601138, 300308, 688981, 920258). On 61 of them `(FREE_SHARES + NON_FREE_SHARES) × 10,000` equals `TOTAL_SHARES` from `RPT_LIFT_STOCK` exactly. The other four are all 688981, where `TOTAL_SHARES` is larger than the sum. That stock is also listed in Hong Kong; we did not check whether that is the reason.
  - The batch size is in `CURRENT_FREE_SHARES`, which `_COLUMNS` requests but `_shape_record` never emits, and in `ABLE_FREE_SHARES`. The two are equal on 56 of the 65 rows. On the 9 rows where they differ, `LIFT_MARKET_CAP` equals `CURRENT_FREE_SHARES` times the report's `NEW` price, and `TOTAL_RATIO` equals `CURRENT_FREE_SHARES × 10,000 / TOTAL_SHARES`. So on those rows `lift_market_cap` and `total_ratio` describe a quantity the record does not carry.
  - Units are not stated anywhere in the output. Share counts are in 10,000 shares, `lift_market_cap` is in 10,000 CNY, and the two ratios are fractions rather than percentages. akshare's `stock_restricted_release_queue_em` reads the same report and multiplies the share counts and the market value by 10,000.

### Observed vs inferred

- **Observed:**
  - The empty `ok: true` envelopes, through MCP and by calling the function from the wheel.
  - The 9501 payload the tool receives, and the same payload from a hand-sent copy of its request.
  - The per-column probes and the `columns=ALL` column lists of both reports.
  - The direct `RPT_LIFT_STAGE` results, and the in-process results above.
- **From code reading:** the path through `raise_for_status`, `_extract_rows` and the envelope, and the line numbers. The `main` comparison is a byte-for-byte diff of the three files named under Version, repeated at `2dfbc03`, and of the skill reference at `e5f7195`. None of the files this issue cites is among those changed between `e5f7195` and `2dfbc03`. The wheel does not ship the test file, so that one was read on `main` only.
- **Inferred from arithmetic on the data:** the reading of `FREE_SHARES`, which share column the market value and ratio follow, and the units.
- **Not known:** whether `RPT_LIFT_STOCK` ever carried these columns. The file has requested the same report and the same columns since it was added in #270 (commit `adfbc82`), so we cannot say whether this once worked and Eastmoney changed, or whether it never worked against the live endpoint. For what it is worth, akshare 1.18.94 reads these six columns from `RPT_LIFT_STAGE`, both per stock and market-wide (`akshare/stock_fundamental/stock_restricted_em.py`).
- **Not tested:** the CLI and Web UI paths. This issue makes no claim about the other tools that read this datacenter endpoint.
- Row counts are as served on 2026-09-19 and will drift as Eastmoney adds records.

### Related

- #1489 (closed) and its PR #1490 (merged on 2026-09-19) were the same class of failure in `get_shareholder_count`: Eastmoney rejects a column with HTTP 200 and `success: false`, and the tool reports it as missing data. That PR changed only `shareholder_count_tool.py`, its test and its skill reference, so it does not fix this tool.
- #1481 (open) and its PR #1484 (open) are about `get_northbound_flow` returning a silent empty envelope, from a different cause.

lorenzozanee:
## Summary

- Switch lockup-expiry queries to Eastmoney's compatible `RPT_LIFT_STAGE` report.
- Surface provider rejection envelopes instead of reporting rejected requests as successful empty schedules.

## Why

Eastmoney rejects the previous report request because several requested columns are no longer available. The HTTP response is valid JSON with `success: false`, but the tool previously treated `result: null` as an empty result.

Fixes #1501

## Changes

- Use `RPT_LIFT_STAGE` in the tool and reference documentation.
- Validate the provider envelope before extracting rows; preserve only the observed `code=9201` and `返回数据为空` response as a genuine empty schedule.
- Add regression coverage for populated responses, provider rejection, genuine empty results, and a rejection using the empty-result message.

## Test Plan

Validated locally with the focused lockup-expiry and shareholder-count tests, compilation, Ruff checks, formatting checks, and `git diff --check`; upstream CI has not been observed.

## Checklist

## Deliberately out of scope

The patch does not reinterpret share-count semantics or add a fallback to the obsolete report.

## Risk / Rollback

Eastmoney's external schema may change again; reverting this single commit restores the prior behavior if needed.


he-yufeng (commented):
Read through the diff mechanism. The rejection envelope handling is the right shape: `success: false` plus code and message surfaces as an error, and the empty-result carve-out is kept deliberately narrow. The third test is the one that proves it: 9501 with the exact same "返回数据为空" message still errors, so the exemption only fires for the code-and-message pair Eastmoney actually uses for empty calendars. Covering the in-between case (rejection code with the empty message) is good adversarial thinking.

One soft spot to be aware of: `_EMPTY_RESULT_MESSAGES` is a single exact string, so if Eastmoney ever varies the wording for empty results (punctuation, locale), a legitimate empty calendar raises as an error instead of returning an empty schedule. That direction fails closed (no fabricated data), which is acceptable; just worth knowing the exemption is exact-match.

Verified at code level only. I did not call the live Eastmoney endpoint from here, so the RPT_LIFT_STAGE column set and the 9201 convention are taken from the PR's evidence rather than re-confirmed against the upstream API.

warren618:
Merged, thanks. One mapping fix on top (bc7abd14), measured live: in `RPT_LIFT_STAGE`, `FREE_SHARES` is the float the unlock is measured against, not the unlock. 603162 on 2026-09-21 returns `FREE_SHARES` 42,489.17 and `CURRENT_FREE_SHARES` 232.88 (万股), and `FREE_RATIO` 0.005481 = 232.88 / 42,489.17. `free_shares` now comes from `CURRENT_FREE_SHARES`, the float is `float_shares`, and the envelope states units.

PR #1515, opened 2026-09-20
Integrate ChatNVIDIA for AI endpoint

Almaqrami-Huthaifa:
## Summary

-

## Changes

-

## Test Plan

## Checklist


warren618:
Thanks for looking into NVIDIA support. NVIDIA NIM is already a built-in provider, through its OpenAI-compatible endpoint: set `LANGCHAIN_PROVIDER=nvidia` and `NVIDIA_API_KEY` (the base URL defaults to `https://integrate.api.nvidia.com/v1`), or pick NVIDIA in the first-run wizard. These three PRs would make `langchain-nvidia-ai-endpoints` a required dependency for every install, construct a `ChatNVIDIA` client for a model called `nvidia/chat-gpt` when `api_server.py` is imported, and import it in `src/providers/`, which needs discussion before any change. So I'm closing all three. If the OpenAI-compatible path is missing something you need from NIM, please open an issue with the model and the failing request.

PR #1517, opened 2026-09-20
fix(grounding): handle locale decimals and precision-aware rounding

zeus229:
## Summary

This PR fixes two related grounding issues in numeric figure handling:

1. locale-aware parsing of decimal commas and mixed thousands/decimal separators;
2. precision-aware matching of rendered figures against observed evidence.

The change is intentionally scoped to numeric parsing and rounding. It does **not** include statistical derivation grammar, market-routing changes, or the tail-risk confidence work in #1444.

### Locale number parsing

The grounding tokenizer can split valid decimal-comma figures into multiple numbers. Examples that fail on pristine `main` include:

* `17,9318145214327%`
* `2,639499655314571`
* `1.234,56`
* `1.234.567,89`

This can create spurious figures and lead to `figure_undeclared` rejections.

The fix keeps preservation cases intact, including:

* `2023,2024`
* `0.500,0.600`
* `1,234`
* `1,234.56`
* `1,234,567`

There is no global comma-to-dot replacement.

### Precision-aware matching

Two different tolerances previously produced contradictory behavior:

* a valid rounded value such as `38.680058... -> 38,68` could fail to bind to its declaration;
* a materially wrong rounded value could still pass through the old flat relative tolerance.

The new matching logic uses the precision actually written by the model for decimal figures.

For example, with an observed value of approximately `30.205165%`:

* `30.20%` is rejected;
* `30.21%` is accepted.

This makes the gate stricter where the written precision is explicit rather than relaxing evidence checks.

Existing behavior is preserved for integer figures where precision is ambiguous and for undeclared price handling that already relies on the legacy tolerance policy.

## Test-driven development

The regression tests were written first.

On pristine upstream `main`:

* 45 new tests total;
* 25 fail;
* 20 preservation/negative cases already pass.

After the implementation:

* all 45 new tests pass;
* grounding / financial-rigor / output-discipline sweep: **886 passed** (841 existing + 45 new).

The full agent tree retains the same environment-dependent failures/errors as pristine `main`:

* baseline: 136 failures, 9 errors;
* branch: the same 136 failures, 9 errors;
* passing count increases by the 45 added tests.

Static checks:

* `ruff`: clean;
* `git diff --check`: clean;
* `py_compile`: clean;
* the changed hunks introduce no new Black formatting drift beyond pre-existing upstream formatting differences.

## Compatibility

Test-merging the current head of #1444 (`792a30bd`) produced no conflicts, and the relevant grounding tests remained green (**898 passed**).

This PR does not duplicate #1444's tail-risk confidence identity changes.

## Scope boundaries

Not included here:

* tail-risk confidence-level handling from #1444;
* structured statistical derivations / series formulas;
* market-data routing changes;
* symbol identity changes;
* any broker- or integration-specific behavior.

## Risk / fail-closed behavior

The rounding change is deliberately precision-aware and can reject values that the old 0.5% relative band accepted. That is intentional.

A model that writes a numerically incorrect rounding may receive another grounding round rather than having the value silently accepted.

No evidence bypass or general relaxation of the grounding gate is introduced.


he-yufeng (commented):
Mechanism review of the two halves.

The rounding rule in `FiguresBlock.match` reads correctly: exact tolerance wins first, then a written decimal covers a declaration only within half a unit of its last written digit, and the `ROUNDED_BAND` relative cap narrows it further on small declared values (so a written "0.01" cannot borrow a 0.014 declaration: 0.4% would pass the half-unit test but the 0.5% relative band rejects it). Returning the nearest candidate when several declarations round to the same figure is the right deterministic tiebreak, and integers are kept out of that path entirely since their precision is unknowable.

The locale heuristics hold up too: a comma followed by four or more digits cannot be a thousands group (grouping is always exactly three), so "2,639499655314571" is unambiguously a decimal, and the "2023,2024" year-pair carve-out is the one lookalike that needed excluding. `_DOTTED_DECIMAL_RE` normalizes "1.234,56" to 1234.56 correctly and the trailing lookahead keeps "1.234,5.6" as a list. One assumption to keep an eye on: `match()` computes the half-unit from `digits.split(".", 1)[1]`, which assumes the caller always hands over plain-decimal digits from the tokenizer; if any future caller can pass a non-plain form the index errors, so a cheap guard or a documented contract on `digits` would make that explicit.

Coordination note for the maintainer: this and #1444 both touch `agent/src/agent/grounding/policies.py` (mine for the identity frame, this one for numeric parsing). The scopes are disjoint by design here, but whichever lands second will need a rebase; the test surfaces (`test_grounding_numeric_locale_rounding` vs `test_grounding_role_hardening`) are complementary and worth running together at that point.

warren618 (changes requested):
Thanks. Several parts are clear gains: `17,9318…%` and `1.410,00 CNY` now ground, an invented `Sharpe 2,7394` is now rejected, and the band only ever tightens. Two parts need to change before this can go into the gate, measured on a two-bar ledger:
1. `_is_long_decimal_comma` (`figures.py:390`) reads two unspaced numbers as one decimal. `1400,1777 元` and `成交量 1000,2500 手`, with 1777 and 2500 invented, are rejected on main and pass here, because `1400,1777` parses to 1400.1777, next to the open.
2. `_is_whole_cell` (`:408`) lets one list-like cell such as `| 5,20 |` switch the whole answer to decimal commas. A correct `234,567` or `1,410` elsewhere then parses as 234.567 or 1.41 and is rejected: a legitimate answer that passes on main fails here.
Could the decimal-comma reading require an unambiguous shape on the number itself (a thousands point before it, or a currency/percent right after), and never flip the document's mode from one cell? Truncation (`30.20%` for 30.2052%) now rejecting is a separate call we can make after that. The gate also changes from 14 to 15 regexes, which is fine since it is shape-only.

zeus229:
Thanks for the review — both regressions are fixed in the current head cae74555.

Changes made:

removed the broad long-comma heuristic, so ambiguous pairs such as 1400,1777 元 and 成交量 1000,2500 手 stay two numbers instead of becoming 1400.1777 / 1000.2500;
removed whole-cell decimal-comma mode detection, so a cell such as | 5,20 | no longer flips the rest of the document into comma-decimal mode;
added regression coverage for those cases while keeping the intended locale forms (17,9318145214327%, 1.234,56, etc.) and the precision-aware rounding behavior.

I re-ran the change in isolated worktrees against current upstream main at cc3e08ee4d4be0d72ad405bdc298792cb222ed24:

the two-bar repros now reject the invented 1777 / 2500 values as on main;
| 5,20 | no longer contaminates 234,567 / 1,410;
17,9318145214327% and 1.234,56 remain fixed;
30.21 vs 30.205165 passes and 30.20 rejects;
690 directed grounding tests passed, plus 170 PR/contract tests;
py_compile and git diff --check passed.

The four PR commits also applied cleanly on top of current upstream main with no conflicts. The relevant figures.py and policies.py files are unchanged between the PR's old base and current main, so there is no hidden behavioral drift from today's upstream update.

I also checked the earlier FiguresBlock.match(..., digits) concern: both current production call sites pass normalized Figure.digits, so it is safe today; I agree it is worth treating that as an explicit contract for future callers.

warren618:
Merged, thanks — both regressions are gone in cae74555. On the same two-bar ledger as before:
- `1400,1777 元`, `1400,1777 CNY` and `成交量 1000,2500 手` (1777 / 2500 invented) are rejected again.
- A legitimate `234,567` or `1,410` next to a `| 5,20 |` cell passes again.
- `17,93…%` and `1.410,00 CNY` still ground, and a written decimal still has to sit within half a unit of its evidence.
Grounding suites: 792 passed; CI is green. One behaviour to know about: a truncated figure (`30.20%` for 30.2052%) is now asked to be corrected, where main let it through.

Issue #1489, opened 2026-09-18
[Bug] get_shareholder_count reports "no disclosure found" for every A-share symbol — the requested report column no longer exists


cgycorey:
**What happened**

`get_shareholder_count` reports that a stock publishes no disclosure, for **every** A-share symbol:

```json
{"ok": false, "error": "no shareholder-count disclosure found for '600519.SH'"}
```

600519.SH (贵州茅台) does publish one — 296,404 holders as of 2026-06-30, against 243,159 the prior quarter. The data is one field name away.

**Root cause — a requested report column no longer exists**

`agent/src/tools/shareholder_count_tool.py:27-30` requests this column list from the Eastmoney datacenter:

```
SECUCODE,SECURITY_CODE,END_DATE,HOLDER_NUM,HOLDER_NUM_CHANGE,HOLDER_NUM_RATIO,AVG_HOLD_AMT,AVG_HOLD_NUM,TOTAL_MARKET_CAP
```

`RPT_HOLDERNUMLATEST` now rejects it. Verbatim, repeatable (4/4 attempts, identical across every symbol tested):

```
HTTP 200  {"success": false, "message": "AVG_HOLD_AMT返回字段不存在", "result": null}
```

The tool's `get_json` only raises on network/HTTP/JSON failures, so this payload arrives intact; `_parse_periods` finds no `result.data`, and the caller is told the **stock** has no disclosure — while the actual cause is that **our query** is invalid. The column list in `agent/src/skills/eastmoney/references/参考数据/股东户数.md` carries the same stale name.

**Evidence**

- Replacing `AVG_HOLD_AMT` with `AVG_MARKET_CAP` returns exactly one row for `600519.SH`, `300308.SZ` and `000001.SZ`; the current list returns `result: null` for all three.
- `AVG_MARKET_CAP` is the same quantity, not a guess: on the live row `TOTAL_MARKET_CAP / HOLDER_NUM` = `1481959237169.49 / 296404` = **4,999,795.00**, and the report's `AVG_MARKET_CAP` is `4999794.99996454` (户均持股市值). `AVG_HOLD_NUM` likewise equals `TOTAL_A_SHARES / HOLDER_NUM`.
- No accepted variant of the old name exists: `AVG_HOLD_AMOUNT`, `HOLD_AMT`, `AVG_AMT`, `AVG_HOLD_VALUE` are each rejected with `...返回字段不存在`.
[...]

PR #1518, opened 2026-09-20
fix(tools): serve shareholder-count history from the per-period report

### LINKED ISSUE #1503: [Bug] get_shareholder_count: max_periods can never return more than one period, and the "quarter-over-quarter" change is measured against a PRE_END_DATE the tool drops (follow-up to #1489)
- **Interface:** MCP Plugin
- **LLM Provider:** n/a. This is the data layer, reproduced through MCP tool calls and by calling the tool class directly.
- **Version:** `main` at `2dfbc03`, which includes PR #1490 (merged on 2026-09-19, merge commit `7586a78`). `agent/src/tools/shareholder_count_tool.py` on `main` is byte-identical to the PR head `2cdf78b` that most of the tests below were run on. Line numbers below are for `main` at `2dfbc03`. The latest release, 0.1.15, predates the merge and still has the #1489 failure. The merge changed neither the report name nor the output fields.

### Description

**Scope of this report.**
On 0.1.15, every call fails with `no shareholder-count disclosure found`, because the request asks `RPT_HOLDERNUMLATEST` for a column it does not have (`AVG_HOLD_AMT`). That is #1489, fixed on `main` by PR #1490. We reproduced it independently on 0.1.15 and have nothing to add to that report.

This issue is about what is left now that the fix is in. #1489 already mentions the first point below as an observation, and #1490 lists it as out of scope. What we add there is a per-symbol check of the per-period report. As far as we can tell, the second point has not been reported.

**What happened:**
The tool reads `RPT_HOLDERNUMLATEST` (`agent/src/tools/shareholder_count_tool.py:24`). For each of the eleven symbols we tried, that report returned exactly one row: the latest disclosure, with the previous one beside it in `PRE_*` columns. Unfiltered, the report has 5,561 rows. Two things follow.

1. **`max_periods` has no effect.**
   - The parameter is documented as "Maximum number of most-recent report periods to return (1-24)" (lines 69-76) and is sent as `pageSize` (line 119).
   - With the column fix applied, as on `main`, `max_periods=4` and `max_periods=24` both return exactly one period, for all eleven symbols. The raw payload says `"count": 1`.
2. **The change fields are not quarter-over-quarter, and the caller cannot tell.**
   - The tool description promises "quarter-over-quarter change (absolute and percent)" (lines 52-58). So does the MCP docstring (`agent/mcp_server.py:2015`).
   - `HOLDER_NUM_CHANGE` and `HOLDER_NUM_RATIO` are measured against the row's `PRE_END_DATE`. In every row we looked at, `HOLDER_NUM_CHANGE` equals `HOLDER_NUM` minus `PRE_HOLDER_NUM`. The tool does not request `PRE_END_DATE` (lines 31-34) and does not return it (lines 230-238).
   - `PRE_END_DATE` is often not the previous quarter-end. Examples, as served on 2026-09-19:

     | Symbol | `END_DATE` | `PRE_END_DATE` | `HOLDER_NUM_RATIO` | Actual interval |
     |---|---|---|---|---|
     | 300750.SZ | 2026-06-30 | 2025-09-30 | +18.71% | nine months |
     | 600958.SH | 2026-06-30 | 2017-06-30 | +21.93% | nine years |
     | 300188.SZ | 2026-09-18 | 2026-09-10 | −1.23% | eight days |

   - For 300750.SZ, Eastmoney's F10 report `RPT_F10_EH_HOLDERNUM` does have rows for 2025-12-31, 2026-02-28 and 2026-03-31, and shows +32.04% for 2026-06-30 against 2026-03-31. The per-period report `RPT_HOLDERNUM_DET` has no rows for those three dates, and the `RPT_HOLDERNUMLATEST` row points back to 2025-09-30. With #1490 applied, the tool returns `holder_count_change_pct: 18.71` for this symbol under a "quarter-over-quarter" description.
   - Across the whole report (5,561 rows, snapshot of 2026-09-19; the numbers move as new disclosures arrive):

     | Rows | `END_DATE` | `PRE_END_DATE` |
     |---|---|---|
     | 3,886 | 2026-06-30 | 2026-03-31 |
     | 196 | 2026-06-30 | later than 2026-03-31 |
     | 20 | 2026-06-30 | earlier than 2026-03-31 |
     | 1,327 | later than 2026-06-30 | any; these are intra-quarter disclosures |
     | 132 | earlier than 2026-06-30 | not broken down |

     So at least 1,543 of 5,561 rows (27.7%) carry a change that is not measured over exactly one quarter.

**What I expected:**
- `max_periods=N` returns up to the N most recent disclosed periods.
- Each period states what its change is measured against. Failing that, the tool should not call it quarter-over-quarter.

### Steps to Reproduce

On 0.1.15 the failure in #1489 hides both points, so the column fix has to be in place first. Our tests predate the merge of #1490 by a few hours, so we got there in two ways, without editing the installed package:

- Executing an in-memory copy of the 0.1.15 module with `AVG_HOLD_AMT` replaced by `AVG_MARKET_CAP`. That is two occurrences, on lines 29 and 194 of the 0.1.15 file.
- Loading a scratch copy of the PR #1490 head file (`2cdf78b`) as a separate module. After the merge we confirmed that this file is byte-identical to the one on `main` at `2dfbc03`, and repeated step 1 on a scratch copy of the `main` file for 002156.SZ, 600519.SH, 300750.SZ and 688208.SH, with the same result.

The eleven symbols were 002156.SZ, 000858.SZ, 600519.SH, 601318.SH, 600958.SH, 300308.SZ, 300750.SZ, 688111.SH, 688208.SH, 920185.BJ and 920799.BJ. The eleven-symbol results below come from the second method. The first was run on 002156.SZ only, and agreed.

1. With the column fix, call `ShareholderCountTool().execute(code="002156.SZ", max_periods=4)`. It returns `ok: true` with one period, `2026-06-30`. `max_periods=24` returns the same single period. So do the other ten symbols, each with its own latest date.
2. Additionally set `_REPORT_NAME = "RPT_HOLDERNUM_DET"` and change nothing else. The same call returns four periods: `2026-06-30`, `2026-03-31`, `2025-12-31`, `2025-09-30`. `max_periods=24` returns 24 periods for ten of the eleven symbols, newest first, with no duplicate dates. 600958.SH returns 14, which is every row the report has for it.
3. For the second point, run the two requests below. The first is the tool's own request, with `PRE_END_DATE` added and the column fixed. The second is the F10 report for the same symbol.

```sh
curl -G 'https://datacenter-web.eastmoney.com/api/data/v1/get' \
  --data-urlencode 'reportName=RPT_HOLDERNUMLATEST' \
  --data-urlencode 'columns=SECUCODE,END_DATE,PRE_END_DATE,HOLDER_NUM,PRE_HOLDER_NUM,HOLDER_NUM_CHANGE,HOLDER_NUM_RATIO' \
  --data-urlencode 'filter=(SECUCODE="300750.SZ")' \
  --data-urlencode 'sortColumns=END_DATE' --data-urlencode 'sortTypes=-1' \
  --data-urlencode 'pageNumber=1' --data-urlencode 'pageSize=24' \

curl -G 'https://datacenter-web.eastmoney.com/api/data/v1/get' \
  --data-urlencode 'reportName=RPT_F10_EH_HOLDERNUM' \
  --data-urlencode 'columns=SECUCODE,END_DATE,HOLDER_TOTAL_NUM,TOTAL_NUM_RATIO' \
  --data-urlencode 'filter=(SECUCODE="300750.SZ")' \
  --data-urlencode 'sortColumns=END_DATE' --data-urlencode 'sortTypes=-1' \
  --data-urlencode 'pageNumber=1' --data-urlencode 'pageSize=6' \
  --data-urlencode 'source=HSF10' --data-urlencode 'client=PC'
```

The first returns `"count": 1`, with `PRE_END_DATE` `2025-09-30`, `HOLDER_NUM_CHANGE` 42557 and `HOLDER_NUM_RATIO` 18.71. The second lists 2026-06-30, 2026-03-31, 2026-02-28, 2025-12-31, 2025-09-30 and 2025-06-30.

The row counts in the table above come from the same endpoint with `reportName=RPT_HOLDERNUMLATEST`, `pageSize=1` and filters such as `(END_DATE='2026-06-30')(PRE_END_DATE<'2026-03-31')`, reading `result.count`. The five buckets add up to the unfiltered count.

### Error Logs

There are none. Both problems are silent: the tool answers `ok: true`.

### Suggested fix

- **Read the history from `RPT_HOLDERNUM_DET`.**
  - In our test it accepted the same `SECUCODE` filter, the same sort and paging, and the corrected column list unchanged. The only change was the report name on line 24.
  - Row counts for the eleven symbols ranged from 14 (600958.SH) to 85 (000858.SZ).
  - akshare's `stock_zh_a_gdhs_detail_em` reads the same report for per-symbol history.
- **Say what the change is measured against.**
  - Request `PRE_END_DATE`, and ideally `HOLD_NOTICE_DATE`, and return them per period, for example as `prev_end_date` and `notice_date`. Both reports accepted both columns.
  - Then either reword "quarter-over-quarter" to "change versus the previous disclosed period", or return the change only when `PRE_END_DATE` is the preceding quarter-end and `null` otherwise.
  - This is needed with either report. In `RPT_HOLDERNUM_DET` a quarter-end row is also measured against whatever row precedes it, which can be a month-end.
- **Reword "quarterly".** The rows are per disclosure, not per quarter.
  - In the 24 most recent `RPT_HOLDERNUM_DET` rows we saw month-end dates such as `2026-02-28` and `2024-02-29` for several symbols.
  - For 000858.SZ the four most recent rows are 2026-08-10, 2026-07-25, 2026-07-24 and 2026-07-07. With this report, `max_periods=4` covers five weeks for that symbol, not four quarters.
- **Tests.** The fixture in `agent/tests/test_shareholder_count_tool.py` has two rows for one symbol and no `success`, `code` or `count` fields. The live report returned one row for every symbol we tried. A fixture recorded from the live endpoint, including `result.count`, would have made the one-row shape visible. #1489 makes the same point about the column name.
- **Known limitation: `RPT_HOLDERNUM_DET` has gaps for some symbols.**
  - For 300750.SZ it lacks the three period-ends listed above (35 rows, against 38 in the F10 report).
  - For 600958.SH it has 14 rows in total and nothing between 2017-06-30 and 2026-06-30. The F10 report has 61 rows for that symbol and shows +0.78% for 2026-06-30 against 2026-03-31.
  - In both cases the `PRE_*` columns of the `RPT_HOLDERNUMLATEST` row match the previous `RPT_HOLDERNUM_DET` row. That is where the nine-month and nine-year intervals come from.
  - `RPT_F10_EH_HOLDERNUM` is fuller, but it uses different column names (`HOLDER_TOTAL_NUM`, `HOLDER_A_NUM`, `HOLDER_H_NUM`, `AVG_HOLD_AMT`, `NOTICE_DATE`), and its total is not the same quantity for dual-listed names. For the two A+H symbols we checked, `HOLDER_TOTAL_NUM` is `HOLDER_A_NUM` plus `HOLDER_H_NUM`, and `HOLDER_A_NUM` equals the `HOLDER_NUM` of the other two reports: 269,979 + 54 = 270,033 for 300750.SZ, and 185,240 + 76 = 185,316 for 600958.SH, both for 2026-06-30. For 002156.SZ, 688111.SH and 920185.BJ all three figures are equal.
  - We have not built or tested a tool variant on the F10 report, so we only point at it. Which report to use is your call.
- **Minor.** The example `830799.BJ` in the parameter description (line 66) returns `返回数据为空` from both reports today, while `920799.BJ` returns data. We did not check whether these are the same company under an old and a new code.

### Observed vs inferred

- **Observed** (2026-09-19, about 17:20 to 18:00 UTC+8, in two separate runs with different symbol sets, and a short re-check on `main` at about 20:05):
  - The #1489 failure on 0.1.15, through the MCP plugin and by calling the tool class directly. The raw payload was `{"version": null, "result": null, "success": false, "message": "AVG_HOLD_AMT返回字段不存在", "code": 9501}` with HTTP 200, for every symbol, including the impossible `999999.SH`.
  - Control requests within a few seconds, identical except for the column name, succeeded each time, including four alternating rounds on 002156.SZ. Requesting the nine columns one at a time, only `AVG_HOLD_AMT` was rejected. So this was not rate limiting.
  - One period with the column fix, and N periods with `RPT_HOLDERNUM_DET`.
  - For 002156.SZ, the patched tool's output matched a direct request to the endpoint field by field (2026-06-30, 561,305 holders), and the 2026-06-30 count is the same in all three reports.
  - The `PRE_END_DATE` values, row counts and F10 columns above.
  - All requests were spaced at least one second apart.
- **Observed on the PR #1490 head, which is the file `main` now carries:** a scratch copy of that file returns one period for `max_periods=4` and for `max_periods=24`, on all eleven symbols. On a scratch copy of the `main` file after the merge: one period for 002156.SZ (`max_periods` 4 and 24) and for 600519.SH, 300750.SZ and 688208.SH (`max_periods=8`).
- **From code reading:** the line references, and that nothing in the tool requests or returns `PRE_END_DATE`.
- **Single source.** Every holder count here comes from Eastmoney. The numbers show what the endpoint returns. We did not check them against company filings.
- **Not tested:**
  - The fixed variants through the MCP server end to end. We ran the tool class directly for those.
  - A full pass over `RPT_HOLDERNUM_DET` for gaps.
  - Any tool variant built on `RPT_F10_EH_HOLDERNUM`.

### Related

- #1489 (closed) and PR #1490 (merged on 2026-09-19) cover the `AVG_HOLD_AMT` failure.
  - **A data point for that thread.** `AVG_HOLD_AMT` is a valid column of `RPT_F10_EH_HOLDERNUM`, so the name may have come from that report's schema. It has been in `_COLUMNS` since the file's first commit (`adfbc82`, #270), and the test has mocked `get_json` since that same commit. We cannot tell whether `RPT_HOLDERNUMLATEST` ever carried it.
  - **A note on the merged check.** `_upstream_rejection` tells an empty result from a rejection by the message text `返回数据为空`. The payload's numeric `code` separates the two cases as well. We saw `9201` for an empty result and `9501` for a rejected request: an unknown column, an unknown report name, and an unsupported `like` filter all gave `9501`. That may be a sturdier test than matching the message text.
- #1481 is a related case on `get_northbound_flow`: the HTTP call succeeds, the payload is not what the parser expects, and the tool answers `ok: true` with an empty result.

he-yufeng:
## Summary

- Serve the shareholder-count history from the per-period detail report (`RPT_HOLDERNUM_DET`): the latest-only report answers exactly one row per symbol regardless of `pageSize`, so `max_periods` was a dead parameter.
- Each returned period now states the interval its change was measured over (`prev_period_end`) instead of the docstring claiming quarter-over-quarter, which #1503 measured as false for 27.7% of rows (nine months to nine years).
- The average-holding and market-cap fields exist only on the latest-only report, so its row joins onto the newest detail row when the end dates agree; symbols absent from the detail report fall back to that single latest row.

## Why

Closes #1503. I verified both claims live against the Eastmoney datacenter before touching code: `RPT_HOLDERNUMLATEST` returns `"count": 1` for 300750.SZ with `pageSize=4`, its PRE_END_DATE (2025-09-30) is nine months from the END_DATE, and `RPT_HOLDERNUM_DET` serves the real per-period history (35 rows there; for 300188.SZ it includes the intra-quarter 2026-09-18 disclosure that the latest-only report surfaces as its latest row).

## Changes

- `shareholder_count_tool.py`: DET for the paged history, LATEST joined for the avg/market-cap fields and as the single-row fallback, `prev_period_end` per row, docstrings/description no longer say quarter-over-quarter.
- `mcp_server.py`: the MCP docstring updated to match.
- Tests: two-report router mock; pins for the multi-row history, the avg-field join, the empty-detail fallback, and the enrichment failure not failing the history.

## Test Plan

- [x] `pytest agent/tests/test_shareholder_count_tool.py -q` — 10 passed; the four new/changed tests fail on the unfixed tree (red-green verified)
- [x] `pytest agent/tests -q -k "tool or market or shareholder"` — 2183 passed
- [x] Live endpoint probes (curl) for 300750.SZ / 300188.SZ on both reports, quoted above

Risk / rollback: response schema gains `prev_period_end` per row and `max_periods` now returns up to N rows (previously always one) — both are the issue's requested behavior. Revert this commit to restore the previous single-row shape. No live-broker or trading paths touched.


warren618:
Merged, thanks. Follow-up (9162dc6a): the latest-only join swallowed its own failure, and a test pinned that, so a retired column (the #1489 failure) read as null averages. A failed join now keeps the history and adds a `warnings` entry, and a rejected latest-only fallback is an error rather than "no disclosure found". The detail report is also asked for one row past `max_periods`, so the oldest period's `prev_period_end` is filled.

PR #1520, opened 2026-09-20
feat(channels): web-based channel configuration with connection test and hot apply

### LINKED ISSUE #1519: [Feature] Configure IM channels from the Web UI: credentials, connection test, enable/disable with hot apply
### Problem or Motivation

The 16-adapter IM channel runtime ships with a **read-only** Settings panel: it shows adapter status and can start/stop the runtime, but the only way to actually configure a channel is hand-editing `channels.<name>` in `~/.vibe-trading/agent.json` and restarting the process.

Three concrete gaps:

1. **No credential surface** — no form, no validation, no way to verify credentials before going live. A typo in `client_secret` is only discovered after a restart and a failed connection.
2. **No hot update** — `_get_channel_runtime()` snapshots the config once at first build, and `ChannelManager._init_channels()` runs in the constructor only. Editing `agent.json` then pressing Start/Stop reuses the same already-built instances, so config edits never apply without a full process restart.
3. **No non-destructive toggle** — disabling a channel today means editing the file (or losing the section), not flipping a switch that keeps credentials.

### Proposed Solution

A generic, contract-driven channel configuration surface, with **DingTalk as the first fully-guided channel** (Stream Mode: Client ID = AppKey, Client Secret = AppSecret, no public callback URL needed):

**Backend**
- `GET /channels/config` — per-channel field metadata (type/secret/required/i18n hint key), non-secret values, and masked secret status (`****` + last 4). Secret values never cross the wire.
- `PUT /channels/config/{name}` — merge-patch one channel section: validate by ephemeral adapter construction **before** any write (422, nothing persisted); auto-verify credentials on the enable transition with an explicit `skip_verify` override; atomic 0600 write to `agent.json`.
- `POST /channels/{name}/test` — probe **unsaved** credentials without persisting (DingTalk: OAuth accessToken exchange), with honest result codes `ok | invalid_credentials | network | unsupported`.
- **Per-channel hot apply** — a running runtime hot-swaps only the affected adapter (`ChannelManager.reload_channel`), no process restart, other channels untouched; stopped/never-built runtimes fall back to a singleton reset so the next Start reads fresh config.

**Web UI**
- Expandable per-channel config panel rendered generically from backend field metadata (adding Feishu/Telegram later = one hints dict entry, no new routes or components).
- Setup guide (DingTalk: create app → bot capability → Stream Mode → copy keys → publish), Test-connection button that works on dirty form values, non-destructive Enable toggle, "applies immediately, no restart" save semantics. 8-locale i18n.

**Adapter fixes found during live end-to-end QA** (each pinned by a regression test):
- DingTalk `_on_message` never passed `is_dm`, so the DM pairing-code flow was dead code (unapproved DMs were silently dropped).
- `stop()` never terminated the SDK stream loop (it swallows the first `CancelledError` and reconnects), leaving a zombie WebSocket after disable/hot-swap and hanging graceful process shutdown.
- Hot-swapping a channel whose `start()` is long-running (reconnect loop) must not await it inline or the config request hangs.
- `start_all()` with zero enabled channels skipped creating the outbound dispatcher, so a channel enabled afterwards could never start.
- New `force_ipv4` option: DingTalk's robot-send API enforces the app egress-IP whitelist, and on dual-stack networks the rotating IPv6 prefix keeps falling out of it; pinning egress to IPv4 makes the whitelist stable.

### Area

Web UI

### Alternatives Considered

- **Full runtime rebuild on every config change** — simpler, but restarts all channels and drops in-flight sessions; kept only as the fallback path if a per-channel reload raises.
- **File-watcher hot reload with a reload plan** (OpenClaw-style `restart | hot | none` classification) — powerful but overkill for v1; explicit apply-on-save (dsh-im style) is simpler and covers the UX.
- **Per-channel hardcoded frontend forms** (Chatwoot-style per-channel components) — rejected in favor of backend-driven field metadata so the long tail of 15 other adapters needs no frontend work.
- **Validate-on-save only** (Chatwoot Telegram `getMe`) — kept for the enable transition, plus an explicit Test button so credentials can be probed before saving (draft-friendly).

### Contribution

- [X] I'm willing to submit a PR for this feature

shadowinlife:
## Summary

- IM channels become configurable from the Web UI: per-channel credential forms rendered from backend field metadata, a **Test connection** button that probes unsaved credentials, and a non-destructive **Enable** toggle — all hot-applied without a process restart.
- DingTalk is the first fully-guided channel (Stream Mode setup guide, accessToken credential probe, `force_ipv4` egress pinning); the other 15 adapters get generic forms from their `default_config()` with zero extra routes or components.
- Live end-to-end QA against a real DingTalk org app surfaced and fixed four pre-existing adapter/runtime defects, each pinned by a regression test.

## Why

Closes #1519.

The Settings panel was read-only: configuring a channel meant hand-editing `~/.vibe-trading/agent.json` **and** restarting, because the runtime singleton snapshots config once and Start/Stop reuse the already-built instances. There was no way to verify credentials before going live, and disabling a channel meant editing the file instead of flipping a switch.

## Changes

**Backend**
- `agent/src/config/writer.py`: atomic `channels.<name>` section writer (raw-dict merge, tmp + fsync + `os.replace`, 0600/0700, JSON-only with `ConfigNotWritableError`); the Feishu QR-login persister now delegates to it.
- `BaseChannel.test_connection()` hook (`supports_connection_test`, honest `ok | invalid_credentials | network | unsupported` envelope) + DingTalk standalone accessToken probe (fresh httpx client, works pre-`start()`, token discarded, detail scrubbed).
- `ChannelManager.reload_channel()`: per-channel hot swap (bounded stop of the old instance, rebuild, start iff the dispatcher is live, status/config snapshot kept in sync — other channels untouched). `start_all()` now always arms the outbound dispatcher (previously zero enabled channels left a later-enabled channel unable to start). `state.reset_channel_runtime()` clears both singleton cache surfaces.
- `agent/src/channels/config_meta.py`: field metadata (hand-written DingTalk hints with i18n help keys; `default_config()`-derived fallback) + fail-safe secret masking (`****`+last4; every secret-heuristic key stripped from `values` unconditionally — a parametrized sweep asserts this across all 16 adapters). Discovery no longer lists the `targets`/`config_meta` helper modules as channels.
- `agent/src/api/channels_config_routes.py` (new sibling module, `channels_routes.py` untouched): `GET /channels/config`, `PUT /channels/config/{name}` (merge-patch; ephemeral-construction validation **before** any write → 422 leaves the file byte-unchanged; enable-transition auto-verify with explicit `skip_verify` override), `POST /channels/{name}/test` (probes unsaved credentials, never persists). Per-channel `asyncio.Lock` serializes save/test/toggle; a global lock guards runtime rebuilds.

**DingTalk adapter fixes found during live QA**
- `_on_message` never forwarded `is_dm` → the DM pairing-code flow was dead code (unapproved DMs were silently denied).
- `stop()` never terminated the SDK stream loop (it swallows the first `CancelledError` and reconnects) → zombie WebSocket after disable/hot-swap and hung graceful shutdown; stop now double-cancels the tracked stream task.
- New `force_ipv4`: the robot-send API enforces the app egress-IP whitelist and a rotating IPv6 prefix on dual-stack networks keeps falling out of it (observed `403 IpNotInWhiteList`); binding egress to IPv4 makes the whitelist stable.

**Frontend**
- `ChannelSettings.tsx` (migrated status table + runtime controls, expandable rows) and a generic `ChannelConfigPanel.tsx`: text/password/bool/list-tag inputs from `fields` metadata, masked-secret placeholders with clear flags, Test on dirty form values, Enable toggle with an "Enable anyway" confirm after a failed verification, collapsible DingTalk setup guide, YAML read-only and missing-SDK banners. Rendering is field-driven, so a secret smuggled into `values` cannot reach the DOM (asserted in tests).
- New keys in all 8 locales; `api.ts` client per the frozen contract.

**Docs**
- "Configure from the Web UI" subsection in the IM channels section of all six READMEs.

## Screenshots

All screenshots use **example credentials** (`dingexampleappkey001` / `****7654`); no real app data appears.

DingTalk config panel — the stored secret is only ever shown masked, never echoed back:

DingTalk config panel

Entering a new Client Secret — the input stays masked while typing; Test connection verifies the unsaved value before anything is written:

Entering AppKey and Secret

Built-in setup guide, Force IPv4 option, and hot-apply save:

DingTalk setup guide

## Test Plan

- [x] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`) — **13904 passed, 98 skipped** on this branch's base
- [x] New tests added — 7 backend test files (writer, probe, reload/reset, metadata sweep, 23 route cases incl. lock serialization and a secret-hygiene sweep over raw response bodies, message-flow regressions) + frontend `ChannelSettings.test.tsx` (11 cases) and migrated Settings tests; frontend suite **644 passed**, `tsc --noEmit` and `npm run build` clean

**Manual end-to-end QA against a real DingTalk enterprise app (Stream Mode):**
1. `POST /channels/dingtalk/test` with valid credentials → `{"ok": true, "code": "ok"}`; with a corrupted secret → `invalid_credentials` carrying the provider's error, scrubbed; neither wrote anything to disk.
2. Saved credentials from the web panel → atomic write to `agent.json` (0600), secret returns only as `****CYzw`.
3. Enable on a running runtime → `applied: "hot_swapped"` in 0.23 s **including the real credential verification**, backend PID unchanged, exactly one stream connection after the swap (zombie fix).
4. Real DM to the bot → inbound through the Stream socket → session runtime → agent reply delivered back to DingTalk.
5. Disable → stopped, credentials preserved; re-enable without re-entry → running again.
6. Send initially failed with `403 IpNotInWhiteList` on the rotating IPv6 egress; enabling **Force IPv4** in the panel (hot-applied) resolved it and the reply landed.
7. Secret hygiene: zero plaintext-secret hits across all response bodies, the rendered DOM, and the server log.

## Checklist

---

Seven atomic commits, each independently green; DCO-signed. Marking as draft for maintainer review — happy to split the adapter fixes into a separate PR if preferred.


he-yufeng:
Heads up on parallel work: this PR and #1521 (mine) both implement #1519. I am leaving the pick to the maintainer, but laying out the factual differences to make that easier:

- This PR is wider: 43 files including the six README locales, frontend i18n, more frontend tests, and it reports live QA against a real DingTalk org app.
- #1521 is narrower: backend routes plus generic forms, and it has already been through a review round. CI is green as of today (the failing suite-isolation case is fixed and re-verified locally).

Whichever base lands, the two test surfaces look complementary rather than conflicting. Happy to port either side's tests over, or to review the other direction in detail if that helps.

warren618:
Merged, thanks. A few things on top:
- `ChannelManager.reload_channel` now holds a per-channel lock itself (cdb17c92). Your route lock covered save/test, but two reloads of one channel through the manager still each stopped the adapter they saw and started their own.
- `stop_all` cancels a reload's start that is still connecting, and the WebSocket channel's logger fix comes from #1521.
- `/agent.json` and its YAML twins are gitignored now (c1122f7e): with `VIBE_TRADING_HOME` pointing at a checkout, this surface would otherwise write channel credentials into the working tree.
- The 2026-09-20 news entry is folded into today's.

PR #1521, opened 2026-09-20
feat(channels): configure IM channels from the Web UI

### LINKED ISSUE #1519: [Feature] Configure IM channels from the Web UI: credentials, connection test, enable/disable with hot apply

he-yufeng:
## Summary

- Configure IM channels from the Web UI: per-channel field contracts (DingTalk fully guided — Client ID / Client Secret / enabled; the other 15 adapters expose the enabled toggle), validated merge-patch writes persisted atomically to `agent.json`, and hot apply without a process restart.
- A credential probe for DingTalk against the real token endpoint (`api.dingtalk.com/v1.0/oauth2/accessToken`), with honest `ok / missing_credentials / invalid_credentials / network / error / unsupported` outcomes instead of silent acceptance.
- A non-destructive enable/disable toggle that keeps credentials while starting/stopping just that channel instance.
- Incidental pre-existing fix (separate concern, one line): the websocket channel called a `logging.Logger` object as a function at start, so it crashed with `TypeError` before serving anything.

## Why

Closes #1519. The 16-adapter IM runtime had a read-only Settings panel: the only way to configure a channel was hand-editing `~/.vibe-trading/agent.json` and restarting the process, because `_get_channel_runtime()` snapshots the runtime once at first build (`api/state.py:99`) and `ChannelManager._init_channels()` only runs in the constructor (`channels/manager.py:62`). Verified on current main before building.

## Changes

- `agent/src/api/channels_config_routes.py` (new): `GET /channels/config-schema`, `GET /channels/config` (secrets masked to the last four; non-guided channels expose only `enabled`), `PUT /channels/config/{name}` (unknown channel 404, unknown field/type 422, adapter config dry-run before write, atomic 0600 persist, non-JSON configs refused with 409), `POST /channels/config/{name}/test`, `POST /channels/config/{name}/toggle`. Models defined locally per the routes-file maintainer rule.
- `agent/src/config/writer.py` (new): the small JSON-only write path (`save_channel_config_section`) — mkstemp + `os.replace` atomic write, 0600; YAML configs and corrupt files are refused loudly instead of rewritten.
- `agent/src/channels/manager.py`: `reconfigure_channel()` — reads fresh config, stops and drops the old instance, rebuilds via the `_init_channels` path, starts detached when the dispatcher is alive (DingTalk's `start()` is a reconnect loop and would hang an inline await). The manager now tracks `_started`: a manager started with zero channels builds no dispatcher, and hot-enabling the first channel brings the dispatcher up too (previously the instance was built but never started — apply-on-next-restart in disguise).
- `agent/src/channels/websocket.py`: one-token fix — use the logger object directly instead of calling it.
- `frontend/src/pages/Settings.tsx` + `api.ts` + 8 locale files: DingTalk credential form (masked save, test-connection result, enabled toggle), other channels get the toggle. All 8 locales because the i18n parity test requires key equality with `en.json`.

## Test Plan

- [x] Backend new tests: `test_channels_config_routes.py` (19) + `test_channel_manager_reconfigure.py` (6) — schema contract, masking, validation, persist+apply, toggle non-destructive, probe outcomes, zero-channel hot-enable
- [x] Regression: channels/api/runtime/media/cli suites 68 passed; config/settings/auth infra 163 passed
- [x] `test_websocket_channel_start.py` (2): the start smoke fails with `TypeError: 'Logger' object is not callable` without the one-line fix, passes with it (red-green)
- [x] Frontend `SettingsChannels.test.tsx` 9/9; `npm run build` passes. Full vitest suite has 38 pre-existing jsdom/localStorage failures identical to pristine upstream/main (compared side by side), unrelated to this change

Not run: live DingTalk with real AppKey/AppSecret (none available); the probe's ok/4xx/network paths are pinned with a fake httpx client. Browser-level rendering was verified through vitest/jsdom, not a manual dev-server clickthrough.

## Risk / rollback

- Writes go through `require_settings_write_auth` like the other settings mutations; reads are `require_local_or_auth`. Secrets are never returned in full (last four only), and a masked echo maps back to the stored value so a form re-save cannot wipe a secret.
- Hot apply touches only the named channel; other instances and the dispatcher are not rebuilt. Revert this PR to restore file-edit-plus-restart configuration.
- No live broker, order, payment, or wallet paths are involved — IM notification channels only.


warren618:
Thanks for this, and for the side-by-side comparison. #1520 and #1529 went in as the base, since #1529 builds on #1520. Two pieces of yours are on main with credit (cdb17c92): `stop_all` now cancels starts still in flight, and the WebSocket logger fix comes with your test. Two notes on this branch, for the record: the DingTalk probe returns the provider's response snippet verbatim in its error detail, so a response that echoed the secret would reach the UI; and `PUT /channels/config/targets` accepted and wrote the `targets` helper as if it were a channel. Closing as superseded.

PR #1522, opened 2026-09-20
fix(ml-strategy): fix single class training crash

Shizoqua:
## Summary
- Walk forward training now skips a retrain window that only contains one label class

## Why
A sustained one directional trend produces a training window with a single class, which crashes fit or predict_proba on every supported model type, aborting signal generation for every remaining symbol in the batch.

## Changes
- Skip the retrain and keep serving the previous model when the window has fewer than 2 classes
- Added a test using real scikit-learn classifiers instead of the stub model

## Test Plan
- [x] Existing tests pass (`pytest tests/test_ml_strategy_skill.py`)
- [x] New tests added
- [x] Tested manually

## Checklist


he-yufeng (commented):
The skip is the right behavior and the `continue` placement matches the too-few-samples case above it: a one-class window trains nothing useful, and keeping the previous model serving is the sane fallback. The test updates are honest — alternating labels keep the two stub-model tests exercising the same fits they pinned before, and the new real-classifier test proves no crash across all three model types.

Two small notes:

1. The "crashes on every model type" line overstates slightly: `LogisticRegression` hard-fails on a single class ("needs samples of at least 2 classes"), but `RandomForestClassifier` and `GradientBoostingClassifier` fit it fine and just predict that class forever. The skip is still right (that model is worthless), the failure mode is just not universal. Worth softening in the comment so a future reader doesn't delete the check thinking it only matters for ridge.

2. The new test proves no-crash but doesn't directly witness the skip: every assertion would also pass if the single-class window had been trained through (the classifiers tolerate it). One assertion that the windows before the label flip produced no fit (e.g. via the stub model's fits count, or that predictions right after the flip come from a model refit on two-class data) would pin the actual branch instead of just the absence of an exception.

warren618:
Merged, thanks. One change on top (950b1247): the `continue` sat inside the per-day loop, so on the retrain day itself the prediction stayed 0.0 rather than coming from the previous model the comment describes. The too-few-rows case had the same shape. Both now skip only the retrain, with a test on a sliding window that turns single-class.

Issue #1463, opened 2026-09-16
[Bug] Alphas still emit values from a window that contains a missing bar — fix once at the registry, not per family


warren618:
Since #1377, `Registry.compute` masks an alpha's output to NaN wherever a declared dependency is missing **on that bar**. It does not mask the bars *after* a gap whose rolling window still contains it, so an alpha that turns a comparison into 0/1, calls `.where(cond, 0)`, or carries an `ewm` through the gap still emits a value that a halt fabricated.

Verified here on current main, through `Registry().compute`:
- qlib158 `cnt{p,n,d}{5,10,20,30,60}`: after one missing close, main emits values for the whole window that follows it. #1459 fixes those 15.
- gtja191_064 shows the same shape; it holds the only `fillna` in the zoo (#1452).

An offline sweep over all 462 alphas (perturbation test: the value moves when the missing input changes, yet is finite when that input is NaN) counted 85 such alphas within their own declared `min_warmup_bars`, and 69 remaining after #1459. Those counts are from that sweep, not re-measured by hand — worth re-running before acting on the exact number, but the class is confirmed by the two cases above.

Why this should not be one PR per family:
- the families share five mechanisms (comparison → 0/1, raw `.rolling`, `.ewm` carry-forward, `.where`, `np.fmax`), not one;
- 141 of 462 modules call pandas `.rolling` / `.ewm` directly, so an operator-layer fix in `factors/base.py` cannot reach them.

Proposed direction, for a decision:
1. A registry-level look-back mask: mask bar `t` when a declared dependency is missing anywhere in `[t − L, t]`, with `L` either `min_warmup_bars` or a new explicit `lookback` field. The sweep suggests this removes fabrication for a bit over half of the offenders, at the cost of extra NaN bars for alphas whose declared warmup exceeds what they actually use.
2. A single shared policy for `ewm` / SMA after a gap, plus corrected warmup declarations for the alphas that emit past their own declared warmup.

Until this lands, per-family NaN PRs should be held (noted on #1459).

warren618:
I re-measured before acting, and the look-back mask proposed above doesn't hold up.

[...]

PR #1523, opened 2026-09-20
fix(factors): fix gtja191_038 NaN propagation

Shizoqua:
## Summary
- gtja191_038 now propagates NaN through a data gap instead of fabricating 0.0

## Why
The 20 day rolling mean is NaN during warmup or after a gap, but the comparison against it evaluates to False rather than NaN, so the fallback branch silently filled in 0.0 for every affected bar.

## Changes
- Mask the output wherever the rolling mean is NaN
- Added regression test with a data gap

## Test Plan
- [x] Existing tests pass (`pytest tests/factors/`)
- [x] New tests added
- [x] Tested manually

## Checklist


warren618:
Merged, thanks. This turned out to be one member of a class: a comparison whose operand is NaN because its window holds the gap reads False and emits a constant. The #1463 sweep perturbed the gap bar by a tick and couldn't see that, because a comparison doesn't move with a tick. Comparing the gapped run with the gap-free one found 22 more alphas beyond these two, fixed together in 52f14773.

PR #1524, opened 2026-09-20
fix(slack): fix missing DM pairing code

Shizoqua:
## Summary
- An unauthorized Slack DM sender now actually receives a pairing code

## Why
Slack overrides is_allowed to always return true since it does its own channel aware check first, but that also silences the base class pairing code branch, so the deliberate call meant to trigger it instead published an empty message to the agent bus and the sender got no code and no denial.

## Changes
- Issue the pairing code directly at the call site instead of relying on the overridden check
- Added regression tests for the unauthorized and authorized paths

## Test Plan
- [x] Existing tests pass (`pytest tests/test_slack_dm_pairing.py tests/test_channels_runtime.py`)
- [x] New tests added
- [x] Tested manually

## Checklist


warren618:
Merged, thanks. CI was red because the test module imported `slack_sdk`, which CI doesn't install. bf67839b moves both Slack test files onto one stub helper that stands down when the SDK is present; the old `SocketModeResponse = object` stub in the table test broke your test whenever it was imported first.

PR #1525, opened 2026-09-20
fix(memory): fix duplicate title index clobber

Shizoqua:
## Summary
- Memory index rows are now matched on filename instead of title alone

## Why
Two memory entries with the same title but different memory types are written to two different files, but the index update matched on title only, so the second entry silently overwrote the first entry's row in the index that gets injected into the system prompt, even though both files exist on disk.

## Changes
- Match on the title and filename together when deciding to update or append
- Added regression test with same title, different memory types

## Test Plan
- [x] Existing tests pass (`pytest tests/test_persistent_memory.py`)
- [x] New tests added
- [x] Tested manually

## Checklist


warren618:
Merged, thanks. Follow-up on the same axis (acd1fb19): with two same-title entries allowed, `remove(title)` deleted whichever it scanned first. It now takes `memory_type`, refuses an ambiguous title, and `forget` passes the type through.

PR #1526, opened 2026-09-20
fix(robinhood): reject malformed portfolio buying power

### LINKED ISSUE #1428: [Bug] Portfolio excludes authenticated Robinhood connection despite working account reads
### What happens
On Vibe-Trading 0.1.15 on macOS (Apple Silicon), Robinhood OAuth succeeds and direct account and position reads succeed when an account is supplied. However, Portfolio → Manage accounts offers no Robinhood connection.

### Reproduction
1. Authenticate the built-in robinhood-live-mcp connector.
2. Verify account and position reads with the selected account.
3. Open Portfolio → Manage accounts → Open connection center.
4. Robinhood is unavailable for portfolio selection.

### Code evidence
agent/src/trading/connections.py: is_portfolio_connection_profile requires profile.readonly to be true.
agent/src/trading/connectors/robinhood/profiles.py: the only Robinhood profile declares account.read and positions.read, but readonly=False.

### Expected behavior
Provide a read-only Portfolio view of an authenticated Robinhood account, or clearly explain the unsupported connection. Preserve the existing separation between portfolio reads and order execution. A dedicated read-only profile may be more appropriate than relaxing the gate. Account selection also needs to be carried through to the read calls.

No account identifiers, balances, tokens, or private logs are included.

lorenzozanee:
## Summary

- Reject malformed Robinhood portfolio `buying_power` responses at the mapping boundary.
- Add a regression test for a non-object `buying_power` payload.

## Why

The Robinhood mapper previously treated a non-object `buying_power` value as an omitted optional field, allowing a partial account summary to continue with `buying_power: None`. This follow-up makes unsupported response shapes fail closed.

The primary account-bound read-only implementation for Issue #1428 is already present on `main`; this PR is limited to the malformed nested-response hardening described above.

Fixes #1428

## Changes

- Require `buying_power` to be an object before mapping it.
- Pass its nested value through the existing finite-decimal validator.
- Add a regression test proving a list-shaped response raises `RobinhoodShapeError`.

## Test Plan

- [ ] Existing tests pass (`pytest --ignore=agent/tests/e2e_backtest --tb=short -q`) — full suite not run.
- [x] New tests added (focused regression and Robinhood portfolio/account-binding coverage).
- [ ] Tested manually — not applicable; validation uses sanitized fixtures and no live broker access.

Focused validation: 65 tests passed; Ruff check, Ruff format check, Python compile, and `git diff --check` passed.

## Checklist

- [x] No changes to protected (`src/agent/`, `src/session/`, `src/providers/`) without prior discussion
- [ ] Documentation updated (if user-facing change) — no interface or documentation change was introduced.


warren618 (changes requested):
Thanks. Two changes, please:
1. The body says "Fixes #1428". Merging would close #1428, which stays open on purpose for the quotes and pagination shapes. "Refs #1428" keeps the link.
2. The change also makes `buying_power` required: an omitted, null, or inner-omitted value used to give `None` and now raises `RobinhoodShapeError`. The totals don't use it, and #1428 never showed whether Robinhood always sends it. Raising only when it's present and not an object matches what the title says. Please also update the Raises docstring.

Issue #1494, opened 2026-09-19
[Bug] QVeris loader: source="qveris" picks tools without regard to price adjustment, mis-maps parameters ("indicators" gets the end date) and cannot parse adj_* responses


jw232:
- **Interface:** MCP Plugin
- **LLM Provider:** n/a
- **Version:** 0.1.15. `agent/backtest/loaders/qveris_loader.py` on `main` is byte-identical.
- **Environment:** macOS (Darwin 25.6.0) / Python 3.14.5

### Description

**What happened:**
`backtest/loaders/qveris_loader.py` picks a QVeris capability at run time and fills its parameters by heuristics. We replayed the loader's own functions offline against real QVeris responses from 2026-09-19. For A-share daily bars, four problems appear.

1. **Selection ignores price adjustment.**
   - `DataLoader._fetch_one` (lines 335-381) searches with `_search_query` (lines 387-389). The query is `"daily OHLCV historical market data for 600519.SH interval 1D"`.
   - It keeps tools that look like OHLCV and ranks them by success rate, then by quoted cost (`_capability_rank`, lines 454-457). It tries up to `_MAX_CANDIDATES = 3`.
   - For that exact query, the top three were FMP endpoints at 24.2 credits each. First was `financialmodelingprep.historical_price_eod.non_split_adjusted.retrieve.v1.4c43e8ed`.
   - The loader never sets or checks an adjustment parameter. Its source has no `cps`, `adjustflag`, `dividend` or `split` handling. `PRICE_CALIBER_BY_SOURCE` has no `qveris` entry, so the caliber of the bars it serves is unknown.
2. **Parameters are matched by substring.**
   - `_build_parameters` (lines 487-511) maps parameter names by substring. `_is_end_param` (lines 538-539) matches `"to"`, so `indicators` counts as an end date.
   - For `cn_financial_pro.history_quotation.v1` it builds `{"codes": "600519.SH", "startdate": "2019-01-02", "enddate": "2019-12-31", "interval": "daily", "indicators": "2019-12-31"}`.
   - As a result, `indicators` receives a date, and `interval` receives `daily` where the tool documents `'D'`. `cps` is not sent, so the tool's stock default `'1'` applies, which is unadjusted.
3. **`adj_*` responses cannot be parsed.**
   - `_FIELD_ALIASES` (line 54) accepts `adj_close` for close but has no `adj_open`, `adj_high` or `adj_low`.
[...]

PR #1527, opened 2026-09-20
fix(qveris): resolve adjusted fields as a complete set, never per field

cgycorey:
## Summary

Problem 3 of #1494: the loader treats adjusted price fields inconsistently, which both discards a paid response and silently corrupts a bar.

`_FIELD_ALIASES` accepted `adj_close` under `close` while `adj_open`/`adj_high`/`adj_low` had no alias at all. That cut two ways, and I reproduced both before changing anything:

- **A response carrying only adjusted fields could never parse.** `qveris_finance.mkt_bars_adjusted` returns `adj_open`/`adj_high`/`adj_low`/`adj_close` plus `adj_factor` and no unadjusted price fields, so `_record_has_ohlc` rejected every row, `_result_to_frame` returned `None`, and the billed call reported "no parseable bars".
- **A response carrying unadjusted open/high/low with only `adj_close` was accepted as a single bar mixing the two series:**

```
              open  high  low  close  volume
trade_date
2024-01-02    1.0   1.2  0.9    1.1     NaN    <- close is the adjusted close; the rest are not
```

## Changes

- Split the alias tables into `_FIELD_ALIASES` (unadjusted) and `_ADJUSTED_FIELD_ALIASES` (adjusted), and removed `adj_close`/`adjusted_close` from the unadjusted `close` aliases.
- Added `_resolve_field_set`, which picks the table supplying a **complete price quartet**: unadjusted when complete (the preference the old lookup had), otherwise adjusted, otherwise `None`. A bar is never assembled from a mix.
- Routed both `_record_has_ohlc` and `_normalize_record` through it, so "is this a bar" and "build this bar" can no longer disagree.

## Test Plan

Tests were written first and watched fail:

- `test_adjusted_only_records_become_bars` — failed before with `assert None is not None`.
- `test_a_bar_never_mixes_unadjusted_and_adjusted_levels` — failed before by returning the mixed bar shown above.
- `test_unadjusted_fields_win_when_both_sets_are_present` — preservation guard for the existing preference; it passes before and after, and pins that this change does not flip it.

```
PYTHONPATH=agent pytest agent/tests/test_qveris_loader.py agent/tests/test_qveris_tool.py \
  agent/tests/test_qveris_mcp.py agent/tests/test_qveris_routes.py \
  agent/tests/test_qveris_minute_granularity_tokens.py agent/tests/test_market_data_tool.py \
  agent/tests/test_metrics.py agent/tests/test_local_source_routing.py \
  agent/tests/test_cli_data.py agent/tests/test_tools_type_value_safety.py -q
# 238 passed

PYTHONPATH=agent pytest agent/tests -q
# 13850 passed, 14 failed, 34 skipped
```

The 14 failures are pre-existing and unrelated (mcp client adapter, mcp oauth schema, oauth token cache, README tool counts, one 35 ms alignment perf test). The same 14 fail on `main` with this change absent — verified by running those six modules on both trees: `14 failed, 180 passed` either way.

`ruff check` passes. `ruff format --check` flags these two files on `main` as well, at hunks outside this diff, so I left unrelated code alone rather than reformat it in here.

## Scope and Risk

Deliberately **not** included, from the same issue:

- **Problem 1 (no `qveris` price caliber).** `PRICE_CALIBER_BY_SOURCE` still has no `qveris` key, so bars served by this source remain `"unknown"` and stay out of mixed-caliber comparison. This PR does not claim the bars are unadjusted, and it does not fix that. Pinning needs a per-capability adjustment choice — which is what your measurements of `cps=2` against `cps=6` establish — and I cannot reproduce those calls without a key.
- **Problem 4 (budget reservation).** `_expected_cost` still takes its number from the quote text. I did not "improve the parse" because the evidence in #1494 shows the quote cannot bound the cost: `cn_financial_pro.history_quotation.v1` quotes `"1 credits/result"` and then billed 9.66 credits for a single stock-year on a per-value rule (244 rows × 30 fields). A cleverer parse of that string would still be a guess. The two honest options are to fail closed on quantity-based quotes, or to pin a per-capability price; both are pricing-policy calls I would rather leave to you.

Problem 2 was already fixed by #1496 (merged), so this PR does not touch `_is_end_param`.

Related: #1494


warren618:
Merged, thanks. Two gaps closed on top (c59e0a06): the family was chosen per record, so a payload with one unadjusted row and one adjusted-only row became a series mixing both levels (closes 11.0 then 1.1); it's now one family per response, and a response whose records disagree yields no bars. And an unadjusted bar no longer takes `adj_volume`, which is not on its scale.

